PR #120, opened 2011-04-06
One pool per redis instance - issue 118

grdmitro:
Reusing implementation this time


HeartSaVioR:
We're now using Commons Pool 2. Marked as Invalid.

HeartSaVioR:
Sorry, wrong explanation.
We agree that we don't treat ShardedJedis as core of Jedis, and for now we don't improve it.
Also it is very old PR so we cannot follow up.

PR #133, opened 2011-04-28
issue 118 - One pool per redis instance

nklasens:
There is already a pull request for the same issue, but to me that implementation could have been simpler. That implementation introduced even more dependecies from the none sharding code (BinaryJedis and Jedis) to the shardinfo implementation. In my opinion. these classes should not know anything about sharding.To return the Jedis instance to the correct pool I decided to use the host and port from the shardinfo and jedis client.


HeartSaVioR:
Since Jedis has changed too much, we can't apply this. 
Furthermore, we agree that we don't treat ShardedJedis as core of Jedis, and for now we don't improve it.
If you think that we really need this feature, please leave an issue and link this so that we don't forget it.
Closing. Thanks!

Issue #134, opened 2011-04-28
NPE in SafeEncoder


vivekhub:
John - found an NPE issue with the SafeEncoder class.  Here is the sample code and the NPE

package bug.redis;

import java.util.HashMap;
import redis.clients.jedis.*;

class simulate_new1{

```
public static void main(String args[])
{
    JedisPool MyPool = new JedisPool("home-host", 6379);

    Jedis myDb = MyPool.getResource();
    String value = null;
    myDb.hset("samplekey", "key", value);

    MyPool.returnResource(myDb);
}
```

}

The NPE is

$ java bug.redis.simulate_new1
Exception in thread "main" java.lang.NullPointerException
    at redis.clients.util.SafeEncoder.encode(SafeEncoder.java:15)
    at redis.clients.jedis.Client.hset(Client.java:134)
    at redis.clients.jedis.Jedis.hset(Jedis.java:670)
    at bug.redis.simulate_new1.main(simulate_new1.java:14)

PR #140, opened 2011-05-11
Added a few overridden methods to support a Set<T> instead of variable length arguments

vivekhub:
the sunion / zunion etc., were all supporting variable length arguments.  I have added additional overridden methods for the same API using a Set<T>.  Hope you can merge it in so that I dont have to maintain my own private version of the library ;)

Also fixed bug #134


HeartSaVioR:
We have already huge methods set, so overriding methods should be taken seriously.
Since we don't get any requests about this, so for now I'm marking this as invalid.

PR #157, opened 2011-06-06
multi get should return a map for easier usage of results

hekoru:
When a list is returned from a multiget request you move to the client the burden of checking what keys were returned and which did not. I just copied the multiget method an added a new one called multigetMap which returns a map instead of a list.

Thanks!


HeartSaVioR:
For now we don't need it now, cause we already have too many methods.
Please comment this if you disagree. Thanks!

PR #174, opened 2011-06-29
Dynamic sharding implementation

yaourt:
Please could you review the following proposal and tell me what do you think about it ?

Currently, Jedis doesn't support dynamic sharding.

If we want to add / remove a Redis server from the available shards, then the only way to achieve this is to stop the application, change the sharding configuration and then restart it.

This implementation allows various things like :
- to add / remove some servers without downtime for the application
- to externalize the Redis connectivity monitoring and adapt the sharding configuration in pseudo-realtime
- to centralize the sharding configuration for a cluster in a DB or something similar
- etc

This proposal delegates the sharding configuration to a component outside of Jedis and allow to dynamically update the sharding configuration without down time.

JedisDynamicShardsProvider class is doing the bridge between this component and Jedis internals.

The external component manages the sharding configuration the way it wants and when an update is required, it only has to call the relevant method add/removeShard( ) or setShards( ) on JedisDynamicShardsProvider.

Upon such a call, JedisDynamicShardsProvider then notify (when absolutely required) Sharded instances about such a change (inspired from the Observable / Observer pattern)

When Sharded handles such a notification, then all the read access on the shard configuration are blocked until it has finished to change the sharding configuration (ReentrantReadWriteLock usage).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,7 @@
 .classpath
+*.iml
+*.ipr
+*.iws
 .project
 .settings/
 .gradle/
--- a/README.md
+++ b/README.md
@@ -48,20 +48,23 @@ You can download the latest build at:
 
 Or use it as a maven dependency:
 
-    <dependency>
-        <groupId>redis.clients</groupId>
-        <artifactId>jedis</artifactId>
-        <version>2.0.0</version>
-        <type>jar</type>
-        <scope>compile</scope>
-    </dependency>
-
+```xml
+<dependency>
+    <groupId>redis.clients</groupId>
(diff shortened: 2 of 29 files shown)


HeartSaVioR:
We agree that we don't treat ShardedJedis as core of Jedis, and for now we don't improve it.
If you think that we really need this feature, please leave an issue and link this so that we don't forget it.
Closing.

phenomax:
Is this feature still under consideration? As a workaround I thought about disconnecting reconnecting the connection to get the new nodes. What do you think about that?

PR #196, opened 2011-08-23
make pubsub threadsafe

smparkes:
Calling subscribe/unsubscribe on a PubSub object from a foreign thread isn't threadsafe because callbacks may also be calling these functions.

This patch simply makes things that might write to redis synchronized while leaving the reader unsynchronized.

It also resets the pipeline count on exit from a subscribe loop since the number isn't used in the pubsub case.


HeartSaVioR:
It is still valid, but we should check implementation again.

smparkes:
I haven't worked on this in a long time, so I'm gonna abandon it.

PR #212, opened 2011-09-26
added @Overrides and import fixes

raszi:
There are missing `@Override` annotations in Jedis code, I fixed them in this patch.


HeartSaVioR:
I agree that we should add @Override annotations to Jedis, but many things were changed so current PR seems to be invalid.
So I'm closing this issue and open new issue addressing on this. Thanks!

PR #213, opened 2011-09-26
provide a reset() method to read all remaining data

raszi:
It would be useful to have a reset() method in Jedis client.

It could be called before returning a shared Jedis client to the pool, or borrowing it again.


HeartSaVioR:
I already introduced resetState() which covers more internal states. 
Marking this as invalid. Closing.

PR #245, opened 2011-12-22
Pull request

nachogiljaldo:
Hi,

please, pull the changes I include. They're basically done to provide support for Unix Domain Sockets (Issue 144). I extended the Jedis class (I thought that was the easiest way to avoid modifying the API).

I also added test for other tickets which were already solved, I think it's good practice to create tests for the bugs already fixed.

Any problem you might have just let me know and I will try to solve it as soon as possible, I am github and I do not discard having made some mistake.

BTW. I'd like to collaborate with you more actively


HeartSaVioR:
I don't believe we can apply it now, cause we don't plan to separate layers.
If you think we really need it, please leave a comment. Closing. Thanks!

Issue #171, opened 2011-06-27
jedisPool.returnRessource doesn't accept BinaryJedis


ib84:
this works:
    private BinaryJedis binaryJedis = pool.getResource();

this doesn't:
    pool.returnResource(binaryJedis);

trivial workaround:
    pool.returnResource((Jedis) binaryJedis);

PR #276, opened 2012-04-17
Issue #171

ivowiblo:


sparkpool:
As i can see  below methods are not present in jedis 2.6.2. Is there any specific reason or there is another to release binary jedis.

public void returnBrokenResource(final BinaryJedis resource) {
+    	returnBrokenResourceObject(resource);
+    }
+    
+    public void returnResource(final BinaryJedis resource) {
+    	returnResourceObject(resource);
+    }

PR #292, opened 2012-05-07
Method getResponse on Queuable renamed to enqueueResponse for improving readability

ivowiblo:


HeartSaVioR:
@xetorthio @marcosnils 
I also agree that enqueueResponse or other names would better to read compared to getResponse().
We cannot apply PR directly, but we can take an ideation.

xetorthio:
Yes. What @ivowiblo suggests is correct.
Lets takw the idea and make the necessary changes. Maybe we can close this
PR and create an issue to analyze this further and decide the right way to
do it.

PR #319, opened 2012-07-12
Add More Variadic Commands

imleon:
Now hdel/rpush/lpush/sadd/srem/zrem support Set/List<String/byte[]>


HeartSaVioR:
For now we don't need it now, cause we already have too many methods. Please comment this if you disagree. Thanks!

PR #320, opened 2012-07-13
 RedisInputStream#readLine can't decode multi-byte characters.

masayuki038:
Since RedisInputStream#readLine can't decode multi-byte characters, changing the way to decode from byte[] to string.


HeartSaVioR:
Now RedisInputStream.readLine() is used for only reading error line.
Since we don't receive reports regarding this, I feel it doesn't really matter.
Please comment if you meet any issues regarding this.
Closing.

PR #336, opened 2012-08-17
I  have correct the format, and just change the zadd method.

sing1ee:
Is that ok?


HeartSaVioR:
Already fixed. Closing.

PR #345, opened 2012-09-15
Make jedis pool return jedis pooled connection object.

mpillar:
And some small build changes.


HeartSaVioR:
There're too much changes on Jedis so I don't think I can apply this now.
Closing.

ps. ```ambiguous reference to overloaded definition``` issue should be reported to another issue.

PR #349, opened 2012-09-16
Async Support

samhendley:
First let me say how impressed I am with this library. Its incredibly well written and made adding a somewhat complex feature surprisingly easy. 

I have a project where we are doing a few relatively simple redis operations on a very fast stream of data. We have reached the limits of the current system and are latency bound waiting for the request to be returned to us. Our project is async in other places so we can get a huge performance win by preparing and making our next request even before we have the answer to the current request. Another way to accomplish the same goal would be to use many clients in parallel but the order of the data matters so its much simpler to increase the performance of a single thread if possible.

The implementation is pretty simple, the key changes were adding a ResponseListener callback to the Response class and a background thread on the Connection that spins reading the responses and calling the Response.set function.

I think the error handling may need a bit of work but it should be usable as is.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@
 target/
 build/
 bin/
+.idea
+out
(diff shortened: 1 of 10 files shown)


HeartSaVioR:
As you already know, we're planning to adopt nonblocking async feature, which could be better to have it.
So for now I'd like to close this first.

PR #361, opened 2012-10-12
ConsistentJedisPool implementation

brianchen2012:
Based on the JedisPool, I put a consistent hash implementation, plus a template style RedisClient to use the ConsistentJedisPool.

The modification allows removing  the failed server and rehashing.  so network or redis server down will have no impact to  client.

The RedisClient make the user no need worry about the resource borrowing and returning.

This solution can dealing with any Redis server down and shard the requests to different Redis servers, easy to cluster the Redis Servers.
System will have no down time when the redis server  goes away and comes back.


HeartSaVioR:
Since we're having ShardedJedis, I'd like to see consistent hash via ShardedJedis.
Sorry for being too late for feedback. Closing.

PR #371, opened 2012-11-27
Added support for binary keys in DEBUG OBJECT command

alexdeleon:
We use especially serialized keys which can not be converted to UTF-8 Strings, so I added support to call a DEBUG OBJECT command using a binary key.

[Review on Reviewable](https://reviewable.io/reviews/xetorthio/jedis/371)

### DIFF
--- a/src/main/java/redis/clients/jedis/DebugParams.java
+++ b/src/main/java/redis/clients/jedis/DebugParams.java
@@ -1,9 +1,11 @@
 package redis.clients.jedis;
 
-public class DebugParams {
-    private String[] command;
+import redis.clients.util.SafeEncoder;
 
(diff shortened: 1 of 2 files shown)


HeartSaVioR:
I think it is still valid. @xetorthio @marcosnils What do you think about?

xetorthio:
Yes. LGTM

sazzad16 (commented):
@gkorland 

> DEBUG OBJECT is a debugging command that **should not be used by clients**. Check the OBJECT command instead.

sazzad16:
https://redis.io/commands/debug/ 

> The `DEBUG` command is an internal command. It is meant to be used for developing and testing Redis.

PR #386, opened 2013-01-24
add FailSafe implementation

huy:
Hi

I have added a simple implementation of Fail Safe Jedis that can detect if Redis Master fails or has been changed and gracefully switch to the new Master (if there is any) in the context of Redis Master/Slave replication cluster.

We are using this code in one of our project, so if you feel  that it is useful for community please incorporate it into master branch. 

I had an intention to do some refactor to reduce a bit of duplication and rewrite sharding implementation to allow combine sharding with fail safe features but  don´t want to touch the original code base without knowing your opinion first.


HeartSaVioR:
Jedis supports Sentinel now. So for now I'm closing this issue.

PR #387, opened 2013-01-30
Execute several TransactionBlock in a unique transaction

enricruiz:
TransactionBlockList allows the execution of several TransactionBlock in the same Transaction

``` java
TransactionBlock transactionBlockA = new TransactionBlock() {
    @Override
    public void execute() throws JedisException {
        set("A", UUID.randomUUID().toString());
    }
};

TransactionBlock transactionBlockB = new TransactionBlock() {
    @Override
    public void execute() throws JedisException
    {
        set("B", UUID.randomUUID().toString());
        set("C", UUID.randomUUID().toString());
    }
};

TransactionBlockList transactionBlockList = new TransactionBlockList();
transactionBlockList.add(transactionBlockA);
transactionBlockList.add(transactionBlockB);

Jedis redis = new Jedis("localhost");
List<Object> results = redis.multi(transactionBlockList);
redis.disconnect();
```

It is useful, for example, if you have abstracted all the persistence operations of an object in a DAO where all modification methods returns a TransactionBlock and you want to execute several calls in a unique transaction.

### DIFF
--- a/src/main/java/redis/clients/jedis/Queable.java
+++ b/src/main/java/redis/clients/jedis/Queable.java
@@ -4,7 +4,7 @@ import java.util.LinkedList;
 import java.util.Queue;
 
 public class Queable {
-    private Queue<Response<?>> pipelinedResponses = new LinkedList<Response<?>>();
+    protected Queue<Response<?>> pipelinedResponses = new LinkedList<Response<?>>();
 
     protected void clean() {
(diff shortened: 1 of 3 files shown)


HeartSaVioR:
I'm sorry, but we're planning to remove TransactionBlock, and actually remove it to master.
Marking as invalid. Closing.

PR #389, opened 2013-02-01
Proper sort (with STORE) implementation for multi

vijaykramesh:
Currently sort with a dstkey inside a multi() transaction raises an exception (java.lang.ClassCastException: java.lang.Long cannot be cast to java.util.List) as redis returns a Long as the response, but jedis is expecting it to return a Set of Strings. Updated this to work correctly - first commit adds a failing test showing the problem, second commit makes that test pass.


HeartSaVioR:
I don't think we should change return type to support optional parameter and break whole things. 
We can't nomalize it while return types are different. It should be treated to new method.
So I'm going to close this issue but open new issue addressing this.

HeartSaVioR:
Never mind. It was fixed.

PR #411, opened 2013-04-04
Intoduce dedicated excpetion for Redis error "busy"

tsc-nemeses:
This exception indicates that Redis is currently busy. A dedicated exception is helpful to recognize operations which can be scheduled for a retry. Hope you agree and add this change to your code base. Thx!


juhrlass:
This would be very useful and is really easy to integrate. Please pull this in.

xetorthio:
As far as I remember redis returns "-BUSY" when it is starting and still loading data.
Are you restarting redis while you app is up & running?
There is no harm in merging this PR (although a test is missing) but I would love to understand better why you are running into this situation.
Usually what people do, is to put your app into a maintenance mode and put it productive as soon as redis answers to a PING command.

marcosnils:
@xetorthio I'd imagine this may become handy in case one if your redis instances restarts and then you'll get the BUSY error until the AOF / DUMP has finished loading. I have some opinions about this

1) You can do this today in a hacky way by catching the JedisDataException and checking the exception message. 
2) I do agree with @xetorthio that if you're aware that your redis instance will return BUSY because you'r manually restarting it, then you should put your apps in some sort of maintenance mode during the lapse of the restart. 
3) Another thing to notice if you're using `JedisPool` (which is the most cases) you can set the `testOnBorrow` option to make the pool only return a connection when `PING` responds. 
4) After all Redis is a cache/db. In most cases if the cache is down you should send your users somwhere else to retrieve the data (until you don't get any more exceptions), and if it's not the case where your system can't work without redis, then "sleeping" a fixed time to wait until it becomes (not-busy) doesn't seem a good strategy to follow.

Saying that, I can't find a scenario where I would catch this specific exception and do something about it. @juhrlass I'd like to hear more about how you'd use this.

juhrlass:
We implemented a Apache Camel Endpoint with Redis Lists and Jedis. The reason for this is that we can partly process our data with lua scripts and trigger further processing with the help of Java Processors. Camel ist our glue between the two worlds.
Now in rare cases (Data migration, external data checks, other administrative changes) it is possible that redis is BUSY while doing processing.
Camel has the concept of "redeliveryPolicyProfiles" where we configured "Just retry when the following exception is thrown". In our case the JedisRedisBusyException.
We do not want to use JedisDataException. We cannot configure JedisDataException with certain message as retry condition.
So we just patched the code to return a dedicated JedisRedisBusyException when redis is BUSY.
For us this has the benifit of:
- configure behaviour in a rare edge case with simple camel rules
- Handle the case in a defined way that is acceptable in our use case
Otherwise we have to go with a patched jedis version or would have to implement a lot more stuff to work around this.

marcosnils:
@juhrlass seems like a good reason to include. 
Can you please rebase and add a test?

@HeartSaVioR thoughts?

sesamzoo:
@marcosnils I rebased and added a test. SlowLogCommandTest fails in the openJDK7 build [1179](https://travis-ci.org/xetorthio/jedis/builds/119957009) , but I think it's unrelated, because the build [1178](https://travis-ci.org/xetorthio/jedis/builds/119946738) succeeded for 45ead5d (the rebase commit w/o the test for JedisRedisBusyException). Could you trigger a re-build of build 1179 in travis?

@tsc-nemeses and @juhrlass FYI: commit history was rewritten due to the rebase of an already published change. If you have any extra commits in addition to the original "Introduce JedisRedisBusyException" in your master branch, you might follow [this advice](http://stackoverflow.com/questions/16773249/git-pulling-a-rebased-branch). Without any extra commits the simplest way is to delete the local master branch (or the entire local repo) and check out again.

sesamzoo:
@marcosnils I found a meaningful change (IMHO) to justify another commit and hence triggered another [build](https://travis-ci.org/xetorthio/jedis/builds/120019117) which succeeded.

src/test/java/redis/clients/jedis/tests/ProtocolTest.java
```diff
@@ -119,4 +120,17 @@ public void nullMultiBulkReply() {
     List<String> response = (List<String>) Protocol.read(new RedisInputStream(is));
     assertNull(response);
   }
-}
\ No newline at end of file
+
+  @Test
```
marcosnils: You can use the `expected` annotation parameter to test for exceptions, no need to try-catch inside test.

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -10,6 +10,7 @@
 import redis.clients.jedis.exceptions.JedisConnectionException;
 import redis.clients.jedis.exceptions.JedisDataException;
 import redis.clients.jedis.exceptions.JedisMovedDataException;
+import redis.clients.jedis.exceptions.JedisRedisBusyException;
```
marcosnils: s/JedisRedis/Jedis/

marcosnils:
@sesamzoo PR looks good overall. Just made some minor comments about conventions. 

After that once @HeartSaVioR approves we can merge!

src/test/java/redis/clients/jedis/tests/ProtocolTest.java
sesamzoo: I didn't use the `expected` annotation parameter because it wouldn't allow to [assert the correct content](https://github.com/xetorthio/jedis/pull/411/files/f399e4fa213fe298e26613d3c03055ab75722d3b#diff-43462f06c65dec4783b0c0abcb502e15R131) of the exception's message.

Do you imply that it should not be tested?

marcosnils: Hmmm.. you're right, even though it's trivial I prefer to leave it this way. Thx for clarifying.

marcosnils:
@HeartSaVioR this is ready to be review. WDYT?

HeartSaVioR:
Sorry for late arrival. LGTM.

marcosnils:
Downmerged to 2.8 and 2.9 respectively. Thx @sesamzoo for the contribution!.

HeartSaVioR:
I got late ;) I was trying 'squash and merge', which I think is good for commit log.
@sesamzoo Thanks for the great work!

marcosnils:
@HeartSaVioR yup, I squashed all commits into one :smile:

HeartSaVioR:
@marcosnils Amazing! You got me. ;)

PR #419, opened 2013-04-19
Make JedisByteHashMap entries Serializable

lmmworko:
As far as I can see it makes no sense for the JedisByteHashMap to be Serializable if its entries are not?
(I hit this when attempting to use the output of Jedis.hgetAll(byte[]) in a wicket page, which has to be serializable)


HeartSaVioR:
It could be still valid, so I'd like @xetorthio @marcosnils to take a look, too.

sazzad16:
Resolved by #2008

PR #422, opened 2013-04-26
add count parameter to srandmember

TioBorracho:
In redis 2.6+ srandmember there is an aditional optional parameter to get multiple random members.

http://redis.io/commands/srandmember


HeartSaVioR:
Already fixed. Closing.

PR #505, opened 2013-12-24
Implement strategy pattern for sharding

### LINKED ISSUE #504: Implement strategy pattern for sharding
It must be a good option to determine custom strategy for sharding logic.
One of the reason to implement this is to maintain same logic for clients implemented on different programming languages.

a4tunado:
FIxes #504


sazzad16:
Sharding feature will be removed.

PR #611, opened 2014-04-03
add SSL support

oranagra:
Adding SSL support.
Tested against stunnel on the server side.
Using:
  System.setProperty("javax.net.ssl.trustStore", "Truststore.jks");
  System.setProperty("javax.net.ssl.trustStorePassword", "mypassword");
  System.setProperty("javax.net.ssl.trustStoreType", "jks");
Once the pull request is accepted, i'll put some how-to in the wiki.


JF9:
Another vote for SSL solely to provide encryption.

pr0met:
Another vote for ssl support.
Successfully tested this change against a Redis server running on Azure.

blicket:
+1 for ssl support.

bryancampbell:
+1 for ssl support

sunnypav:
+1 have tested this feature and it is working fine. When it gets merged in the code.

sunnypav:
When this pull reuqest gets merged we are in dire need of ssl proxy support in jedis client api.

sunnypav:
Please let us know when jedis considers this request

SenadI:
+1 need this.

svella:
+1. I really need something like this as well, though I would recommend that ability to specify a SocketFactory would be a much better approach because it allows you fine grained control over the trust management process. Managing trust through system properties is problematic when connections to different things have different trust requirements and in things like web applications where you may be sharing the JVM with other applications.

damiri-ts:
Trying to keep this alive with a new pull request (https://github.com/xetorthio/jedis/pull/1173) which builds on the code by adding the ability to specify an SSLSocketFactory, SSLParams, and a HostnameVerifier. Credit for the original code that i built on top of goes to RedisLabs.

I believe the ability to specify these three items ensures that the user can create an ssl connection with whatever hostname or trust verification they see fit. Hopefully, this will spike enough interest from the Jedis folks that they might seriously consider SSL/TLS support.

marcosnils:
@damiri-ts thanks for the contribution. We'll review the PR and continue on #1173 accordingly. I'm closing this for the moment.

PR #647, opened 2014-05-24
Change Old IO (Stream) to NIO (Buffer & Channel)

HeartSaVioR:
Hello!

As programming paradigm has been changing, we should eventually support non-blocking.
I've worked first step of supporting non-blocking - change Stream based IO to NIO.
Although it's still "blocking" and there're no special performance gain, we can change it incrementally.
I'm just beginner of NIO, so we can tune it for more faster.
- replace IO Stream things to NIO utils I have made
  - RedisInputStream -> SocketChannelReader
  - RedisOutputStream -> SocketChannelWriter
- modify relevent classes : Connection, Protocol
- non blocking is not yet present (currently blocking)
- modify Protocol unit test for testing with NIO utils

@xetorthio @marcosnils Please review and comment!
Thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/xetorthio/jedis/647)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,12 +59,6 @@
 			<type>jar</type>
 			<scope>test</scope>
 		</dependency>
-        <dependency>
-            <groupId>org.mockito</groupId>
-            <artifactId>mockito-core</artifactId>
-            <version>1.9.5</version>
-            <scope>test</scope>
-        </dependency>
 		<dependency>
 			<groupId>org.apache.commons</groupId>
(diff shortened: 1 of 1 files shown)


srirama-rayaprolu:
Hi
When this will be available for community?

uglide:
Closing as stale. We do not plan to switch to NIO in Jedis at this point, because it's going to break Keep-Alive support which was introduced in NIO only for Java 11+

PR #671, opened 2014-07-06
Jedis Cluster functionalities improved (cluster-revised branch)

HeartSaVioR:
Since there're some PRs on JedisCluster and it will be merged on 3.0.0, we can incubate cluster-revised branch for improving JedisCluster.

Currently these PRs are merged to cluster-revised branch (updated continuously)

- #617 
  - Remove BasicCommand implementation from JedisCluster
- #618 
  - BinaryJedisCommands, ScriptingCommands, BinaryScriptingCommands


src/main/java/redis/clients/jedis/BinaryJedisCluster.java
```diff
@@ -0,0 +1,1260 @@
...
+  public BinaryJedisCluster(Set<HostAndPort> jedisClusterNode, int timeout, int maxRedirections,
+      final GenericObjectPoolConfig poolConfig) {
+    this.connectionHandler = new JedisSlotBasedConnectionHandler(jedisClusterNode, poolConfig);
+    this.timeout = timeout;
+    this.maxRedirections = maxRedirections;
+  }
+
+  @Override
+  public String set(final byte[] key, final byte[] value) {
+    return new JedisClusterCommand<String>(connectionHandler, timeout, maxRedirections) {
```
nykolaslima: I don't really like these `new JedisClusterCommand` repetition everywhere but I also don't see any alternative to it. Maybe if we could use some lambda here?

src/test/java/redis/clients/jedis/tests/ShardedJedisTest.java
```diff
@@ -20,9 +20,8 @@
   private static HostAndPort redis2 = HostAndPortUtil.getRedisServers().get(1);
 
   /**
-   * Test for "Issue - BinaryShardedJedis.disconnect() may occur memory leak".
```
nykolaslima: Those formatting diffs could conflict with `removing ShardedJedis PR`, but if you don't mind with it, it's ok

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
HeartSaVioR: As you know, it's command pattern.
It's better to represent it with lambda, but it's available only JDK8.
I can't also find any alternatives. Maybe we can implement it with lambda, and convert it to support JDK7. There's a library which can do the trick.

src/test/java/redis/clients/jedis/tests/ShardedJedisTest.java
HeartSaVioR: OK, I can upmerge these branches(PRs) to minimize conflicts.

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
nykolaslima: Do you think it worth the effort?

HeartSaVioR: Library is it's https://github.com/orfjackal/retrolambda

I can't sure it worth the effort. Having JDK 8 as requirement to at compilation level could be disadvantages, too.
(We still need to test Jedis with JDK6 and 7 but we need to build with JDK8. So complicated.)
Maybe we're stuck with Language so it could be natural to just stick it.

nykolaslima: I think we could keep your original implementation instead of using this retrolambda

HeartSaVioR: OK. :)

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
```diff
@@ -0,0 +1,1260 @@
...
+  }
+
+  @Override
+  public byte[] echo(final byte[] arg) {
+    return new JedisClusterCommand<byte[]>(connectionHandler, timeout, maxRedirections) {
+      @Override
+      public byte[] execute(Jedis connection) {
+        return connection.echo(arg);
+      }
+    }.runBinary(null);
```
christophstrobl: Calling `runBinary(null)` leads to `JedisClusterException` due to the null / length check there. 
Would it be possible to dispatch the command to an arbitrary cluster node as a dedicated slot is not required for this operation?

HeartSaVioR: @christophstrobl 
Thanks for finding missed spot! I'll fix it.

Btw, I think echo should belong to BasicCommand, as all commands with no key belong to BasicCommand, and echo doesn't have actual operation about key space.

@xetorthio @marcosnils @nykolaslima What do you think?

HeartSaVioR:
We should upmerge to fix build.

HeartSaVioR:
It's better to remove all deprecated methods and upmerge since it's easier than handling deprecated methods.

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
```diff
@@ -0,0 +1,1261 @@
...
+  @Override
+  public List<byte[]> blpop(final byte[] arg) {
+    return new JedisClusterCommand<List<byte[]>>(connectionHandler, timeout, maxRedirections) {
+      @Override
+      public List<byte[]> execute(Jedis connection) {
+        return connection.blpop(arg);
+      }
+    }.runBinary(arg);
+  }
+
```
christophstrobl: would be nice to have `cluster.blpop(byte[] arg, int timeout)` as well.

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
```diff
@@ -0,0 +1,1261 @@
...
+  @Override
+  public List<byte[]> brpop(final byte[] arg) {
+    return new JedisClusterCommand<List<byte[]>>(connectionHandler, timeout, maxRedirections) {
+      @Override
+      public List<byte[]> execute(Jedis connection) {
+        return connection.brpop(arg);
+      }
+    }.runBinary(arg);
+  }
+
```
christophstrobl: would be nice to have `cluster.brpop(byte[] arg, int timeout)` as well.

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
```diff
@@ -0,0 +1,1261 @@
...
+    return new JedisClusterCommand<byte[]>(connectionHandler, timeout, maxRedirections) {
+      @Override
+      public byte[] execute(Jedis connection) {
+        return connection.echo(arg);
+      }
+    }.runBinary(arg);
+  }
+
+  @Override
+  public Long move(final byte[] key, final int dbIndex) {
```
christophstrobl: `move` is not allowed in cluster mode since there's only db at index `0` available. Leads to `(error) ERR MOVE is not allowed in cluster mode`

christophstrobl:
Hey,
just curious if there are any plans to support cluster commands such as `CLUSTER FORGET` directly on `BinaryJedisCluster`?

What I mean by that would be something like:

```java
public void clusterForget(ClusterNode node) {

  // determine set of nodes and pre remove the one to be forgotten.
  Set<ClusterNode> nodes = readClusterNodesFromDriver();
  nodes.remove(node);
  
  // parallel run command on remaining cluster nodes.
  runCommandAsyncOnNodes(new JedisCommandCallback<Void>() {

    @Override
    public Void doInJedis(Jedis jedis) {

      jedis.clusterForget(node.getId());
      return null;
    }
  }, nodes);
}
```

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
HeartSaVioR: @christophstrobl 
Thanks for reporting!
Btw, maybe we need to define new interfaces for single key operations with cluster, as I'm trying to do at #687.

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
HeartSaVioR: @christophstrobl 
Thanks for reporting!
First of all, brpop(byte[] arg) should be removed, and in reviewing now via #902.
Regarding to brpop(byte[] arg, int timeout), we should add List<byte[]> brpop(byte[] arg, int timeout) to BinaryJedisCommands in order to add method.

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
HeartSaVioR: @christophstrobl 
Thanks for reporting!
First of all, blpop(byte[] arg) should be removed, and in reviewing now via #902.
Regarding to blpop(byte[] arg, int timeout), we should add List<byte[]> blpop(byte[] arg, int timeout) to BinaryJedisCommands in order to add method.

HeartSaVioR:
@christophstrobl 
Currently Jedis is focused to provide APIs which are matched to each Redis commands.
So it's not planned, but we can take it into account.
Pull Request is more welcomed!

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
HeartSaVioR: @christophstrobl 
Oh sorry, I have mistaken. ```cluster.blpop(final int timeout, final byte[]... keys)``` will be introduced by #687.

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
HeartSaVioR: @christophstrobl 
Oh sorry, I have mistaken. ```cluster.brpop(final int timeout, final byte[]... keys)``` will be introduced by #687.

christophstrobl:
@HeartSaVioR sorry for late reply. Thanks for clarification. Can we somehow help moving this PR forward? would be nice to have it in master.

benkris1:
@HeartSaVioR  when do  you want  merge the branch to master

HeartSaVioR:
@christophstrobl @benkris1 
I'm sorry but by the rule it needs to be reviewed by other collaborators.

benkris1:
@HeartSaVioR  hope the progress can be speed up,you know that the redis 3.0 has already released

benkris1:
@marcosnils @Jonathan Leibiusky

marcosnils:
@HeartSaVioR if you can fix the PR so it's mergable and tests pass i'll promise to take a look at it today :)

HeartSaVioR:
So many random tests failure occurred... We should have an issue tracing these and try to fix / discard tests.

HeartSaVioR:
@marcosnils 
I've merged multi key ops PR into this PR, and upmerged.
We still don't have tests, and I wonder we can test commands for normal, pipeline, transaction, cluster, etc without much duplicated code.

HeartSaVioR:
Please note that I spent another over 1 hours to upmerge #687 and this.
I REALLY don't want to try it again.

HeartSaVioR:
@christophstrobl 
I introduced new interfaces for BinaryJedisCluster and JedisCluster.
BinaryJedisCluster and JedisCluster don't have move() now. :)

christophstrobl:
thanks @HeartSaVioR!

marcosnils:
@HeartSaVioR exceptional work, I don't have preliminary observations. This PR really takes us really close to have a 3.0 version!.

HeartSaVioR:
@marcosnils Thanks for reviewing huge PR! Great work!

benkris1:
@marcosnils @HeartSaVioR  @christophstrobl  you guys do a great job.  great programer! Thank all of you!

bpoweski:
Is the intent to remove any common protocol between Jedis/JedisCluster for the 3.0 version?

marcosnils:
@bpoweski what do you mean exactly by "removing common protocols"?

bpoweski:
Apologies, I meant interfaces!

In the 2.7.3 branch `JedisCluster` and `Jedis` have the following signatures.

```java
public class JedisCluster implements JedisCommands, BasicCommands, Closeable 

public class Jedis extends BinaryJedis implements JedisCommands, MultiKeyCommands,
    AdvancedJedisCommands, ScriptingCommands, BasicCommands, ClusterCommands, SentinelCommands 
```

In master they now have:

```java
public class JedisCluster extends BinaryJedisCluster implements JedisClusterCommands,
        MultiKeyJedisClusterCommands, JedisClusterScriptingCommands

public class Jedis extends BinaryJedis implements JedisCommands, MultiKeyCommands,
        AdvancedJedisCommands, ScriptingCommands, BasicCommands, ClusterCommands, SentinelCommands
```

While there are various caveats for using Redis cluster, the underlying API is mostly the same.   Having a common interface between the two clients is useful.  For example, we run our application using both a Redis Cluster and a non clustered Redis configuration.  We currently use `JedisCommands` to abstract most of the differences between the two connections.

HeartSaVioR:
@bpoweski 
There was some comments about removing caveats at compile time, and I think it is still important.
I also think having same interface is awesome for abstraction, but filling method body to UnsupportedOperationException is also bad kind of implementation.

tl;dr. There's pros and cons about each approach, and I just pick first one. It could be changed anytime when there's elegant way to resolve.

Also please note that Jedis 3.0 should be breaking changes, so we're free to REWRITE whole codes though I think it's hard to be occured for now.

bpoweski:
Have you considered rolling up shared signatures into a common interface and then using `extends`.  For example, changing from:

```java
public interface JedisCommands {
  String set(String key, String value);
  ..
}

public interface JedisClusterCommands {
  String set(String key, String value);
  ..
}
```
to:

```java
public interface SingleKeyCommands {
  String set(String key, String value);
}

public interface JedisCommands extends SingleKeyCommands {
  ..
}

public interface JedisClusterCommands extends SingleKeyCommands, OtherSpecificNonClusterAllowedCommands {
  ..
}
```

An added benefit is that it would facilitate decomposing the larger `JedisCommands` and `JedisClusterCommands` interfaces into ones grouped by function.  I suspect you'd reduce the line count by doing so given there is a so much overlap.

HeartSaVioR:
@bpoweski Maintaining OtherSpecificNonClusterAllowedCommands should be annoying.
I'm thinking another approach, breaks interface into data type, though it is just an idea for now.

HeartSaVioR:
@bpoweski Forgot to say "Thanks for giving your opinion!" :)
Please feel free to tell us your idea when there's better approaches.

bpoweski:
@HeartSaVioR, thank you for your work on this project.

lcy362:
when will 3.0.0 release?

PR #687, opened 2014-08-05
Supports Multi Key commands to JedisCluster (revised of #673)

HeartSaVioR:
(It's revised PR of #673.)

Hello.

As supporting Redis Cluster work, I implemented Cluster multi key commands based on previous interfaces.

During implementing, I found that some commands are fine with Redis, but maybe some commands are not fine with Redis Cluster.
(Actually there're no documentation from Redis describing cluster-able commands.)

So I removed some suspicious commands from interface, and introduced new interfaces for Jedis Cluster.
(Please let me know when you think something is wrong.)

Below is the commands I removed from previous interfaces.

BinaryJedisClusterCommands (copied from BinaryCommands)

List<byte[]> blpop(byte[] arg);
List<byte[]> brpop(byte[] arg);
Long move(byte[] key, int dbIndex);
MultiKeyBinaryJedisClusterCommands (copied from MultiKeyBinaryCommands)

List<byte[]> blpop(byte[]... args);
List<byte[]> brpop(byte[]... args);
Set<byte[]> keys(byte[] pattern);
String watch(byte[]... keys);
String unwatch();
byte[] randomBinaryKey();
JedisClusterCommands (copied from JedisCommands)

List<String> blpop(String arg);
List<String> brpop(String arg);
Long move(String key, int dbIndex);
ScanResult<Map.Entry<String, String>> hscan(final String key, int cursor);
ScanResult<String> sscan(final String key, int cursor);
ScanResult<Tuple> zscan(final String key, int cursor);
MultiKeyJedisClusterCommands (copied from MultiKeyCommands)

String watch(String... keys);
String unwatch();
String randomKey();
ScanResult<String> scan(int cursor);
ScanResult<String> scan(final String cursor);
Set<String> keys(String pattern);
List<String> blpop(String... args);
List<String> brpop(String... args);
And I introduce some utils and exception, you can find it from commit log.

Please review and comment so that this PR goes to cluster-revised.
Thanks!


HeartSaVioR:
I'm curious that we can reuse current interfaces, and throw UnsupportedOperationException if method seems not fit to Redis Cluster.
@xetorthio @marcosnils @nykolaslima 
What do you think? Is it acceptable about creating new interfaces for only JedisCluster?

nykolaslima:
I believe that it's better to have new interfaces for `JedisCluster`, but we can reuse some of the existing if it fits.
The reasons why I believe in that are:
1. Maybe we could have methods JedisCluster exclusive. If we share the same interface, we will have to add methods there that will only be used if it's JedisCluster
2. Using the `UnsupportedOperationException` approach, the user could only have those exceptions in QA/Production. IMO, it's better to being able to verify that operation isn't supported in the compilation time, that's one of the main advantages of Java and I wouldn't like to discard it.

HeartSaVioR:
Since we're in progress of reviewing #671, it should be included to cluster-revised branch.

HeartSaVioR:
I'll merge this to cluster-revised branch to prepare reviewing.

hegc:
@rupatel ,for example this{foo}.bar1 and another{foo}.bar2 are guaranteed to be in the same hash slot, and I want get all keys like '{foo}.xxx', but there is no jedisCluster function like 'keys', so you have idea or some jedis solutions now ? Because of your Q&A above, could you give me some suggestion?

rupatel:
Hey, We have to wait until they implement and place this feature in new release. Other option is to implement feature, I tried by using one of the api they had exposed to fetch the slot in which the current pattern is stored, but it is not a good design to iterate  all nodes in  cluster and fetch the keys since we need to handle failover and other cases.

So I have not yet cracked the problem, if you find any solution please let me know.

PR #713, opened 2014-08-26
Support non-blocking (Async) APIs

HeartSaVioR:
On #647, we switched I/O from Stream based Old I/O to New I/O.
Based on New I/O, finally I can introduce non-blocking APIs to Jedis!

I'm not familiar with Java New I/O, and our clients may have Java 6, so there're some restrictions on implementation.
I'm trying to implement it simply, so I try to avoid multi-thread things as many as possible.
(It may lose some performances.)
1 dispatcher handles 1 connection's request/response with only one thread.

Current APIs doesn't support Sharded, Redis Cluster, Sentinel, scripting, transaction, pipeline.
You can't select DB index, and you should present AUTH password with AsyncJedis constructor.

I wish NIO experts would improve my base codes to gain high performance.

ps. It's upmerged.
ps2. Failing test is not related to async, it's related to https://github.com/antirez/redis/issues/1561.
We should fix it by create another issue.

---
This change is [Reviewable](https://reviewable.io/reviews/xetorthio/jedis/713)

### DIFF
--- a/src/main/java/redis/clients/jedis/BuilderFactory.java
+++ b/src/main/java/redis/clients/jedis/BuilderFactory.java
@@ -53,7 +53,7 @@ public class BuilderFactory {
     }
   };
 
-    public static final Builder<byte[]> BYTE_ARRAY = new Builder<byte[]>() {
+  public static final Builder<byte[]> BYTE_ARRAY = new Builder<byte[]>() {
     public byte[] build(Object data) {
       return ((byte[]) data); // deleted == 1
     }
--- a/src/main/java/redis/clients/jedis/async/AsyncBinaryJedis.java
+++ b/src/main/java/redis/clients/jedis/async/AsyncBinaryJedis.java
@@ -19,8 +19,8 @@ import static redis.clients.jedis.Protocol.Command.*;
 import static redis.clients.jedis.Protocol.toByteArray;
 
 public class AsyncBinaryJedis implements AsyncBasicCommands, AsyncBinaryJedisCommands,
-    AsyncMultiKeyBinaryCommands, AsyncAdvancedBinaryJedisCommands,
-    AsyncBinaryScriptingCommands, Closeable {
+    AsyncMultiKeyBinaryCommands, AsyncAdvancedBinaryJedisCommands, AsyncBinaryScriptingCommands,
+    Closeable {
   public static final int BUFFER_SIZE = 8192;
 
   protected final AsyncDispatcher processor;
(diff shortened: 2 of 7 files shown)


HeartSaVioR:
Upmerged to current master.

I revised current benchmark and ran to compare Normal vs Async vs Pipeline.
Please refer https://groups.google.com/d/msg/jedis_redis/7ZJPFp_qTY0/VVY3p1D5EzcJ for details.

HeartSaVioR:
Nils Kilden-Pedersen pointed out from Google Groups.

- Method AsyncDispatcher.registerRequest is synchronized. Since it only calls LinkedBlockingDeque.add and Selector.wakeup that should be unnecessary, as both are safe to call concurrently, and I don’t believe those two method calls need to be atomic.
- Same class, method handleConnectionException, while loop does a peek then a poll, which is redundant, since poll returns null if empty. Probably not super important since it’s during an exceptional condition, but still…
- Java7 introduced ConcurrentLinkedDeque, which is a non-blocking deque implementation. This makes it better than LinkedBlockingDeque under load, but is supposedly slightly less performant in the non-contended case. I think for this scenario, ConcurrentLinkedDeque offers better scalability.

I'll try to adopt these one by one and check it would help.

HeartSaVioR:
We're talking about changing our async implementation to Netty, but it's not confirmed so I'll try both version of async, NIO and Netty.

HeartSaVioR:
Compared to using Netty, lack of ReplayingDecoder we need to handle response by struggling with byte.
When we decide to stick NIO, we need to implement custom ReplayingDecoder and let ResponseHandler simplified.

darkcrawler01:
@HeartSaVioR :+1: 
Any ETA on this pull request getting merged to master ?

src/main/java/redis/clients/util/SocketChannelReader.java
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        sb.append((char) curr);
+        sb.append((char) next);
+      } else {
+        sb.append((char) curr);
+      }
+    }
+
+    String reply = sb.toString();
```
Alwayswithme: what about
```java
String reply = new String(this.readLineBytes());
```

jeremyblog:
why not yeah merge，i expect it long time

AaronHarris:
Is this still a priority to add Async apis to Jedis? Currently looking a Lettuce for this reason alone, although I prefer the lightweight Jedis.

moadi:
@HeartSaVioR sorry to bother you, do you happen to know if there are still any plans to support async features in jedis?

Htermotto:
bump! Any plans for this to get merged?

jiangxiaoqiang:
does jedis support reactive right now?

sazzad16:
> does jedis support reactive right now?

@jiangxiaoqiang No.

marcosnils:
If you're looking for an async client you can checkout lettuce (
https://lettuce.io/)

uglide:
As pointed out before, there are no plans to add async/reactive support to Jedis at this point. If you need async or reactive client please use https://github.com/redis/lettuce

PR #749, opened 2014-09-25
Alternative implementation of Jedis Cluster commands using Java Proxy

causton81:
I wrote a small app to set N string keys to a constant value using the existing JedisCluster and my new ClusterInvocationHandler.  This was very limited testing, but with 10,000 and 50,000 keys the new implementation performed as well (+/- 100ms) as the current implementation; sometimes a bit faster, sometimes a bit slower.


zts1993:
wrong implementation at processing JedisAskDataException .

sazzad16:
@HeartSaVioR @marcosnils Close?

Issue #726, opened 2014-09-02
JedisSentinelPool will keep pinging the redis-sentinel which blocks my service.


airpingu:
Supposing my redis-sentinel is not running, when I call:

```
HashSet<String> sentinels = new HashSet<String>();
sentinels.add("127.0.0.1:1234");
JedisSentinelPool jedisPool = new JedisSentinelPool("mymaster", sentinels);
```

It will endlessly output the following log:

Sep 02, 2014 8:18:20 PM redis.clients.jedis.JedisSentinelPool initSentinels
INFO: Trying to find master from available Sentinels...
Sep 02, 2014 8:18:20 PM redis.clients.jedis.JedisSentinelPool initSentinels
WARNING: Cannot connect to sentinel running @ 127.0.0.1:1234. Trying next one.
Sep 02, 2014 8:18:20 PM redis.clients.jedis.JedisSentinelPool initSentinels
SEVERE: All sentinels down, cannot determine where is mymaster master is running... sleeping 1000ms.
Sep 02, 2014 8:18:21 PM redis.clients.jedis.JedisSentinelPool initSentinels
INFO: Trying to find master from available Sentinels...
Sep 02, 2014 8:18:21 PM redis.clients.jedis.JedisSentinelPool initSentinels
WARNING: Cannot connect to sentinel running @ 127.0.0.1:1234. Trying next one.
Sep 02, 2014 8:18:21 PM redis.clients.jedis.JedisSentinelPool initSentinels
SEVERE: All sentinels down, cannot determine where is mymaster master is running... sleeping 1000ms.
Sep 02, 2014 8:18:22 PM redis.clients.jedis.JedisSentinelPool initSentinels
INFO: Trying to find master from available Sentinels...
Sep 02, 2014 8:18:22 PM redis.clients.jedis.JedisSentinelPool initSentinels
WARNING: Cannot connect to sentinel running @ 127.0.0.1:1234. Trying next one.
Sep 02, 2014 8:18:22 PM redis.clients.jedis.JedisSentinelPool initSentinels
SEVERE: All sentinels down, cannot determine where is mymaster master is running... sleeping 1000ms.
... (the following is the same)

I'm very confused by this phenomenon because I cannot catch any exceptions to do fallback when the redis-sentinel goes down or is not available. It just keep pinging the sentinel which blocks my Web service.

Does anyone know how to handle this case?

Cheers,
Gene

fulnLee:
[...]

PR #752, opened 2014-09-27
Fail fast while initializing JedisSentinelPool

HeartSaVioR:
It's related to #726.

I modified JedisSentinelPool's behavior to fail fast on
- wrong master name is specified
- all sentinels are down or not reached

Actually I also think it should fail fast and let users know about it instead of keep retrying.

Please review and comment! Thanks!

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisSentinelPool.java
+++ b/src/main/java/redis/clients/jedis/JedisSentinelPool.java
@@ -127,23 +127,21 @@ public class JedisSentinelPool extends Pool<Jedis> {
 	    try {
 		jedis = new Jedis(hap.getHost(), hap.getPort());
 
-		if (master == null) {
-		    List<String> masterAddr = jedis
-			    .sentinelGetMasterAddrByName(masterName);
+		List<String> masterAddr = jedis
+			.sentinelGetMasterAddrByName(masterName);
 
-		    // connected to sentinel...
-		    sentinelAvailable = true;
+		// connected to sentinel...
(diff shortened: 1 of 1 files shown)


src/main/java/redis/clients/jedis/JedisSentinelPool.java
```diff
@@ -112,52 +113,53 @@ private HostAndPort initSentinels(Set<String> sentinels,
...
-		e.printStackTrace();
+	if (master == null) {
+	    if (sentinelAvailable) {
+		// can connect to sentinel, but master name seems to not monitored
+		throw new JedisException("Can connect to sentinel, but " + masterName
+			+ " seems to be not monitored...");
+	    } else {
+		throw new JedisConnectionException("All sentinels down, cannot determine where is "
+			+ masterName + " master is running...");
 	    }
```
namangala: How can I possibly catch this exception if I do not want my application to fail on starup ? Can you please guide ? I am using spring boot.

marcosnils: @namangala I think it's best if you ask this question in the Jedis mailing list.

namangala: @marcosnils : Can you please guide me to the right place where I can post an issue or file a ticket please ?

marcosnils: http://groups.google.com/group/jedis_redis

PR #782, opened 2014-11-03
MultiKeyPipelineBase and PipelineBase made public

thenobody:
MultiKeyPipelineBase and PipelineBase made public in order to make Pipeline and transaction mockable with Mockito


nykolaslima:
LGTM, I think we can merge it.

marcosnils:
Hmmm I'm not entirely sure about this PR. We should find a generic solution for this type of situations because otherwise we'll have to make all relevant classes public so they can be mocked by mockito and it just doesn't feel right. 

It's been a while since this PR. @ZephroC, how did you move on along the problem?

thenobody:
@marcosnils as far as I can tell apart from the these two classes (and a couple of private inner classes, which are in turn enclosed in explicitly public classes) there's only JedisFactory which is restricted to its package. What's the reasoning behind keeping these classes non-public? Thanks

marcosnils:
@thenobody I guess the idea to make the protected is that nobody outside Jedis extends or use those closes to do something as by themselves they're pretty much useless. 
 
You're right that no other classes are protected so I guess we can go ahead and merge this PR.

HeartSaVioR:
Since it doesn't break binary compatibility, we can apply it to 2.7.1.

https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html
```
13.4.3. public Classes

Changing a class that is not declared public to be declared public does not break compatibility with pre-existing binaries.

If a class that was declared public is changed to not be declared public, then an IllegalAccessError is thrown if a pre-existing binary is linked that needs but no longer has access to the class type; such a change is not recommended for widely distributed classes.
```

HeartSaVioR:
Merged into master and 2.7 branch respectively.
@thenobody Thanks!

thenobody:
Thank you guys, appreciated

PR #835, opened 2015-01-02
adding support to choose databases when using ShardedJedis

### LINKED ISSUE #820: There is no way to select database in ShardedJedis
There is no way to select database with `ShardedJedis`?
```
//With Jedis class, it works
jedis.select(0);
//With ShardedJedis, there is no method select
shardedJedis.select(0);
```

nykolaslima:
Closes #820

Also changing `JedisURIHelper``getDBIndex(uri)` to return an `int` instead of `Integer`. This way we save some bytes and we don't need it to be an `Integer`, since if there is no `db` in the given `URI` we return the default DB: `0`.
Also removing unnecessary null checks in `getDBIndex(uri)` calls.

@marcosnils @HeartSaVioR

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisShardInfo.java
+++ b/src/main/java/redis/clients/jedis/JedisShardInfo.java
@@ -4,7 +4,7 @@ import java.net.URI;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-import redis.clients.jedis.exceptions.JedisConnectionException;
+import redis.clients.jedis.exceptions.JedisException;
 import redis.clients.util.JedisURIHelper;
 import redis.clients.util.ShardInfo;
 import redis.clients.util.Sharded;
@@ -112,7 +112,7 @@ public class JedisShardInfo extends ShardInfo<Jedis> {
     Jedis jedis = new Jedis(this);
     try {
       jedis.select(db);
-    } catch(JedisConnectionException e) {
+    } catch(JedisException e) {
       log.log(Level.SEVERE, "Can't select database due a Redis connection problem", e);
     }
       
--- a/src/main/java/redis/clients/jedis/JedisShardInfo.java
+++ b/src/main/java/redis/clients/jedis/JedisShardInfo.java
@@ -2,10 +2,12 @@ package redis.clients.jedis;
 
 import java.net.URI;
 
+import redis.clients.util.JedisURIHelper;
 import redis.clients.util.ShardInfo;
 import redis.clients.util.Sharded;
 
 public class JedisShardInfo extends ShardInfo<Jedis> {
+  
   public String toString() {
     return host + ":" + port + "*" + getWeight();
   }
@@ -15,6 +17,8 @@ public class JedisShardInfo extends ShardInfo<Jedis> {
   private int port;
   private String password = null;
   private String name = null;
+  //Default Redis DB
+  private int db = 0;
 
   public String getHost() {
     return host;
@@ -30,7 +34,8 @@ public class JedisShardInfo extends ShardInfo<Jedis> {
     if (uri.getScheme() != null && uri.getScheme().equals("redis")) {
       this.host = uri.getHost();
       this.port = uri.getPort();
-      this.password = uri.getUserInfo().split(":", 2)[1];
+      this.password = JedisURIHelper.getPassword(uri);
+      this.db = JedisURIHelper.getDBIndex(uri);
     } else {
       this.host = host;
       this.port = Protocol.DEFAULT_PORT;
@@ -69,7 +74,8 @@ public class JedisShardInfo extends ShardInfo<Jedis> {
     super(Sharded.DEFAULT_WEIGHT);
     this.host = uri.getHost();
     this.port = uri.getPort();
-    this.password = uri.getUserInfo().split(":", 2)[1];
+    this.password = JedisURIHelper.getPassword(uri);
+    this.db = JedisURIHelper.getDBIndex(uri);
   }
 
   public String getPassword() {
@@ -91,9 +97,16 @@ public class JedisShardInfo extends ShardInfo<Jedis> {
   public String getName() {
     return name;
   }
+  
+  public int getDb() {
+    return db;
+  }
 
   @Override
   public Jedis createResource() {
-    return new Jedis(this);
+    Jedis jedis = new Jedis(this);
+    jedis.select(db);
+    return jedis;
   }
+
 }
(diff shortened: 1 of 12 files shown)


marcosnils:
@nykolaslima did you run the tests before sending the PR?. Travis is complaining about a broken build.

nykolaslima:
@marcosnils I can't run the test environment here in my machine, can you help me?
```
$ make start
rm -vf /tmp/redis_cluster_node*.conf 2>/dev/null
rm dump.rdb appendonly.aof - 2>/dev/null
make: [cleanup] Error 1 (ignored)
echo "$REDIS1_CONF" | redis-server -
echo "$REDIS2_CONF" | redis-server -
echo "$REDIS3_CONF" | redis-server -
echo "$REDIS4_CONF" | redis-server -
echo "$REDIS5_CONF" | redis-server -
echo "$REDIS6_CONF" | redis-server -
echo "$REDIS7_CONF" | redis-server -
echo "$REDIS_SENTINEL1" > /tmp/sentinel1.conf && redis-server /tmp/sentinel1.conf --sentinel
echo "$REDIS_SENTINEL2" > /tmp/sentinel2.conf && redis-server /tmp/sentinel2.conf --sentinel
echo "$REDIS_SENTINEL3" > /tmp/sentinel3.conf && redis-server /tmp/sentinel3.conf --sentinel
echo "$REDIS_SENTINEL4" > /tmp/sentinel4.conf && redis-server /tmp/sentinel4.conf --sentinel
echo "$REDIS_CLUSTER_NODE1_CONF" | redis-server -

*** FATAL CONFIG FILE ERROR ***
Reading the configuration file, at line 3
>>> 'cluster-node-timeout 50'
Bad directive or wrong number of arguments
make: *** [start] Error 1
```

marcosnils:
@nykolaslima Jedis master branch is tested against the last unstable redis version. You need to checkout redis code from github and do a `make && make install` so you can get the latest jedis version running

nykolaslima:
@marcosnils looking at the tests who failed, looks like the test is wrong.

The Redis server configured in `Makefile` at port `6379` is configured with password `foobared`.
But in test `testMD5Sharding`, the connection is configured without password: `shards.add(new JedisShardInfo("localhost", Protocol.DEFAULT_PORT));`.

That's why the test is failing. What do you think?

marcosnils:
@nykolaslima the thing is that several tests failed, not just one. I believe that maybe the change that you make that automatically selects the database when trying to establish the connection is not taking into account that redis may be password protected. This is something important that we should validate beforehand.

nykolaslima:
@marcosnils the problem is that with my changes, we attempt to hit the database when we are creating the Jedis resource, because we are selecting the database.

Before my changes, Jedis only hit the database when executing one command, like a `get`.

I think we have two solutions:
1. fix the tests. If we configure the password in the broken tests they will pass
2. change the select database method, making it kind of `lazy` to just select the database before executing the first command.

What do you think?

nykolaslima:
Or we can just check if it's possible to connect to the database before selecting it and if it isn't, we can just don't select the database.

nykolaslima:
@marcosnils now the build is passing. What do you think about this solution?

marcosnils:
@nykolaslima let me check the complete PR and i'll get back to you later

nykolaslima:
Ok. Thank you very much for your help!

src/main/java/redis/clients/jedis/JedisShardInfo.java
```diff
@@ -91,9 +102,21 @@ public void setTimeout(int timeout) {
...
   }
+  
+  public int getDb() {
+    return db;
+  }
 
   @Override
   public Jedis createResource() {
-    return new Jedis(this);
+    Jedis jedis = new Jedis(this);
```
HeartSaVioR: @nykolaslima 
How about handling selecting db by constructor of  BinaryJedis, as authentication takes place?

HeartSaVioR: @nykolaslima We can change BinaryClient to add setDBIndex() so initial DB index can be set, as password did.

nykolaslima: I've thought that. But this is a 'big' change. If you agree with that I can do this.

HeartSaVioR: @marcosnils How do you think about it? It seems to be a hard to maintain if place of setting password and db index differs.

HeartSaVioR: @nykolaslima We can try it with @marcosnils review. 
Keeping changeset small is important, but I think not breaking convention/consistency is more important.

nykolaslima: I agree. With this solution we can keep it simple but it looks like a 'workaround'

nykolaslima: @HeartSaVioR actually we will need to create a `setDb` in `BinaryClient`. This way we can configure the db and let the `connect` method use it.
I'm thinking about changing `BinaryJedis` constructor to receive the `db` and then call `setDb`. 
It will look like this:
```
public class BinaryJedis {
 public BinaryJedis(final JedisShardInfo shardInfo, int db) {
    client = new Client(shardInfo.getHost(), shardInfo.getPort());
    client.setTimeout(shardInfo.getTimeout());
    client.setPassword(shardInfo.getPassword());
    client.setDb(db);
  }
}
```

What do you think about it @HeartSaVioR?

marcosnils: I agree with @HeartSaVioR solution. It's more elegant to set the DB at BinaryJedis level where initialization takes place rather to make it elsewhere. @nykolaslima please go ahead and make the changes and let us know once we can review.

nykolaslima: Ok @marcosnils. Did you see the comment above? Can I implement this way?

marcosnils: @nykolaslima. Yes, go ahead and create the setter for the db field in BinaryClient.

nykolaslima: Ok! I'll do this.

marcosnils:
@nykolaslima @HeartSaVioR changes look good. Please go ahead and merge.

HeartSaVioR:
@nykolaslima Thanks, I've merged it into master, 2.6, 2.7 branches.

nykolaslima:
Thank you guys.

codeHui:
I wonder if this function is added in jedis? in what version,thanks!

codeHui:
and if this function(changed the source code)  may have a bad effect if I use Redis Sentinel also

PR #838, opened 2015-01-06
add getters for pool stats

2bitoperations:
move the getter for numActive to the Pool<...> base class, and add a few other pool statistics getters.

we'd love to have these exposed to allow us to have metrics on the pool, and alert if we detect that we're leaking connections


HeartSaVioR:
@2bitoperations Thanks for your effort. Merged your PR to master, 2.6, 2.7 branches.

PR #842, opened 2015-01-07
adding so timeout configuration

### LINKED ISSUE #799: Allow setting the connection timeout and so-timeout.
The Jedis constructor allows users to supply a timeout value.  The supplied value is used for the connection timeout (socket.connection()) as well as the read timeout (socket.setSoTimeout()) w/in
redis.clients.jedis.Connection.connect().

Adding the property with the same default would maintain backward compatibility for those use cases where users do not supply a value.  For those scenarios where users constructed a Jedis object w/ a timeout both setters could be called:
```
    public BinaryJedis(final String host, final int port, final int timeout) {
	client = new Client(host, port);
	client.setConnTimeout(timeout);
	client.setSoTimeout(timeout);
    }
```

A new constructor could then be added for Jedis/BinaryJedis:
```
    public BinaryJedis(final String host, final int port, final int connTimeout, final int soTimeout) {
	client = new Client(host, port);
	client.setConnTimeout(connTimeout);
	client.setSoTimeout(soTimeout);
    }
```
Where the following property refactors (with associated accessors) are made w/in redis.clients.jedis.Connection:
```
-     private int timeout = Protocol.DEFAULT_TIMEOUT;
+     private int connTimeout = Protocol.DEFAULT_TIMEOUT;
+     private int soTimeout = Protocol.DEFAULT_TIMEOUT;
```
... then the connect() method would be changed to take advantage of these properties separately:
```
                socket.connect(new InetSocketAddress(host, port), connTimeout);
                socket.setSoTimeout(soTimeout);
```

nykolaslima:
closes #799

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryJedis.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedis.java
@@ -45,12 +45,20 @@ public class BinaryJedis implements BasicCommands, BinaryJedisCommands, MultiKey
 
   public BinaryJedis(final String host, final int port, final int timeout) {
     client = new Client(host, port);
-    client.setTimeout(timeout);
+    client.setConnectionTimeout(timeout);
+    client.setSoTimeout(timeout);
+  }
+  
+  public BinaryJedis(final String host, final int port, final int connectionTimeout, final int soTimeout) {
+    client = new Client(host, port);
+    client.setConnectionTimeout(connectionTimeout);
+    client.setSoTimeout(soTimeout);
   }
 
   public BinaryJedis(final JedisShardInfo shardInfo) {
     client = new Client(shardInfo.getHost(), shardInfo.getPort());
-    client.setTimeout(shardInfo.getTimeout());
+    client.setConnectionTimeout(shardInfo.getConnectionTimeout());
+    client.setSoTimeout(shardInfo.getSoTimeout());
     client.setPassword(shardInfo.getPassword());
     client.setDb(shardInfo.getDb());
   }
@@ -61,7 +69,14 @@ public class BinaryJedis implements BasicCommands, BinaryJedisCommands, MultiKey
 
   public BinaryJedis(final URI uri, final int timeout) {
     initializeClientFromURI(uri);
-    client.setTimeout(timeout);
+    client.setConnectionTimeout(timeout);
+    client.setSoTimeout(timeout);
+  }
+  
+  public BinaryJedis(final URI uri, final int connectionTimeout, final int soTimeout) {
+    initializeClientFromURI(uri);
+    client.setConnectionTimeout(connectionTimeout);
+    client.setSoTimeout(soTimeout);
   }
 
   private void initializeClientFromURI(URI uri) {
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -1,12 +1,5 @@
 package redis.clients.jedis;
 
-import redis.clients.jedis.Protocol.Command;
-import redis.clients.jedis.exceptions.JedisConnectionException;
-import redis.clients.jedis.exceptions.JedisDataException;
-import redis.clients.util.RedisInputStream;
-import redis.clients.util.RedisOutputStream;
-import redis.clients.util.SafeEncoder;
-
 import java.io.Closeable;
 import java.io.IOException;
 import java.net.InetSocketAddress;
@@ -15,6 +8,13 @@ import java.net.SocketException;
 import java.util.ArrayList;
 import java.util.List;
 
+import redis.clients.jedis.Protocol.Command;
+import redis.clients.jedis.exceptions.JedisConnectionException;
+import redis.clients.jedis.exceptions.JedisDataException;
+import redis.clients.util.RedisInputStream;
+import redis.clients.util.RedisOutputStream;
+import redis.clients.util.SafeEncoder;
+
 public class Connection implements Closeable {
 
   private String host = Protocol.DEFAULT_HOST;
@@ -22,7 +22,8 @@ public class Connection implements Closeable {
   private Socket socket;
   private RedisOutputStream outputStream;
   private RedisInputStream inputStream;
-  private int timeout = Protocol.DEFAULT_TIMEOUT;
+  private int connectionTimeout = Protocol.DEFAULT_TIMEOUT;
+  private int soTimeout = Protocol.DEFAULT_TIMEOUT;
   private boolean broken = false;
 
   public Connection() {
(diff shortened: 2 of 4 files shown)


marcosnils:
@nykolaslima the change looks good. As it introduces a new functionality it should be scheduled for the next MINOR release (2.7.x). I'm wondering if we should write a test which actually verifies that timeouts are being set at the connect and read level.

@HeartSaVioR @xetorthio I'd like your thoughts about this last part..

nykolaslima:
@marcosnils How can we test it? Getting `Connection`'s `socket` and checking the configured `soTimeout`?

HeartSaVioR:
@marcosnils @nykolaslima @xetorthio 
I don't know how we can determine timeout to conn timeout or read timeout.
My sketched idea is making two unit tests which one is too short conn timeout and long enough read timeout, and another one is long enough conn timeout and too short read timeout.
If it would be root of random failure, we should remove those tests.

Or we can just check two variables are properly set.

How do you think?

HeartSaVioR:
Anyway, changeset looks good to me.

nykolaslima:
I don't think we need to check if values are set. This way we are only
testing the constructor and we do not do this.

HeartSaVioR:
I'll try and report it. :)

HeartSaVioR:
@nykolaslima It would be great to add Jedis constructor that handles connection timeout and socket timeout. (Same to BinaryJedis's new constructor) 
Normal case users are encouraged to use Jedis, not BinaryJedis directly.

HeartSaVioR:
On related tests, I realized it would become tests for testing which seems not good.
So when @nykolaslima applies my latest comment, I'll merge this.

nykolaslima:
@HeartSaVioR done!

HeartSaVioR:
Merged into master, 2.7 branches.
@nykolaslima Thanks!

BlackRider97:
Hi,
I write a very basic test to verify soTimeout with `jedis-2.7.3.jar` and found that it is not working.
To simulate long time taking query in redis, I wrote a simple LUA script which simply keeps server busy for given amount of time.
 
I am also attching my whole java program here
```
import java.util.ArrayList;
import java.util.List;

public class TestRedisTimeout
{
    private static final String luaScript = "local aTempKey = \"a-temp-key\" "+" "+
                                            "local cycles" +" " +
                                            "redis.call(\"SET\",aTempKey,\"1\")"+" "+
                                            "redis.call(\"PEXPIRE\",aTempKey, ARGV[1])"+" "+
                                            "for i = 0, 999999999, 1 do"+" "+
                                            "    local apttl = redis.call(\"PTTL\",aTempKey)"+" "+
                                            "    cycles = i;"+" "+
                                            "    if apttl <= 0 then"+" "+
                                            "        break;"+" "+
                                            "    end"+" "+
                                            "end"+" "+
                                            "return cycles"+" ";
   
    private static long runLuaScriptForGivenTimeout(Jedis redis, int timeOutInMillisecons) throws Exception
    {
        List<String> keys = new ArrayList<String>(1);
        keys.add("");
        List<String> args = new ArrayList<String>(4);
        args.add(timeOutInMillisecons+"");

        try
        {
            Object returnValue = redis.eval(luaScript, keys, args);
            return (long)returnValue;
        }
        catch (JedisConnectionException e) 
        {
            throw new Exception("Error in eval - " + e.getMessage(), e);
        }
        catch (Exception e)
        {
            throw new Exception("Error in eval - " + e.getMessage(), e);
        }
        finally
        {
            if(redis != null)
            {
                redis.close();
            }
        }
    }
    
    public static void main(String[] args)
    {
        int connectionTimeout = 100; // 100 milliseconds
        int soTimeout = 10; // 10 milliseconds
        Jedis j = new Jedis("127.0.0.1", 6379, connectionTimeout, soTimeout);
        long start = System.currentTimeMillis();
        try
        {
            long cycles = runLuaScriptForGivenTimeout(j, 2000);
            System.out.println("total cycles="+cycles);
        }
        catch (Exception e)
        {
            e.printStackTrace();
            System.out.println("Caught exception:"+e.getMessage());
        }
        long timetaken = System.currentTimeMillis() - start;
        System.out.println("time taken (in milli seconds)="+timetaken); 
    }
}
```
**Expected output**
```
Caught exception: redis.clients.jedis.exceptions.JedisConnectionException: java.net.SocketTimeoutException: Read timed out
```

**Actual output**
```
total cycles=667500
time taken (in milli seconds)=2018
```

Am I doing something wrong here ?

HeartSaVioR:
@BlackRider97 
Jedis changes timeout to infinite during executing lua script in case of long running script. So you may need to find alternative way.

BlackRider97:
@HeartSaVioR 
Thanks for pointing out issue. I fixed my test case after this so now I am using two connection - one for keeping server busy with long running lua script and another one to test read timeout
```
import java.util.ArrayList;
import java.util.List;
import java.util.concurrent.Executor;

import com.bsb.hike.util.concurrent.HikeThreadPoolExecutor;

public class TestRedisTimeout
{
    private static final String luaScript = "local aTempKey = \"a-temp-key\" "+" "+
            "local cycles" +" " +
            "redis.call(\"SET\",aTempKey,\"1\")"+" "+
            "redis.call(\"PEXPIRE\",aTempKey, ARGV[1])"+" "+
            "for i = 0, 999999999, 1 do"+" "+
            "    local apttl = redis.call(\"PTTL\",aTempKey)"+" "+
            "    cycles = i;"+" "+
            "    if apttl <= 0 then"+" "+
            "        break;"+" "+
            "    end"+" "+
            "end"+" "+
            "return cycles"+" ";

    private static long runLuaScriptForGivenTimeout(Jedis redis, int timeOutInMillisecons) throws Exception
    {
        List<String> keys = new ArrayList<String>(1);
        keys.add("");
        List<String> args = new ArrayList<String>(4);
        args.add(timeOutInMillisecons+"");

        try
        {
            Object returnValue = redis.eval(luaScript, keys, args);
            return (long)returnValue;
        }
        catch (JedisConnectionException e) 
        {
            throw new Exception("Error in eval - " + e.getMessage(), e);
        }
        catch (Exception e)
        {
            throw new Exception("Error in eval - " + e.getMessage(), e);
        }
    }

    public static void main(String[] args)
    {
        int connectionTimeout = 1000; // 100 milliseconds
        int soTimeout = 50; // 50 milliseconds
        final Jedis j1 = new Jedis("127.0.0.1", 6379, connectionTimeout, soTimeout);
        final Jedis j2 = new Jedis("127.0.0.1", 6379, connectionTimeout, soTimeout);
        Executor executor = new HikeThreadPoolExecutor(1);
        // keep redis server busy for certain time but not more than 5 seconds as server will kill long running LUA script 
        executor.execute(new Runnable()
        { 
            @Override
            public void run()
            {
                try
                {
                    runLuaScriptForGivenTimeout(j1, 2000);
                }
                catch (Exception e)
                {
                }
            }
        });
        // wait for sometime to start running LUA script
        try
        {
            Thread.sleep(100);
        }
        catch (InterruptedException e1)
        {
        }

        long start = System.currentTimeMillis();
        try
        {
            j2.get("key");
        }
        catch (Exception e)
        {
            System.out.println("Caught exception:"+e.getMessage());
        }
        long timetaken = System.currentTimeMillis() - start;
        System.out.println("time taken (in milli seconds)="+timetaken); 
        if(j1 != null)
        {
            j1.close();
        }
        if(j2 != null)
        {
            j2.close();
        }
    }
}
```
** Actual output ** ( Expected output)
```
Caught exception:java.net.SocketTimeoutException: Read timed out
time taken (in milli seconds)=53
```

PR #843, opened 2015-01-07
first try to select the database. Then, if it was successfully, update the db index

### LINKED ISSUE #723: select execute twice
## Case 1

Code goes like this:

``` java

class Demo {
		Jedis j = new Jedis("localhost");
		j.select(1);
		j.keys("*");
	}
}
```

Monitor Redis and you will get this:

```
1409563316.698655 [0 127.0.0.1:63894] "SELECT" "1"
1409563316.699014 [1 127.0.0.1:63894] "SELECT" "1"
1409563316.709793 [1 127.0.0.1:63894] "KEYS" "*"
```

## Case 2

Just modify the code:

``` java

class Demo {
		Jedis j = new Jedis("localhost");
		j.keys("*");
		j.select(1);
	}
}

```

Output changes too:

```
1409563394.991312 [0 127.0.0.1:63912] "KEYS" "*"
1409563394.991910 [0 127.0.0.1:63912] "SELECT" "1"
```

## Confused

In case 1, `select` executed twice. Why?

Hope to get some help. :beers:

nykolaslima:
closes #723

> Only one we should take care of is, what state is proper when sendCommand(SELECT) fail.
It shouldn't save db index? Or it should save db index?

If there was a problem when trying to select the database index, then we should not update `db` variable value.
This also avoid Jedis from executing the `select` command twice if the connection was not opened.

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryClient.java
+++ b/src/main/java/redis/clients/jedis/BinaryClient.java
@@ -163,7 +163,6 @@ public class BinaryClient extends Connection {
   }
 
   public void select(final int index) {
-    db = index;
     sendCommand(SELECT, toByteArray(index));
   }
 
--- a/src/main/java/redis/clients/jedis/BinaryJedis.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedis.java
@@ -77,6 +77,7 @@ public class BinaryJedis implements BasicCommands, BinaryJedisCommands, MultiKey
     if (dbIndex > 0) {
       client.select(dbIndex);
       client.getStatusCodeReply();
+      client.setDb(dbIndex);
     }
   }
 
@@ -366,7 +367,10 @@ public class BinaryJedis implements BasicCommands, BinaryJedisCommands, MultiKey
   public String select(final int index) {
     checkIsInMulti();
     client.select(index);
-    return client.getStatusCodeReply();
+    String statusCodeReply = client.getStatusCodeReply();
+    client.setDb(index);
+    
+    return statusCodeReply;
   }
 
   /**
--- a/src/main/java/redis/clients/jedis/MultiKeyPipelineBase.java
+++ b/src/main/java/redis/clients/jedis/MultiKeyPipelineBase.java
@@ -385,7 +385,10 @@ abstract class MultiKeyPipelineBase extends PipelineBase implements
 
 	public Response<String> select(int index) {
 		client.select(index);
-		return getResponse(BuilderFactory.STRING);
+		Response<String> response = getResponse(BuilderFactory.STRING);
+		client.setDb(index);
+		
+		return response;
 	}
 
 	public Response<Long> bitop(BitOP op, byte[] destKey, byte[]... srcKeys) {
--- a/src/test/java/redis/clients/jedis/tests/JedisTest.java
+++ b/src/test/java/redis/clients/jedis/tests/JedisTest.java
@@ -14,6 +14,7 @@ import redis.clients.jedis.JedisShardInfo;
 import redis.clients.jedis.Protocol;
 import redis.clients.jedis.exceptions.JedisConnectionException;
 import redis.clients.jedis.exceptions.JedisDataException;
+import redis.clients.jedis.exceptions.JedisException;
 import redis.clients.jedis.tests.commands.JedisCommandTestBase;
 import redis.clients.util.SafeEncoder;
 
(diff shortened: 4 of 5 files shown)


HeartSaVioR:
@nykolaslima Looks good to me.
Could we make a test which tries to select wrong db index (maybe Max integer or negative?) and check db index variable would be changed or not?

nykolaslima:
Sure we can! I'll do this.

nykolaslima:
@HeartSaVioR tests seems failing because of our datatype change in `db` attribute. But it's not merged on master.
I also changed the `select` method in `BinaryClient` to retrieve the result. We need this to check if the command was successfully executed in order to update `db` index. But I'm wondering if it will break `MultiKeyPipelineBase`. What do you think @HeartSaVioR?

nykolaslima:
@HeartSaVioR I did the rebase with master branch, but Travis is failing in `ObjectCommandsTest.objectEncoding`.
I couldn't figure out why this is failing. Can you give me a help? @marcosnils if you can help me too.

marcosnils:
@nykolaslima tests are failing due to #845.

nykolaslima:
That's why on my machine it worked, I was not running on latest unstable Redis.

Can I fix it in this same PR?

marcosnils:
I'm uploading the fix at this moment. Once it get's merged I'll re-run travis tests

nykolaslima:
@marcosnils ok! Can you tell me when you got it merged? Then I need to rebase my branch.

marcosnils:
Sure,  will let u know.

marcosnils:
@nykolaslima it's merged. I've also re-run the tests for this PR

nykolaslima:
@marcosnils now it's all passing!

Can you guys review, please? @marcosnils @HeartSaVioR

marcosnils:
@nykolaslima. Question: I've seen the test you added but I don't get why 999 is an invalid DB. Seems like I'm missing something and I can't see it.

marcosnils:
I guess I've figured it out. Select will fail because Redis will be expecting for an AUTH right?. If that is the case then I guess we should rename the test or add a comment or something. If you don't know beforehand that the redis instance is expecting for an auth command you'll never know why that test is passing because 999 is a __valid__ redis db number.

nykolaslima:
@marcosnils in the `redis.conf` we can configure the maximum number of databases.

By default, it's 16
http://download.redis.io/redis-stable/redis.conf
```
# Set the number of databases. The default database is DB 0, you can select
# a different one on a per-connection basis using SELECT <dbid> where
# dbid is a number between 0 and 'databases'-1
databases 16
```

nykolaslima:
Oh, you find a bug in my code. We supposed to pass the auth. I'll fix it.

marcosnils:
I guess we should clarify the test somehow, you even didn't realize that the test was passing because of the AUTH proble. I guess i'd name the test something like `testDontUpdateDbIndexIfSelectFails` and add a comment why it's failing so the person who reads the text understand the context because select may fail for many reasons (connection, auth, invalid db, etc) and not just miss configuration.

nykolaslima:
Actually it was passing here, because I was using a local redis. But I liked the test name suggestion.
About be more explicit, I think it already are because we've extracted a variable that represents an invalid db index `invalidDb` and if no error occurs, we are failing with the text: `Should throw an exception if tried to select invalid db`.

nykolaslima:
I did some changes, what do you think @marcosnils? https://github.com/xetorthio/jedis/commit/285841ce7b6eb9ee536ed5babb9901e2f3dcaef1

nykolaslima:
@HeartSaVioR I changed the value of invalid db index to -1. I think it's more explicit this way.

Do you have other suggestion? @HeartSaVioR @marcosnils

marcosnils:
@nykolaslima LGTM.

HeartSaVioR:
@nykolaslima OK, I will not bother about it because it's just a unit test. :) LGTM.

HeartSaVioR:
@nykolaslima If you don't mind, could you reorganize your commits?
It's too many upmerged commits, and contains wrong commit.
It would be hard to cherry-pick to apply it to bugfix / minor release branch.
Thanks in advance!

nykolaslima:
@HeartSaVioR Done!

HeartSaVioR:
@nykolaslima Thanks, merging in progress.

HeartSaVioR:
Merged into master, 2.6, 2.7 branches.

PR #854, opened 2015-01-13
validating URI before connecting

### LINKED ISSUE #850: NullPointerException initializing JedisPool with URI with no path
When constructing a JedisPool using the URI constructor, if you supply a minimalist basic URI without path, such as `localhost:6379`, the result is an NPE in the `JedisURIHelper`:

```java
  public static Integer getDBIndex(URI uri) {
    String[] pathSplit = uri.getPath().split("/", 2);  // <-- Exception here, because uri.getPath() is null
    if (pathSplit.length > 1) {
    // ...
    }
  }
```

nykolaslima:
closes #850 

@HeartSaVioR @marcosnils

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisPool.java
+++ b/src/main/java/redis/clients/jedis/JedisPool.java
@@ -27,7 +27,7 @@ public class JedisPool extends Pool<Jedis> {
 
   public JedisPool(final String host) {
     URI uri = URI.create(host);
-    if (uri.getScheme() != null && uri.getScheme().equals("redis")) {
+    if (JedisURIHelper.isValid(uri)) {
       String h = uri.getHost();
       int port = uri.getPort();
       String password = JedisURIHelper.getPassword(uri);
@@ -78,9 +78,7 @@ public class JedisPool extends Pool<Jedis> {
   }
 
   public JedisPool(final GenericObjectPoolConfig poolConfig, final URI uri, final int timeout) {
-    super(poolConfig, new JedisFactory(uri.getHost(), uri.getPort(), timeout,
-        JedisURIHelper.getPassword(uri),
-        JedisURIHelper.getDBIndex(uri), null));
+    super(poolConfig, new JedisFactory(uri, timeout, null));
   }
 
   @Override
--- a/src/main/java/redis/clients/jedis/JedisShardInfo.java
+++ b/src/main/java/redis/clients/jedis/JedisShardInfo.java
@@ -2,6 +2,7 @@ package redis.clients.jedis;
 
 import java.net.URI;
 
+import redis.clients.jedis.exceptions.InvalidURIException;
 import redis.clients.util.JedisURIHelper;
 import redis.clients.util.ShardInfo;
 import redis.clients.util.Sharded;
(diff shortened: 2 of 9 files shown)


marcosnils:
@nykolaslima LGTM. @bennesher what do you think?

bennesher:
@nykolaslima LGTM too!

src/main/java/redis/clients/jedis/JedisPool.java
```diff
@@ -27,7 +27,7 @@ public JedisPool(String host, int port) {
 
   public JedisPool(final String host) {
     URI uri = URI.create(host);
```
HeartSaVioR: Seems like we can call this(new GenericObjectPoolConfig(), URI.create(host)); instead of extracting partial URL values to remove duplication.
Could you please check? :)

HeartSaVioR:
I've commented one thing about removing duplication, but it's trivial thing. 
LGTM, too.
After verifying my comments are valid, and applying if it's valid, I'll merge this.

src/main/java/redis/clients/jedis/JedisShardInfo.java
```diff
@@ -31,7 +32,7 @@ public int getPort() {
   public JedisShardInfo(String host) {
     super(Sharded.DEFAULT_WEIGHT);
```
HeartSaVioR: Seems like we can call this(URL.create(host)) instead of extracting partial URL values to remove duplication.
Could you please check? :)

src/main/java/redis/clients/jedis/JedisPool.java
nykolaslima: @HeartSaVioR we can call `this(new GenericObjectPoolConfig(), URI.create(host))` because call another constructor must be the first command inside the constructor.

We could change the entire constructor to delegate to `this(new GenericObjectPoolConfig(), URI.create(host))`, but doing this, if the URI is invalid we will thrown one exception instead of trying to build the `GenericObjectPoll` with host string. I think that's not the desired action.

src/main/java/redis/clients/jedis/JedisShardInfo.java
nykolaslima: Same here. If we delegate to another constructor, we will thrown a `InvalidURIException` instead of setting the host string with the default port.

HeartSaVioR: @nykolaslima OK, I can see "else" state. 
We can remove duplication by adding new setup method which is like BinaryJedis.initializeClientFromURI(), but current state is acceptable.

HeartSaVioR:
LGTM, too. I'll merge this in.

src/main/java/redis/clients/jedis/JedisPool.java
HeartSaVioR: @nykolaslima I see, thanks for explanation. :)

HeartSaVioR:
@nykolaslima Thanks for your effort. Merged into master, 2.6, 2.7 branches.

nykolaslima:
No problem @HeartSaVioR. Thank you for supporting me :)

PR #865, opened 2015-01-22
Fixes #864

### LINKED ISSUE #864: Class BinaryShardedJedis does not have method "public String set(final byte[] key, final byte[] value, final byte[] nxxx, final byte[] expx, final long time)"
As you see, could not find the method, is there a reason for this?

auhig:
Fixes #864
 Add method "public String set(byte[] key, byte[] value, byte[] nxxx, byte[] expx, long time)"

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryJedisCommands.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedisCommands.java
@@ -11,6 +11,8 @@ import java.util.Set;
 public interface BinaryJedisCommands {
   String set(byte[] key, byte[] value);
 
+  String set(byte[] key, byte[] value, byte[] nxxx, byte[] expx, long time);
+
   byte[] get(byte[] key);
 
   Boolean exists(byte[] key);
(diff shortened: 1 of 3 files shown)


marcosnils:
@auhig LGTM. Once another review checks it we'll merge.

nykolaslima:
It would be nice to add one test to `AllKindOfValuesCommandsTest`. What do you guys think @marcosnils @auhig?

HeartSaVioR:
LGTM, too. But I also think it would be better to have unit test.
@auhig Could you apply it? Thanks in advance!

auhig:
Unit testing is good, but I could not find some classes for testing commands of BinaryShardedJedis, since these classes use the Jedis instance. Is it necessary to create a new UT class?

nykolaslima:
@auhig you should create the test in this class: `AllKindOfValuesCommandsTest `.

This class already has a `jedis` instance inherited from `JedisCommandTestBase`;

nykolaslima:
LGTM!

marcosnils:
:+1:

HeartSaVioR:
LGTM, too.

marcosnils:
I'll merge.

marcosnils:
Merged to master and 2.7

PR #871, opened 2015-01-31
Pexpire at

Tavio:
Adding the pexpireAt command to interfaces that missed it (and their corresponding implementations). I'm not implementing the pexpire command here because it is already present in a recent PR by @nykolaslima.

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryJedisCommands.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedisCommands.java
@@ -24,6 +24,8 @@ public interface BinaryJedisCommands {
   Long expire(byte[] key, int seconds);
 
   Long expireAt(byte[] key, long unixTime);
+  
+  Long pexpireAt(byte[] key, long millisecondsTimestamp);
 
   Long ttl(byte[] key);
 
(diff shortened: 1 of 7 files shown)


nykolaslima:
LGTM. What do you think @HeartSaVioR @marcosnils?

HeartSaVioR:
LGTM, too.

HeartSaVioR:
It adds missed spots so I labeled it with 2.6.3.
Seems like if we release minor version faster, I can label this to 2.7.0 too.

nykolaslima:
@marcosnils @xetorthio Do you guys have any consideration on this? Can we merge it?

xetorthio:
The implementation is fine. But it is missing tests, right?

nykolaslima:
As @HeartSaVioR commented on https://github.com/xetorthio/jedis/pull/869, we don't have tests for `ShardedJedis` so I think it's fine. After all, it only delegates to `Jedis` that is already tested.

HeartSaVioR:
Yes, it is not only this PR's issue. 
We need to place 'under construction' and do a numerous changes in order to achieve.
So unless we decide to go on, I think it's good to merge, at least 2.x version.

marcosnils:
Upmerged to 2.7 and 2.6

PR #875, opened 2015-02-03
Remove ShardedJedis in order to focus core functionality (3.x)

HeartSaVioR:
https://groups.google.com/d/msg/jedis_redis/7ZJPFp_qTY0/mIZ9qZ8ZgdkJ

We're decided to stop supporting ShardedJedis starting from 3.x. (with leaving it to 2.x)

I removed ShardedJedis with relevant codes and try to clean codes to make Jedis clearer.

At first, many classes are related to ShardedJedis including utilities, so we can get rid of it.
And we can simplify Pipeline and relevant interfaces because there's no need to split operations which one is single key and other one is multi keys.

In addition, I've modified some interfaces name cause to have consistent naming convention.
#630 also removes ShardedJedis, but it removes only ShardedJedis and relevant codes, not modifying interfaces, and Pipeline.

---
This change is [Reviewable](https://reviewable.io/reviews/xetorthio/jedis/875)


nykolaslima:
@HeartSaVioR I don't know too much about Pipeline part, but the PR looks good to me.

I liked the names' refactoring. But I think that after merging it, we can begin to refactoring the packages. We have a single `jedis` package and many classes inside it. We can split it in their specific domains.

Great job removing all this :+1:

HeartSaVioR:
Regarding Pipeline, if we remove ShardedJedis, we are no longer need to separate single key and multi key operations by Pipeline/Transaction.
We are still need to separate single key and multi key operations by normal situation because of JedisCluster.
Some multi-key operations seems not fit to JedisCluster, so we can't use same interfaces, which we need to discuss how to solve.

HeartSaVioR:
@marcosnils @sazzad16 
I still think removing ShardedJedis brings benefits on codebase (as you can see the count of deleted lines from the diff). Redis introduces Redis Cluster from 3.0 and Redis even reached 4.0. Let's recommend Redis Cluster and just remove ShardedJedis.

sazzad16:
Resolved by #2693

PR #878, opened 2015-02-06
refactoring set command to use SetParams for the optional parameters

### LINKED ISSUE #877: Being able to use #set with only expire time
We currently have the method: `String set(String key, String value, String nxxx, String expx, long time)`

But if we want to only pass the `ex` or `px` timeout parameter, we are not able to because `nx` or `xx` parameter is required.

I'm thinking in two possible solutions:
1. create a new set method, with this signature: `String set(String key, String value, String expx, long time)`
2. accept null value in the current method, so we can call this way: `jedis.set("myKey", "value", null, "ex", 60)`

I believe that the 1. is the better solution, what do you guys think @HeartSaVioR @marcosnils ?

nykolaslima:
It closes #877 

1. I let the `binary` layer receiving all the possible parameters. The `binary` layer was receiving expire time in `long` instead of `byte[]`, so I also change it.

2. I removed the `set` methods that receive only a few optional parameters.

3. The other layers(command interfaces and implementations) will receive the `SetParams`.


xetorthio:
I think that checking arguments is not necessary and actually wrong.
Redis allows to send both EX and PX on the same command.
And even EX and NX. If you check, redis doesn't return an error and doesn't complain. Why should we?

nykolaslima:
@xetorthio in my opinion, Redis should also complain about it. If we set both, EX and PX, what expire time should be used? And if we set both EX and NX, no value will be returned.

We know that those values are wrong, so why don't warning the user about it? It can avoid wrong use of the parameters. Actually, I'll ask this in Redis repo too.

nykolaslima:
Actually, after we talked, I agree with you! I did remove the checking.

xetorthio:
So LGTM!

nykolaslima:
Thanks @xetorthio. @heartsavior and @marcosnils what do you think about it?

src/main/java/redis/clients/jedis/BinaryClient.java
```diff
@@ -100,9 +102,10 @@ public void set(final byte[] key, final byte[] value) {
     sendCommand(Command.SET, key, value);
   }
 
-  public void set(final byte[] key, final byte[] value, final byte[] nxxx, final byte[] expx,
-      final long time) {
-    sendCommand(Command.SET, key, value, nxxx, expx, toByteArray(time));
+  public void set(final byte[] key, final byte[] value, final byte[] nxxx, final byte[] expx, final byte[] time) {
```
HeartSaVioR: @nykolaslima 
Even BinaryClient uses ZParams, so we can change its signature to public void set(final byte[] key, final byte[] value, final SetParams params).
Same things can be applied to BinaryJedis.

@xetorthio @marcosnils What do you think about it?

nykolaslima: I agree with that.

This way we will be more flexible if `*Command*Parameters` have any changes. I'll fix it.

HeartSaVioR:
@nykolaslima 
Looks like Redis complains Syntax Error. Could you check again?
https://travis-ci.org/xetorthio/jedis/builds/50217281

marcosnils:
@nykolaslima any updates about this?

nykolaslima:
@marcosnils and @HeartSaVioR sorry about my delay, I'm really busy this week but I'll try to finish it today!

HeartSaVioR:
@nykolaslima PING

nykolaslima:
@HeartSaVioR @marcosnils @xetorthio Guys, I'm sorry for taking so long on this. I was very busy during these days.

Please, take a look at the PR. Now we got all tests passing. If you have any considerations, please let me know.

By the way, congratulations on the new versions releases :tada:

HeartSaVioR:
@nykolaslima Well done! When other collaborators take a look and OK then I'll merge.

HeartSaVioR:
I'm +1, and since any collaborators review such a long time, it would be good to merge now.

HeartSaVioR:
@nykolaslima Thanks for the great effort! I merged this to master.

marcosnils:
:+1:

nykolaslima:
Thank you guys. And i'm sorry for being so away those times..

HeartSaVioR:
@nykolaslima Not at all! We have many issues to resolve, and I'd be happy if you lend a hand. :)

PR #886, opened 2015-02-12
Fix blpop and brpop in JedisCluster

antoaneta-marinova:
Before when trying to execute RedisCluster's method blpop or brpop (for example redisCluster.blpop(0, QUEUE)) a JedisClusterException was thrown with the meassage "No way to dispatch this command to Redis Cluster.".
The first commit proposes a fix to this issue.
The second adds a test for the method.

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisCluster.java
+++ b/src/main/java/redis/clients/jedis/JedisCluster.java
@@ -1095,7 +1095,7 @@ public class JedisCluster implements JedisCommands, BasicCommands, Closeable {
       public List<String> execute(Jedis connection) {
         return connection.blpop(arg);
       }
-    }.run(null);
+    }.run(arg);
   }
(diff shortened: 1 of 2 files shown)


marcosnils:
@antoaneta-marinova thanks for the contribution. The test is not really necessary as it's Jedis Cluster is just a wrapper of Jedis. Usually when we find bugs like these is because we forget to pass over the arguments from JedisCluster to Jedis. That's why we haven't implemented any commands in JedisCluster tests because there's not really much to test there rather than possible typos we may have made. 

Once you remove the test I'll merge the PR.

antoaneta-marinova:
Thank you for the review @marcosnils. 
I reverted the commit which added the test.

marcosnils:
Downmerged to 2.6 and 2.7 respectively.

PR #888, opened 2015-02-20
Add readonly command to cluster-revised

taer:
Add the READONLY command so you can use a redis.clients.jedis.Jedis instance to query a slave that happens to be a cluster member.


taer:
It's failing due to cluster-revised missing PR https://github.com/xetorthio/jedis/pull/846
Should I add that minor test update to this PR?

taer:
Re-opening due to accidental close

marcosnils:
@taer is there any reason why this can't be merged in the current master branch?. I'm wondering why you opened the PR against the cluster-revised.

taer:
@marcosnils There is a bit of divergence on cluster-revised vs master in where commands are implemented. JedisCluster on the cluster-revised no longer implements BasicCommands, which is where things like echo, ping, and now readonly exist.

I could base this off master fairly easy, but that means the JedisCluster class will now get a readonly() implementation. I could create that PR fairly easily. 

How are y'all handling merges to cluster-revised? This PR won't merge cleanly.

taer:
#892 is against master. I'll  close this one and deal with merge at a later time.

taer:
@marcosnils  in case you would rather have this against cluster-revised for now.

HeartSaVioR:
Since we don't support READONLY from JedisCluster, I think adding it to cluster-revised branch is more clearer way to add READONLY, too.

taer:
@HeartSaVioR Want me to close the one against master?

HeartSaVioR:
Let us hold your two PRs and select one. We'll take care of it.

HeartSaVioR:
Closing it since cluster-revised branch is merged to master.
We should deal with conflict, but #892 is the right place.

PR #889, opened 2015-02-20
make pexpire in BinaryJedis was take a byte[] instead of String

taer:
Updated pexpire in BinaryJedis to be byte[] key. Wasn't consistent with rest of binaryJedis


marcosnils:
@taer as this change breaks backwards compatibility you need to split your PR in two parts. 

1. @deprecate the current pexpire method in BinaryJedis and add the new one. 

2. Submit a new PR which removes the @deprecated method before.

Let me know if it's unclear.

PR #890, opened 2015-02-21
Add BinaryJedis.pexpire which take a byte[] instead of String

taer:
Added pexpire in BinaryJedis with byte[] key. Wasn't consistent with rest of binaryJedis.

Deprecated the String version.

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryJedis.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedis.java
@@ -337,6 +337,7 @@ public class BinaryJedis implements BasicCommands, BinaryJedisCommands, MultiKey
   }
 
   /**
+   * @deprecated use BinaryJedis.pexpire(byte[], long) or Jedis.pexpire(String,long)
    * Set a timeout on the specified key. After the timeout the key will be automatically deleted by
    * the server. A key with an associated timeout is said to be volatile in Redis terminology.
    * <p>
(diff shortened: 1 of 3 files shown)


HeartSaVioR:
+1, LGTM
Could you please post relevant PR to remove wrong methods? Thanks!

marcosnils:
@HeartSaVioR shouldn't this be included in 2.7?.

HeartSaVioR:
@marcosnils Oh, you're right. I'll fix it.

Btw, classifying version but not planning to release seems like a kind of postponement. How can we settle it? Or it's not an issue?

marcosnils:
Hmm I guess the best option would be to schedule releases (if there are actually something to release) regularly. Another thing that we can do is to make daily snapshots of the latest versions, that way if anyone want's to have the latest version without an official release they can.

taer:
@HeartSaVioR  My Pr would be based against master, which doesn't yet include this. So it's diffs show this PR's commit as well. I assume the PR will automatically update when this PR one gets merged.

HeartSaVioR:
@taer That's what I ask for. ;)
Btw, sorry I missed one thing, please add @deprecated to BinaryJedis's pexpire, too.

HeartSaVioR:
@taer I think it's good to merge. Thanks!

HeartSaVioR:
Merged into master, and 2.7 branch.

PR #891, opened 2015-02-21
Add readonly command

taer:
Add the READONLY command so you can use a redis.clients.jedis.Jedis instance to query a slave that happens to be a cluster member.

Accidentally deleted branch backing pullReq #888, which closes it. Same patch


taer:
Sorry for the extra pull. Just found the re-open option on the one I accidentally closed

PR #892, opened 2015-02-22
Add readonly command to master branch

taer:
A port of #888 based against master, with unit tests. The Cluster unit test utilized static instances, which got wonky. So I setup and tore down the slave in the added test method.


src/main/java/redis/clients/jedis/JedisCluster.java
```diff
@@ -1127,6 +1127,15 @@ public String execute(Jedis connection) {
       }
     }.run(null);
   }
+  @Override
+  public String readonly() {
```
marcosnils: What's the purpose of adding the readonly command here?. JedisCluster will throw an error because there's no key supplied in the `run` method.

HeartSaVioR: Readonly is fit for BasicCommand, and JedisCluster now implements BasicCommand.
Seems like we should merge cluster-revised PR to master first.

taer: It's only there due to JedisCluster implementing BasicCommands in master. The cluster-revised option is much cleaner. In the meantime, we could stub it out to return OK. Since JedisCluster will always send that to a master node, there is really no reason to call readonly. But if someone did, a OK stub would result in their next command working as expected.

HeartSaVioR:
@taer 
So sorry for waiting too long. 
We merged cluster-revised branch to master, so I'd like to take this PR, but needs upmerge.
If you don't mind, could you do some works to upmerge?

marcosnils:
@taer can you please rebase?

Thx!.

taer:
Dropped the readonly command from JedisCluster. It didn't make sense there. This should be good to go now.

src/main/java/redis/clients/jedis/BasicCommands.java
```diff
@@ -41,4 +41,6 @@
   String configResetStat();
 
   Long waitReplicas(int replicas, long timeout);
+
+  String readonly();
```
marcosnils: Wouldn't it make sense to add this command to the `ClusterCommands` interface?. As `readonly` it's available only under clustering conditions.

src/test/java/redis/clients/jedis/tests/JedisClusterTest.java
```diff
@@ -21,12 +22,14 @@
...
   private static Jedis node3;
   private static Jedis node4;
+  private static Jedis nodeSlave2;
   private String localHost = "127.0.0.1";
 
   private HostAndPort nodeInfo1 = HostAndPortUtil.getClusterServers().get(0);
   private HostAndPort nodeInfo2 = HostAndPortUtil.getClusterServers().get(1);
   private HostAndPort nodeInfo3 = HostAndPortUtil.getClusterServers().get(2);
   private HostAndPort nodeInfo4 = HostAndPortUtil.getClusterServers().get(3);
+  private HostAndPort nodeInfoSlave2 = HostAndPortUtil.getClusterServers().get(4);
```
marcosnils: For what I see this is not a `slave` node of the cluster. It's another master node...

src/main/java/redis/clients/jedis/BasicCommands.java
taer: It's really of use for standalone connections to a redis cluster. If you connect to a slave redis custer node with a normal Jedis connection(that has a copy of the slot) and perform a query, it will still redirect you. If you set readonly, it will serve the answer it has. I'm not sure how this would apply to cluster, since Jedis cluster will always send you to the master for that slot.

src/test/java/redis/clients/jedis/tests/JedisClusterTest.java
taer: In the test readonly method below, I set this node as a slave of the first now. I was simply using an existing connection for the test.

src/main/java/redis/clients/jedis/BasicCommands.java
marcosnils: @taer I'm not saying that `JedisCluster` should have the `readonly` command. I'm suggesting to move this `readonly` signature to `ClusterCommands` interface which is implemented by `Jedis` class not `JedisCluster`

src/test/java/redis/clients/jedis/tests/JedisClusterTest.java
marcosnils: Oh, I missed the part where you set the node as slave. Then I guess it's ok.

src/main/java/redis/clients/jedis/BasicCommands.java
taer: Ah, makes sense. Will update

HeartSaVioR:
I agree ClusterCommands seems to be more proper place to readonly(). Thanks @marcosnils!
Looks good to me overall. Thanks @taer!

marcosnils:
thanks @HeartSaVioR I'll re-run the tests and merge.

marcosnils:
Merged to 3.0 and 2.7 respectively. Thanks for the contribution @taer

Issue #896, opened 2015-02-28
RENAME functionality has changed in Redis Unstable


ianjoneill:
The test ```renameOldAndNewAreTheSame()``` in ```AllKindOfValuesCommandsTest``` fails following antirez/redis#2392. Renaming a key to itself is now essentially a noop.

PR #897, opened 2015-02-28
PEXPIRE should take a long parameter

ianjoneill:
Deprecates ```pexpire(String, int, String)``` in favour of ```pexpire(String, long, String)```. Note that test failures are due to #896. This change is in the same vein as #576.


ianjoneill:
Named commit wrong, meant psetex

PR #902, opened 2015-03-03
Remove deprecated methods (blpop, brpop, pexpire)

HeartSaVioR:
Removed all deprecated methods in master branch.

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisCluster.java
+++ b/src/main/java/redis/clients/jedis/JedisCluster.java
@@ -1089,26 +1089,6 @@ public class JedisCluster implements JedisCommands, BasicCommands, Closeable {
   }
 
   @Override
-  public List<String> blpop(final String arg) {
-    return new JedisClusterCommand<List<String>>(connectionHandler, timeout, maxRedirections) {
-      @Override
-      public List<String> execute(Jedis connection) {
-        return connection.blpop(arg);
-      }
-    }.run(arg);
-  }
-
-  @Override
-  public List<String> brpop(final String arg) {
-    return new JedisClusterCommand<List<String>>(connectionHandler, timeout, maxRedirections) {
-      @Override
-      public List<String> execute(Jedis connection) {
-        return connection.brpop(arg);
-      }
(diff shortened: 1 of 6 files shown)


src/main/java/redis/clients/jedis/JedisCluster.java
```diff
@@ -1089,26 +1089,6 @@ public Long execute(Jedis connection) {
   }
 
   @Override
-  public List<String> blpop(final String arg) {
```
nykolaslima: This guys weren't marked as `@Deprecated` but I believe there is no problem to delete it since `JedisCommands` marked it as `@Deprecated`. :+1:

nykolaslima:
Just added a note but it looks good to me. :+1:

src/main/java/redis/clients/jedis/JedisCluster.java
HeartSaVioR: Oh, that's right. Thanks for finding!
Btw, we can do the trick, adding @Deprecated to methods and apply it to 2.7.

nykolaslima: I agree with that. Do you want me to do it or you gonna do by yourself?

HeartSaVioR: Oh, that would be nice. Could you do it if you don't mind?
Thanks in advance!

nykolaslima: Sure. Doing it.

Jungtaek Lim <notifications@github.com> escreveu no dia ter, 3/03/2015 às
11:23:

HeartSaVioR: Resolved by #903

HeartSaVioR:
We should apply this to (Binary)JedisCluster when #671 will be merged into mater.

marcosnils:
@HeartSaVioR can you please rebase so it can be merged?

Thx!.

HeartSaVioR:
@marcosnils Sure! Here it is. :)

Issue #909, opened 2015-03-06
Jedis returning wrong values after long command execution


mpaoletta:
We're having a severe app/driver bug on our heavily concurrent app. 

This is a single redis installation with over 4M keys. In order to reproduce the problem, we execute a long running command from redis-cli, like a KEYS * or a DEBUG SLEEP (this is just to reproduce the problem), our client code (Jedis based) starts to throw errors, first socket read timeouts, which are expected and then it throws typical concurrency errors in Jedis, like trying to cast a string into a binary. The worst scenario is that sometimes it seems that it returns wrong keys. These errors keep occurring even after the long running command finishes. 

We were using a Jedis instance in each akka actor instance, so it shouldn't be a concurrency issue in our code (we were fairly sure that the redis connection wasn't shared between threads) but we changed to JedisPool nevertheless. But the problem remained (we are using Jedis 2.6.2).

The usual pattern is (this is Scala):

		    val redis = MicrositeActor.jedisPool.getResource()
		    try {
		    	redis.getSet(token + ":sent", "sent") == null
		    }
		    finally {
		      MicrositeActor.jedisPool.returnResource(redis)
		    }  

We use the following commands:

get, exists, hgetAll, getSet, expire, ttl, multi/exec, set, del.

Are you aware of any concurrency issue on any of these commands?

We are currently evaluating other drivers, but we would really appreciate any tips regarding this.

TIA,
Martin Paoletta

mpaoletta:
I closed the original redis issue because it seems it is not related to redis but to Jedis (about to upload a simple case to reproduce it)

nykolaslima:
It would be nice if you could provide us a small test case about this problem.

I know that `Jedis` isn't thread safe, but you said that you are using one for each akka actor.

I'll try to take a look at it.

mpaoletta:
We reproduced the issue just with GET commands on a single redis instance. About to upload sample test code that reproduces the issue.

nykolaslima:
Ok, thank you. I'll wait for this.

mpaoletta:
This is simple groovy code that reproduces the issue:

package main.groovy

import redis.clients.jedis.*

import java.util.concurrent.*


static main(args) {


    def THREADS = 5

    def jedisPool = new redis.clients.jedis.JedisPool(new redis.clients.jedis.JedisPoolConfig(maxTotal:10), 'localhost')

// This populates the DB
    j = jedisPool.resource
[...]

PR #910, opened 2015-03-08
deprecating returnBrokenResource

nykolaslima:
We will use only returnResource and the broken logic will be in it.

This is related to #909 in order to simplify the return resource to pool logic.

@HeartSaVioR @marcosnils

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisPool.java
+++ b/src/main/java/redis/clients/jedis/JedisPool.java
@@ -88,6 +88,10 @@ public class JedisPool extends Pool<Jedis> {
     return jedis;
   }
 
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
When deprecating method, we should provide a way to not using it right now.
So returnResource should be changed with this PR, too.

And hey wait, we don't end discussion about #909. ;)

marcosnils:
We agreed that we're going to deprecate both `returnBrokenResource` and `returnResource` in favor of making usage of `jedis.close()`. This PR doesn't apply anymore.

nykolaslima:
:+1:

PR #912, opened 2015-03-09
Deprecates JedisPool returnResource and returnBrokenResource

marcosnils:
Per #909 as we're planning not to support these methods in the future to avoid confusions.

PR #913, opened 2015-03-09
Change visibility of returnBrokenResource and returnResource

marcosnils:
Following #912 we're changing the visibility of the specified methods. 

I've also found a bug in `jedis.quit()` which I'd like @nykolaslima or @HeartSaVioR to review. Please check this commit  aa9a5e2. If you think it's ok we should downmerge that commit to 2.6 and 2.7


src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
```diff
@@ -210,16 +210,6 @@ public void rename() {
   }
 
   @Test
-  public void renameOldAndNewAreTheSame() {
```
HeartSaVioR: Is it a reason to be removed?

HeartSaVioR:
#912 again.
Could we apply same rule to JedisSentinelPool and ShardedJedisPool respectively?

src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
marcosnils: @HeartSaVioR my bad. I wasn't running the latest unstable version of redis in my computer and the `rename` option changed a couple of days ago. 

I'll rollback this test.

nykolaslima:
@marcosnils When we call `quit`, there is some reason to connection keep opened?

marcosnils:
@nykolaslima no there isn't  because redis actually closes the socket. That's what my fix does, because if you issue a `quit` then Jedis will throw an IOException because the other side ended the connection.

nykolaslima:
So it wouldn't be better if we call `disconnect` inside `quit`?

marcosnils:
@nykolaslima that's what I did. I'm calling `client.disconnect()` inside `jedis.quit()` method. I don't understand where you're going.

nykolaslima:
As we talked, we can't move the `disconnect` call into `BinaryClient`'s `quit` because this way we can't read the result in `BinaryJedis`.

The PR looks good to me. (Travis is falling, but Marcos said we just need to re-run it)

HeartSaVioR:
Forgot to tell about quit. Great! Looks good to me. :)

marcosnils:
@nykolaslima @HeartSaVioR I've already changed the visibility to SentinelPool and ShardedPool.
Please review it one last time and merge when you feel like it :)

HeartSaVioR:
@marcosnils LGTM!
Btw, it's easier to merge sequentially, first #912, and #913.
I'll merge this after merging #912.

HeartSaVioR:
Upmerged, and merged into master.
Thanks all!

nykolaslima:
:+1:

HeartSaVioR:
@Xiaoshuai Thanks for reporting! We're completely missed, and I'll apply same strategy to Pool.returnResourceObject().

PR #945, opened 2015-04-09
Continue renewSlotCache if exception is encountered

pcchan:
If it encounters a JedisMovedDataException, JedisClusterCommand.runWithRetries will trigger renewSlotCache. However, renewSlotCache's behavior isn't currently like initializeSlotsCache and will fail out if it happens to connect to a node that is inaccessible. 

I encountered this problem while testing out cluster node failures under load.

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisClusterConnectionHandler.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterConnectionHandler.java
@@ -72,6 +72,8 @@ public abstract class JedisClusterConnectionHandler {
         jedis = jp.getResource();
         cache.discoverClusterSlots(jedis);
         break;
+      } catch (JedisConnectionException e) {
+        // try next nodes
       } finally {
         if (jedis != null) {
           jedis.close();


marcosnils:
@pcchan your commit makes totally sense. Any chance you can write a test so we make sure we don't break this in the future?. 

Thanks for the contribution!

HeartSaVioR:
LGTM, Thanks for finding missed spot!

Btw, I'm afraid that it would be something complicated to make test since 

- first value of cache.getNodes().values() (naming to first node) should be alive at initializing cluster
-- actually I'm not sure it should be, maybe it isn't
- and we should invalidate all connections from JedisPool which points to first node
-- or kill first node but make sure that cluster doesn't go to DOWN (We should take care so that it doesn't affect other tests)
- and we should trigger MOVED to let JedisCluster call renewSlotCache's

marcosnils:
@HeartSaVioR what about doing this: 

1) Set up the test and initialize the cluster (this is already done in JedisClusterTest)
2) disconnect a node manually by doing jc.getClusterNodes and closing the pool of one of the nodes
3) Re-assign slots manually in some node to trigger the MOVED condition
4) Try to get a key to the moved slot so JedisCluster executes `renewSlotCache's` method
5) Assert that exception is not raised

I believe the first part of the test should be something like this 

```java
  public void testRenewSlot() {
    jc.getClusterNodes().get("127.0.0.1:7380").close();
  }
```
Only the MOVED triggering is missing which we already have in some other tests.

What do you think?

HeartSaVioR:
@marcosnils Totally makes sense.
@pcchan Could you craft a unit test if you don't mind? 
If you're having hard time to implement, just let us know that we can implement it later.

pcchan:
I'll give it a shot and update the commit.

marcosnils:
I've submitted #949 with a test case. Let's continue there.

marcosnils:
Merged to master and 2.7 respectively.

Issue #847, opened 2015-01-09
Possible resource leak in disconnect()


allanwax:
In `Connection`

```java
	public void disconnect() {
		if (isConnected()) {
			try {
				inputStream.close();
				if (!socket.isClosed()) {
					outputStream.close();
					socket.close();
				}
			}
			catch (IOException ex) {
				broken = true;
				throw new JedisConnectionException(ex);
			}
		}
	}
```

If either `inputStream` or `outputStream` fails during close, the socket is never closed.  Sockets are one of the few things that if they are not closed the kernel will keep them 'open' and deny reuse of the resource.

Better would be separate `try ... catch` clauses which will eventually get to closing the socket.

allanwax:
Something like this.  apologies, I just quickly scrathed this out.

```java
public void disconnect() {
		if (isConnected()) {
			try {
				try {
					inputStream.close();
				}
				catch (IOException ex) {
					broken = true;
					throw new JedisConnectionException(ex);
				}
				finally {
					if (!socket.isClosed()) {
						try {
							outputStream.close();
						}
						catch (Exception e2) {
							broken = true;
							throw new JedisConnectionException(e2);
						}
					}
				}
			}
			catch (Exception ex) {
				broken = true;
				throw new JedisConnectionException(ex);
			}
			finally {
				if (!socket.isClosed()) {
					try {
						socket.close();
					}
					catch (Exception e1) {
						broken = true;
						throw new JedisConnectionException(e1);
					}
				}
			}
		}
	}
```

allanwax:
This is also one of those rare times when a finalizer to close the socket would be useful.

HeartSaVioR:
At first, thanks for finding it!
It should ensure that it will try to close socket. You're right.

Btw, We can't rely on Java's finalizer, you know already.
We can add ShutdownHook, but it doesn't work with kill -9, and hey, Jedis is just a driver for Redis, not framework or sth.
We provide Closeable already (it means users can close it at shutdown phase), and it should be enough.

allanwax:
Having Closeable does not mean it will be closed so I don't think it is enough just to have it.  The shutdown hook is usefull but only at the normal/TERM exit of the program.  kill -KILL makes all of this meaningless.  So a finalizer is usefull in the case where someone just throws away the object.

[...]

PR #946, opened 2015-04-12
Release all resources from disconnect regardless of IOException

HeartSaVioR:
Sorry I'm too late on applying #847. (Thanks @allanwax!)

When IOException occurs on disconnect, it forces to close others with finally statement.
IMO it would be safe to use first IOException on disconnect to report JedisConnectionException.
I'm not 100% sure it's the best way, so any opinions are greatly appreciated.

### DIFF
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -169,6 +169,12 @@ public class Connection implements Closeable {
       } catch (IOException ex) {
         broken = true;
         throw new JedisConnectionException(ex);
+      } finally {
+        closeIOResourceQuietly(inputStream);
+        if (!socket.isClosed()) {
+          closeIOResourceQuietly(outputStream);
+          closeIOResourceQuietly(socket);
+        }
       }
     }
   }
@@ -271,5 +277,16 @@ public class Connection implements Closeable {
       }
     }
     return responses;
+  }
+
+  private void closeIOResourceQuietly(Closeable resource) {
+    // It's same thing as Apache Commons - IOUtils.closeQuietly()
+    if (resource != null) {
+      try {
+        resource.close();
+      } catch (IOException e) {
+        // pass
+      }
+    }
   }
 }
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -162,18 +162,18 @@ public class Connection implements Closeable {
   public void disconnect() {
     if (isConnected()) {
       try {
-        inputStream.close();
+        outputStream.close();
         if (!socket.isClosed()) {
-          outputStream.close();
+          inputStream.close();
           socket.close();
         }
       } catch (IOException ex) {
         broken = true;
         throw new JedisConnectionException(ex);
       } finally {
-        IOUtils.closeQuietly(inputStream);
+        IOUtils.closeQuietly(outputStream);
         if (!socket.isClosed()) {
-          IOUtils.closeQuietly(outputStream);
+          IOUtils.closeQuietly(inputStream);
           IOUtils.closeQuietly(socket);
         }
       }
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -11,6 +11,7 @@ import java.util.List;
 import redis.clients.jedis.Protocol.Command;
 import redis.clients.jedis.exceptions.JedisConnectionException;
 import redis.clients.jedis.exceptions.JedisDataException;
+import redis.clients.util.IOUtils;
 import redis.clients.util.RedisInputStream;
 import redis.clients.util.RedisOutputStream;
 import redis.clients.util.SafeEncoder;
@@ -170,10 +171,10 @@ public class Connection implements Closeable {
         broken = true;
         throw new JedisConnectionException(ex);
       } finally {
-        closeIOResourceQuietly(inputStream);
+        IOUtils.closeQuietly(inputStream);
         if (!socket.isClosed()) {
-          closeIOResourceQuietly(outputStream);
-          closeIOResourceQuietly(socket);
+          IOUtils.closeQuietly(outputStream);
+          IOUtils.closeQuietly(socket);
         }
       }
     }
@@ -277,16 +278,5 @@ public class Connection implements Closeable {
       }
     }
     return responses;
-  }
-
-  private void closeIOResourceQuietly(Closeable resource) {
-    // It's same thing as Apache Commons - IOUtils.closeQuietly()
-    if (resource != null) {
-      try {
-        resource.close();
-      } catch (IOException e) {
-        // pass
-      }
-    }
   }
 }
--- a/src/main/java/redis/clients/util/IOUtils.java
+++ b/src/main/java/redis/clients/util/IOUtils.java
@@ -1,0 +1,32 @@
+package redis.clients.util;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.net.Socket;
+
+public class IOUtils {
+  private IOUtils() {}
+
+  public static void closeQuietly(Closeable resource) {
+    // It's same thing as Apache Commons - IOUtils.closeQuietly()
+    if (resource != null) {
+      try {
+        resource.close();
+      } catch (IOException e) {
+        // pass
+      }
+    }
+  }
+
+  public static void closeQuietly(Socket resource) {
+    // It's same thing as Apache Commons - IOUtils.closeQuietly()
+    if (resource != null) {
+      try {
+        resource.close();
+      } catch (IOException e) {
+        // pass
+      }
+    }
+  }
+
+}


HeartSaVioR:
I'm not aware of JDK1.6, fixing now... ;(

marcosnils:
@HeartSaVioR have you thought about a way to test this?

HeartSaVioR:
@marcosnils 
Sorry I didn't think about test cause it is based on possibility, not having met this situation.
Could you confirm below scenario could make sense?

1. make a connection -- (A)
2. another connection sends client kill to let Redis kills connection (A)
3. call (A).disconnect() with catching JedisConnectionException
4. check if socket is closed -- Jedis.getClient().isConnected() should be true before applying, and should be false after applying

I'm not sure that 3. raises IOException every time, so above scenario may not make sense.

What do you think about it?

HeartSaVioR:
Unfortunately I can't produce failure with unit test.

```
  @Test
  public void disconnectShouldReleaseAllResourcesRegardlessOfIOException() {
    Jedis jedis = new Jedis(hnp.getHost(), hnp.getPort());
    jedis.auth("foobared");
    // ensure that jedis instance is connected to Redis
    jedis.clientSetname("disconnect");

    assertTrue(jedis.getClient().isConnected());

    Jedis jedis2 = new Jedis(hnp.getHost(), hnp.getPort());
    jedis2.auth("foobared");

    ClientKillerUtil.killClient(jedis2, "disconnect");

    jedis.disconnect();
    assertTrue(jedis.getClient().getSocket().isClosed());
  }
```

jedis.disconnect() never throws JedisConnectionException. ;(

@allanwax Could you produce it?

allanwax:
I can see if I can produce it.

All I can say for now is that I've seen the situation happen where the socket is never closed because during the close, closing of either the input or output stream failed and the existing code just throws the exception rather than finishing up.

allanwax:
While I'm looking into this, I also suggest trying the change in #847 and the following in Jedis.java (below).  But I haven't tried this out yet.  The idea is to close everything down _then_ return the resource.  But I don't know right now if that is the right order to do stuff in.  Close() calls disconnect() and with the code I outlined in #847 it will attempt to close input, output, socket always.

```java
@Override
	public void close() {
		try {
			client.close();
		}
		finally {
			if (dataSource != null) {
				if (client.isBroken()) {
					this.dataSource.returnBrokenResource(this);
				}
				else {
					this.dataSource.returnResource(this);
				}
			}
		}
	}
```

allanwax:
Is this a proper test?

```java
package test;

import java.net.InetAddress;
import java.security.SecureRandom;
import java.util.ArrayList;
import java.util.HashMap;
import java.util.HashSet;
import java.util.List;
import java.util.Map;
import java.util.Map.Entry;
import java.util.Random;
import java.util.Set;
import java.util.concurrent.Callable;
import java.util.concurrent.Future;
import java.util.concurrent.TimeUnit;
import java.util.concurrent.atomic.AtomicInteger;

import org.apache.log4j.ConsoleAppender;
import org.apache.log4j.Level;
import org.apache.log4j.Logger;

public class JedisBlitz {
	private static Logger log;
	private static int THREAD_COUNT = 1000;
	private static final ExecutorService executor = Executors.newFixedThreadPool(THREAD_COUNT);

	private static String redisHost = null;
	private static JedisCluster jc = null;

	private static AtomicInteger activeCount = new AtomicInteger(0);

	static {
		try {
			redisHost = InetAddress.getByName("test2").getHostAddress();
			Set<HostAndPort> jedisClusterNodes = new HashSet<HostAndPort>();
			jedisClusterNodes.add(new HostAndPort(redisHost, 18000));
			jc = new JedisCluster(jedisClusterNodes);
		}
		catch (Exception e) {
			log.fatal("Could not get address for redis host");
			System.exit(1);
		}
	}

	protected static ThreadLocal<Random> random = new ThreadLocal<Random>() {
		@Override
		public Random initialValue() {
			return new SecureRandom();
		}
	};

	static {
		Logger rootLogger = Logger.getRootLogger();
		if (!rootLogger.getAllAppenders().hasMoreElements()) {
			rootLogger.setLevel(Level.INFO);
			rootLogger.addAppender(new ConsoleAppender(new org.apache.log4j.EnhancedPatternLayout("%d{yyyy-MM-dd HH:mm:ss.SSS} [%t; %C{1}] %-5p -- %m%n")));
		}

		log = Logger.getLogger(JedisBlitz.class);

		Runtime.getRuntime().addShutdownHook(new Thread(new Runnable() {
			// Clean up at exit
			@Override
			public void run() {
				log.info(JedisBlitz.class.getSimpleName() + " shutdown");

				executor.shutdownNow();
				jc.close();
			}
		}));
	}

	static int loopSize = 10000;

	private static void pumpData() {
		try {
			Map<String, String> entries = new HashMap<>();
			for (int i = 0; i < loopSize; i++) {
				String key = "XXX:" + random.get().nextInt(100000);
				//String key = "{" + runID + "}" + random.get().nextInt(1000000);
				entries.put(key, key);
			}

			try {
				log.info("Start to pump data.  Active pumpers=" + activeCount.incrementAndGet());
				for (Entry<String, String> entry : entries.entrySet()) {
					String key = entry.getKey();
					String value = entry.getValue();

					try {
						jc.set(key, value);
					}
					finally {
						jc.expire(key, 120);
					}
				}

				log.info("Pumped " + loopSize + " keys. Active pumpers=" + activeCount.decrementAndGet());

				//				for (Entry<String, String> entry : entries.entrySet()) {
				//					String key = entry.getKey();
				//					String value = entry.getValue();
				//					String v = jc.get(key);
				//					if (!value.equals(v))
				//						overwrites++;
				//				}
			}
			catch (Exception e) {
				log.error(e);
			}
			finally {}
		}
		finally {

		}
	}

		ArrayList<Callable<Boolean>> callableList = new ArrayList<>();
		for (int i = 0; i < THREAD_COUNT; i++) {
			callableList.add(new Callable<Boolean>() {
				@Override
				public Boolean call() throws Exception {
					pumpData();
					return Boolean.TRUE;
				}
			});
		}

		List<Future<Boolean>> futures = null;
		try {
			log.info("InvokeAll()");
			futures = executor.invokeAll(callableList, 1, TimeUnit.MINUTES);
		}
		catch (Exception e) {
			log.fatal("invokeAll() failed", e);
			System.exit(1);
		}

		log.info("Start cancelling randomly");
		int f = 0;
		for (Future<Boolean> future : futures) {
			f++;
			if (random.get().nextInt(THREAD_COUNT) <= (THREAD_COUNT * 0.1)) {
				log.info("Cancelling a task; #" + f);
				future.cancel(true);
			}
			try {
				Thread.sleep(100L);
			}
			catch (Exception e) {
				// ignore
			}
		}

		log.info("Checking futures");
		for (Future<Boolean> future : futures) {
			Boolean bool = null;
			try {
				bool = future.get();
			}
			catch (Exception e) {
				log.fatal("future.get() failed", e);
			}
			if (bool == null || future.isCancelled()) {
				log.error("pump data failed");
			}
		}

		log.info("Exiting");
		System.exit(0);
	}
}

```

HeartSaVioR:
@allanwax Thanks for following up. It isn't a junit test form but I'll give it a try.
Your comment about Jedis.close() seems valid. 
When there's a JedisConnectionException on disconnect(), Jedis seems not take care of Jedis instance to invalid to Pool. It's leaked resource.
@marcosnils Please confirm @allanwax comments to verify if I'm misunderstanding. Thanks!

allanwax:
I'm not sure of the change to close() in Jedis.  But I am _sure_ of the change in disconnect().

The current logic only close()s the connections if the datasource is null.  The new disconnect() will shut down the connections and throw an exception if there is one, guaranteeing in the process that in/out/socket get closed.  But do we always want to close()?  Is returning the resource back to the pool with open connections what we want?  Maybe.

disconnect() sets isBroken if there is an issue and that value is check upon return to the pool if you close first.

More thoughts needed.

HeartSaVioR:
@allanwax @marcosnils 
I'm Sorry for confusing, but current Jedis.close() is valid. We shouldn't call discard() on pooled resource.

allanwax:
Thanks for clearing that up.  I was suspicious of the change to close I mentioned anyway.  However the change to disconnect() is the right thing to do.

marcosnils:
@HeartSaVioR @allanwax I've just went through this and I agree with you to change the `disconnect`. I'm thinking of a simple way to test it. If we can't come up with a proper test then we can merge as It's a very specific scenario

allanwax:
Looks like even my proposed code needs to be tweaked to care of the case where some of the stuff is closed but not others.  *Note that closing the socket also closed input/output*

```java
public boolean isConnected() {
    return socket != null && socket.isBound() && !socket.isClosed() && socket.isConnected()
        && !socket.isInputShutdown() && !socket.isOutputShutdown();
  }
```
If the input is shutdown (or never opened) but the socket is open, this returns false and we never close the socket.

Maybe we need (formatting sucks here):
```java
public void disconnect() {
try {
        if (socket != null && !socket.isClosed())  {
            try {
                    socket.close();
            }
            catch (Exception e) {
                 broken = true;
                 throw new JedisConnectionException(e.fillInStacktrace());
             }
             finally {
                 socket = null;
             }
       }
}
catch (Exception e1) {
   // something happened when trying to check the socket status
   broken = true; // ??? needed here ???
   throw new JedisConnectionException(e1);
}
```
and potentially `socket` should be declared `volatile` since multiple threads may try this, but the locks may prevent this, so it doesn't matter.

HeartSaVioR:
http://stackoverflow.com/questions/3428127/what-is-the-difference-between-closing-input-outputstream-and-closing-socket-dir

According to Javadoc, there's no chance that InputStream is closed but relevant Socket is open.
InputStream, OutputStream, Socket are coupled and whole things become closed if we close anything. 

{code}
  public void disconnect() {
    if (isConnected()) {
      try {
        inputStream.close();
        if (!socket.isClosed()) {
          outputStream.close();
          socket.close();
        }
      } catch (IOException ex) {
        broken = true;
        throw new JedisConnectionException(ex);
      } finally {
        IOUtils.closeQuietly(inputStream);
        if (!socket.isClosed()) {
          IOUtils.closeQuietly(outputStream);
          IOUtils.closeQuietly(socket);
        }
      }
    }
  }
{code}

If it runs normally, outputStream.close() and socket.close() are never called.

HeartSaVioR:
I've changed sequence of closing IO resource, OutputStream first.

marcosnils:
@HeartSaVioR LGTM!.

HeartSaVioR:
Merged into master and 2.7 branch. Thanks all!

allanwax:
Again, why not just flush and close the socket, taking care of everything.

allanwax:
If we are _absolutely_ going to close the socket at this point, disconnect is as simple as:
```java
 public void disconnect() {IOUtils.closeQuietly(socket);}
```
and no exception will ever be thrown

HeartSaVioR:
@allanwax 
Sorry I forgot your comment. You're right. As I stated again, only thing we should close is socket.
I'll reflect it.

allanwax:
And if you want to be super-paranoid in the case where disconnect is never called (perhaps due to an error) and the object is dropped, add a finalizer to call close.  The socket will stay open until a GC happens where memory is needed, but in a _real_ system the finalizers will eventually be called.  I know that people are against finalizers and in most cases I agree but as previously mentioned, sockets are one of the few things that actually lives in the kernel and takes up resources there.

HeartSaVioR:
http://stackoverflow.com/questions/158174/why-would-you-ever-implement-finalize

allanwax:
I've already lived through this.  I don't think other people have and it's a bitch to debug. If you don't close sockets you'll eventually run out of them.  Other than kernel resources, you don't need finalizers.

PR #949, opened 2015-04-12
Add test case for #945

marcosnils:


marcosnils:
@HeartSaVioR can u please review?

HeartSaVioR:
@marcosnils 
I copied your test into current master and see 'success'.
Redis sends 'ASK' from node1 to node3 instead of 'MOVED' so it doesn't call renewSlotCache().
(I ran test 5 times and all succeed.)

Btw, I found other issue, we're not sure about first value of getNodes() since it is HashMap and it's generated per each getNodes() call.
So ```jc.getClusterNodes().get("127.0.0.1:7380").close();``` doesn't make sense.

I'm not sure how we can force 'MOVED', and new issue I've stated. When first value of getNodes() is node1 or node3, we cannot close Pool since we should send command there.

tl;dr. I'm not sure we can make unit test accordingly.

marcosnils:
@HeartSaVioR I've updated the PR, can you check now?

Thx!

HeartSaVioR:
@marcosnils 
Great work about triggering MOVED instead of ASK!
But I still think we should clarify (without hard-coding) what node to close JedisPool, or comment it if it's hard to clarify.

marcosnils:
@HeartSaVioR I've added a comment before closing the pool to clarify the test. Do you think it's more understandable now?

HeartSaVioR:
@marcosnils 
Please check below thing and comment.

```
public void renewSlotCache() {
    for (JedisPool jp : cache.getNodes().values()) {
      Jedis jedis = null;
      try {
        jedis = jp.getResource();
        cache.discoverClusterSlots(jedis);
        break;
        // try next nodes but don't raise exception
      } finally {
        if (jedis != null) {
          jedis.close();
        }
      }
    }
  }
```

In test, node2 is the first value of cache.getNodes().values(), which isn't guaranteed. (It's a HashMap.)
It seems that test is having a great luck to avoid closing node1 and node3 which node's JedisPool shouldn't be closed. (Maybe you gave a try and found this combination.)
In other words, if we change our cluster environment, first value of cache.getNodes().values() can be changed, which makes test invalid.

marcosnils:
@HeartSaVioR I see what you're saying. What if we change the test like this: 

```java
  @Test
  public void testAvoidConnectionExceptionOnRenewSlot() throws InterruptedException {
    //We close first node connection so we ensure it's the same that renewCacheSlots uses
    jc.getClusterNodes().values().iterator().next().close();
    int gamma_slot = JedisClusterCRC16.getSlot("gamma");
    node1.clusterDelSlots(gamma_slot);
    node3.clusterDelSlots(gamma_slot);
    node3.clusterAddSlots(gamma_slot);
    JedisClusterTestUtil.waitForClusterReady(node1, node2, node3);
    try {
      jc.get("gamma");
      fail("renewCacheSlots shouldn't throw JCE");
    }
  }
```

That way we'll always make sure that we're closing the **same** connection independently of the cluster configuration.

HeartSaVioR:
@marcosnils 
What if first node connection is node1 or node3? Then test always fails. It's not that we want.
We can let test pass when first node connection is node1 or node3, but then test may always pass with that situation.
To tell the truth, this test is limited, so it may be better to discard test or describe limitation.
But I'm also fine to go on with adding defensive logic that when first node connection is node1 or node3, test always pass.

marcosnils:
@HeartSaVioR if connection order changes is because we either add or remove a node from the cluster tests. If that is the case then pretty much all JedisCluster tests will fail because slots and nodes are hardcoded and manually used pretty much everywhere in JedisClusterTest.

I understand and agree that the test is limited, but as long as the test cluster is not modified we shouldn't experience any problems

HeartSaVioR:
@marcosnils I agree about current state of tests.
Then I think it's better to add assertion so that test always fails on first node is node1 or node3 with assertion, not JedisConnectionException. It'll prevent confusion about test.

marcosnils:
@HeartSaVioR I've added a new assertion and comments to clarify the test. Check it out and let me know your comments

marcosnils:
@HeartSaVioR seems like we've run in the situation that you previously described. Travis is failing because for the java8 tests the cluster organizes the slots differently. I'll check if there's a way to improve this otherwise I'll discard the test.

HeartSaVioR:
@marcosnils Is it in progress?

marcosnils:
@HeartSaVioR I give up. Let's go ahead and merge #945

PR #950, opened 2015-04-16
Configure connection timeout in JedisPool and JedisSentinelPool

maxd:
I have separated timeout (`timeout` variable) to connection timeout (`connectionTimeout` variable) and read timeout (`soTimeout` variable) because in some cases need to specify short connection timeout and long read timeout.

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisFactory.java
+++ b/src/main/java/redis/clients/jedis/JedisFactory.java
@@ -15,33 +15,32 @@ import redis.clients.util.JedisURIHelper;
  */
 class JedisFactory implements PooledObjectFactory<Jedis> {
   private final AtomicReference<HostAndPort> hostAndPort = new AtomicReference<HostAndPort>();
-  private final int timeout;
+  private final int connectionTimeout;
+  private final int soTimeout;
   private final String password;
   private final int database;
   private final String clientName;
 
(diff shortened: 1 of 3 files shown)


maxd:
One build job has been failed but other three is passed. So, I don't think that this is related to my changes.

HeartSaVioR:
LGTM.

marcosnils:
@maxd thanks for the contribution. I'm wondering if we should also change `JedisCluster` and `Jedis` constructors to allow multiple timeouts as well. 

@HeartSaVioR what do you think?

HeartSaVioR:
@marcosnils 
Jedis constructors already have conn timeout and sock timeout.
JedisCluster constructors don't have it. I'm sure that it's better to have it. :)

marcosnils:
@HeartSaVioR I didn't notice about `Jedis`. I guess adding it to `JedisCluster` would be the best as you said.

HeartSaVioR:
@marcosnils Adding connection timeout to JedisCluster requires this PR, and I can take care of it later.
Please review it first. :)

marcosnils:
@HeartSaVioR  LGTM!

HeartSaVioR:
Merged into master and 2.7 respectively.

PR #955, opened 2015-04-20
Adds binary zcount to pipeline and transaction. Fixes #954

### LINKED ISSUE #954: zcount(byte[] key, byte[] min, byte[] max) not available on PipelineBase and Transaction.
would be nice if `PipelineBase` and `Transaction` would expose the binary variant of `zcount` as offered by `BinaryJedis`.

marcosnils:


marcosnils:
As it's a trivial PR I'll merge once test passes

PR #970, opened 2015-05-04
Added a unit test case

nvharikrishna:
Changes:
1. Added one unit test case to JedisPool. Testing if it is allowing returning of same object twice.
2. Remove unused import from Connection.java
3. Ran 'make format' from command line. It adjusted few documentation comments.
Hope this small PR helps :)

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryJedis.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedis.java
@@ -980,8 +980,8 @@ public class BinaryJedis implements BasicCommands, BinaryJedisCommands, MultiKey
    * <b>Out-of-range indexes</b>
    * <p>
    * Indexes out of range will not produce an error: if start is over the end of the list, or start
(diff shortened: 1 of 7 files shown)


marcosnils:
@nvharikrishna thanks for the contribution. We've previously deprecated `returnXXX` methods of Jedis in favor of using `jedis.close()` directly. I believe we've missed the `returnResourceObject` from JedisPool. 

What about making that method protected and encourage users only to return resources by calling `jedis.close()`?. Having multiple ways of doing the same thing it's quite confusing.

@HeartSaVioR any comments?

HeartSaVioR:
Thanks for contribution @nvharikrishna but I agree with @marcosnils.
We just missed changing Pool.returnResourceObject() to protected. It's better to fix it and after then PR becomes invalid.

HeartSaVioR:
With #971 and #972 it is not an issue now.

PR #971, opened 2015-05-04
Deprecate Pool.returnResourceObject()

HeartSaVioR:
We missed this from #912 and #913.
I'll make another PR to change visibility to protected.


marcosnils:
@HeartSaVioR :+1:

HeartSaVioR:
Merged into master and 2.7 respectively. Thanks @marcosnils for reviewing!

PR #973, opened 2015-05-05
BUG FIX: release connection before throw

yinmingjun:
I review the JedisClusterCommand.java's runWithRetries code, find it throw exception before release connection.

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisClusterCommand.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterCommand.java
@@ -119,14 +119,16 @@ public abstract class JedisClusterCommand<T> {
 
       return execute(connection);
     } catch (JedisConnectionException jce) {
-      if (tryRandomNode) {
-        // maybe all connection is down
-        throw jce;
-      }
-
+      
+      //release connection before throw exception
       releaseConnection(connection, true);
       connection = null;
-
+      
+      if (tryRandomNode) {
+      // maybe all connection is down
+        throw jce;
+      }
+      
       // retry with random connection
       return runWithRetries(key, redirections - 1, true, asking);
     } catch (JedisRedirectionException jre) {


marcosnils:
I was looking at this and I felt a code smell. It doesn't feel right to have the `releaseConnection` everywhere. IMHO we should only use the `releaseConnection` in the `finally` blocks and move the `broken` checking inside the same method. 

I also see that sometimes we're setting the connection to `null` (https://github.com/xetorthio/jedis/blob/master/src/main/java/redis/clients/jedis/JedisClusterCommand.java#L145) and sometimes we don't (https://github.com/xetorthio/jedis/blob/master/src/main/java/redis/clients/jedis/JedisClusterCommand.java#L89) which also makes me thing that this can be improved.

I was thinking that we could do something like: 

```java
  private void releaseConnection(Jedis connection) {
    if (connection != null) {
      if (connection.getClient().isBroken()) {
        connectionHandler.returnBrokenConnection(connection);
      } else {
        connectionHandler.returnConnection(connection);
      }
      connection = null;
    }
  }
```

And then just call `releaseConnection` in all `finally` blocks. I've made the changes locally and tests pass but I'd like @HeartSaVioR to review it as we worked in JedisCluster also. 

@yinmingjun what do you think?

marcosnils:
I've crafted a PR here: #974

HeartSaVioR:
@marcosnils Maybe we can simplify these logics to use Jedis.close(). 
No need to check broken, we just need to check it is null or not.

marcosnils:
@HeartSaVioR you're right `jedis.close()` is already aware of the connection state. I'll update my PR per your recommendations

marcosnils:
@HeartSaVioR done, you can check it out in #974

yinmingjun:
Hi, @marcosnils 

I just have a first view of this code, and my intuition feeling we should release broken connection before throw exception. 

I agree with should release connection in finally block. That is a beautiful way. The key is this exception is aware by Client class.  Otherwise we return an invalid connection into pool.

The way that work around with exception it's more simple. When we got a exception, then drop a connection from pool.

marcosnils:
@yinmingjun I've created PR #974 with the proposed changes. If we release the connection in the `finally` block then we're sure that connections will be released before the client traps the exception. I guess that change fixes the scenario you were proposing right?

HeartSaVioR:
With #974 we seems to resolve this issue, so now I'm closing this issue.
@yinmingjun If you feel something is wrong please reopen this issue. Thanks for your effort!

yinmingjun:
I take the code from  master (#974):
  
private void releaseConnection(Jedis connection) {  
       if (connection != null) {  
                  connection.close();  
                  connection = null;  
       }  
     }  
================================
No one feel something wrong? or this code is designed to call mutiple times?
@marcosnils  @marcosnils

marcosnils:
@yinmingjun this method is called multiple times from the `JedisClusterCommand` class. I don't quite understand your question....

yinmingjun:
@marcosnils  change parameter connection to null is no sense. It can't change the connection's value that call releaseConnection's connection. 
that is:
releaseConnection(conn); //connection's value is change to null, conn still it's old value.

HeartSaVioR:
Yes, actually I know it but it is too trivial and have no effect so I didn't mention it.

marcosnils:
@yinmingjun you're right. As it's a trivial change I guess we can leave it this way

yinmingjun:
I had viewed the code about jedis connection and client under the hood, know that jedis‘s connection now aware connection state by intercept exceptions.  It is a right way. Thanks for your work!
@marcosnils  @HeartSaVioR

HeartSaVioR:
Yeah, it was just misunderstanding. :)

PR #977, opened 2015-05-06
Introduce SetFromList for efficient Set result

itugs:
Hi, I face a similar issue with https://github.com/xetorthio/jedis/issues/907.
This PR is my solution to eliminate overhead of Set creation from List.

The core idea is introducing SetFromList class. For this we can
* Eliminate Set creation overhead including both memory and time
* Fast set iteration by backed ArrayList

Here is result. 
You can see test code [HERE](https://gist.github.com/itugs/2fb948f0a340e8ff9a29) 
```
LinkedHashSet creation Timing avg : 86.44 ms
SetFromList creation Timing avg : 0.0 ms

LinkedHashSet iteration Timing avg : 8.0 ms
SetFromList iteration Timing avg : 2.55 ms
```


src/main/java/redis/clients/collections/SetFromList.java
```diff
@@ -0,0 +1,102 @@
...
+
+  public String toString() {
+    return list.toString();
+  }
+
+  public int hashCode() {
+    return list.hashCode();
+  }
+
+  public boolean equals(Object o) {
```
HeartSaVioR: It would be better to check o is null, since we can meet NPE at Line 74 or 75.

src/test/java/redis/clients/jedis/tests/collections/SetFromListTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+    // contains
+    assertTrue(cut.contains("B"));
+    assertFalse(cut.contains("A"));
+
+    cut.add("C");
+    cut.add("D");
+
+    // containsAll
+    cut.containsAll(cut);
```
HeartSaVioR: It seems not affect whole tests. Maybe you missed something to compare, or just missed assertion.

HeartSaVioR:
I've commented some notes, but LGTM overall.
Thanks for your effort!

Btw, it's trade off between speed of creation / iteration and speed of membership check / deletion.
If we really need to provide O(1) on contains() we should stick HashSet. 
But our method signature don't promise it, and IMO users may use return value to iterate more than checking membership, so it would be fine to change it.

@xetorthio @marcosnils 
I'd like to hear your opinion. Thanks!

HeartSaVioR:
Please note that it doesn't check duplication on constructor. It heavily relies on Redis response, so we should use this to only proper places.

src/test/java/redis/clients/jedis/tests/collections/SetFromListTest.java
itugs: Oops, I missed assert statement.

src/main/java/redis/clients/collections/SetFromList.java
itugs: Yes I should do that.

itugs:
@HeartSaVioR
Thank you for you review.

For that trade off(creation/iteration vs. membership/deletion) you mentioned, that's my assumtions.
I write it down here.

1. Iteration of Set is dominant than membership check. eg) deserialize byte[] to specific type
2. Deletion is rare
3. If membership check is big problem, then user can create HashSet again. Although it looks little complicate, but overall cost is same as before.

HeartSaVioR:
Yes. Important thing is "We don't promise that it returns HashSet.". So focusing Set's basic features is enough.

Btw, maybe it would be better to change its name to more descriptive way, like SetFromNonDuplicateList or SetFromUniqueElementsList, or more proper name.
My last concern is improper usage of SetFromList. It never checks origin list, so we should explain its limitation, and maybe comment couldn't be enough.

src/main/java/redis/clients/jedis/BinaryJedis.java
```diff
@@ -2209,7 +2203,11 @@ public Long zcount(final byte[] key, final byte[] min, final byte[] max) {
   private Set<Tuple> getBinaryTupledSet() {
     checkIsInMulti();
     List<byte[]> membersWithScores = client.getBinaryMultiBulkReply();
-    Set<Tuple> set = new LinkedHashSet<Tuple>();
+    if (membersWithScores.size() == 0) {
+      return Collections.emptySet();
+    }
+    Set<Tuple> set = new LinkedHashSet<Tuple>(
+        CollectionUtils.mapCapacity(membersWithScores.size() / 2));
```
marcosnils: @itugs I don't totally understand this. Why do we need to calculate an "optimal" capacity depending on the growth if the resulting map will never grow?. We already have the final map size from the `membersWithScores` variable, so why just instantiating a `LinkedHashSet<Tuple>(membersWithScores.size())` isn't enough?. We know for sure that the map will be that size and it will not grow.

itugs: @marcosnils I do that to avoid resizing because some JDKs do recalculate by loadFactor.
http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/7-b147/java/util/HashMap.java?av=f#203

But because Oracle JDK use power of two size, my intention was wrong.
Just do `new LinkedHashSet<Tuple>(membersWithScores.size() / 2, 1.0)` is enough.
` / 2` is neccessary because a Tuple consume two elements of `membersWithScores`.

itugs:
@HeartSaVioR As naming we can change the class name but as you said someone can still use it improper way. So I think naming is not a big deal but I wanna listen other collaboratos' opinions.

src/main/java/redis/clients/jedis/BinaryJedis.java
marcosnils: @itugs now it's better. Go ahead and change the initialization to `new LinkedHashSet<Tuple>(membersWithScores.size() / 2, 1.0)`. We can remove the CollectionUtil class also

itugs: @malomalo I added a commit for that. Thanks!

marcosnils:
@itugs @HeartSaVioR PR LGTM!. 

@HeartSaVioR please give a last review and feel free to merge!.

HeartSaVioR:
@marcosnils How about its name? I commented below earlier, and want to know your opinion. Thanks!

> Btw, maybe it would be better to change its name to more descriptive way, like SetFromNonDuplicateList or SetFromUniqueElementsList, or more proper name.
My last concern is improper usage of SetFromList. It never checks origin list, so we should explain its limitation, and maybe comment couldn't be enough.

marcosnils:
@HeartSaVioR IMHO the class name it's fine. Maybe we can clarify it's limitations through javadoc and make it an InnerClass from BinaryJedis so nobody can use it incorrectly outside Jedis. As long as we control merges then we shouldn't encounter any proble,s 

If we make it an inner class then we should see how to test it as it won't be accessible from the tests.

HeartSaVioR:
I'm OK to leave as it is when we can't find a way to test inner (private) class with javadoc. 
Or we can make class package private and test with same package, but I also think it doesn't look good.

I'll give it a try some more, and merge as it is if all things seems not good.

itugs:
@HeartSaVioR One quick way to test private inner class is by reflection.
It is impossible to cast instance as `SetFromList`, by casting to `Set` testing itself is possible.
(And yes, it is not so pretty.)
```
Constructor<?> con = Class.forName("redis.clients.jedis.BinaryJedis$SetFromList").getDeclaredConstructor(List.class);
con.setAccessible(true);

Set cut = (Set)con.newInstance(new ArrayList<String>());
```

As @marcosnils said, the collaboratos can review and control PRs, I don't think this is necessary.

HeartSaVioR:
@itugs That class can be used outside of Jedis. I talked with Marcos and he agree we would be better to hide class.

HeartSaVioR:
@itugs Maybe I misunderstood you. We're agreed to hide SetFromList, but test matters now. 
If you mean test is not necessary, I'm OK.

itugs:
@HeartSaVioR OK, than I'll hide `SetFromList` to inside of `BinaryJedis` and fix my test case.

itugs:
@HeartSaVioR @marcosnils I move `SetFromList` to inside of BinaryJedis.

marcosnils:
@HeartSaVioR @itugs I like it much better now. LGTM.

HeartSaVioR:
LGTM, I'll merge it.
@itugs Thanks for you effort!

itugs:
Thank you!

Issue #981, opened 2015-05-12
Attemp read InputStream in connection is closed/reseted case for reasoning


itugs:
This is related with #906

When Redis close client connection, it write ERR Resp message to client.
It is worth if we can read it and use as exception message for reasoning.

error case example
```
09:17:27.740553 IP localhost.53053 > localhost.6379: Flags [P.], seq 1:8191, ack 1, win 12759, options [nop,nop,TS val 404996303 ecr 404994609], length 8190
E. 2..@.@............=..0....9mX..1..'.....
.#...#.1*1048578  <---- exceed limitation
$5
HMSET
$14
RedisHmsetTest
$18
100001569758228599
$18

09:17:27.740797 IP localhost.6379 > localhost.53053: Flags [P.], seq 1:48, ack 8191, win 12503, options [nop,nop,TS val 404996303 ecr 404996303], length 47
E..cg.@.@..............=.9mX0.:...0..W.....
.#...#..-ERR Protocol error: invalid multibulk length  <-- protocol error

... omitted ...

09:17:27.740865 IP localhost.6379 > localhost.53053: Flags [R], seq 238644616, win 0, length 0
E..(..@.@..............=.9m.....P.......  <-- RST
```

marcosnils:
Close by #985

PR #985, opened 2015-05-14
Attempt to read RESP Errors message

itugs:
Following #981.

Attempt to read RESP Errors message from input stream when IOException occurs during `sendCommand`.
I need some reviews about introducing `Protocol.readLine(inputStream)` method.
I made it at `Protocol` class for consistency of read/write operation but there is a chance of unintended usage of it. Another option is directly using like `inputStream.readLine()` at `Connection.sendCommand` method.

Exception message in stack trace
* AS-IS
```
redis.clients.jedis.exceptions.JedisConnectionException: java.net.SocketException: Broken pipe
	at redis.clients.jedis.Connection.sendCommand(Connection.java:107)
Caused by: java.net.SocketException: Broken pipe
	at java.net.SocketOutputStream.socketWrite0(Native Method)

```

* TO-BE
```
redis.clients.jedis.exceptions.JedisConnectionException: -ERR Protocol error: invalid multibulk length
	at redis.clients.jedis.Connection.sendCommand(Connection.java:107)
Caused by: java.net.SocketException: Broken pipe
	at java.net.SocketOutputStream.socketWrite0(Native Method)
```

### DIFF
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -104,8 +104,6 @@ public class Connection implements Closeable {
        * before close connection. We try to read it to provide reason of failure.
        */
       try {
-        // Set soTimeout 100ms to avoid long blocking
-        socket.setSoTimeout(100);
         String errorMessage = Protocol.readErrorLineIfPossible(inputStream);
         if (errorMessage != null && errorMessage.length() > 0) {
           ex = new JedisConnectionException(errorMessage, ex.getCause());
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -8,7 +8,6 @@ import java.net.SocketException;
 import java.util.ArrayList;
 import java.util.List;
 
-import redis.clients.jedis.Protocol.Command;
 import redis.clients.jedis.exceptions.JedisConnectionException;
 import redis.clients.jedis.exceptions.JedisDataException;
 import redis.clients.util.IOUtils;
@@ -17,6 +16,8 @@ import redis.clients.util.RedisOutputStream;
 import redis.clients.util.SafeEncoder;
 
 public class Connection implements Closeable {
+
+  private static final byte[][] EMPTY_ARGS = new byte[0][];
 
   private String host = Protocol.DEFAULT_HOST;
   private int port = Protocol.DEFAULT_PORT;
@@ -88,24 +89,26 @@ public class Connection implements Closeable {
     return sendCommand(cmd, bargs);
   }
 
+  protected Connection sendCommand(final ProtocolCommand cmd) {
+    return sendCommand(cmd, EMPTY_ARGS);
+  }
+
   protected Connection sendCommand(final ProtocolCommand cmd, final byte[]... args) {
     try {
       connect();
       Protocol.sendCommand(outputStream, cmd, args);
       return this;
     } catch (JedisConnectionException ex) {
-      // Any other exceptions related to connection?
-      broken = true;
-      throw ex;
-    }
-  }
-
-  protected Connection sendCommand(final ProtocolCommand cmd) {
-    try {
-      connect();
-      Protocol.sendCommand(outputStream, cmd, new byte[0][]);
-      return this;
-    } catch (JedisConnectionException ex) {
+      try {
+        // Try to read single line for error message from Redis
+        String errorMessage = Protocol.readLine(inputStream);
+        if (errorMessage != null && errorMessage.length() > 0) {
+          // if possible replace exception
+          ex = new JedisConnectionException(errorMessage, ex.getCause());
+        }
+      } catch (Exception e) {
+        // Ignore
+      }
       // Any other exceptions related to connection?
       broken = true;
       throw ex;
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -99,15 +99,23 @@ public class Connection implements Closeable {
       Protocol.sendCommand(outputStream, cmd, args);
       return this;
     } catch (JedisConnectionException ex) {
+      /*
+       * When client send request which formed by invalid protocol, Redis send back error message
+       * before close connection. We try to read it to provide reason of failure.
+       */
       try {
-        // Try to read single line for error message from Redis
-        String errorMessage = Protocol.readLine(inputStream);
+        // Set soTimeout 100ms to avoid long blocking
+        socket.setSoTimeout(100);
+        String errorMessage = Protocol.readErrorLineIfPossible(inputStream);
         if (errorMessage != null && errorMessage.length() > 0) {
-          // if possible replace exception
           ex = new JedisConnectionException(errorMessage, ex.getCause());
         }
       } catch (Exception e) {
-        // Ignore
+        /*
+         * Catch any IOException or JedisConnectionException occurred from InputStream#read and just
+         * ignore. This approach is safe because reading error message is optional and connection
+         * will eventually be closed.
+         */
       }
       // Any other exceptions related to connection?
       broken = true;
--- a/src/main/java/redis/clients/jedis/Protocol.java
+++ b/src/main/java/redis/clients/jedis/Protocol.java
@@ -117,6 +117,15 @@ public final class Protocol {
     throw new JedisDataException(message);
   }
 
+  public static String readErrorLineIfPossible(RedisInputStream is) {
+    final byte b = is.readByte();
+    // if buffer contains other type of response, just ignore.
+    if (b != MINUS_BYTE) {
+      return null;
+    }
+    return is.readLine();
+  }
+
   private static String[] parseTargetHostAndSlot(String clusterRedirectResponse) {
     String[] response = new String[3];
     String[] messageInfo = clusterRedirectResponse.split(" ");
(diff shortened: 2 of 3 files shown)


HeartSaVioR:
I'm afraid that we can't read a line from InputStream when (socket / read) timeout has been occurred and it could be blocked.
1.

HeartSaVioR:
IMO we have to answer these questions to make it really works.

1. Could we peek InputStream and read bytes which are current available?
2. Is it better to show response which may be partial?
3. Is it better to show partial response which doesn't start with -ERR?

itugs:
@HeartSaVioR 
Yes, I think so, I'm testing about availability of InputStream. I'll add some commits within after few days.
Number 2/3, how about to check `\r\n`? REDIS protocol spec said Redis RESP Error ends with `\r\n`.

HeartSaVioR:
@itugs OK, to wrap it up, when JedisConnectionException occurs 
1. Check availability of InputStream, and read available "bytes" (and convert to String)
2. Check that it starts with -ERR. Use origin JedisConnectionException if it doesn't.
3. Check that it contains "\r\n". If it does, pick that line and use it for wrapping JedisConnectionException's message.

itugs:
@HeartSaVioR Good. Let's try.

marcosnils:
@itugs @HeartSaVioR have thought about a way to test this?. I think there might be a way of doing it. I'll give it a thought and comment here.

HeartSaVioR:
@marcosnils Currently we already know various ways how to let Redis disconnects connection. :)

marcosnils:
@HeartSaVioR yeahh but this case is different because redis needs to send the error before disconnecting. So I guess it's going to be a tricky one but I believe it can be done.

HeartSaVioR:
@marcosnils We can use Redis limitation which @itugs found at networking.c though it can be changed. Maybe it would be hard to change.

itugs:
@HeartSaVioR @marcosnils At first I'll use that limitation and searching for other cases.
Would you let me know if there are any existed test cases for broken write connection due to server RST?

One another point I need to test is read timeout when try to read RESP Error message, 
but think behavior of Redis will 1) send RESP Error message, 2) close connection.
So we could think that RESP Error message is always present in input buffer. 
If then there is no need to test timeout. How do you think?

HeartSaVioR:
@itugs You can refer PublishSubscribeCommandTest.handleClientOutputBufferLimitForSubscribeTooSlow() but it's slightly complicated cause we need to settle precondition (change some configurations to Redis).

itugs:
@HeartSaVioR Thanks, I'll check it.

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -88,24 +89,26 @@ protected Connection sendCommand(final ProtocolCommand cmd, final String... args
...
       return this;
     } catch (JedisConnectionException ex) {
+      try {
+        // Try to read single line for error message from Redis
+        String errorMessage = Protocol.readLine(inputStream);
+        if (errorMessage != null && errorMessage.length() > 0) {
+          // if possible replace exception
+          ex = new JedisConnectionException(errorMessage, ex.getCause());
+        }
+      } catch (Exception e) {
```
marcosnils: Why / When can we get an exception here?. What's the purpose of ignoring it?

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -196,6 +196,10 @@ public static Object read(final RedisInputStream is) {
     return process(is);
   }
 
+  public static String readLine(RedisInputStream inputStream) {
```
marcosnils: What's the purpose of this new method? Can't we just do inputStream.readLine() in https://github.com/xetorthio/jedis/pull/985/files#diff-99ff5b9f483253e37476a66788c13ec6R104

marcosnils:
@itugs @HeartSaVioR I came up with something that tests it:

```java
  @Test
  public void getErrorAfterConnectionReset() throws IOException {
	class TestConnection extends Connection {
		@Override
		protected Connection sendCommand(ProtocolCommand cmd, byte[] ...args) {
			return super.sendCommand(cmd, args);
		}
		
	}
	
	TestConnection conn = new TestConnection();
	byte[][] params = new byte[1024 * 1024 +1][100];
	
	try {
		conn.sendCommand(Command.HMSET,params);
		assertEquals("-ERR Protocol error: invalid multibulk length", jce.getMessage());
		return;
	}
	fail("Should throw exception");
  }
```

I'd like your opinions

itugs:
@marcosnils @HeartSaVioR I improve codes and add test cases.
@marcosnils Your first question is commented in codes.
Second one, the reason of new method of `Protocol` class is written at my initial PR comment.
In shortly, I made it for consistency. I thought `Protocol` class is where have responsibilities of read/write.

During this PR I feel that better location of `process*/read*/send*` is inside of `Connection` because these methods are tightly coupled with `Input/OutStream`.

marcosnils:
@itugs thanks for the updates. I've made a second review to the PR and I still have some suggestions / questions. I'll comment the code shortly

itugs:
@marcosnils Any updates for this?

src/test/java/redis/clients/jedis/tests/ConnectionTest.java
```diff
@@ -48,4 +50,32 @@ public void checkCloseable() {
     client.connect();
     client.close();
   }
-}
\ No newline at end of file
+
+  @Test
+  public void getErrorAfterConnectionReset() throws Exception {
+    brokenReason(new byte[1024 * 1024 + 1][0], "ERR Protocol error: invalid multibulk length");
+    brokenReason(new byte[1][512 * 1024 * 1024 + 1], "ERR Protocol error: invalid bulk length");
```
marcosnils: I don't think we should test two scenarios here. We'e not testing redis behaviour, we just want to test that Jedis handles connection reset with remaining body correct. With just one test case it's more than enough to validate that code.

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -88,24 +89,34 @@ protected Connection sendCommand(final ProtocolCommand cmd, final String... args
...
       return this;
     } catch (JedisConnectionException ex) {
+      /*
+       * When client send request which formed by invalid protocol, Redis send back error message
+       * before close connection. We try to read it to provide reason of failure.
+       */
+      try {
+        // Set soTimeout 100ms to avoid long blocking
+        socket.setSoTimeout(100);
+        String errorMessage = Protocol.readErrorLineIfPossible(inputStream);
```
marcosnils: Do we really need the `readErrorLineIfPossible` method? Can't we just call `Protocol.read(inputStream)` and catch the JedisDataException and convert it into a JedisConnectionException?. I believe the code in the `readErrorLineIfPossible` is pretty much like what `Protoco.read` does when an error is returned.

marcosnils:
@itugs sorry for the delay. I've made some comments to your changes. Let me know if something needs clarification.

src/main/java/redis/clients/jedis/Connection.java
marcosnils: Just to test I've changed this code to something like this and test passes:

```java
  */
      try {
        // Set soTimeout 100ms to avoid long blocking
        socket.setSoTimeout(100);
        Protocol.read(inputStream);
      } catch (JedisDataException jde) {
    	  throw new JedisConnectionException(jde.getMessage());
      } catch (Exception e) {
        /*
         * Catch any IOException or JedisConnectionException occurred from InputStream#read and just
         * ignore. This approach is safe because reading error message is optional and connection
         * will eventually be closed.
         */
      }
```

I believe it's much cleaner and it avoids duplicated code. What do you think?

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -88,24 +89,34 @@ protected Connection sendCommand(final ProtocolCommand cmd, final String... args
...
+      Protocol.sendCommand(outputStream, cmd, args);
       return this;
     } catch (JedisConnectionException ex) {
+      /*
+       * When client send request which formed by invalid protocol, Redis send back error message
+       * before close connection. We try to read it to provide reason of failure.
+       */
+      try {
+        // Set soTimeout 100ms to avoid long blocking
+        socket.setSoTimeout(100);
```
marcosnils: I'm also wondering about this parameter. Did you come up with the scenario where the code was hanging due to this?. I've removed the `setSoTimeout` option and it doesn't hang for me.

src/main/java/redis/clients/jedis/Connection.java
itugs: I didn't use `Protocol.read(inputStream)` because it throw expection.
At this context read error message from InputStream is normal case, so your approach is work but not properly expose intention.
Though there is duplication between `processError` and `readErrorLineIfPossible` but I thought it's worth.
What do you think?

src/main/java/redis/clients/jedis/Connection.java
itugs: This was for safety. 
I rechecked Redis codes and confirmed that Redis ensure `REDIS_CLOSE_AFTER_REPLY` at input buffer processing phase. So this is not necessary.

src/test/java/redis/clients/jedis/tests/ConnectionTest.java
itugs: OK I got it.

marcosnils:
@itugs once you make the changes we can merge!.

itugs:
@marcosnils I updated PR. Thanks.

HeartSaVioR:
LGTM.

marcosnils:
LGTM too!

HeartSaVioR:
Marcos merged this into master, 2.7, 2.8 branches.

itugs:
Thanks!

PR #993, opened 2015-05-20
Fixes #992. 

### LINKED ISSUE #992: Jedis should not throw an exception on close()
This is about Jedis version 2.7.2.

The class `Jedis` implements `java.io.Closeable`.  The Java doc for the `Closeable.close()` clearly states "If the stream is already closed then invoking this method has no effect.".  However, this is not true with `Jedis.close()`.  An IllegalStateException is wrapped into a JedisException and is thrown to the caller, which is a surprise.  I would suggest Jedis swallow the IllegalStateException thrown by GenericObjectPool instead of passing it to the caller.

marcosnils:
Jedis throws an exception calling close multiple times if resource was obtained from a pool

### DIFF
--- a/src/main/java/redis/clients/jedis/Jedis.java
+++ b/src/main/java/redis/clients/jedis/Jedis.java
@@ -3157,6 +3157,7 @@ public class Jedis extends BinaryJedis implements JedisCommands, MultiKeyCommand
       } else {
         this.dataSource.returnResource(this);
       }
+      dataSource = null;
     } else {
       client.close();
     }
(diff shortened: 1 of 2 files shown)


HeartSaVioR:
I'm afraid it is not idempotent since Jedis.close() closes connection when we call Jedis.close() twice in a pool environment.

marcosnils:
@HeartSaVioR you're right. Any suggestions about what to do about this?.

Shall we catch `IllegalStateException` and

PR #1000, opened 2015-05-22
Fix race condition in RedisClusterCommand

atty303:
This is fix for below problem that happened in my environment.

1-a. borrow `connection A` from `pool A`.
1-b. execute on `connection A` throws JedisConnectionException.
1-c. make recursive call with tryRandomNode.
2-a. borrow `connection B` from `pool B`.
2-b. execute on `connection B` throws JedisRedirectionException because `pool B` is not slot owner.
2-c. make recursive call for `MOVED` handling.
3-a. try to borrow `connection C` from `pool A`.  But we still holds `connection A`. It's never returned to pool before `connection C` is borrowed. for instance pool's maxTotal is 1, `connection C` couldn't borrow.

If release a connection before make recursive call, this problem is gone.

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisClusterCommand.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterCommand.java
@@ -123,9 +123,17 @@ public abstract class JedisClusterCommand<T> {
         throw jce;
       }
 
+      // release current connection before recursion
+      releaseConnection(connection);
+      connection = null;
+
       // retry with random connection
       return runWithRetries(key, redirections - 1, true, asking);
     } catch (JedisRedirectionException jre) {
+      // release current connection before recursion or renewing
+      releaseConnection(connection);
+      connection = null;
+
       if (jre instanceof JedisAskDataException) {
         asking = true;
         askConnection.set(this.connectionHandler.getConnectionFromNode(jre.getTargetNode()));
@@ -146,7 +154,6 @@ public abstract class JedisClusterCommand<T> {
   private void releaseConnection(Jedis connection) {
     if (connection != null) {
       connection.close();
-      connection = null;
     }
   }
 


xetorthio:
Nice catch!
Seems like made the tests to fail.
Also we should add a test to this before merging.

HeartSaVioR:
Great catch! We should concentrate while implementing recursion... Thanks!
@atty303 Could you make an unit test representing your deadlock scenario?

marcosnils:
@atty303 just come with a test that reproduces your case

```java
    @Test
    public void testRetryRaceCondition() throws InterruptedException {
        config.setMaxTotal(1);
        JedisCluster jc = new JedisCluster(jedisClusterNode, config);

        Jedis j = jc.getClusterNodes().get("127.0.0.1:7380").getResource();
        ClientKillerUtil.tagClient(j, "DEAD");
        ClientKillerUtil.killClient(j, "DEAD");
        j.close();

        node3.clusterSetSlotMigrating(15363,JedisClusterTestUtil.getNodeId(node2.clusterNodes()));
        node2.clusterSetSlotImporting(15363,JedisClusterTestUtil.getNodeId(node3.clusterNodes()));

        try {
            //This connection has been killed
            jc.get("test");
        }catch (JedisConnectionException jce) {
            System.out.println("Here");
            //Will trigger a move to previous node
            jc.get("51");
        }
    }

```

Here's the code for the ClientKillUtil new helper function: 

```java
public static void tagClient(Jedis jedis, String name) {
    jedis.clientSetname(name);
  }
```

Can you go ahead and verifies that your code fixes that scenario?

Thx!

marcosnils:
Hmmm I've just realized that my test is not entirely deterministic. There are several scenarios here which are very complex and I believe need different testing strategies (if possible)

1) The first scenario to test that JedisCluster actually releases the connection if JCE occurs. This is the test for that scenario

```java
   @Test
    public void testReturnConnectionOnJedisConnectionException() throws InterruptedException {
        config.setMaxTotal(1);
        JedisCluster jc = new JedisCluster(jedisClusterNode, config);

        Jedis j = jc.getClusterNodes().get("127.0.0.1:7380").getResource();
        ClientKillerUtil.tagClient(j, "DEAD");
        ClientKillerUtil.killClient(j, "DEAD");
        j.close();

        jc.get("test");
    }
```

Here's the code for the `tagClient` method:

```java
  public static void tagClient(Jedis jedis, String name) {
    jedis.clientSetname(name);
  }
```

If you run this test without @atty303 patch the code will hang waiting as `runWithRetries` will not be able to retry the operation as the pool ran out of connections. 

2) This is more complex to test because `Jedis` strategy when JCE occurs is to try the operation against a random connection and It's impossible to know beforehand which connection it will choose.

marcosnils:
Just realized that we don't need to test the complex scenario I've described in the 2nd point before. 

If we add the following test:

```java
  @Test(expected = JedisClusterMaxRedirectionsException)
  public void testReturnConnectionOnRedirection() {
    config.setMaxTotal(1);
    JedisCluster jc = new JedisCluster(jedisClusterNode,0, 2, config);

    //This will cause an infinite redirection between node 2 and 3
    node3.clusterSetSlotMigrating(15363, JedisClusterTestUtil.getNodeId(node2.clusterNodes()));
    jc.get("e");
  }
```

Then we'll be testing that upon redirections Jedis releases the connection successfully. With the two specific tests I've provided we handle all possible connection release scenarios

HeartSaVioR:
@marcosnils 
Amazing tests!
I confirmed that your tests fail on current master, and succeed on master + current PR.
Then how can we handle your tests? Can you merge this PR and apply your tests later?

marcosnils:
@HeartSaVioR I was hoping @atty303 could add the tests but seems like he might not be around. I'll open a new PR with the complete changeset

marcosnils:
Closed in favor of #1005

atty303:
Sorry for late response.
Thank you for write tests and merge. @marcosnils

PR #1001, opened 2015-05-23
modified for Redis single node support.

ran-jit:
The issue is, Using protected access specifier we can't return the jedis resources from our custom redis cache util class in single node jedis pool creation. So, I have changes this from protected to public for both cluster and single node jedis pool creation supportability.

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisPool.java
+++ b/src/main/java/redis/clients/jedis/JedisPool.java
@@ -100,13 +100,13 @@ public class JedisPool extends JedisPoolAbstract {
     return jedis;
   }
 
-  protected void returnBrokenResource(final Jedis resource) {
+  public void returnBrokenResource(final Jedis resource) {
     if (resource != null) {
       returnBrokenResourceObject(resource);
     }
   }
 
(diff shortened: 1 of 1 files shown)


marcosnils:
@ran-jit can you provide a snippet of why you need to change the method visibility?. I didn't understand what you're trying to do

ran-jit:
Hello marcosnils,

http://www.programcreek.com/java-api-examples/index.php?api=redis.clients.jedis.JedisPool

Please look at the example in the above link, for set/reterive data into single redis master (non cluster), we need a public visibility of those methods.

marcosnils:
@ran-jit code is outdated now if you call `jedis.close()` connection will be returned to the pool automatically. There's no need to use those methods anymore outside Jedis

marcosnils:
You can check our tests or benchmarks https://github.com/xetorthio/jedis/blob/master/src/test/java/redis/clients/jedis/tests/benchmark that they don't use those public methods anymore.

ran-jit:
Thanks @marcosnils.

PR #1002, opened 2015-05-23
Add missing set command to JedisCluster.

### LINKED ISSUE #999: JedisCluster should  add a method : set(final String key, final String value, final String nxxx) {
public String set(final String key, final String value, final String nxxx) {
      return new JedisClusterCommand<String>(connectionHandler, maxRedirections) {
          @Override
          public String execute(Jedis connection) {
              return connection.set(key, value, nxxx);
          }
      }.run(key);
  }

marcosnils:
Fixes #999

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryJedisCluster.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedisCluster.java
@@ -86,6 +86,16 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


marcosnils:
This PR applies only to 3.0. For 2.7 and 2.8 I'll provide a different PR.

HeartSaVioR:
@marcosnils We already have #878 for reviewing, and when we merge #878 this PR becomes invalid.
So I'd like you to review #878 first and make a decision which one we choose. :)

marcosnils:
@HeartSaVioR I forgot about #878. You're right.

PR #1008, opened 2015-05-26
Improves performance when joining params. 

### LINKED ISSUE #380: BinaryClient joinParameters doesn't use System.arraycopy
It strikes me as odd that BinaryClient#joinParameters loops over the parameters, when it would be easier (and likely more performant) to use System.arraycopy:

```java
byte[][] result = new byte[rest.length + 1][];
result[0] = first;
System.arraycopy(rest, 0, result, 1, rest.length);
return result;
```

marcosnils:
Closes #380


HeartSaVioR:
Great job for finding a missed spot!

marcosnils:
Merged to master, 2.7 and 2.8

PR #491, opened 2013-12-04
ShardedJedis can take optional parameters for h/s/z scan commands

HeartSaVioR:
Hello.
Jonathan pushed *scan commands to Jedis master branch, it's amazing!

I wish to add some functionalities to new commands.
Since sharding algorithm is based on key, h/s/z scan using option parameters can make sense to ShardedJedis.

---
This change is [Reviewable](https://reviewable.io/reviews/xetorthio/jedis/491)

PR #1022, opened 2015-06-10
add ShardedJedis hscan match params

feilaoda:


HeartSaVioR:
Hi!
It's duplicated to #491, though #491 needs to upmerge to use cursor param as String.
Actually we start discussing to not maintaining ShardedJedis, so I'm not sure #491 is still valid.
Please leave comment to #491 if you think Jedis really needs this feature.

Anyway, thanks for your effort!

PR #1025, opened 2015-06-12
2.8

xsjie:
add CustomizeClusterPipeline.java

### DIFF
--- a/src/main/java/redis/clients/jedis/CustomizeClusterPipeline.java
+++ b/src/main/java/redis/clients/jedis/CustomizeClusterPipeline.java
@@ -1,0 +1,156 @@
+/**
+ * 
+ */
+package redis.clients.jedis;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Queue;
+
+import redis.clients.util.JedisClusterCRC16;
+import redis.clients.util.SafeEncoder;
+
+/**
+ * @ClassName: CustomizeClusterPipeline
+ * @Description: TODO 自定义 CustomizeClusterPipeline ,增加 cluster pipeline 功能
+ * @author xiongshengjie 382202220@qq.com
+ * @date 2015年6月9日 下午3:28:40
+ * 
+ */
+public class CustomizeClusterPipeline extends MultiKeyPipelineBase {
+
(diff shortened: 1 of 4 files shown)


marcosnils:
@xsjie seems like you need to review your PR. This change is including things which are actually in Jedis master branch. Specifically commits https://github.com/xsjie/jedis/commit/669a57d3c87c9a041596fba12f6e47c23554ede7 and https://github.com/xsjie/jedis/commit/e3a5b7400554f2f25907a5dd5a61f424fc65bbc6

Can you please reopen properly?

xsjie:
What meaning？

marcosnils:
@xsjie can you elaborate?

xsjie:
in order to use pipeline on redis cluster model,I add CustomizeClusterPipeline.java, I test use CustomizeClusterPipeline set() 1000000 keys request , use I extends CustomizeClusterPipeline 10-100 Times faster then cluster set()

HeartSaVioR:
Please follow https://groups.google.com/d/msg/jedis_redis/u6j8slokO3E/Dh5Q94TRjJUJ in order to understand why I couldn't introduce Pipeline for JedisCluster.

tl;dr. Two things block me.
1) JedisCluster has to take care of slot moving transparently. 
2. Some of requests can trigger redirection, and if JedisCluster collects MOVED / ASK when collecting responses and requests again to proper node, we already break order of requests.

HeartSaVioR:
Btw, we can still implement batched requests from JedisCluster, please refer  https://groups.google.com/d/msg/jedis_redis/u6j8slokO3E/Uff1p70ZXToJ, though it would be hard to implement.

xsjie:
@HeartSaVioR  when use JedisCluster pipeline, I can try{...} catch(Exception e ){ ..} , I keep the batched is atomic ， if some of request  be redirected , I catch() exception then try ...

HeartSaVioR:
@xsjie 
Seems like you don't understand pipeline mode.
In pipeline mode, each command is sent to Redis, and Redis executes it immediately. Just we delegate to get responses from Redis one by one, and get whole responses at once.
Please don't forget - all requests are executed if it is not inside transaction (I mean MULTI).

Based on this, while pipeline mode, if Redis responses that some of requests need redirecting, it also means other requests are already executed.
If we try to send failed requests again, we break the order. 

Only selectable answer could be "don't retry automatically", which seems not good enough.

PR #1018, opened 2015-06-03
Pool.close() should call destory() to shutdown derived classes properly

HeartSaVioR:
Closes #1017 

Before then, MasterListener threads aren't terminated although JedisSentinelPool.close() is called.

You can confirm that modified JedisSentinelPoolTest.checkCloseableConnections() fails on current master and succeeds on PR.

Please review and comment. Thanks!

PR #1028, opened 2015-06-13
Pool.close() should call destory() to shutdown derived classes properly

### LINKED ISSUE #1017: Pool.close() should call destroy() in order to clean up any threads / resources
Currently, Pool.close() calls closeInternalPool() to just clean up internal pool.
https://github.com/xetorthio/jedis/blob/master/src/main/java/redis/clients/util/Pool.java#L22-L24

It may be sufficient to JedisPool, but not JedisSentinelPool, since it launches MasterListener Threads and only JedisSentinelPool.destroy() shutdowns these threads, leading process to be alive forever.

https://github.com/xetorthio/jedis/blob/master/src/main/java/redis/clients/jedis/JedisSentinelPool.java#L83-L89

It may be great to set MasterListener Threads to daemon so process could be terminated when users forget to close / destroy pool.

HeartSaVioR:
Closes #1017

Before then, MasterListener threads aren't terminated although JedisSentinelPool.close() is called.
I was submitted PR #1018 but it also introduced random test failures, so I discard tests and re-submit.

In addition, I changed MasterListener thread to daemon to avoid MasterListener thread preventing  process to be not terminated. I also fix socket leak from MasterListener.

Please review and comment again. Thanks!

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisSentinelPool.java
+++ b/src/main/java/redis/clients/jedis/JedisSentinelPool.java
@@ -168,6 +168,8 @@ public class JedisSentinelPool extends JedisPoolAbstract {
     for (String sentinel : sentinels) {
       final HostAndPort hap = toHostAndPort(Arrays.asList(sentinel.split(":")));
       MasterListener masterListener = new MasterListener(masterName, hap.getHost(), hap.getPort());
+      // whether MasterListener threads are alive or not, process can be stopped
+      masterListener.setDaemon(true);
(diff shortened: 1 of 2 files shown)


HeartSaVioR:
@xetorthio @marcosnils Could you take a look?

marcosnils:
@HeartSaVioR lgtm!

HeartSaVioR:
@marcosnils Thanks for reviewing!

I merged to master, 2.7, 2.8 branches respectively.

Issue #1035, opened 2015-06-18
Wrap ClastCastException inside of Jedis to JedisConnectionException


HeartSaVioR:
While discussing #1030 , @ichernev and me agreed that wrapping ClassCastException to some familiar Exception is necessary, since users don't know why CCE occurs.
It actually notifies that Jedis instance is in bad state, JedisConnectionException would be fine.

I have other question, maybe root issue of #1030.
Why Jedis instance allows sending operations while Pipeline is opened? IMO It shouldn't. It is completely same to multi.

diamondo25:
ClassCastException*

I've been seeing this error in production for no particular reason. We've been using Jedis in combination with Redis for more than 2 year now, and it seems to increase.

I wonder what I should do when a ClassCastException occurs; is there a way to reset the client's state?

ichernev:
@diamondo25 just recreate the client.

I think I've also seen it after a timeout exception. In general it happens when the client is in an invalid state, so making a new client will for sure fix it.

HeartSaVioR:
@diamondo25 
There's no way to reset broken connection now. 
As @ichernev said, recreate the client (or borrow other instance from Pool) makes sense.

PR #1038, opened 2015-06-25
Implement clear() and Closeable to Pipeline and Transaction

HeartSaVioR:

PR #1040, opened 2015-06-28
Throw exception when trying to use jedis instance in Pipeline

marcosnils:
To call same jedis instance which is in pipline makes no sense as
commands already sent to the pipeline are waiting to be processes. This
will cause inconsistencies whenever trying to execute commands.

A JedisDataException will be raised if client tries to use an instance
which is in Pipeline state. Aditionally `resetState` method can be
called to reset jedis instance and use it again normally.

Fixes something that @HeartSaVioR mentioned in #1035

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryJedis.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedis.java
@@ -1586,9 +1586,12 @@ public class BinaryJedis implements BasicCommands, BinaryJedisCommands, MultiKey
   }
 
   protected void checkIsInMultiOrPipeline() {
-    if (client.isInMulti() || pipeline != null) {
+    if (client.isInMulti()) {
       throw new JedisDataException(
-          "Cannot use Jedis when in Multi or Pipeline. Please use Transation o Pipeline instead.");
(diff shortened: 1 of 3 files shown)


marcosnils:
@HeartSaVioR plz review.

HeartSaVioR:
@marcosnils LGTM. Great Job!

HeartSaVioR:
@marcosnils 
I've merged this into master. 
But I cannot merge this into 2.7, cause build would be failed unless we apply #1038, and it's merged to only 2.8.

HeartSaVioR:
Merged into master and 2.8.
We have to apply #1038 into 2.7 branch before we merge this into 2.7 branch.
@marcosnils What do you think about it? Do we have plan to release 2.7.3 before 2.8.0?

marcosnils:
@HeartSaVioR probably. 
Let's go ahead and merge #1038 and this into 2.7 :smile:

HeartSaVioR:
@marcosnils Complete!

marcosnils:
thx @HeartSaVioR

HeartSaVioR:
Not at all! Thanks for great work. :)

PR #1041, opened 2015-07-01
Introduce toString for Slowlog

Raul-:
This is to resolve issue 334. https://github.com/xetorthio/jedis/issues/334

### DIFF
--- a/src/main/java/redis/clients/util/Slowlog.java
+++ b/src/main/java/redis/clients/util/Slowlog.java
@@ -8,6 +8,7 @@ public class Slowlog {
   private final long timeStamp;
   private final long executionTime;
   private final List<String> args;
+  private static final String COMMA = ",";
 
   @SuppressWarnings("unchecked")
   public static List<Slowlog> from(List<Object> nestedMultiBulkReply) {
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
@Raul- 
Do you have some outputs which print slowlog after applying your PR? I'd like to see actual output. 
Making unit test is more appreciated way, but I wonder it is easy to reproduce slowlog.

Thanks in advance!

Raul-:
Printing slowlog after the PR change: 107,1435827586,3,[SET, foo2, bar2] which is basically "id,timeStamp,executionTime,args".

I am not sure what kind of test would add value since it is a straightforward change. Do you have something particular in mind?

HeartSaVioR:
Not at all. It seems to be enough.
LGTM.
Please wait for other collaborators to review. Thanks in advance!

marcosnils:
LGTM too I re ran the tests so once they pass we can merge

HeartSaVioR:
Thanks @Raul- I merged it to master and 2.8 branches.

Raul-:
Thank you!

PR #1042, opened 2015-07-02
Provides execute method;

violetgo:
Provides execute method;
you can use the Jedis to execute the command like in redis-cli, cover most test case;
i need this method and hope can help some people;


HeartSaVioR:
- Unintended file mode change
- Unintended repos change for build status URL

Please leave only intentional change and post PR. Thanks!

IMO, it can be existed as external utils aside of Jedis. But it does't represent Jedis team's opinion.

PR #1044, opened 2015-07-02
Support setting client name from sentinel pool

tgyurci:


marcosnils:
@tgyurci this change will break backwards compatibility as it changes a current constructor. I guess it's better to create a new constructor with the new field to prevent this situation.

tgyurci:
I have created a new constructor with same arguments and it calls the new one.

marcosnils:
@tgyurci you're right github diff tricked me. 

One another collaborator reviews it we can merge. Thanks for contributing!.

HeartSaVioR:
@tgyurci LGTM. It is already included to JedisPool, so adding it to JedisSentinelPool is totally make sense.

HeartSaVioR:
Thanks @tgyurci for contributing, I merged this to master, 2.7, 2.8 branches respectively.

PR #1045, opened 2015-07-03
Execute function

violetgo:
hi;
Also Provides execute method;
you can use the Jedis to execute the command like in redis-cli, cover most test case;
I need this method and hope can help some people;

and only commit modify file and through travis-ci.org build;

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryClient.java
+++ b/src/main/java/redis/clients/jedis/BinaryClient.java
@@ -279,6 +279,10 @@ public class BinaryClient extends Connection {
     sendCommand(HVALS, key);
   }
 
+  public String execute(final String command){
+    return sendCommand(command);
+  }
+  
   public void hgetAll(final byte[] key) {
     sendCommand(HGETALL, key);
   }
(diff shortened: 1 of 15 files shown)


marcosnils:
@violetgo can you explain please what are you using it for?. I can't find a situation where I'd use this method instead of the already provided Jedis functions.

violetgo:
Well, Jedis provides all of the  basically functions;
But I have a scene;
We are developing an automated test system for QA;
QA can directly edit the tools for the redis initialization and verification on the page;
The QA does not want to increase the cost of learning java;
I hope to provide a page to QA directly on the page to write redis script;
But I don't want to parse the script into the Jedis function;
So I want to have a similar feature;

marcosnils:
@violetgo this won't probably get merged because of two things mainly. 

- It doesn't follow the Jedis way of providing a simple API. We try as much as possible to maintain a simple API were inputs and outputs are typesafe and predictable to users. I can see your `excute` method returns a string; I believe this might bring confusion as there are a bunch of commands where a `string` response is just not the best way of representing its output

- It adds another layer of complexity when maintaining Jedis as we need to change the SimpleProtocolCommand class everytime one command is added to Redis. 

I still believe this would be useful to many users which are in the same situation as you so I'd strongly recommend you to set-up another project, use Jedis as a dependency and release it as a Jedis utility. 
I'd like to see something like "jedis-cli" where you can send command to Redis dynamically.

PR #1052, opened 2015-07-13
compareTo() method for case of equal scores

antonio-tomac:
Reason:
compareTo() method violates it's contract, and it is not transitive

Reprodicing:
		List<Tuple> tuples = new LinkedList<>();
		//creating unsorted list with elements with duplicate scores
		for (int i = 0; i < 1000; i++) {
			tuples.add(new Tuple("elem_" + i, (i % 10) * 1.));
		}
		Collections.sort(tuples);

results in:
 java.lang.IllegalArgumentException: Comparison method violates its general contract!
	at java.util.ComparableTimSort.mergeLo(ComparableTimSort.java:714)
	at java.util.ComparableTimSort.mergeAt(ComparableTimSort.java:451)
	at java.util.ComparableTimSort.mergeCollapse(ComparableTimSort.java:376)
	at java.util.ComparableTimSort.sort(ComparableTimSort.java:182)
	at java.util.ComparableTimSort.sort(ComparableTimSort.java:146)
	at java.util.Arrays.sort(Arrays.java:472)
	at java.util.Collections.sort(Collections.java:155)

I know that Jedis.zrangeByScoreWithScores() returns Set<Tuple> which is already sorted.
My use case when I got this error is getting elements from different ZSETs with different ranges, and merging them into List and then sorting it.


marcosnils:
@antonio-tomac seems fine. Maybe I'd compare first by the score as it's much faster than performing `Array.Equal`. That way we don't need to go through the whole array if the score is already the same. 

What do you think?

antonio-tomac:
Yes, that would be faster

marcosnils:
@antonio-tomac perfect. Go ahead and make the changes. Once another collaborator approves we can merge. 

@HeartSaVioR ?

HeartSaVioR:
@antonio-tomac @marcosnils  
Looks good to me. 
Seems like it can be easily tested. How about creating unit test? It would be same as you already posted.

HeartSaVioR:
After adding unit test I'll merge.

marcosnils:
@HeartSaVioR good catch. Also the unit test doesn't need to create 1000 different elements. By just comparing two tuples should be more than enough.

HeartSaVioR:
@antonio-tomac 
Build failure is not on your test so that you don't need to manage the build issue.
I can see that we may want to apply first 3 commits since last 4 commits are not having changes, am I right?

antonio-tomac:
It seems that there was an issue on Travis CI, since some other test failed with SocketTimeoutException, and couple hours later I simply made some change to re-run tests again and it passed.

HeartSaVioR:
@antonio-tomac Collaborators can re-run tests from Travis CI's webpage so you don't worry about it. :)

HeartSaVioR:
@antonio-tomac 
If you don't mind, could you create pull request which have only first 3 commits? Rebasing current PR is fine for me.

I'm fine to cherry-pick your 3 commits, so let me know if you want to. :)

marcosnils:
@antonio-tomac @HeartSaVioR as soon as commits are fixed we can merge.

antonio-tomac:
@HeartSaVioR Sorry, but I'm not really quite sure how to do that on git.

HeartSaVioR:
I merged into master, 2.7, 2.8 branches respectively.

Thanks @antonio-tomac for great work.

PR #1053, opened 2015-07-13
Set Jedis timeout longer while PipeliningTest

HeartSaVioR:
It was too short (0.5 sec), and it periodically triggers build failure on Travis CI, JDK8.


HeartSaVioR:
Since it is just adjusting timeout value, once Travis CI build pass I'll merge it.

PR #1063, opened 2015-07-17
Expose pool borrow stats.

jayanderson:
I'd like to get at these borrow statics to help inform how we should setup our connection pool.

### DIFF
--- a/src/main/java/redis/clients/util/Pool.java
+++ b/src/main/java/redis/clients/util/Pool.java
@@ -95,7 +95,7 @@ public abstract class Pool<T> implements Closeable {
   }
 
   public int getNumActive() {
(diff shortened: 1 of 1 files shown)


marcosnils:
@jayanderson seems reasonable. I don't believe that tests are needed as it's pretty much straightforward and Mex/Mean pool borrow doesn't seem easy to est. 

@HeartSaVioR what do you think?

HeartSaVioR:
@jayanderson @marcosnils 
Looks great. Since it just uses Commons Pool 2 feature, I think we don't need to make test.

HeartSaVioR:
Merged into master, 2.7, 2.8 branches respectively.
@jayanderson Thanks for contribution!

jayanderson:
@HeartSaVioR, @marcosnils Great thanks!

PR #1077, opened 2015-08-04
Add @Override annotation to missing methods

### LINKED ISSUE #1075: Add @Override annotations to implementations
We have been received similar kinds of issues which points out that there's an implementation but interface is missing.

Maybe we can prevent it by adding @Override annotations to all of implementations as JedisCluster does it now.

marcosnils:
Fixes #1075.

Signed-off-by: Marcos Lilljedahl <marcosnils@gmail.com>

### DIFF
--- a/src/main/java/redis/clients/jedis/Jedis.java
+++ b/src/main/java/redis/clients/jedis/Jedis.java
@@ -1307,14 +1307,14 @@ public class Jedis extends BinaryJedis implements JedisCommands, MultiKeyCommand
    * Return the difference between the Set stored at key1 and all the Sets key2, ..., keyN
    * <p>
    * <b>Example:</b>
-   *
+   * 
    * <pre>
    * key1 = [x, a, b, c]
    * key2 = [c]
    * key3 = [a, d]
    * SDIFF key1,key2,key3 =&gt; [x, b]
    * </pre>
-   *
+   * 
    * Non existing keys are considered like empty sets.
    * <p>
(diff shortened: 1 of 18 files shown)


marcosnils:
This PR will be included from 3.0 on wards it's quite big to make down merge it.

src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -2894,33 +3055,28 @@ public Long pttl(final String key) {
...
   }
 
+  @Override
   public String set(final String key, final String value, final String nxxx) {
     checkIsInMultiOrPipeline();
     client.set(key, value, nxxx);
     return client.getStatusCodeReply();
   }
 
-  public String set(final String key, final String value, final String nxxx, final String expx,
```
HeartSaVioR: @marcosnils :+1: Great work catching this!

HeartSaVioR:
@marcosnils 
I'm wondering we can backport the changes on interfaces to 2.7 and 2.8 to expand the interfaces earlier.
Actually it is missed thing, which can be handled as a bug.
What do you think?

HeartSaVioR:
Forgot to tell you did a amazing work! :100:

HeartSaVioR:
@marcosnils Never mind. I will take care of it.

sazzad16:
@HeartSaVioR @marcosnils I took a stab and downmerged this!

PR #1078, opened 2015-08-04
Add missing methods to interfaces and implementations to 2.7 (and 2.8)

HeartSaVioR:
Taking a look at #1077 , I found that there're some methods which are missed from interfaces.

This PR adds missing methods to interfaces, and also adds implementations if needed.

Btw, I don't add @Override annotation to methods, which could take some time to finish.
If we think we should add this annotation to feel safe, I'm happy to. :)


marcosnils:
@HeartSaVioR LGTM!.

HeartSaVioR:
Merged to 2.7 and 2.8 branches.
@marcosnils Thanks for reviewing!

PR #1080, opened 2015-08-04
Handles JedisException from JedisSentinelPool.initSentinels() (Closes #1036)

### LINKED ISSUE #1036: 'ERR max number of clients reached' in JedisSentinelPool.initSentinels breaks the pool initialization
If, during the sentinel initialization, one of the sentinels rejects the connection attempt with 

    redis.clients.jedis.exceptions.JedisDataException: ERR max number of clients reached

the whole pool initialization fails.

It seems, that in the process of sentinel pool initialization, every sentinel is tried, regardless of whether master has been found or not. And, the initialization only catches JedisConnectionException, while `max number of clients reached` throws JedisDataException. So even if one of three sentinels returns this error during initialization, the whole process fails.

https://github.com/xetorthio/jedis/blob/master/src/main/java/redis/clients/jedis/JedisSentinelPool.java#L145

Changing the catch clause to `JedisException` should fix the problem, but I am not sure that is the right way to fix things.

HeartSaVioR:
Please refer #1036 for more details.

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisSentinelPool.java
+++ b/src/main/java/redis/clients/jedis/JedisSentinelPool.java
@@ -157,8 +157,11 @@ public class JedisSentinelPool extends JedisPoolAbstract {
         master = toHostAndPort(masterAddr);
         log.fine("Found Redis master at " + master);
         break;
-      } catch (JedisConnectionException e) {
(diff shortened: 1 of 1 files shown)


marcosnils:
Merged to master and 2.x branches!.

Issue #978, opened 2015-05-06
Improve javadocs


marcosnils:
Javadocs are mainly a copy/paste of the redis documentation. We need to change them to represent proper java types and exception handling. 

There are also methods which doesn't have any javadoc at all.

kag0:
Just ran into #976 as well and finding the docs lacking.
Whats the status on this issue?

marcosnils:
@kag0 we're working on some important core features of Jedis at this moment and we're only two active collaborators. Any help would be much appreciated :smile:

HeartSaVioR:
I thought about Javadoc a little.

As Marcos said, we're only two active collaborators.
I'm wondering we could perfectly maintain source codes with javadoc. It was already nearly unmaintained, and I think it can't be changed unless we have many active contributors or collaborators.

Btw, we can move javadoc from implementations to interfaces, and use @inheritDoc tag to reduce amount of duplicated comments.
@marcosnils What do you think?

marcosnils:
@HeartSaVioR moving Javadocs to interfaces seems the right way to go. I don't have time to do that kind of work at this moment, hopefully some new collaborator would help :)

HeartSaVioR:
OK. I also think we may just replace the explanations for the command to link to Redis doc, since it is actively maintained with many contributors.

marcosnils:
If we replace javadocs with redis links then users without internet connection won't be able to see the docs (which is kind of sad). Redis docs are stored in a JSON format in the redis page repo. 

Maybe we can annotate the javadocs in some way which allows us to run a simple program which takes the commands description form the redis json doc and replaces it into Jedis. I'm not sure about this exactly but I agree with you that maintaining the documentation is quite messy as we don't have the enough collaborators to do it ATM.

kag0:
javadocs support html in the comments, so you could potentially do something there with the redis docs. Wouldn't fix issues like #976 though.
I wouldn't be opposed to helping to get some native docs in place, I just don't know when I'd have time. At bare minimum it shouldn't be hard to get default javadocs up on gh-pages as a starter.

marcosnils:
@kag0 :+1:

kag0:
@marcosnils can you please create a gh-pages branch (off of master) so I can open a PR into it?

HeartSaVioR:
@kag0 Done.

PR #1083, opened 2015-08-05
Support new ZADD options introduced at Redis 3.0.2

### LINKED ISSUE #1067: New ZADD Options
Redis recently added new options to the [ZADD](http://redis.io/commands/ZADD) command as of 3.0.2.  Is anyone already planning on adding these?  I took a look at it, but it gets a bit hairy because the INCR option changes the return data type... Is it desired to add a new method interface for every combination of options?

HeartSaVioR:
Closes #1067 

We discussed many things from #1067, so please refer that issue for more details.

I didn't add javadoc for added methods for some reasons.  
I'll address what thing prevents me to add javadoc to #978.

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryClient.java
+++ b/src/main/java/redis/clients/jedis/BinaryClient.java
@@ -406,7 +406,7 @@ public class BinaryClient extends Connection {
     sendCommand(ZADD, params.getByteParams(key, toByteArray(score), member));
   }
 
-  public void zaddBinary(final byte[] key, final Map<byte[], Double> scoreMembers) {
+  public void zadd(final byte[] key, final Map<byte[], Double> scoreMembers) {
     ArrayList<byte[]> args = new ArrayList<byte[]>(scoreMembers.size() * 2 + 1);
     args.add(key);
     args.addAll(convertScoreMembersToByteArrays(scoreMembers));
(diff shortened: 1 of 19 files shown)


jamespedwards42:
Looks great.  Thanks for handling all of this, sorry that it had to turn into such a fuss...  Looking forward to the *Param args as well.

src/main/java/redis/clients/jedis/BinaryClient.java
```diff
@@ -399,15 +401,15 @@ public void zadd(final byte[] key, final double score, final byte[] member) {
     sendCommand(ZADD, key, toByteArray(score), member);
   }
 
-  public void zaddBinary(final byte[] key, final Map<byte[], Double> scoreMembers) {
+  public void zadd(final byte[] key, final double score, final byte[] member,
+      final ZAddParams params) {
+    sendCommand(ZADD, params.getByteParams(key, toByteArray(score), member));
+  }
 
+  public void zaddBinary(final byte[] key, final Map<byte[], Double> scoreMembers) {
```
marcosnils: I know it has nothing to do with this PR. But I was looking at the `zaddBinary` method and I was thinking: "why is it called zaddBinary if zadd is already binary?". Then I realized that the difference is that `zaddBinary` accepts multiple score members. 

Maybe we can take the opportunity to rename this method to just `zadd`?. In Client.java the method that calls `zaddBinary` is called `zadd`.

marcosnils:
@HeartSaVioR PR is perfect. Let's see what we want to do with the `zaddBinary` thing and then merge :)

src/main/java/redis/clients/jedis/BinaryClient.java
HeartSaVioR: @marcosnils 
Great finding! Renamed zaddBinary to zadd.
Maybe it was just a workaround if Java compiler or JVM could not distinguish methods if only param's Generic is different.
For now, all tests passed.

marcosnils:
@HeartSaVioR LAGTM (Looks amazingly good to me??) :+1:

HeartSaVioR:
@jamespedwards42 You don't need to be sorry. It was me who can't stand with that decision. :)

HeartSaVioR:
Merged into master and 2.8 branch respectively.
Thanks all!

PR #1090, opened 2015-08-14
Fix logic of refreshing slot table when MOVED redirection occurred.

be-hase:
Currently, When refresh slot map table, Jedis select node(host:port) from beginning of map.  
But Redis cluster's configuration does not propagate at the same time.

So, we should use node which send MOVED redirection at first.
And if fail, we choose randomly.

Please review. :D
thanks.


HeartSaVioR:
@be-hase 
LGTM, Thanks for the great contribution!
@marcosnils Please review and comment. Thanks!

marcosnils:
@HeartSaVioR @be-hase with this approach we'll be re-discovering the node that sent the MOVED twice right?.

marcosnils:
:+1: . Let's merge!.

marcosnils:
Merged to 2.7 and 2.8 accordingly.  Thanks for contributing!

be-hase:
Thanks !! :smile:

PR #1096, opened 2015-08-21
validate the password in JedisFactory constructor

ancanglee:
comment please ref:
https://github.com/xetorthio/jedis/pull/1095

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisFactory.java
+++ b/src/main/java/redis/clients/jedis/JedisFactory.java
@@ -26,7 +26,11 @@ class JedisFactory implements PooledObjectFactory<Jedis> {
     this.hostAndPort.set(new HostAndPort(host, port));
     this.connectionTimeout = connectionTimeout;
     this.soTimeout = soTimeout;
-    this.password = password;
+    if(password != null && ! password.trim().isEmpty()){
+      this.password = password;
+    }else{
+      this.password = null;
+    }
     this.database = database;
     this.clientName = clientName;
   }


HeartSaVioR:
@ancanglee Could you add unit test to check functionality? Thanks in advance!

marcosnils:
Why do we need to validate the password?. Can you please explain which situation you're facing that you need to introduce this validation?

Thanks!.

HeartSaVioR:
@marcosnils @ancanglee 
It is not a validation, but just convert empty string to null.

Actually I got into same issue some month ago.
We should call different constructor when password is present (or not persent), but it is not easy thing for spring xml configuration.
So I have to make another utility factory method to initialize JedisPool.

But I also found that I can set null value to default value instead of empty string. It should help to initialize JedisPool, so we may think it is not necessary to merge this.

@ancanglee 
Could you try this? http://stackoverflow.com/a/17470732/3480122

HeartSaVioR:
Another thought is we can merge this safely, cause redis-cli doesn't allow spaces to "auth" parameter, which implies redis-server doesn't allow spaces to password.

src/main/java/redis/clients/jedis/JedisFactory.java
```diff
@@ -26,7 +26,11 @@ public JedisFactory(final String host, final int port, final int connectionTimeo
     this.hostAndPort.set(new HostAndPort(host, port));
     this.connectionTimeout = connectionTimeout;
     this.soTimeout = soTimeout;
-    this.password = password;
+    if(password != null && ! password.trim().isEmpty()){
```
marcosnils: trim is not necessary as if the string has white spaces, `isEmpty` returns false.
Please fix braces indentation.

marcosnils:
@HeartSaVioR seems reasonable. I left some minor fixes and after they're applied we can merge.

src/main/java/redis/clients/jedis/JedisFactory.java
HeartSaVioR: Maybe it means that If string has only whitespaces, it should be interpreted as null.

marcosnils: @HeartSaVioR guess it makes sense.

HeartSaVioR: One thing I'm thinking about is, redis-cli trims parameter, so when redis-server's password is set to "foobared", I can get OK with typing "auth foobared     ".
I'm wondering it is just a redis-cli's behavior or redis-server's password is trimmed.

ancanglee:
@HeartSaVioR ,@marcosnils. Sorry for this late replying since I'm busy recently. HeartSaVioR has got my purpose. I met the problem in my job.In my dev environment, I don't need redis password, in my product environment, I need use password, so I want my code can work on these two env. I rewrite the JedisPool, add the password judgement. HeartSaVioR give  me  a good solution for  the default null value. thx

marcosnils:
Let's close this to avoid any problems that may appear when trimming the password. As @HeartSaVioR suggested you can set it at the spring level.

ancanglee:
OK.

PR #1098, opened 2015-08-21
Fix infinite loop from sentinel pool when failover occurred (for 2.7)

### LINKED ISSUE #1093: Endless loop happens in JedisSentinelPool.getResource()
Using JedisSentinelPool, after failover, any borrowed instances connected to old master will return to pool. If testOnReturn and testOnBorrow are using default setting(false), thread calling getResource will run into endless loop because old instances borrowed and returned repeatly.

```
import java.util.HashSet;
import java.util.Set;

import redis.clients.jedis.JedisSentinelPool;


public class JedisSentinelPoolTest {
	
	public static void main(String args[]) throws InterruptedException {
		String master = "shard1";
		Set<String> sentinels = new HashSet<String>();
		sentinels.add("10.11.202.186:26380");
		sentinels.add("10.11.202.186:26381");
		JedisSentinelPool pool = new JedisSentinelPool(master, sentinels);
		
		// after borrowed instance from pool
		Jedis jedis = pool.getResource();
		// if failover happen here
		Thread.sleep(60000);
		jedis.close();
		
		// endless loop happen, same jedis instance borrowed and returned repeatly
		jedis = pool.getResource();
		try {
			jedis.set("key", "value");
			e.printStackTrace();
		} finally {
			jedis.close();
		}
		pool.close();
	}
}
```

HeartSaVioR:
Fixes #1093 

* After failover, any borrowed instances connected to old master could be returned to pool
* Thread calling JedisSentinelPool.getResource() will run into infinite loop because old instances borrowed and returned repeatedly
* Fix JedisSentinelPool.getResource() to return instance as broken when instance is not connected to current master

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisPool.java
+++ b/src/main/java/redis/clients/jedis/JedisPool.java
@@ -102,9 +102,10 @@ public class JedisPool extends Pool<Jedis> {
   }
 
   /**
-   * @deprecated starting from Jedis 3.0 this method won't exist. Resouce cleanup should be done
-   *             using @see {@link redis.clients.jedis.Jedis#close()}
+   * @deprecated starting from Jedis 3.0 this method will not be exposed.
+   * Resource cleanup should be done using @see {@link redis.clients.jedis.Jedis#close()}
    */
+  @Override
   @Deprecated
   public void returnBrokenResource(final Jedis resource) {
     if (resource != null) {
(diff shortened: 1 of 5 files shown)


HeartSaVioR:
I'll cherry-pick this to 2.8 when it is labeled as READY TO MERGE.

marcosnils:
@HeartSaVioR amazing. We probably forgot to downmerge :)

marcosnils:
Merged to 2.7 and 2.8 respectively.

PR #1076, opened 2015-08-03
Handle local host reference automatically.

allanwax:
Handle IPV6 local host references.

Issue #1023, opened 2015-06-11
Automatically change localhost references in HostAndPort to the address of the current machine


allanwax:
First, apologies as I'm still sick (today=2015-06-11) so if there are stupid errors below, please forgive.

Many people get caught specifying localhost/127.0.0.1 in configuration values for jedis cluster.  It is fine for jedis.  I myself got caught when I first started using it.  I propose to help by automatically replacing localhost references to the IP address of the local machine.  Thus as information gets passed around it will contain an absolute address rather than a relative one.

Here's the suggested new class.  NOTE `169.254` is a Microsoft special address sometimes used if the machine can't get to the internet.  I.e. the connections window will show a connection to the router but none to the internet.

``` java
package redis.clients.jedis;

import java.net.InetAddress;

public class HostAndPort {
    public static final String LOCALHOST_STR ;

    static {
        try {
            LOCALHOST_STR = InetAddress.getLocalHost().getAddress();
        }
        catch (Exception e) {
            LOCALHOST_STR = "localhost";
        }
    }

    private String host;
    private int port;

    public HostAndPort(String host, int port) {
        this.host = host;
        this.port = port;
    }

    public String getHost() {
        return host;
    }

    public int getPort() {
        return port;
    }

    @Override
    public boolean equals(Object obj) {
                if (obj == null) return false;

        if (obj instanceof HostAndPort) {
            HostAndPort hp = (HostAndPort) obj;
            String thisHost = convertHost(host);
            String hpHost = convertHost(hp.host);
            return port == hp.port && thisHost.equals(hpHost);
        }

        return false;
    }

    @Override
    public int hashCode() {
        return 31 * convertHost(host).hashCode() + port;
    }

    @Override
    public String toString() {
        return host + ":" + port;
    }

    private String convertHost(String host) {
         if (host.equals("127.0.0.1") || host.startsWith("localhost") || host.equals("0.0.0.0") || host.startsWith("169.254"))
             return LOCALHOST_STR;
         else if (host.startsWith("::1") || host.startsWith("0:0:0:0:0:0:0:1"))
            return LOCALHOST_STR;

        return host;
    }
}
```

marcosnils:
@allanwax you already did all the hard work. Why not submitting a PR?.
I agree it's something it will definitely help Jedis users.

allanwax:
[...]

PR #1099, opened 2015-08-24
Allanwax localhost fix

marcosnils:
This PR includes the functionality of #1076 and implements #1023 

@HeartSaVioR @allanwax plz review.


src/main/java/redis/clients/jedis/BinaryJedis.java
```diff
@@ -175,11 +175,11 @@ public String quit() {
   }
 
   /**
-   * Test if the specified keys exist. The command returns the number of keys existed
-   * Time complexity: O(N)
+   * Test if the specified keys exist. The command returns the number of keys existed Time
```
HeartSaVioR: Maybe formatter breaks its intention.

src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -108,8 +108,8 @@ public String get(final String key) {
   }
 
   /**
-   * Test if the specified key exists. The command returns the number of keys existed
-   * Time complexity: O(N)
+   * Test if the specified key exists. The command returns the number of keys existed Time
```
HeartSaVioR: Maybe formatter breaks its intention.

HeartSaVioR:
Overall looks fine.
One thing I'm not sure is actual ip is completely same as loopback. Does it work when redis-server is bound to 127.0.0.1?

allanwax:
'''
One thing I'm not sure is actual ip is completely same as loopback. Does it work when redis-server is bound to 127.0.0.1?
'''
I don't know the answer to this.  The fix I checked in only applies to usage of the class `HostAndPort` which is what people usage to set up a cluster.  I _think_ that the binding will work fine since that is how the outside world gets to the machine, not the setup of the cluster.  But again, I don't know.

I also don't know how it will affect `redis-trib` in creation of the cluster since I don't know that code at all.

marcosnils:
> One thing I'm not sure is actual ip is completely same as loopback. Does it work when redis-server is bound to 127.0.0.1?

Yes, it will work when redis is bound to loopback. The ip that `InetAddress.getLocalHost().getHostAddress()` is a loopback address.

src/main/java/redis/clients/jedis/Jedis.java
marcosnils: Hmm... do we need to fix the formatter?

allanwax: I didn't format HostAndPort before checking it in.  However if we're going to change the formatter, it's much easier to read (at least for me) with 4 spaces instead of two.  When it gets very nested, its more difficult to to, at least for me.

HeartSaVioR: I don't think it becomes a reason to fix the formatter. But if we can fix the formatter to not modify comments, it may be also fine to me.

HeartSaVioR:
@marcosnils OK. Got it.

marcosnils:
I'll merge this locally before merging here as I'm worried about how much impact it will require to downmerge.

marcosnils:
This PR is super hard to merge because it contains lots of "fix" commits and also applies some heavy unnecessary reformatting. I'll make some cleanup and re-open a more friendly mergable one.

marcosnils:
Closed by #1342

Issue #1087, opened 2015-08-11
Feature Request: Configuration option for marking connections as "broken" after a JedisDataException


terrysuereth-glu:
Background information: In our use-case, redis redundancy is accomplished by running a single "master" node and multiple read-only replica nodes.  When a failure occurs on the master, it is demoted from master status (becoming read-only) and one of the replicas is then promoted (gaining writability) to become the new master.  This master switch is flipped by using CNAME name records, i.e., clients connecting to the redis cluster at "redis.host.name" will subsequently be pointed to whichever node is the current master.  (I do not particularly like this setup, but the infrastructure is out of my and my colleagues' hands.)

The problem we encounter in this case is when an issue occurs on the master node that _does not sever active connections._  The master node becomes demoted and read-only, but any clients which were previously connected to it are still connected after the demotion.  (Or, alternatively, a race condition may occur where the client is able to reconnect to this redis node before the demotion event is effective.)  Then when a client attempts a write operation on this node, it is knocked back with a JedisDataException due to the redis node being read-only.

If this was a JedisConnectionException, then the connection would automatically be marked as broken by the catch block in Connection.sendCommand, and the connection resource would be cleaned up once the Jedis object is scoped out.  However, since a JedisDataException won't trigger this broken flag, our client's connection will return to the pool and be re-used - continuing to trigger the same error/exception - indefinitely.

In other words, in our use-case, a **JedisDataException is a _de facto_ connection error.**  We must react to it by closing the connection and creating a new one, which should hit the new master redis node.

The workaround in our use-case is to explicitly catch both JedisConnectionException and JedisDataException, and to use **releaseBrokenResource (now deprecated)** in both cases to mark the connection as broken.  This also makes us unable to use a try-with-resources pattern to initialize and auto-close a Jedis connection, since in the error case, the connection would be closed twice.

I realize, with lament, that this use-case is highly specific and the behavior I want is not generally ideal.  So my feature request is to add a configuration option to JedisPool to enable a "mark connection as broken after a JedisDataException" behavior.  E.g. if a Connection has "tolerateDataExceptions" set to false, a JedisDataException in Connection.sendCommand will cause the Connection to be marked as broken.

HeartSaVioR:
@terrysuereth-glu 
Thanks for reporting!

I skimmed your document, and feel it is some kind of specific use-case, so I wonder we can handle it as general option.

[...]

PR #1101, opened 2015-08-24
Introduces ConnectionBrokenDeterminer

HeartSaVioR:
It is an sketched version of ```custom exception handler for marking connection as broken```, as I stated from #1087 

I'm focusing on showing how to introduce custom exception handler, so some details could be missing.
We can start improving from this version - rename new interface/classes to have better naming, introduce new package, change the way how to setup custom exception handler, or maybe find another ways and rewrite it completely.

ps. Since I'm modifying JedisPoolAbstract, it would be hard to introduce this to 2.8. I'll re-work to make this compatible to 2.x, and post a PR again.

### DIFF
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -28,6 +28,7 @@ public class Connection implements Closeable {
   private int connectionTimeout = Protocol.DEFAULT_TIMEOUT;
   private int soTimeout = Protocol.DEFAULT_TIMEOUT;
   private boolean broken = false;
+  private ConnectionBrokenDeterminer connBrokenDeterminer = new ConnectionBrokenDeterminer();
 
   public Connection() {
   }
@@ -39,6 +40,10 @@ public class Connection implements Closeable {
   public Connection(final String host, final int port) {
     this.host = host;
     this.port = port;
+  }
+
+  public void setConnectionBrokenDeterminer(final ConnectionBrokenDeterminer determiner) {
+    this.connBrokenDeterminer = determiner;
   }
 
   public Socket getSocket() {
@@ -264,9 +269,13 @@ public class Connection implements Closeable {
   protected Object readProtocolWithCheckingBroken() {
     try {
       return Protocol.read(inputStream);
-    } catch (JedisConnectionException exc) {
-      broken = true;
-      throw exc;
+    } catch (RuntimeException e) {
+      if (connBrokenDeterminer.determine(e)) {
+        broken = true;
+        throw new JedisConnectionException(e);
+      } else {
+        throw e;
+      }
     }
   }
 
--- a/src/main/java/redis/clients/jedis/ConnectionBrokenPattern.java
+++ b/src/main/java/redis/clients/jedis/ConnectionBrokenPattern.java
@@ -1,0 +1,5 @@
+package redis.clients.jedis;
+
+public interface ConnectionBrokenPattern {
+  boolean determine(final RuntimeException throwable);
+}
--- a/src/main/java/redis/clients/jedis/JedisPoolAbstract.java
+++ b/src/main/java/redis/clients/jedis/JedisPoolAbstract.java
@@ -6,6 +6,7 @@ import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
 import redis.clients.util.Pool;
 
 public class JedisPoolAbstract extends Pool<Jedis> {
+  public ConnectionBrokenDeterminer connBrokenDeterminer;
 
   public JedisPoolAbstract() {
     super();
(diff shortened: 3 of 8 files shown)


HeartSaVioR:
It also opens a way to resolve #1035 partly since we can introduce new pattern handling CCE.
Why I talk it can resolve #1035 "partly" is that it doesn't cover all kind of type casting.
Currently there're too many paths which incurs type casting, so I want to think about it later.

Spikhalskiy:
It's what I mostly want and talking about in Issue #1252

HeartSaVioR:
@marcosnils @sazzad16 
I think this is worth to revisit. Maybe we would want to do ping-pong instead of checking pattern as @Spikhalskiy pointed out, but I would like to get consensus on adopting this idea.

src/main/java/redis/clients/jedis/JedisPoolAbstract.java
```diff
@@ -6,6 +6,7 @@
 import redis.clients.util.Pool;
 
 public class JedisPoolAbstract extends Pool<Jedis> {
+  public ConnectionBrokenDeterminer connBrokenDeterminer;
```
sazzad16: `public`?

sazzad16: `public`?

src/main/java/redis/clients/jedis/ConnectionBrokenPattern.java
```diff
@@ -0,0 +1,5 @@
+package redis.clients.jedis;
+
+public interface ConnectionBrokenPattern {
+  boolean determine(final RuntimeException throwable);
```
sazzad16: Remove `final`.

sazzad16:
@HeartSaVioR Implementation is goot. I can see the usage and I'm okay with it. But, before taking any decision, let's wait sometime whether @xetorthio and @marcosnils have anything to say.

src/main/java/redis/clients/jedis/ConnectionBrokenPattern.java
HeartSaVioR: OK.

src/main/java/redis/clients/jedis/JedisPoolAbstract.java
HeartSaVioR: Nice finding.

HeartSaVioR:
@sazzad16 
What if we just remove all the complexities and do PING-PONG instead of examining the pattern? Examining the pattern may be error-prone and PING-PONG would be the best way to check the connection.

HeartSaVioR:
@sazzad16 And please note that this patch has been waiting on more than 2 years.

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -264,9 +269,13 @@ protected void flush() {
   protected Object readProtocolWithCheckingBroken() {
     try {
       return Protocol.read(inputStream);
-    } catch (JedisConnectionException exc) {
-      broken = true;
-      throw exc;
+    } catch (RuntimeException e) {
+      if (connBrokenDeterminer.determine(e)) {
+        broken = true;
+        throw new JedisConnectionException(e);
```
sazzad16: `e` is likely to be JedisConnectionException, at least by default. So why not just throw `e` without wrapping?

sazzad16:
@HeartSaVioR PING-PONG is definitely a better idea. But there is a chance that some user may prefer not to do an extra network operation. So either it can't be default or there should be an easy option to skip that.

src/main/java/redis/clients/jedis/Connection.java
HeartSaVioR: There's a chance for `e` to be `JedisDataException`  like #1087, or even `ClassCastException`.

sazzad16: **At this line, by default,** there is no chance `e` would be `JedisDataException` or `ClassCastException`.

By default, `ConnectionBrokenDeterminer` has only `JedisConnectionExceptionAsConnectionBroken` which just checks if provided `RuntimeException` is an instance of `JedisConnectionException`. So it is either `JedisConnectionException` or `JedisNoReachableClusterNodeException`. So, if the condition of `if` is true, for the default `ConnectionBrokenDeterminer`, `e` is `JedisConnectionException` (unless it is cluster mode and every node in cluster is down). And wrapping a most likely `JedisConnectionException` within another `JedisConnectionException` doesn't seem to be a good idea to me.

HeartSaVioR: OK then does it make sense to wrap `e` with JedisConnectionException only when `e` is not an instance of JedisConnectionException? 
Also you may want to be aware that ConnectionBrokenDeterminer is designed as a plug-in, and the purpose of ConnectionBrokenDeterminer is to classify some non-JedisConnectionExceptions as a JedisConnectionException.

HeartSaVioR:
I realized why I didn't pass Jedis instance as parameter to implement PING-PONG idea: the logic should be laid on Connection layer, so Jedis instance is not available.
As an alternative, we may be able to provide implementation class based on Connection which just implements `BasicCommands`, or even just supports `ping()` which looks like to be best-effort.

src/main/java/redis/clients/jedis/Connection.java
sazzad16: Oh! The way I saw is that ConnectionBrokenDeterminer is to help defining the connection as broken. Which, to simply put, is to set the variable `broken = true`. Our ideas are somehow related, but seem to be different.

How is wrapping other exception as JedisConnectionException helpful? What if we don't?

HeartSaVioR: Users could also have try-catch statement in their codebase, and expect connection related exception to be thrown as JedisConnectionException. We have reported too many issues for ClassCastException in Jedis codebase (users may feel really odd and strange) which was due to connection messed up (so the response of Redis is not expected type), and throwing CCE as it is doesn't help for users.

sazzad16: Understood. Thanks!

uglide:

PR #1105, opened 2015-08-28
Fix socket leak on makeObject.

### LINKED ISSUE #1104: Socket leak when commands failed at makeObject()
There's a potential socket leak issue at makeObject() when using JedisPool here:

https://github.com/xetorthio/jedis/blob/master/src/main/java/redis/clients/jedis/JedisFactory.java#L87

After establishing the connection using ```.connect()```, if commands like ```.select()``` failed and exception thrown there, then this socket would hang since there's no way to call ```returnBrokenResource``` or ```close()``` on it in higher level logics. 

If Redis is not configured to close idle connections from server side, such connections would stay idle there.

marcosnils:
Close Jedis connection if error is raised when trying to initialize
a connection in JedisFactory.

Fixes #1104

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisFactory.java
+++ b/src/main/java/redis/clients/jedis/JedisFactory.java
@@ -8,6 +8,7 @@ import org.apache.commons.pool2.PooledObjectFactory;
 import org.apache.commons.pool2.impl.DefaultPooledObject;
 
 import redis.clients.jedis.exceptions.InvalidURIException;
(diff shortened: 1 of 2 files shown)


HeartSaVioR:
LGTM.

HeartSaVioR:
Merged into master, 2.7, 2.8 branches! Thanks @marcosnils!

marcosnils:
:+1: anytime!.

PR #1106, opened 2015-08-30
Support geo commands

HeartSaVioR:
* geoadd, geodist, geohash, geopos, georadius, georadiusbymember
* implementations matrix (2 * 3 = 6 kinds of implementation)
  * binary, string
  * normal, cluster, pipeline
* side updates: move RedisPipeline to proper package

During implementing I realized we can't apply patch to 2.x automatically.
Cherry-picking should be harder than manually copy and pasting and fix conflicts.
(I don't want to work on both things, though.)

Geo commands are still in unstable, and it will be introduced to Redis 3.2. 
I think if we start to work to complete Jedis 3.0.0, we don't need to backport this patch.

@xetorthio @marcosnils 
Please review and comment.
If you think we should go on 2.8 before starting to work on 3.0, please change milestone to 2.8.0.
Thanks in advance!


src/main/java/redis/clients/jedis/BinaryClient.java
```diff
@@ -421,8 +424,7 @@ public void zadd(final byte[] key, final Map<byte[], Double> scoreMembers) {
     sendCommand(ZADD, argsArray);
   }
 
-  public void zadd(final byte[] key, final Map<byte[], Double> scoreMembers,
-      final ZAddParams params) {
+  public void zadd(final byte[] key, final Map<byte[], Double> scoreMembers, final ZAddParams params) {
```
marcosnils: @HeartSaVioR is this format change consequence of `make format`?. If not, every time we run `make format` again it will revert this.

src/main/java/redis/clients/jedis/BinaryClient.java
```diff
@@ -1213,6 +1215,57 @@ public void readonly() {
     sendCommand(READONLY);
   }
 
+  public void geoadd(byte[] key, double longitude, double latitude, byte[] member) {
+    sendCommand(GEOADD, key, toByteArray(longitude), toByteArray(latitude), member);
+  }
+
+  public void geoadd(byte[] key, Map<byte[], GeoCoordinate> memberCoordinateMap) {
```
marcosnils: @HeartSaVioR question: Is there a specific reason you decided to use a `Map<byte, Geocoordinate>` instead of just making it a `Set<GeoCoordinate>` and include the member inside the GeoCoordinate bean?. Reading the `geoadd` command documentation it says the redis stores the coordinates as a sorted set internally. That doesn't mean Jedis should do the same thing, but I was just thinking what would be the difference between using a Map or Set for us.

marcosnils:
@HeartSaVioR made some minor comments which are not critical, overall LGTM!.

src/main/java/redis/clients/jedis/BinaryClient.java
HeartSaVioR: @marcosnils Yes, it's the result of ```make format```.

src/main/java/redis/clients/jedis/BinaryClient.java
HeartSaVioR: @marcosnils 
Yes, geo stores datas to sorted set internally. That's why I uses Map<byte[], GeoCoordinate> to represent multiple members of geoadd.
zadd() also uses Map<byte[], GeoCoordinate> cause score is changeable and only member name is unique. 
Set<GeoCoordinate> allows duplication of same member with different coordinate, and we can't guess which thing will be stored effectively.

HeartSaVioR:
@marcosnils 
Why don't we include ```backporting cluster-revised branch``` into 2.8 and release?
We may need to postpone issues which are not yet finished, but I think backporting cluster-revised branch is ready for review, though it may need to be upmerged.

src/main/java/redis/clients/jedis/BinaryClient.java
marcosnils: @HeartSaVioR this is depending on how we implement the `hashcode` and `equals` on the GeoCoordinate bean. If we add the `member` attribute to that class and we modify the mentioned methods accordingly, we could use Set<GeoCoordinate> as an input to the `geoadd` command. 

I'm just thinking about which would be the simplest API / code. If either using Map or Set.

marcosnils:
@HeartSaVioR sure, let's merge the cluster-revised branch and release. One thing that I was thinking though is that cluster-revised would break backwards compatibility right?. Mainly because JedisCluster interface hierarchy changes. 

Can you confirm this?

src/main/java/redis/clients/jedis/BinaryClient.java
HeartSaVioR: @marcosnils 
1. GeoCoordinate is used by geopos and GeoRadiusResponse. 
It represents only coordinate (point) itself, not including member. So if we change GeoCoordinate to include member, we need another class to cover coordinate.

2. Let's think about GeoCoordinate itself.
Assume we have two GeoCoordinate objects which members are same but coordinates are different. Are two objects same, and it is natural to understand without explanation?

3. It would be minor, but in real world, relation of place name and coordinate is not permanent.

HeartSaVioR:
@marcosnils 
Please refer #1107. From #1107, I don't modify interfaces to accomplish backward compatibility.

marcosnils:
@xetorthio you have two machines now. You shouldn't have an excuse to review this PR :)

xetorthio:
@marcosnils @HeartSaVioR Checked the code and it all seems amazing! Lets get this merge :) Thanks and amazing work!

marcosnils:
Merged to 2.8 respectively. We're good ok to release :tada: :tada: :tada: :tada:

HeartSaVioR:
@xetorthio @marcosnils Great! Thanks for reviewing, and let's release 2.8.0 to the world!

PR #1115, opened 2015-09-10
Add constructors in JedisCluster for just one node

ricardorames:
As there is nodes auto discovering in JedisCluster, I believe to be useful a constructor who receives only one node. Something like this:
```
JedisCluster cluster = new JedisCluster(new HostAndPort("10.10.10.10", 7777));
```

Best regards!


HeartSaVioR:
LGTM.

Issue #1147, opened 2015-11-10
setTimeoutInfinite() in eval()


panchenko:
At the moment eval is implemented like this

``` java
  public Object eval(String script, int keyCount, String... params) {
    client.setTimeoutInfinite();
    try {
      client.eval(script, keyCount, params);
      return getEvalResult();
    } finally {
      client.rollbackTimeout();
    }
  }
```

I am not sure if there are cases when eval execution takes a long time, one of our use cases is "distributed locks" as described in http://redis.io/topics/distlock and eval-ed script is pretty simple

``` lua
if redis.call("get",KEYS[1]) == ARGV[1] then
    return redis.call("del",KEYS[1])
else
    return 0
end
```

Unfortunately sometimes it hangs because of network issues.

In our case definitely continue with a default timeout should be OK.
If someone is doing a lot of computations in redis it would probably make sense to set some different timeout, but running without timeout could result is some vary long waits in case of network failures.

Thoughts?

marcosnils:
@panchenko I have a question. What happens if when you get these network issues you send a `SCRIPT KILL` to redis. Does the application continue processing?

The reason why I'm asking this is because I've checked other clients (including the redis-cli code) and I've seen that none of them actually sets a timeout for the `eval` command. In addition do that, the Redis official docs state the following:

>SCRIPT KILL
This command is the only way to interrupt a long-running script that reaches the configured maximum execution time for scripts. The SCRIPT KILL command can only be used with scripts that did not modify the dataset during their execution (since stopping a read-only script does not violate the scripting engine's guaranteed atomicity). See the next sections for more information about long running scripts.

So I'm wondering if this is really necessary.

panchenko:
I filed this after I've got alerts from tomcat on some stuck threads (i.e. requests taking >5 minutes) with the stack trace like the following:

```
at java.net.SocketInputStream.socketRead0(Native Method)
at java.net.SocketInputStream.read(SocketInputStream.java:152) 
at java.net.SocketInputStream.read(SocketInputStream.java:122)
at java.net.SocketInputStream.read(SocketInputStream.java:108) 
at redis.clients.util.RedisInputStream.ensureFill(RedisInputStream.java:195) 
at redis.clients.util.RedisInputStream.readByte(RedisInputStream.java:40) 
at redis.clients.jedis.Protocol.process(Protocol.java:132) 
at redis.clients.jedis.Protocol.read(Protocol.java:196) 
[...]

PR #1148, opened 2015-11-11
Jedis#eval with the specified timeout

panchenko:
I would suggest this change in order to be specify what timeout for eval, see #1147.
This change is backwards compatible, however I still think infinite timeout is not a good idea.


marcosnils:
Pending on https://github.com/xetorthio/jedis/issues/1147#issuecomment-157783207

sazzad16:
Closing #1147 and this due to inactivity.

sazzad16:
Closed by #2408

PR #1151, opened 2015-11-13

liujg:


liujg:
i sent a new pull request,please check the newest...

PR #1155, opened 2015-11-17

liujg:

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
@@ -2,8 +2,10 @@ package redis.clients.jedis;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


marcosnils:
@liujg is this PR correct?. The files changed only shows a nonsense change.

liujg:
@marcosnils I found All checks have failed,so i just test,but i find a nonsense change,this code can also not compile....why?https://travis-ci.org/xetorthio/jedis/builds/91599783

marcosnils:
@liujg I still don't see this PR changes. This is the change proposed: 

image

liujg:
@marcosnils, I donot known why the code cannot compile success,i just do nothing for testing. So I submmit a no changes codes,but it alse 

Add more commits by pushing to the master branch on liujg/jedis.

Hide all checks
All checks have failed
1 failing check
Details continuous-integration/travis-ci/pr — The Travis CI build failed
This branch is up-to-date with the base branch
Only those with write access to this repository can merge pull requests.

PR #1156, opened 2015-11-18
Malicious code vulnerability - Public static method may expose internal representation by returning array

cyburs:
This pull request is focused on resolving occurrences of findbugs rule findbugs:MS_EXPOSE_REP - “Malicious code vulnerability - Public static method may expose internal representation by returning array”. 
You can find more information about the issue here: https://dev.eclipse.org/sonar/rules/show/findbugs:MS_EXPOSE_REP

### DIFF
--- a/src/main/java/redis/clients/jedis/Protocol.java
+++ b/src/main/java/redis/clients/jedis/Protocol.java
@@ -207,7 +207,7 @@ public final class Protocol {
   }
 
   public static final byte[] toByteArray(final boolean value) {
-    return value ? BYTES_TRUE : BYTES_FALSE;
+    return value ? BYTES_TRUE.clone() : BYTES_FALSE.clone();
   }
 
   public static final byte[] toByteArray(final int value) {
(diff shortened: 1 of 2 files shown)


Chnkr:
:-1: I don't see how this is a issue, considering `jedis` is a library.

marcosnils:
@cyburs as @Chnkr says, Jedis is a library that will be used by `your` application. So end users shouldn't have access to this method to exploit the mentioned vulnerability. Can you please explain how someone would exploit this?.

cyburs:
@marcosnils Even though this is a library, it should ensure that its internal functionality cannot be changed by external code. By returning original array, the internal functionality of the code can be changed as demonstrated in the test case:

Chnkr:
`it should ensure that its internal functionality cannot be changed by external code`

Why? It's a library, if people want to do crazy things, let them be.

marcosnils:
@cyburs I agree with you that it's a vulnerability, but in this case as @Chnkr says, it's only exploitable if the application using Jedis exposes it as such. It's the application responsibility to call the `clone` method on the byte array if you're using the `Protocol.toByteArray`. We won't fix this as it penalizes the library performance for something that it's only exploitable if the application using Jedis does it wrong.

PR #1172, opened 2015-12-03
Added support for binary script eval/evalsha

janssk1:
You still need to add tests. Did not get the testsuite working


marcosnils:
@janssk1 I see that you didn't make PipelineBase implement BinaryScriptingCommands. Is there a reason for that?

marcosnils:
@janssk1 can you please craft the PR against the master branch so we can have the latest Jedis version working?. We'll take care to down merge it to the older versions.

Thx!.

janssk1:
On master, MultiKeyPipelineBase already supports binary script invocations.

marcosnils:
@janssk1 you're right. The proper way to proceed is to downmerge 3a1b3a21631a38241d56e00a22526285c8becfc7 to 2.7 and 2.8 respectively. I'll try to get it done by today, keep watching the thread.

PR #1173, opened 2015-12-04
Support for ssl/tls using defaults or user specified SSLSocketFactory, SSLParameters, and HostnameVerifier.

damiri-ts:
This pull request provides support for SSL/TLS with the option to specify an SSLSocketFactory, SSLParameters, and HostnameVerifier.

I am essentially resuscitating the content from the pull request submitted by RedisLabs (https://github.com/xetorthio/jedis/pull/611/files). The majority of the credit for these changes belongs to them. For my part, I added the ability for the user to explicitly set an SSLSocketFactory, SSLParameters, and HostnameVerifier which should alleviate any concerns about missing or sub-optimal hostname verification when using either Java 6 or 7.

I have tested this with the Azure's Redis cache service.

### DIFF
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -8,6 +8,11 @@ import java.net.SocketException;
 import java.util.ArrayList;
 import java.util.List;
 
+import javax.net.ssl.HostnameVerifier;
+import javax.net.ssl.SSLParameters;
+import javax.net.ssl.SSLSocket;
+import javax.net.ssl.SSLSocketFactory;
+
 import redis.clients.jedis.commands.ProtocolCommand;
 import redis.clients.jedis.exceptions.JedisConnectionException;
 import redis.clients.jedis.exceptions.JedisDataException;
@@ -28,6 +33,10 @@ public class Connection implements Closeable {
   private int connectionTimeout = Protocol.DEFAULT_TIMEOUT;
   private int soTimeout = Protocol.DEFAULT_TIMEOUT;
   private boolean broken = false;
+  private boolean ssl;
+  private SSLSocketFactory sslSocketFactory;
+  private SSLParameters sslParameters;
+  private HostnameVerifier hostnameVerifier;
 
   public Connection() {
   }
@@ -39,6 +48,23 @@ public class Connection implements Closeable {
   public Connection(final String host, final int port) {
     this.host = host;
     this.port = port;
+  }
+
+  public Connection(final String host, final int port, final boolean ssl) {
+    this.host = host;
+    this.port = port;
+    this.ssl = ssl;
+  }
+
+  public Connection(final String host, final int port, final boolean ssl,
+      SSLSocketFactory sslSocketFactory, SSLParameters sslParameters,
+      HostnameVerifier hostnameVerifier) {
+    this.host = host;
+    this.port = port;
+    this.ssl = ssl;
+    this.sslSocketFactory = sslSocketFactory;
+    this.sslParameters = sslParameters;
+    this.hostnameVerifier = hostnameVerifier;
   }
 
   public Socket getSocket() {
@@ -155,6 +181,23 @@ public class Connection implements Closeable {
 
         socket.connect(new InetSocketAddress(host, port), connectionTimeout);
         socket.setSoTimeout(soTimeout);
+
+        if (ssl) {
+          if (null == sslSocketFactory) {
+            sslSocketFactory = (SSLSocketFactory)SSLSocketFactory.getDefault();
+          }
+          socket = (SSLSocket) sslSocketFactory.createSocket(socket, host, port, true);
+          if (null != sslParameters) {
+            ((SSLSocket) socket).setSSLParameters(sslParameters);
+          }
+          if ((null != hostnameVerifier) &&
+              (!hostnameVerifier.verify(host, ((SSLSocket) socket).getSession()))) {
+            String message = String.format(
+                "The connection to '%s' failed ssl/tls hostname verification.", host);
+            throw new JedisConnectionException(message);
+          }
+        }
+
         outputStream = new RedisOutputStream(socket.getOutputStream());
         inputStream = new RedisInputStream(socket.getInputStream());
       } catch (IOException ex) {
--- a/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
@@ -7,6 +7,10 @@ import java.util.Map;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 
+import javax.net.ssl.HostnameVerifier;
+import javax.net.ssl.SSLParameters;
+import javax.net.ssl.SSLSocketFactory;
+
 import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
 
 import redis.clients.util.ClusterNodeInformation;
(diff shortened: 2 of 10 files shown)


marcosnils:
@damiri-ts we really appreciate this contribution. SSL is a feature we wanted to introduce some time ago but we couldn't due to our workload. At a first glance I see that this PR doesn't have any tests, can you think of a way to test this with the current Jedis testing strategy?. 

I'd also like if @HeartSaVioR and @xetorthio can help to review.

damiri-ts:
@marcosnils Thanks for response. I apologize for not having any tests at this stage but I wanted to submit the pull request anyway for two reasons:
1. It was unclear to me from the discussion in the previous pull request whether SSL/TLS support was a feature that the project was interested in.
2. I wanted to have someone at least validate the approach and at least comment on whether the new changes address the issues in the older pull request.
3. The only SSL/TLS enabled server I have to test with is my own Azure Redis cache.

That being said, so long as there is interest in this feature, and consensus that the implementation is acceptable, then I'd be happy to figure out a way to test this in CI.

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -155,6 +181,23 @@ public void connect() {
 
         socket.connect(new InetSocketAddress(host, port), connectionTimeout);
         socket.setSoTimeout(soTimeout);
+
+        if(ssl) {
```
HeartSaVioR: Nit: space between if and (

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -155,6 +181,23 @@ public void connect() {
...
+          if (null == sslSocketFactory) {
+            sslSocketFactory = (SSLSocketFactory)SSLSocketFactory.getDefault();
+          }
+          socket = (SSLSocket) sslSocketFactory.createSocket(socket, host, port, true);
+          if (null != sslParameters) {
+            ((SSLSocket) socket).setSSLParameters(sslParameters);
+          }
+          if((null != hostnameVerifier) &&
```
HeartSaVioR: Nit: space between if and (

damiri-ts:
@HeartSaVioR Thanks for pointing out the missing spaces. These are addressed with commits squashed to keep things nice and tidy.

damiri-ts:
@marcosnils @HeartSaVioR @xetorthio Just wanted to followup with you guys on this pull request. Any thoughts? Do you feel it addresses the shortcomings of the original pull request for ssl/tls support? I can't emphasize enough how important it is to support ssl/tls in production environments, especially when working with a PaaS layer where all services are exposed to the internet.

HeartSaVioR:
@damiri-ts 
Sorry for the long delay.
I'm interested in, but I don't have a chance to learn about SSL/TLS, and any other committers seems to be same. We need to maintain the codebase so at least one of active committer should know about background knowledge and understand codes of pull request.
If you can provide some materials I'm happy to learn.
Thanks in advance!

bbroerman30:
I'm going to have to look at this as well.. I just started working on an ssl enabled version of jedis fir use with my ssl enabled Redis.

damiri-ts:
@HeartSaVioR @bbroerman30 
* For background on SSL/TLS I recommend reading Bulletproof SSL and TLS by Ivan Ristic (https://www.feistyduck.com/books/bulletproof-ssl-and-tls).
* For reference on JSSE, the Java SSL/TLS implementation, see https://docs.oracle.com/javase/6/docs/technotes/guides/security/jsse/JSSERefGuide.html, https://docs.oracle.com/javase/7/docs/technotes/guides/security/jsse/JSSERefGuide.html, and https://docs.oracle.com/javase/8/docs/technotes/guides/security/jsse/JSSERefGuide.html.
* For a discussion on hostname verification see the previous pull request: https://github.com/xetorthio/jedis/pull/611. Especially, https://stackoverflow.com/questions/18139448/how-should-i-do-hostname-validation-when-using-jsse/18174689#18174689 and https://stackoverflow.com/questions/17972658/sslsocket-ignores-domain-mismatch/17979954#17979954.

FWIW this is not the first time I've added SSL/TLS support to Java software. One advantage that Java has in this department is that it comes with a stable and mature SSL/TLS implementation. Rest assured that the changes in this pull request are not re-inventing the wheel, rather they provide SSL/TLS functionality in accordance with Java best practices for creating SSL sockets and performing hostname verification.

Finally, if you're looking for a spare-time contributer to maintain SSL/TLS related functionality I'd be happy to do what's necessary.

HeartSaVioR:
@damiri-ts 
That's great news that you want to contribute to Jedis by maintaining SSL/TLS!

Btw, it seems that what I need to understand is just this code blocks.

```
if (ssl) {
  if (null == sslSocketFactory) {
    sslSocketFactory = (SSLSocketFactory)SSLSocketFactory.getDefault();
  }
  socket = (SSLSocket) sslSocketFactory.createSocket(socket, host, port, true);
  if (null != sslParameters) {
    ((SSLSocket) socket).setSSLParameters(sslParameters);
  }
  if ((null != hostnameVerifier) &&
      (!hostnameVerifier.verify(host, ((SSLSocket) socket).getSession()))) {
    String message = String.format(
        "The connection to '%s' failed ssl/tls hostname verification.", host);
    throw new JedisConnectionException(message);
  }
}
```

Technote materials and clear code block convinces me. I also want to become sponsor of the code for SSL/TLS. LGTM.

Could you please try to craft unit test for SSL/TLS feature? Thanks!

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -105,7 +109,37 @@ public void setNodeIfNotExist(HostAndPort node) {
...
-          connectionTimeout, soTimeout, null, 0, null);
+          connectionTimeout, soTimeout, null, 0, null, false, null, null, null);
+      nodes.put(nodeKey, nodePool);
+    } finally {
+      w.unlock();
+    }
+  }
+
+  public void setNodeIfNotExist(HostAndPort node, boolean ssl) {
+    w.lock();
```
tomfitzhenry: You could replace this method body with:
`setNodeIfNotExist(node, ssl, null, null, null)`

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -105,7 +109,37 @@ public void setNodeIfNotExist(HostAndPort node) {
       if (nodes.containsKey(nodeKey)) return;
 
       JedisPool nodePool = new JedisPool(poolConfig, node.getHost(), node.getPort(),
-          connectionTimeout, soTimeout, null, 0, null);
+          connectionTimeout, soTimeout, null, 0, null, false, null, null, null);
```
tomfitzhenry: You could replace this method body with:
`setNodeIfNotExist(node, false, null, null, null)`

damiri-ts:
@HeartSaVioR In plain English the code would read like this.
1. If the user did not set an sslSocketFactory, then use the system default socket factory set by the JVM
2. Use the available sslSocketFactory to create an ssl socket.
3. If the user specified a set of non null ssl parameters, apply those to the sockets.
4. If the user specified a non null hostname verifier and the hostname verifier did not like the hostname of the peer thrown an exception.

I'll see what I can do about writing a useful unit test. My first thoughts on that task is that the unit test would add little value because the logic itself is fairly straightforward and pretty standard usage of JSSE. An integration test would likely be more useful but the difficulty with that is figuring out how to start an SSL/TLS protected Redis server for the tests. I'd be willing to do that if it is deemed useful.

damiri-ts:
@HeartSaVioR I've been able to create an environment where it is possible to run SSL/TLS integration tests. Essentially, all that is necessary is to configure stunnel as per: http://bencane.com/2014/02/18/sending-redis-traffic-through-an-ssl-tunnel-with-stunnel

With this stunnel configuration for /etc/stunnel/redis-server.conf
```
cert = /etc/stunnel/private.pem
pid = /var/run/stunnel.pid
[redis]
accept = 127.0.0.1:6380
connect = 127.0.0.1:6379
```

An SSL/TLS port of 6380 is setup right alongside the standard 6379 listening port. Then a simple test like this one:
```
public class SslJedisTest extends JedisCommandTestBase {

  @BeforeClass
  public static void setupTrustStore() {
    String trustStoreFilePath = "src/test/resources/truststore.jceks";
    Assert.assertTrue(String.format("Could not find trust store at '%s'.", trustStoreFilePath),
        new File(trustStoreFilePath).exists());
    System.setProperty("javax.net.ssl.trustStore", trustStoreFilePath);
    System.setProperty("javax.net.ssl.trustStoreType", "jceks");
  }

  @Test
  public void connectWithShardInfo() {
    final URI uri = URI.create("rediss://localhost:6380");
    final SSLSocketFactory sslSocketFactory = (SSLSocketFactory) SSLSocketFactory
        .getDefault();
    // These SSL parameters ensure that we use the same hostname verifier used for HTTPS.
    // Note: this options is only available in Java 7.
    final SSLParameters sslParameters = new SSLParameters();
    sslParameters.setEndpointIdentificationAlgorithm("HTTPS");

    JedisShardInfo shardInfo = new JedisShardInfo(uri, sslSocketFactory, sslParameters, null);
    shardInfo.setPassword("foobared");

    Jedis jedis = new Jedis(shardInfo);
    jedis.get("foo");
    jedis.close();
  }
}
```

Passes successfully. If I can work with someone that can modify your current CI build environment to install and configure stunnel, then I can start checking in integration tests for the various SSL/TLS modes of operation. Let me know what you think.

HeartSaVioR:
@damiri-ts @marcosnils 
We may refer https://github.com/mp911de/lettuce to setup stunnel for Travis CI.
I don't have machine for develop for now (maybe during this week), so unfortunately it takes some more days to try.
I'm really appreciated if anyone try it first. :)

marcosnils:
@heartsavior. Don't worry, I'll review it tomorrow

sent from mobile

marcosnils:
I've added travis support as well as some basic SSL tests to #1192. Let's continue the discussion there.

PR #1174, opened 2015-12-04
2.8 pipeline binary

### LINKED ISSUE #1169: Can not invoke binary scripting commands from pipeline
I'm using jedis 2.7.2

I have a script that needs a binary parameter. However, when i invoke the script in a pipeline, jedis does not allow me to pass binary parameters. (Transaction does not implement BinaryScriptingCommands)
If the script is invoked directly on Jedis,without a pipeline, binary parameters are accepted.

marcosnils:
Fixes #1169. 

@HeartSaVioR can you please review?. You only need to check f8f19ba8a8a3336813e9747b6cbd734ff41c1518, the other stuff is just formatting.

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryClient.java
+++ b/src/main/java/redis/clients/jedis/BinaryClient.java
@@ -1304,7 +1304,7 @@ public class BinaryClient extends Connection {
     sendCommand(GEODIST, key, member1, member2, unit.raw);
   }
 
-  public void geohash(byte[] key, byte[]...members) {
+  public void geohash(byte[] key, byte[]... members) {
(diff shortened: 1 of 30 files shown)


marcosnils:
Note: I didn't change the current `Response<String> evalxxx` methods as I didn't want to break backwards compatibility.

HeartSaVioR:
@marcosnils 
LGTM.
I feel we would like to continue our works about 3.0.0 since we can't fix several issues because of semver.

marcosnils:
@HeartSaVioR I agree. We should release 3.0 soon so we can start working on a stable version and leave behind all this complicated merge stuff.

PR #1178, opened 2015-12-08
Fix connection leak in JedisSlotBasedConnectionHandler

### LINKED ISSUE #1168: JedisCluster leaks connections
When redis cluster gets problems (e.g., some nodes crashed and are being restarted), we get some exceptions like this:

redis.clients.jedis.exceptions.JedisException: Could not return the resource to the pool
        at redis.clients.util.Pool.returnBrokenResourceObject(Pool.java:100)
        at redis.clients.jedis.JedisPool.returnBrokenResource(JedisPool.java:112)
        at redis.clients.jedis.JedisPool.returnBrokenResource(JedisPool.java:12)
        at redis.clients.jedis.Jedis.close(Jedis.java:3355)
        at redis.clients.jedis.JedisClusterCommand.releaseConnection(JedisClusterCommand.java:160)
        at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:154)
        at redis.clients.jedis.JedisClusterCommand.runBinary(JedisClusterCommand.java:59)
        at redis.clients.jedis.BinaryJedisCluster.get(BinaryJedisCluster.java:98)
        at com.jnetx.redis.RedisClient.get(RedisClient.java:92)
        at com.jnetx.perffw.RedisPerf$ReadWriteThread.doOperation(RedisPerf.java:78)
        at com.jnetx.perffw.ThreadBase.run(ThreadBase.java:173)
Caused by: java.lang.IllegalStateException: Invalidated object not currently part of this pool
        at org.apache.commons.pool2.impl.GenericObjectPool.invalidateObject(GenericObjectPool.java:643)
        at redis.clients.util.Pool.returnBrokenResourceObject(Pool.java:98)
        ... 10 more

The reason why it appears is currently unclear to us.

Then we get some exceptions like this:

redis.clients.jedis.exceptions.JedisDataException: LOADING Redis is loading the dataset in memory
	at redis.clients.jedis.Protocol.processError(Protocol.java:117)
	at redis.clients.jedis.Protocol.process(Protocol.java:151)
	at redis.clients.jedis.Protocol.read(Protocol.java:205)
	at redis.clients.jedis.Connection.readProtocolWithCheckingBroken(Connection.java:297)
	at redis.clients.jedis.Connection.getStatusCodeReply(Connection.java:196)
	at redis.clients.jedis.BinaryJedis.ping(BinaryJedis.java:105)
	at redis.clients.jedis.JedisSlotBasedConnectionHandler.getConnection(JedisSlotBasedConnectionHandler.java:42)
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:113)
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:131)
	at redis.clients.jedis.JedisClusterCommand.runBinary(JedisClusterCommand.java:59)
	at redis.clients.jedis.BinaryJedisCluster.get(BinaryJedisCluster.java:98)
	at com.jnetx.redis.RedisClient.get(RedisClient.java:83)
	at com.jnetx.perffw.RedisPerf$ReadWriteThread.doOperation(RedisPerf.java:82)
	at com.jnetx.perffw.ThreadBase.run(ThreadBase.java:173)

The connections leak here!

As I see in sources of jedis 2.8.0 and jedis master branch, when such exception occurs, method JedisSlotBasedConnectionHandler.getConnection() does not return the connection back to the pool since it catches JedisConnectionException, but not JedisDataException.

Also we get many exceptions like this:

redis.clients.jedis.exceptions.JedisClusterMaxRedirectionsException: Too many Cluster redirections?
        at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:97)
        at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:131)
        at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:152)
        at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:131)
        at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:152)
        at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:131)
        at redis.clients.jedis.JedisClusterCommand.runBinary(JedisClusterCommand.java:59)
        at redis.clients.jedis.BinaryJedisCluster.get(BinaryJedisCluster.java:98)
        at com.jnetx.redis.RedisClient.get(RedisClient.java:92)
        at com.jnetx.perffw.RedisPerf$ReadWriteThread.doOperation(RedisPerf.java:78)
        at com.jnetx.perffw.ThreadBase.run(ThreadBase.java:173)

In this case, it seems that jedis returns connections to the pool, but I am not sure.

And some later, we see all the generator threads blocked with the following stack trace:

Unsafe.park(boolean, long) line: not available [native method]	
LockSupport.park(Object) line: 186	
AbstractQueuedSynchronizer$ConditionObject.await() line: 2043	
LinkedBlockingDeque<E>.takeFirst() line: 583	
GenericObjectPool<T>.borrowObject(long) line: 442	
GenericObjectPool<T>.borrowObject() line: 363	
JedisPool(Pool<T>).getResource() line: 48	
JedisPool.getResource() line: 99	
JedisSlotBasedConnectionHandler.getConnection() line: 36	
JedisSlotBasedConnectionHandler.getConnectionFromSlot(int) line: 65	
BinaryJedisCluster$3(JedisClusterCommand<T>).runWithRetries(byte[], int, boolean, boolean) line: 115	
BinaryJedisCluster$3(JedisClusterCommand<T>).runBinary(byte[]) line: 59	
RedisClient$MyCluster(BinaryJedisCluster).get(byte[]) line: 98	
RedisClient.get(String) line: 83	
RedisPerf$ReadWriteThread.doOperation() line: 82	
RedisPerf$ReadWriteThread(ThreadBase).run() line: 173	

-------------------

JedisCluster instance is created using the following code:
Set<HostAndPort> nodes = new LinkedHashSet<>();
....... fill nodes .......
JedisCluster jedisCluster = new JedisCluster(nodes);

All data accesses are performed via calls to jedisCluster.get(byte[]) and jedisCluster.set(byte[], byte[]) in muliple threads.

serge-groshev:
Fix #1168

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisSlotBasedConnectionHandler.java
+++ b/src/main/java/redis/clients/jedis/JedisSlotBasedConnectionHandler.java
@@ -42,13 +42,10 @@ public class JedisSlotBasedConnectionHandler extends JedisClusterConnectionHandl
 
         if (result.equalsIgnoreCase("pong")) return jedis;
 
-        pool.returnBrokenResource(jedis);
+        jedis.close();
       } catch (JedisException ex) {
         if (jedis != null) {
-          try {
-            pool.returnBrokenResource(jedis);
-          } catch (JedisException silent) {
-          }
+          jedis.close();
         }
       }
     }
--- a/src/main/java/redis/clients/jedis/JedisSlotBasedConnectionHandler.java
+++ b/src/main/java/redis/clients/jedis/JedisSlotBasedConnectionHandler.java
@@ -1,13 +1,12 @@
 package redis.clients.jedis;
 
-import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
-import java.util.Set;
-
 import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
 
+import java.util.Set;
+
 import redis.clients.jedis.exceptions.JedisConnectionException;
+import redis.clients.jedis.exceptions.JedisException;
 
 public class JedisSlotBasedConnectionHandler extends JedisClusterConnectionHandler {
 
@@ -44,9 +43,12 @@ public class JedisSlotBasedConnectionHandler extends JedisClusterConnectionHandl
         if (result.equalsIgnoreCase("pong")) return jedis;
 
         pool.returnBrokenResource(jedis);
-      } catch (JedisConnectionException ex) {
+      } catch (JedisException ex) {
         if (jedis != null) {
-          pool.returnBrokenResource(jedis);
+          try {
+            pool.returnBrokenResource(jedis);
+          } catch (JedisException silent) {
+          }
         }
       }
     }


HeartSaVioR:
@serge-groshev 
Sorry for the delay.
Looking #1178, I got another leak point. Actually, connection is leaked when it is `not` broken.
I'll try to fix and submit new pull request.
Thanks for pointing out!

HeartSaVioR:
@serge-groshev Sorry, ignore previous comment. I was mistaken.

HeartSaVioR:
I can now see the scenario how connection could be leaked. `ping` could throw other Exception, especially JedisException.
LGTM.

@marcosnils 
Btw, could we replace `ping-pong` manually with `setTestOnBorrow(true)` ? Actually we don't need to ping if Pool takes care of validation.

serge-groshev:
@HeartSaVioR 
Both methods getConnectionFromSlot(int slot) and getConnection() call JedisPool.getResource(), but method getConnectionFromSlot(int slot) does not ping the borrowed connection. If you replace ping-pong with setTestOnBorrow(true), it will ping connections as well.

HeartSaVioR:
@serge-groshev 
Actually not pinging would be more appreciated because it drops the performance.
@marcosnils Sorry for confusing. I really want to have is that we would want to remove pinging from getConnection(), too.

serge-groshev:
getConnection() is called when getConnectionFromSlot() fails. Pinging may be appropriate in such case.

HeartSaVioR:
@serge-groshev 
You're right. It was my intention and I forgot it. ;(
PR itself looks good for merging.

marcosnils:
I'll review tomorrow

sent from mobile

marcosnils:
@serge-groshev @HeartSaVioR PR LGTM!. I'm wondering if there's a way to actually test this. If we can't think of anything I guess we can merge

HeartSaVioR:
@marcosnils I think making unit test for this scenario is some kind of hard thing since we should trigger JedisDataException.

src/main/java/redis/clients/jedis/JedisSlotBasedConnectionHandler.java
```diff
@@ -44,9 +43,12 @@ public Jedis getConnection() {
         if (result.equalsIgnoreCase("pong")) return jedis;
 
         pool.returnBrokenResource(jedis);
-      } catch (JedisConnectionException ex) {
+      } catch (JedisException ex) {
         if (jedis != null) {
-          pool.returnBrokenResource(jedis);
+          try {
```
marcosnils: Why is this catch block here?. This shouldn't raise an exception. 
Additionally I believe it can be replaced by `jedis.close()`. @HeartSaVioR WDYT?

marcosnils:
@HeartSaVioR @serge-groshev I made some comments. Please check them while I think about a way to test this.

src/main/java/redis/clients/jedis/JedisSlotBasedConnectionHandler.java
HeartSaVioR: @marcosnils 
When JedisDataException is occurred, the connection is not marked to be broken. So jedis.close() doesn't work on this situation.
And if we want to consume any Exceptions during returning, we can even define it to consume Throwable since it is more safer.

marcosnils: @HeartSaVioR in this case it's ok if the connection is not marked broken because in the original issue #1168 @serge-groshev was getting a JedisDataException because Redis was still loading the dataset into memory, so it's OK to re-use the same connection afterwards when retrieving another connection. 

Regarding the IllegalArgumentException as far as I can remember it can only happen when you try to return an resource which doesn't belong to the pool which can't be this case.

serge-groshev: Method returnBrokenResource() is able to throw JedisException. In bug #1168 we sometimes got exception like this:

redis.clients.jedis.exceptions.JedisException: Could not return the resource to the pool
at redis.clients.util.Pool.returnBrokenResourceObject(Pool.java:100)
at redis.clients.jedis.JedisPool.returnBrokenResource(JedisPool.java:112)
at redis.clients.jedis.JedisPool.returnBrokenResource(JedisPool.java:12)
at redis.clients.jedis.Jedis.close(Jedis.java:3355)
...
Caused by: java.lang.IllegalStateException: Invalidated object not currently part of this pool
at org.apache.commons.pool2.impl.GenericObjectPool.invalidateObject(GenericObjectPool.java:643)
at redis.clients.util.Pool.returnBrokenResourceObject(Pool.java:98)
...

I am not sure whether we can get such exception here, but throwing it further up seems inappropriate, so I added the try-catch block.

serge-groshev:
How to reproduce that issue:
1. Create a cluster and fill it with gigabytes of persistent data
2. Kill and restart all cluster nodes (in our case, I used more sofisticated environment: redis servers, some kind of chaos monkey that eventually killed some random servers, handmade monitoring tool that performed failovers and restarted crashed servers, and a client that continuously worked with the redis cluster)
3. Try to access redis server that is still starting.

With proposed fix, everything works fine. Without it, we ended with all the client threads blocked on JedisPool.getResource()

src/main/java/redis/clients/jedis/JedisSlotBasedConnectionHandler.java
serge-groshev: Regarding connection is not marked broken: please note that if ping() does not throw, but returns anything else than "pong", the same method returnBrokenResource() is invoked.
And if ping() returns null, we get uncaught NPE there.
I suppose, both cases should not occur with sane redis server, but anything may happen in the wild :)

marcosnils: > I am not sure whether we can get such exception here, but throwing it further up seems inappropriate, so I added the try-catch block.

@serge-groshev I prefer not to through it up because otherwise we could be hiding bugs that can be difficult to trace. defensive programming usually is not a good practice. That `try` block __shoudln't__ be necessary there, if it throws an exception is because something it's wrong and something else needs to be done. 

Also, the `returnBrokenResource` exception in #1168 comes from a different place, not from the `getConnection` method, so that's another smell that makes me think we shouldn't do anything here. Just to make sure, can you remove that `try-catch` block and run your tests again to see what we get?


> please note that if ping() does not throw, but returns anything else than "pong", the same method returnBrokenResource() is invoked.

Ping shoudn't return anything but "pong" or an exception. If it does, then it's a bug in the code which should be treated as such. Having said that, I guess we can change the whole codeblock to something like this:

```java
    for (JedisPool pool : pools) {
      Jedis jedis = null;
      try {
        jedis = pool.getResource();

        if (jedis == null) {
          continue;
        }

        String result = jedis.ping();

        if (result.equalsIgnoreCase("pong")) return jedis;

      } finally {
        if (jedis != null) {
          jedis.close();
        }
      }
    }
```
Since we'll always return the connection to the pool, we can just remove the exception handling and add a `finally` clause.

> And if ping() returns null, we get uncaught NPE there.

Again, defensive programming here. In theory should not happen, and if it does, it should be something we need to fix somewhere else, but not here.

serge-groshev: OK, let's remove that nested try-catch if needed.
I am not so familiar with that code to tell whether replacing old call to returnBrokenResource() with call to close() is good or bad, and why.
I can say no pros or cons for these two suggested changes, but replacing try-catch with try-finally seems to be a very bad idea: earlier that code catched JedisConnectionException from ping(), now it shall catch any JedisException as well, but you suggest to let that exception be thrown further up instead of catching.

HeartSaVioR: @marcosnils  @serge-groshev 

First of all, the code block Marcos suggested is having a bug that it returns instance to pool which is return value of the method.
(I was confused at the same point, so be attention all!)

This method MUST ensure that it returns one Jedis instance if available.
(regarding to antirez's ruby implementation)
In order to achieve, it intentionally consumes any exceptions. 
Actually it can even be changed to consume any RuntimeException in defensive programming's eyes.

IMHO why @serge-groshev meets the bug is that we were not defensive. We just forgot to check other than JedisConnectionException.

marcosnils: > I can say no pros or cons for these two suggested changes, but replacing try-catch with try-finally seems to be a very bad idea: earlier that code catched JedisConnectionException from ping(), now it shall catch any JedisException as well, but you suggest to let that exception be thrown further up instead of catching.


@serge-groshev you're right, we shouldn't remove the catch block as otherwise the exception would be propagated incorrectly. I guess we can refactor the block as follows:

```java
    for (JedisPool pool : pools) {
      Jedis jedis = null;
      try {
        jedis = pool.getResource();

        if (jedis == null) {
          continue;
        }

        String result = jedis.ping();

        if (result.equalsIgnoreCase("pong")) return jedis;
        jedis.close();
      } catch (JedisException je) {
        if (jedis != null) {
          jedis.close();
        }
      }
    }
```
@serge-groshev  @HeartSaVioR WDYT?

serge-groshev: I am not sure about the difference between `pool.returnBrokenResource(jedis)` and `jedis.close()`
All the rest looks OK for me.

marcosnils: @serge-groshev `jedis.close()` automatically returns the `Jedis` instance to it's corresponding pool accordingly. If the Jedis instance is dirty (an error raised trying to use it) the `close` method will take care of returning the broken resource automatically. 

This way you don't have to deal with knowing which state the connection has before returning it to the pool as it's being done automatically.

HeartSaVioR: @marcosnils @serge-groshev 
What we're trying to do is same to validating object in JedisFactory. So we can just follow how it is implemented. Maybe we can refactor via `extract method` and expose new method, and call it.

```
  @Override
  public boolean validateObject(PooledObject<Jedis> pooledJedis) {
    final BinaryJedis jedis = pooledJedis.getObject();
    try {
      HostAndPort hostAndPort = this.hostAndPort.get();

      String connectionHost = jedis.getClient().getHost();
      int connectionPort = jedis.getClient().getPort();

      return hostAndPort.getHost().equals(connectionHost)
          && hostAndPort.getPort() == connectionPort && jedis.isConnected()
          && jedis.ping().equals("PONG");
    } catch (final Exception e) {
      return false;
    }
  }
```

marcosnils: @HeartSaVioR is that validateObject method correct?. In case of exception shouldn't we return the jedis instance to the pool?

HeartSaVioR: @marcosnils It is just for checking validation of instance, and rest works are done by commons pool. We may have to do rest works by hand, but we can reuse validation method from here.

marcosnils: @HeartSaVioR you're right. Commons-pool takes care of the resource as we're overriding a method of the library.

marcosnils:
@serge-groshev I spoke with @HeartSaVioR about this and we agreed that the last snippet from the comments should be necessary to fix this issue. Can you please apply it to your local environment and check if the errors you were seeing are now gone?. 

If they do, please submit the changes so we can marge. Thanks very much for taking the time for this!.

serge-groshev:
Which snippet is the last? I see something with `validateObject()` method.

I will try with this code:
``` java
     for (JedisPool pool : pools) {
      Jedis jedis = null;
      try {
        jedis = pool.getResource();

        if (jedis == null) {
          continue;
        }

        String result = jedis.ping();

        if (result.equalsIgnoreCase("pong")) return jedis;
        jedis.close();
      } catch (JedisException je) {
        if (jedis != null) {
          jedis.close();
        }
      }
    }

**Comment by serge-groshev** [comment]
I have checked with those close()s.
Killing all servers in redis cluster with gigabytes of persistent data, I got several thousands exceptions

```
redis.clients.jedis.exceptions.JedisDataException: LOADING Redis is loading the dataset in memory
        at redis.clients.jedis.Protocol.processError(Protocol.java:117)
        at redis.clients.jedis.Protocol.process(Protocol.java:151)
        at redis.clients.jedis.Protocol.read(Protocol.java:205)
        at redis.clients.jedis.Connection.readProtocolWithCheckingBroken(Connection.java:297)
        at redis.clients.jedis.Connection.getStatusCodeReply(Connection.java:196)
        at redis.clients.jedis.BinaryJedis.ping(BinaryJedis.java:105)
        at redis.clients.jedis.JedisSlotBasedConnectionHandler.getConnection(JedisSlotBasedConnectionHandler.java:42)  <-------------
        at redis.clients.jedis.JedisSlotBasedConnectionHandler.getConnectionFromSlot(JedisSlotBasedConnectionHandler.java:67)
        at com.jnetx.redis.RedisClient.getPooledJedis(RedisClient.java:302)
```

while the cluster was being recovering, and then the test went on.

serge-groshev:
I have added the commit to the master branch. Please merge.
We would like to see those change in the freshest 2.x branch (2.8 or 2.9?). Do I have to create a similar commit for a branch?

marcosnils:
@serge-groshev Once CI get's green I will merge. You don't need to create additional PR's we'll take care of downmerging to corresponding branches. 

Thanks a lot for the contribution!.

marcosnils:
merged to master and 2.8 respectively!.

HeartSaVioR:
@serge-groshev @marcosnils Thanks for following up and merge!

PR #1185, opened 2015-12-21
Adding onProcess callback for subscribing more channels while working

kak2:
Hi xetorthio,

I have a little change on JedisPubSub class, adding onProcess callback. With it, Jedis will allow us to subscribe channels easily while working.
What do you think about it?


kak2:
It is replaced by #1186 (https://github.com/xetorthio/jedis/pull/1186)

PR #1186, opened 2015-12-21
Adding onProcess callback for subscribing more channels while working

kak2:
Using onProcess callback to subscribe more channel. There are two changes:
- Create onProcess callback function in RedisPubSub
- Make getRawObjectMultiBulkReply to non-blocking function via setSoTimeout with default timeout is 200ms.

### DIFF
--- a/src/main/java/redis/clients/jedis/Protocol.java
+++ b/src/main/java/redis/clients/jedis/Protocol.java
@@ -22,7 +22,7 @@ public final class Protocol {
   public static final String DEFAULT_HOST = "localhost";
   public static final int DEFAULT_PORT = 6379;
   public static final int DEFAULT_SENTINEL_PORT = 26379;
-  public static final int DEFAULT_TIMEOUT = 2000;
+  public static final int DEFAULT_TIMEOUT = 200;
   public static final int DEFAULT_DATABASE = 0;
 
   public static final String CHARSET = "UTF-8";
--- a/src/main/java/redis/clients/jedis/BinaryJedis.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedis.java
@@ -3056,7 +3056,7 @@ public class BinaryJedis implements BasicCommands, BinaryJedisCommands, MultiKey
 
   @Override
   public void subscribe(BinaryJedisPubSub jedisPubSub, byte[]... channels) {
-    client.setTimeoutInfinite();
+    client.setSoTimeout(Protocol.DEFAULT_TIMEOUT);
     try {
       jedisPubSub.proceed(client, channels);
     } finally {
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -10,6 +10,7 @@ import java.util.List;
 
 import redis.clients.jedis.commands.ProtocolCommand;
 import redis.clients.jedis.exceptions.JedisConnectionException;
+import redis.clients.jedis.exceptions.JedisConnectionTimeOutException;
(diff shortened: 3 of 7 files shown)


HeartSaVioR:
@kak2 
Sorry for the delay.

First of all, Please minimize your change when submitting PR. 
```JedisPubSub.onProcess()``` is only related modification regarding to your PR's title.
I don't see why we treat JedisConnectionException and JedisConnectionTimeoutException differently.
And modifying default timeout would break backward compatibility and stability. I don't think we're OK to do.

And what users can do when onProcess() is called? Please share your use case and code snippet regarding onProcess().

**Comment by kak2** [comment]
Hi @HeartSaVioR 

Firstly, with onProcess callback, the users can subscribe more channels here. You can see my use case at [RedisHandler class](https://gist.github.com/kak2/2987e4959784b13e0b03)

Secondly, I will talk about the difference between JedisConnectionException and JedisConnectionTimeoutException. Let's take a look in readProtocolWithCheckingBroken() function in Connection.java file, Protocol.read(inputstream) is blocking function in the original version. It just returns when having data from redis server or JedisConnectionException is thrown (In this case, there're some problem with networking, and you decided to close connection). But in my patch, Protocol.read(inputstream) is still blocking, but only 200ms - DEFAULT_TIMEOUT or JedisConnectionTimeoutException is thrown (and I just skip this exception, don't close connection)

In addition, the original DEFAULT_TIMEOUT (2000ms = 2s) is too big for me, so I decided to make it smaller. I can revert this value if you accept this PR. But I think the best solution is adding timeout parameter to subscribe function. (In original version, the timeout is infinitive).

PS: Sorry if my English is not good enough. I have tried to explain my idea. If exist something is not clear, please tell me know. Thanks.

**Comment by HeartSaVioR** [comment]
Why we treat SocketTimeoutException to JedisConnectionException is, we can't sure whether timeout is occurred from sending, or receiving. It would be better to safely close the connection to prevent reading  response of previous request.

**Comment by kak2** [comment]
In Java 1.7 docs, they said SocketTimeoutException has just thrown on [socket read or accept](https://docs.oracle.com/javase/7/docs/api/java/net/SocketTimeoutException.html)

**Comment by 2bitoperations** [comment]
This PR would really help w/ ElastiCache in AWS as well. Several times now, we've had Amazon do maintenance on our hosted Redis instances that cause them to go down (stop responding) without killing the sockets open to .subscribe(). This means our subscriptions wait indefinitely to receive data that will never come - better behavior would be to timeout and attempt a reconnect like the PR proposes. (We end up having to restart the JVM to get our application happy again.)

What's remaining to get this merged? How can I help progress along?

**Comment by 2bitoperations** [comment]
I've tried a slightly different approach with PR #1275. Trying a different way to fix same issue.

File: src/main/java/redis/clients/jedis/Protocol.java [hunk 1]
```diff
@@ -22,7 +22,7 @@
   public static final String DEFAULT_HOST = "localhost";
   public static final int DEFAULT_PORT = 6379;
   public static final int DEFAULT_SENTINEL_PORT = 26379;
-  public static final int DEFAULT_TIMEOUT = 2000;
+  public static final int DEFAULT_TIMEOUT = 200;
```
marcosnils: ?? Change the default timeout configuration used in Jedis overall?. This will definitely won't happen

marcosnils:
I don't like this approach. Changing the default timeout (or specifying a new one) and making the process constantly time out I don't believe is the best approach here.

I'm not sure which would be the best approach for this case, but I'd like to see @2bitoperations proposal as well.

sazzad16:
Closed due to inactivity.

PR #1189, opened 2016-01-05
Pool IO buffers

mauhiz:
This is a simple implementation that enables pooling the byte buffers used in Jedis IO filtered stream classes. It is non intrusive and easy to disable, yet should improve memory allocation pressure on large applications.

*to use* : nothing special to do by default. If you use a custom buffer size, call `BufferPool.setPooledBufferSize` with this buffer size.
*to completely disable* : call `BufferPool.setPooledBufferSize(-1)`

### DIFF
--- a/src/main/java/redis/clients/util/BufferPool.java
+++ b/src/main/java/redis/clients/util/BufferPool.java
@@ -1,0 +1,79 @@
+package redis.clients.util;
+
+import org.apache.commons.pool2.BasePooledObjectFactory;
+import org.apache.commons.pool2.PooledObject;
+import org.apache.commons.pool2.impl.DefaultPooledObject;
+import org.apache.commons.pool2.impl.GenericObjectPool;
+
+public class BufferPool {
+  private static final class BufferFactory extends BasePooledObjectFactory<byte[]> {
+    static final BufferFactory INSTANCE = new BufferFactory();
+
+    @Override
+    public byte[] create() {
+      return new byte[pooled_buffer_size];
+    }
+
+    @Override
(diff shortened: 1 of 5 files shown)


HeartSaVioR:
Cool idea!
I'd like to see the result of benchmark, memory usages, GC count, ops, etc.
@mauhiz Do you mind to do benchmark between before vs. after this PR, or I'll do it?

mauhiz:
@HeartSaVioR if you have a test harness ready, by all means please go ahead.

HeartSaVioR:
@mauhiz 
Looking into your changes deeply, BufferPool.obtainBuffer() is used just for RedisInputStream and RedisOutputStream, so actually it only helps when you use much of Jedis instances.
(In other words, it doesn't help when you use JedisPool with proper configuration.)

Could you elaborate your use case?
Thanks in advance!

mauhiz:
@HeartSaVioR You are right, I was under the wrong impression that `RedisInputStream` and `RedisOutputStream` were transient. I should have been looking at my head dump histogram more closely. :bow: 

Therefore, I don't think the current implementation has issues with regards to buffer management. Sorry and thanks for your time.

HeartSaVioR:
@mauhiz Thanks for following up!
When we use string version of Jedis methods, we need to initialize byte[] for each operations.
If we can pool it, it should be helpful.

mauhiz:
What do you mean? The `String` to `byte[]` conversion in `SafeEncoder.encode` ?

HeartSaVioR:
@mauhiz Yes, right. Actually its size is not deterministic, so it's just an idea regarding speed vs space.

PR #1204, opened 2016-01-28
using cluster slots instead of cluster nodes

charsyam:
antirez tweet(https://twitter.com/antirez/status/691993742606274560)
"Redis Cluster client developer? Please if you are using “CLUSTER NODES” switch to “CLUSTER SLOTS” API ASAP. Thanks!"

so I changed discoverClusterNodesAndSlots to use clusterSlots not clusterNodes


src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -49,14 +51,31 @@ public void discoverClusterNodesAndSlots(Jedis jedis) {
...
-            jedis.getClient().getHost(), jedis.getClient().getPort()));
+      List<Object> slots = jedis.clusterSlots();
 
-        HostAndPort targetNode = clusterNodeInfo.getNode();
-        setNodeIfNotExist(targetNode);
-        assignSlotsToNode(clusterNodeInfo.getAvailableSlots(), targetNode);
+      for (Object slotInfoObj : slots) {
+        List<Object> slotInfo = (List<Object>) slotInfoObj;
+
+        if (slotInfo.size() <= 2) {
```
marcosnils: Shouldn't we use the MASTER_NODE_INDEX here as well?

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -49,14 +51,31 @@ public void discoverClusterNodesAndSlots(Jedis jedis) {
...
+        int size = slotInfo.size();
+        for (int i = MASTER_NODE_INDEX; i < size; i++) {
+          List<Object> hostInfos = (List<Object>) slotInfo.get(i);
+          if (hostInfos.size() <= 0) {
+            continue;
+          }
+
+          HostAndPort targetNode = generateHostAndPort(hostInfos);
+          setNodeIfNotExist(targetNode);
+          if (i == MASTER_NODE_INDEX) {
```
marcosnils: We weren't validating this before, why now?

I've also realized that `assingSlotsToNode` already checks if node exists and calls `setNodeIfNotExist` appropriately so maybe we can remove it from here in favor of code readability?

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -49,14 +51,31 @@ public void discoverClusterNodesAndSlots(Jedis jedis) {
...
+
+        if (slotInfo.size() <= 2) {
+          continue;
+        }
+
+        List<Integer> slotNums = getAssignedSlotArray(slotInfo);
+
+        // hostInfos
+        int size = slotInfo.size();
+        for (int i = MASTER_NODE_INDEX; i < size; i++) {
```
marcosnils: Do we need to iterate through all the cluster nodes necessarily?. Why would we want the `slaves` if we don't do anything with them?

I've realized we were doing this before also, any ideas why?. @HeartSaVioR

HeartSaVioR:
I don't understand why antirez makes his mind to break the public API again...

Maybe we will have a chance to stuck when antirez makes CLUSTER NODES stable and make a note, "CLUSTER SLOTS are changing, please use CLUSTER NODES instead".
We may want to think about supporting multiple formats (if possible) as @mp911de said via twitter.

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
HeartSaVioR: Yes, let's use MASTER_NODE_INDEX.
@charsyam

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
HeartSaVioR: @marcosnils 
I thought it would be safer to discover as many nodes as we can, even though they're slaves because of failover.

Let's assume that Jedis knows about only masters.
Let's think about failover situation: we have 3 masters and 3 slaves, and we didn't notice that 3 masters are dead.
In a normal situation, 3 slaves should be promoted to master so there's no issue from Redis Cluster. But Jedis can't access any random nodes for redirection because all nodes Jedis knows are dead.

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
HeartSaVioR: Please see above comment.

HeartSaVioR:
@charsyam @marcosnils 
What about fixing discoverClusterSlots() to clear node informations (`this.nodes.clear();`) and discover slaves (but not assign to slot)?
I believe we can replace discoverClusterNodesAndSlots() with discoverClusterSlots() then.

charsyam:
@marcosnils @HeartSaVioR Thanks for reviews.
Actually, I wanted to get functional compatibility. so It 
will return same result compared to before

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
marcosnils: @HeartSaVioR you're right. It's an extreme situation where all the masters should die at exactly the same time, so Jedis won't be able to connect to any. It's __very__ unlikely to happen but not impossible :wink:. We can leave it as it is then.

HeartSaVioR: @marcosnils Yeah, Redis never pushes updated cluster information to Jedis so long absence of request could increase possibilities a bit.

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
marcosnils: @HeartSaVioR I didn't see anything in #1204 about this. What I'm asking is if we can remove the `setNodeIfNotExist` here as it's being called also in the `assignSlotsToNode` method.

marcosnils:
> What about fixing discoverClusterSlots() to clear node informations (this.nodes.clear();) and discover slaves (but not assign to slot)?
> I believe we can replace discoverClusterNodesAndSlots() with discoverClusterSlots() then.

Hmm I believe that assigning slots to slave is a good choice because whenever a master dies it's __very__ likely that the slave will take over the slots that the master previously had, correct?

marcosnils:
> Maybe we will have a chance to stuck when antirez makes CLUSTER NODES stable and make a note, "CLUSTER SLOTS are changing, please use CLUSTER NODES instead".

@HeartSaVioR maybe we can merge this in MASTER so tests keep passing and finally think about releasing Jedis 3.0 with this change. WDYT?

HeartSaVioR:
@marcosnils 
Hmm... Thinking it again, and I've got one idea for slave slot information.
When we fail to connect node, we can query one of slaves (candidate to master) instead of pick random node.
If it throw JedisConnectionException, try next slave (call discoverClusterSlots() when we tried all slaves)
If query doesn't throw MOVED, picked slave is promoted to master. 
If query throw MOVED, find destination from slaves and that's promoted to master.
If destination of MOVED is not on slaves, discoverClusterSlots() will be called to rebuild (node, slot) relation.

marcosnils:
@HeartSaVioR is there a reason why you think we can't merge this now?. We're not using the nodeId in Jedis, so migrating to `CLUSTER NODES` should be trivial for us, correct?

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
HeartSaVioR: @marcosnils 
For now, assignSlotsToNode() overwrites slot to node information, which means slave information can overwrite master information.
But as I stated we'd be better to store slaves as nodes, so slaves should be added to nodes but without slot assignments.

HeartSaVioR:
@marcosnils @charsyam 
Sorry to respond it too late. There's holiday in South Korea so I was logged off some days.
Seems like there's miscommunication between us. I don't mean to block this pull request, just complaining about Redis API changes.
IMO, comments regarding existing implementation could be address to separate issue if we think it's wrong.
Then non-addressed review is only about MASTER_NODE_INDEX, and we can merge when @charsyam addresses it.

charsyam:
@marcosnils @HeartSaVioR 
Hi. I removed ClusterNodeInformation and parser and tests.
and I also removed testReadonly in JedisClusterTest.
To support old version cluster slots, there is no way to get cluster node id except using "cluster nodes" command to run testReadonly testcase

src/test/java/redis/clients/jedis/tests/JedisClusterTest.java
```diff
@@ -579,4 +555,19 @@ private boolean isAnyNodeHandshaking(Jedis node) {
     return false;
   }
 
+  private HostAndPort generateHostAndPort(List<Object> hostInfos) {
+    return new HostAndPort(SafeEncoder.encode((byte[]) hostInfos.get(0)),
+            ((Long) hostInfos.get(1)).intValue());
+  }
+
+  private String getNodeId(Jedis node) {
```
marcosnils: @charsyam I believe we're not using this `getNodeId` method. Why was it added?

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
marcosnils: @charsyam remember to change this `2` value to use the MASTER_NODE_INDEX constant.

src/test/java/redis/clients/jedis/tests/JedisClusterTest.java
charsyam: @marcosnils I will remove it. first, I tried to get node id from info command but it failed.

charsyam:
@marcosnils @HeartSaVioR I applied reviews

marcosnils:
@charsyam I don't want to be annoying but is there any chance you can change the last commit message to something more meaningful like: "Removed unused method and use constant instead of fixed value?.  We try to keep commit history as meaningful and clean as possible.

Besides from that LGTM. @HeartSaVioR any final thoughts?

src/test/java/redis/clients/jedis/tests/JedisClusterTest.java
```diff
@@ -579,4 +555,8 @@ private boolean isAnyNodeHandshaking(Jedis node) {
     return false;
   }
 
+  private HostAndPort generateHostAndPort(List<Object> hostInfos) {
```
HeartSaVioR: @charsyam Seems like unused method. Could you remove it also?

HeartSaVioR:
LGTM once new comment from me is addressed.
Squashing commits into one would be nice since it is not huge modification.

src/test/java/redis/clients/jedis/tests/JedisClusterTest.java
marcosnils: @HeartSaVioR I believe it's used here https://github.com/xetorthio/jedis/pull/1204/files#diff-834ad1f566a8da02a00da75ce4e9c3a6R69

HeartSaVioR: @marcosnils JedisClusterInfoCache already has generateHostAndPort() as private method. It's private method. ;)

marcosnils: @HeartSaVioR weird?. Even though the PR shows that the method has been added, I only see it once here: https://github.com/charsyam/jedis/blob/feature/using_cluster_slots/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java

HeartSaVioR: @marcosnils 
If Github isn't lying to me, generateHostAndPort() is here in JedisClusterTest.
https://github.com/charsyam/jedis/blob/feature/using_cluster_slots/src/test/java/redis/clients/jedis/tests/JedisClusterTest.java

charsyam:
@marcosnils @HeartSaVioR I removed unused method generatedHostAndPort in JedisClusterTest(I added it for testReadonly that is removed)
and squashing commits :) Thanks.

HeartSaVioR:
@charsyam Great! LGTM. Thanks for the great contribution. :)

charsyam:
@marcosnils I added TestReadonly again :)

marcosnils:
@charsyam I'll merge once tests pass. Amazing job!. Thanks for the contribution!.

charsyam:
@marcosnils You're welcome :) It's my plesure

marcosnils:
@charsyam @HeartSaVioR as change works with all current Redis versions I've applied it to 2.8 and 2.9 respectively. 

Thx!

PR #1209, opened 2016-02-07
Bump commons-pool2 to latest version

### LINKED ISSUE #1208: New version of apache commons pool cause problems with class loader.
Due to this lines of code:
https://github.com/apache/commons-pool/blob/trunk/src/main/java/org/apache/commons/pool2/impl/BaseGenericObjectPool.java#L605-L606

Jedis may throw 
`java.lang.ClassNotFoundException: org.apache.commons.pool2.impl.DefaultEvictionPolicy`
error when creating JedisPool in class with other class loader than thread, like when we have master app and addons for it (and this apache library is shaded into addon), simple work around is just override context class loader:
`Thread.currentThread().setContextClassLoader(this.getClass().getClassLoader());`
but this may cause other problems depending on used software. (Other apps can change this loader etc)

Any possible fixes?

GotoFinal:
Fixes #1208 
@marcosnils

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -62,7 +62,7 @@
 		<dependency>
 			<groupId>org.apache.commons</groupId>
 			<artifactId>commons-pool2</artifactId>
-			<version>2.3</version>
+			<version>2.4.2</version>
 			<type>jar</type>
 			<scope>compile</scope>
 		</dependency>


marcosnils:
LGTM. @HeartSaVioR WDYT?

marcosnils:
@GotoFinal Jedis is tested against the latest Redis version from unstable branch. I think some changes were made with the cluster commands that we need to address in order for tests to pass.  I'll try to take a look tomorrow and ask you to rebase so we can get this merged.

HeartSaVioR:
LGTM, too.

marcosnils:
@HeartSaVioR @GotoFinal  we need to wait on #1211 to rebase and make tests pass.

HeartSaVioR:
We merged #1204 instead of #1211, so we're OK to go.

HeartSaVioR:
Merged into master, 2.8, 2.9 respectively.

Issue #1024, opened 2015-06-11
redis.clients.jedis.exceptions.JedisConnectionException: no reachable node in cluster


arulanandsp:
Hi,

I created a ElasticCache Redis Cluster with 1 Master node. When I tried to connect I am getting 
redis.clients.jedis.exceptions.JedisConnectionException: no reachable node in cluster

Here is the code

Set<HostAndPort> jedisClusterNodes = new HashSet<HostAndPort>();
//Jedis Cluster will attempt to discover cluster nodes automatically
jedisClusterNodes.add(new HostAndPort("xxxxx.0001.use1.cache.amazonaws.com", 6379));
JedisCluster jc = new JedisCluster(jedisClusterNodes);
jc.set("foo", "bar");
String value = jc.get("foo");

Please let me know if I am missing something

marcosnils:
@arulanandsp aws ElasticCache has it's own concept of a "Redis cluster" which is not the same as the original proposal of redis (http://redis.io/topics/cluster-tutorial).

You can't use `JedisCluster` with ElasticCache because ElasticCache doesn't support native "Redis clustering" yet.

PR #1211, opened 2016-02-10
Use cluster slots instead of cluster nodes

marcosnils:
@antirez suggested that all cluster libraries should move to `cluster
slots` for better API compatibility.

https://twitter.com/antirez/status/691993742606274560

@HeartSaVioR  @xetorthio plz review?

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
@@ -49,10 +49,9 @@ public class JedisClusterInfoCache {
       this.nodes.clear();
       this.slots.clear();
 
-      String localNodes = jedis.clusterNodes();
-      for (String nodeInfo : localNodes.split("\n")) {
-        ClusterNodeInformation clusterNodeInfo = nodeInfoParser.parse(nodeInfo, new HostAndPort(
-            jedis.getClient().getHost(), jedis.getClient().getPort()));
+      List<Object> localNodes = jedis.clusterSlots();
+      for (Object nodeInfo : localNodes) {
+        ClusterNodeInformation clusterNodeInfo = nodeInfoParser.parse((List<Object>) nodeInfo);
 
         HostAndPort targetNode = clusterNodeInfo.getNode();
         setNodeIfNotExist(targetNode);
--- a/src/main/java/redis/clients/util/ClusterNodeInformation.java
+++ b/src/main/java/redis/clients/util/ClusterNodeInformation.java
@@ -8,26 +8,14 @@ import redis.clients.jedis.HostAndPort;
 public class ClusterNodeInformation {
   private HostAndPort node;
   private List<Integer> availableSlots;
-  private List<Integer> slotsBeingImported;
-  private List<Integer> slotsBeingMigrated;
 
   public ClusterNodeInformation(HostAndPort node) {
     this.node = node;
     this.availableSlots = new ArrayList<Integer>();
-    this.slotsBeingImported = new ArrayList<Integer>();
-    this.slotsBeingMigrated = new ArrayList<Integer>();
   }
 
   public void addAvailableSlot(int slot) {
     availableSlots.add(slot);
-  }
-
-  public void addSlotBeingImported(int slot) {
-    slotsBeingImported.add(slot);
-  }
-
-  public void addSlotBeingMigrated(int slot) {
-    slotsBeingMigrated.add(slot);
   }
 
   public HostAndPort getNode() {
@@ -38,11 +26,4 @@ public class ClusterNodeInformation {
     return availableSlots;
   }
 
-  public List<Integer> getSlotsBeingImported() {
-    return slotsBeingImported;
-  }
-
-  public List<Integer> getSlotsBeingMigrated() {
-    return slotsBeingMigrated;
-  }
 }
--- a/src/main/java/redis/clients/util/ClusterNodeInformationParser.java
+++ b/src/main/java/redis/clients/util/ClusterNodeInformationParser.java
@@ -2,71 +2,30 @@ package redis.clients.util;
 
 import redis.clients.jedis.HostAndPort;
 
+import java.util.List;
+
 public class ClusterNodeInformationParser {
-  private static final String SLOT_IMPORT_IDENTIFIER = "-<-";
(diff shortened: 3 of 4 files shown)


HeartSaVioR:
@marcosnils It seems to be in conflicts with #1204. Do you want to close #1204 and go on with this pull request?

marcosnils:
@HeartSaVioR @charsyam I totally forgot about #1024 and re did the whole stuff this weekend (yes, I'm burned out). TBH I'm not sure about which of the implementations we should follow. One thing that  I like about this PR is that we're still using the `ClusterNodeInformation`, a test has been added and that all the unnecessary code goes away.

Not sure how to proceed though, WDYT?

HeartSaVioR:
@marcosnils 
Since #1204 is also good patch, I'd like to let @charsyam get a credit, and moving forward.

Btw, ClusterNodeInformationParser is for parsing "cluster nodes", so I'd like to let ClusterNodeInformationParser supports old & new output formats of "cluster nodes".
("cluster slots" doesn't show informations for 'migration', so someone still wants to use "cluster nodes".)

When we think we also have to have parser for "cluster slots", how about adding ClusterSlotInformationParser?

HeartSaVioR:
FYI, output of `cluster nodes` / `cluster slots` for current unstable

```
127.0.0.1:7379> cluster slots
1) 1) (integer) 10923
   2) (integer) 16383
   3) 1) "127.0.0.1"
      2) (integer) 7381
      3) "9b9338b01eac0293c8e40828c0d461426b69c9f9"
   4) 1) "127.0.0.1"
      2) (integer) 7384
      3) "a1364c31ee52aeaad963605992b296036941d13d"
2) 1) (integer) 0
   2) (integer) 5460
   3) 1) "127.0.0.1"
      2) (integer) 7379
      3) "b9ccaf6db964fcff2cb0029c30627fcb0fbece39"
   4) 1) "127.0.0.1"
      2) (integer) 7382
      3) "b88ca1055e5a338cd7547f457fb93943fe617fbe"
3) 1) (integer) 5461
   2) (integer) 10922
   3) 1) "127.0.0.1"
      2) (integer) 7380
      3) "4f33016add3f90da209669c7f2dd88ba178c21d2"
   4) 1) "127.0.0.1"
      2) (integer) 7383
      3) "9143145f9231d7f1cd7245e4763c86b6bae642a1"
```

```
127.0.0.1:7379> cluster nodes
9b9338b01eac0293c8e40828c0d461426b69c9f9 127.0.0.1:7381@17381 master - 0 1455077974871 3 connected 10923-16383
b88ca1055e5a338cd7547f457fb93943fe617fbe 127.0.0.1:7382@17382 slave b9ccaf6db964fcff2cb0029c30627fcb0fbece39 0 1455077974871 4 connected
b9ccaf6db964fcff2cb0029c30627fcb0fbece39 127.0.0.1:7379@17379 myself,master - 0 0 1 connected 0-5460
a1364c31ee52aeaad963605992b296036941d13d 127.0.0.1:7384@17384 slave 9b9338b01eac0293c8e40828c0d461426b69c9f9 0 1455077974871 6 connected
4f33016add3f90da209669c7f2dd88ba178c21d2 127.0.0.1:7380@17380 master - 0 1455077974871 2 connected 5461-10922
9143145f9231d7f1cd7245e4763c86b6bae642a1 127.0.0.1:7383@17383 slave 4f33016add3f90da209669c7f2dd88ba178c21d2 0 1455077974871 5 connected
```

HeartSaVioR:
If we can't support both format for `cluster nodes`, we may have to **drop supporting "cluster nodes"**.
Then we just remove ClusterNodeInformationParser, ClusterNodeInformation, and relevant tests.

marcosnils:
@HeartSaVioR I'd only stick with `cluster slots` for the moment. @antirez mentioned that the `cluster nodes` API will probably change in the future, so it doesn't make sense to support both strategies if we'll be using only `cluster slots` from now on. 

@charsyam do you mind making the changes in #1024 to entirely remove `ClusterNodeInformationParser` and all unnecessary code / tests?

charsyam:
@marcosnils @HeartSaVioR I think we should remain thise codes and tests. some users can use "cluster nodes" directly. so It is still meaningful for cluster commands. so if command is changed. at that time we can change tests

marcosnils:
@charsyam I don't understand what you're saying. For node discovery we'll be relying on `cluster slots` as antirez said it should be how clients need to query the cluster status. I'm not saying to remove `clusternodes` from JedisCluster, I'm just saying to remove it from the nodes discovery implementation.

HeartSaVioR:
@charsyam 
Passing the test is actually meaningless if output formats vary among the versions.
Test is running with unstable branch for convenience, but it doesn't mean that we're fine to forget about managing backward compatibility.

@marcosnils 
I mean to remove `clusterNodes` from JedisCluster since its output will be changed unexpected ways. We don't know how many, when, why, and how. It would be troublesome to follow up that changes. 
If we can't remove the command, we should at least remove `ClusterNodeInformationParser` and `ClusterNodeInformation` and users have to parse raw value themselves according to their Redis version.

HeartSaVioR:
FYI, below commit is for Spring Data Redis (maintainer of lettuce works at Pivotal now) to support both formats of `cluster nodes`.

https://github.com/spring-projects/spring-data-redis/commit/f18def013acc4f3706a7dd57059e23d67f36982b

marcosnils:
> I mean to remove clusterNodes from JedisCluster since its output will be changed unexpected ways. We don't know how many, when, why, and how. It would be troublesome to follow up that changes.

@HeartSaVioR Jedis currently returns a `string` for the `cluster nodes` command. So our response type is backwards compatible in terms of Jedis information. If Jedis users are using that command for something they'd have to parse it according to their Redis version but that's something that we can't handle and we don't care. 

The reason why lettuce did that is because it parses the command internally to return an object with all the populated information. I vote for keeping the command with a `string` response until the API becomes stable and then maybe we can return a representative object as lettuce does.

WDYT?

HeartSaVioR:
@marcosnils Ah, thanks! I missed it. Then I also vote for keeping the return value. Just removing parser and information class should be enough.

marcosnils:
@HeartSaVioR :+1: .

@charsyam WDYT?, can you make the remaining changes so we can merge this?

PR #1223, opened 2016-02-22
Disable JMX of GenericObjectPoolConfig for speed

RobinDong:
When running 32 threads and every thread has a JedisCluster(), the
program will spend more than 30 seconds to init 32 JedisCluster()
instances.
The reason is in commons-pool-2.4 :

    GenericObjectPool() --> BaseGenericObjectPool() --> jmxRegister()

the jmxRegister() will spend a lot of time to register MBean because
it always use the same Instance name and have to loop many turns.

Because users can't affort minutes to init multi-instances of
JedisCluster(), I suggest to defaultly disable JMX in JedisCluster().


marcosnils:
@RobinDong JedisCluster is multi-threaded as it uses JedisPool internally. Is there a reason why you're using one instance per thread?

We can't change this default behavior because there might be many Jedis users that are expecting JMX to be enabled by default (as it is today). As you still have the ability to disable JMX for your app only, I don't see the reason to make it a default.

PR #1224, opened 2016-02-24
Master/slave and Elasticache pool implementation

PatrickSauts:
Master/Slave implementation to be able to read from slaves with autodiscovery of master every second.

First Implemented to use with Elasticache but works also with master/slave redis

The configuration also works with a single node.

Trying to write to a slave will trigger an error -> READONLY You can't write against a read only slave. (tested on Redis server v=3.0.7)

```
  String[] cacheAddresses = new String[]{"node1:6379,node2:6379,node3:6379"};
  JedisMasterSlavePool jms = new JedisMasterSlavePool(cacheAddresses, poolConfig, REDIS_TIMEOUT, null, DB_NUM);

  String key = "theTestKey";
  //Writing to the master
  try (Jedis j = jms.getResource()) {
      j.setex(key.getBytes(), ttl, "test".getBytes());
    } catch (Exception e) {
      LOG.error("Error writing to redis", e);
    }

    // Reading from a slave
    try (Jedis j = jms.getReadOnlyResource()) {
      byte[] bs = j.get(key.getBytes());
    } catch (Exception e) {
      LOG.error("Error reading from redis key: {}", key, e);
    }

jms.close()
```

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisElasticachePool.java
+++ b/src/main/java/redis/clients/jedis/JedisElasticachePool.java
@@ -1,0 +1,284 @@
+package redis.clients.jedis;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+
+import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
+
+import redis.clients.jedis.exceptions.JedisException;
+
+public class JedisElasticachePool extends JedisPoolAbstract {
+
+  protected GenericObjectPoolConfig poolConfig;
+
+  protected int connectionTimeout = Protocol.DEFAULT_TIMEOUT;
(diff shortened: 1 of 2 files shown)


marcosnils:
skimming through the code seems like this doesn't need any changes in Jedis at all, so it can easily be extracted it as a library. Please check comments here https://github.com/xetorthio/jedis/issues/1219#issuecomment-187968000

PatrickSauts:
You're absolutely right, that how I use it. But having an already implemented pool inside the driver itself seemed like a good idea to me.

PatrickSauts:
Actually JedisFactory should be removed from the code

marcosnils:
@PatrickSauts what I'm suggesting is for you to create a new library. That way you can decide how to better implement it, which dependencies to use and how it should be tested.

marcosnils:
@PatrickSauts the reason why I'm closing is because even though I'd like to implement as many features as possible in Jedis, the truth is that we're very few contributors and we try to keep it as simple and lightweight as possible. I really like the idea you're bringing, but I strongly believe that if you publish it as a separate library it will be better for the project as it doesn't depend on Jedis at all.

PR #1237, opened 2016-03-10
Don't return JedisConnectionException if pool is exahusted

### LINKED ISSUE #1236: `JedisConnectionException: no reachable node in cluster` error when all objects in pool are in use
### Expected behavior

My primary use case for redis cluster at the moment is as an in-memory cache. Ideally, if redis wasn't available, or if the jedis pool was too busy, we'd just fail immediately and fall back to reading data from our persistent data store.

In particular, I'm running with:

```
// ... stuff
config.setBlockWhenExhausted(false);
```

My hope was that in this configuration, if my java app was already using all of its connections, the new attempt to use the connection would error immediately, and I could catch that error and treat it as if the value didn't exist in the cache and just serve it from our persistent store.

### Actual behavior

What is actually happening is that when there aren't any available `Jedis` instances available, the underlying GenericObjectPool pool is doing `throw new NoSuchElementException("Pool exhausted");`, and Jedis is catching that exception and wrapping it in a `JedisConnectionException: "Could not get a resource from the pool"`. _That_ exception is then caught at `JedisClusterCommand:~120`, and is treated the same as if the cluster is unreachable, or if a MOVED response was received. The `NoSuchElementException`, and `Could not get a resources from the pool` exceptions are also caught without logging, which made it more difficult to debug this issue.

I don't think this issue will show up with the default JedisPoolConfig object, because it has `blockWhenExhausted=true` and `maxWaitMillis=-1`, so it will just wait indefinitely. 

### Steps to reproduce:
For a guaranteed repro, set JedisPoolConfig:
```
config.setBlockWhenExhausted(false);
config.setMaxTotal(1);
```

and then issue calls to jedis in parallel.

(I don't have a working reduced repro case handy, but if it would be helpful I'm happy to whip one up.)

#### Jedis version:
3.0.6
#### Redis version:
2.8.0
#### Java version:
JDK 8

Thanks!

marcosnils:
Fixes #1236 

@roblg @xetorthio @HeartSaVioR PTAL.

### DIFF
--- a/src/main/java/redis/clients/util/Pool.java
+++ b/src/main/java/redis/clients/util/Pool.java
@@ -1,6 +1,7 @@
 package redis.clients.util;
 
 import java.io.Closeable;
+import java.util.NoSuchElementException;
 
 import org.apache.commons.pool2.PooledObjectFactory;
 import org.apache.commons.pool2.impl.GenericObjectPool;
@@ -46,6 +47,8 @@ public abstract class Pool<T> implements Closeable {
   public T getResource() {
     try {
       return internalPool.borrowObject();
+    } catch (NoSuchElementException nse) {
+      throw new JedisException("Could not get a resource from the pool", nse);
     } catch (Exception e) {
       throw new JedisConnectionException("Could not get a resource from the pool", e);
     }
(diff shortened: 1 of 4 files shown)


roblg:
LGTM. :+1: Thanks for the quick turnaround!

marcosnils:
@roblg 2.8.1 will be released tomorrow probably. You can go ahead and build the 2.8.1 branch yourself in case you need this fixed right away.

Issue #1229, opened 2016-03-01
JedisClusterConnectionHandler.initializeSlotsCache can result in two entries for a single redis instance


roblg:
### Expected behavior

It seems like `JedisCluster` should always have exactly one `Jedis` instance per redis server. It should compute its lists of nodes from `cluster slots` (or `cluster nodes` -- I forget which one is recommended now), and it shouldn't add the start nodes, since they should already be in the list anyway.

(i.e., I think the behavior would be correct if these lines were removed from JedisClusterConnectionHandler line ~50:

```
    for (HostAndPort node : startNodes) {
      cache.setNodeIfNotExist(node);
    }
```

### Actual behavior

`JedisClusterConnectionHandler.initializeSlotsCache` initializes the slots cache by querying the redis server, and _also_ adds whatever `startNodes` it was given. If the start nodes were given as hostnames, this will result in two entries for each start node: one for the host name, and one for the IP address reported by `cluster slots`. Queries against the cluster continue to work, but iterating over the nodes in the cluster (e.g., for aggregating per-node stats) also includes these bogus entries.

### Steps to reproduce:

Our setup:

1. Two redis instances running on separate machines, named `redis-1` and `redis-2`
2. Java application running on a third machine: `jobs-1`. Specifies cluster start nodes as `redis-1:6379`, `redis-2:6379`.
3. Iterate through pool nodes with:
```java
for (JedisPool pool : cluster.getClusterNodes().values()) {
    try (Jedis jedis = pool.getResource()) {
        // do stuff
    } catch (Exception e) {
        // we hit this exception any time `pool.getResource()` returns one of the "phantom" nodes w/ hostname instead of IP 
    }
}
```

We have a workaround that allows us to specify per-machine hostnames (we just us `InetAddress` to do a DNS lookup of the IP before we send it to jedis), BUT it would be really nice to be able to pass the host name of a load-balancer pool to Jedis for service discovery. It could make an initial query to that machine, learn all the node info from `cluster slots`, and that way we wouldn't need any hard-coded IP address or machine names in our configuration.

If this seems like a reasonable approach, LMK and I'll attempt a PR.

Thanks!

#### Jedis version:

2.8.0

#### Redis version:

3.0.x

#### Java version:

8.x

marcosnils:
@roblg thanks for reporting this issue. I see the situation you're facing and I totally agree that it's something that we need solve from the Jedis perspective. 

[...]

PR #1239, opened 2016-03-14
Attempted fix for #1229 - multiple entries in node list for same host

roblg:
Hey all, this is my attempt at a fix for #1229. It sounded like the approach seemed reasonable, so I thought I'd take a stab at it.

Tests pass for me locally against the unstable branch of redis. I tried to add a new test for my specific case where equality check in HostAndPort fails, so multiple nodes end up in the list, but there's already code to convert `127.0.0.1` --> `localhost` in `HostAndPort`, so it wasn't clear how I'd be able to test my particular situation safely in many different test environments.

I based my fix on the 2.8 branch, but I can switch it to the 2.9 or master branches if that makes more sense.

Thanks for considering!


marcosnils:
@roblg thanks for submitting a patch. I believe it's heading in the proper direction. In order to test this you might try to setup JedisCluster like this:

```java
    jedisClusterNode.add(new HostAndPort("localhost", 7379));
    jedisClusterNode.add(new HostAndPort("some_other_host", 7379));
    // Assert on jc.getClusterNodes().size()
```
With the previous code JedisCluster will populate all the hosts in the cluster plus "some_other_host" (even though it's invalid). Your test needs to assert that supplied hosts are not added into the node list.

roblg:
@marcosnils good call. I realized while I was writing the "invalid host" test that I was mistaken about how the 127 vs. localhost equality was happening. I actually got the "broken" behavior if I created a `new HostAndPort("localhost", 7379)`, so I included a test for that as well.

PR updated.

marcosnils:
@roblg LGTM. I'd like @HeartSaVioR or @xetorthio to review before merging. 

I'd also like to schedule this for 2.9 as it changes the way JedisCluster is initializing. Even though I believe it "shouldn't" cause any problems, I propose to move it to a minor release just in case

HeartSaVioR:
@roblg 
LGTM. Thanks for the work!
Btw, it would be more sense to move to master branch. We can take care of backporting if necessary.

roblg:
@HeartSaVioR @marcosnils would you like me to update the PR to be based on master?

marcosnils:
@roblg yes please. We'll take care to downmerge to 2.9

HeartSaVioR:
Yes, if you don't mind. Thanks in advance!

HeartSaVioR:
@marcosnils You're really fast! lol

marcosnils:
@HeartSaVioR :tongue: going to bed. Can you please merge once changes are made?. Thx!.

HeartSaVioR:
@roblg Ah, one thing we didn't say... PR should be re-created against master. It's a limitation of Github PR.

roblg:
@HeartSaVioR yeah that kind of surprised me. I've been using Bitbucket at work for several years. Got it sorted out though (I hope).

Issue #1248, opened 2016-04-01
Jedis cluster info rediscovering queueing on high load


Spikhalskiy:
Jedis queues cluster rediscovering using lock currently. It could cause unpredictable latencies in calls to Jedis + useless increased load to redis nodes. We got 5-6 seconds with soTimeout 5ms and huge amount of threads stucked in rediscovering and acquiring lock.

### Expected behavior

If we have thousands requests to JedisCluster and we got MOVED from Redis Cluster on 500 of them in a very short time - we should rediscover cluster only once.  

### Actual behavior

It's possible that we will rediscover cluster 500 times and everything will wait.
We had like 5 seconds waiting with redis in local network for sure.

### Steps to reproduce

Perhaps, create fresh cluster and start fill it with very high volume with large number of threads from multiple client to stimulate slots moving - 

#### Jedis version:

2.8.1 and current master

#### Redis version:

3.0.0

#### Code investigation

Let's imagine, we have 1000 threads writing with Jedis and 200 threads got MOVED on the same time. After that all of them will call JedisClusterCommand#137 ```this.connectionHandler.renewSlotCache(connection);``` and all of them got to JedisClusterInfoCache#discoverClusterSlots. As a result, all of them one by one will require lock and rebuild cluster info. Threads that will try to do something even with non-moved slots will wait on acquiring read lock in JedisClusterInfoCache.

#### Proposed solution

1) Change lock structure in JedisClusterInfoCache#discoverClusterNodesAndSlots and JedisClusterInfoCache#discoverClusterSlots to something like...

```
AtomicBoolean rebuilding = new AtomicBoolean();

public void discoverClusterSlots(Jedis jedis) {
    if (rebuilding.get()) {
      r.lock(); //wait end of rebuilding by another process
      r.unlock();
    } else {
      w.lock();
      rebuilding.set(true);
      try {
        //old actual discovering code 
      } finally {
        rebuilding.set(false);
        w.unlock();
      }
    }
}
```

2) Use information from MOVED and move only one slot; don't make additional external calls to redis and rediscover full cluster info in acquired lock.

marcosnils:
@Spikhalskiy nice finding. Would you like to craft a PR with the solution you proposed?. I believe it's the right way to proceed.

Spikhalskiy:
@marcosnils Yes, if you think proposal looks fine, I would prepare PR.

marcosnils:
[...]

PR #1249, opened 2016-04-03
Issue #1248: Avoid queueing multiple cluster discovers of the same type

Spikhalskiy:
Address issue with queueing multiple JedisCluster state rediscovering from Issue #1248

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
@@ -22,6 +22,7 @@ public class JedisClusterInfoCache {
   private final ReentrantReadWriteLock rwl = new ReentrantReadWriteLock();
   private final Lock r = rwl.readLock();
   private final Lock w = rwl.writeLock();
+  private volatile boolean rediscovering;
   private final GenericObjectPoolConfig poolConfig;
 
   private int connectionTimeout;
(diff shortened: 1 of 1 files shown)


src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -22,6 +22,7 @@
   private final ReentrantReadWriteLock rwl = new ReentrantReadWriteLock();
   private final Lock r = rwl.readLock();
   private final Lock w = rwl.writeLock();
+  private volatile int currentRediscoverProcessType;
```
marcosnils: Can we make this a private ENUM so we can better name the discovering type?

marcosnils:
@Spikhalskiy does it make sense to have two discovering types? Because `discoverClusterNodesAndSlots` only happens on JedisCluster initialization. After that only `discoverClusterSlots` is called

Spikhalskiy:
I just think that code with flag only is very error prone. Somebody could add new discovering on any other case (or make discoverClusterNodesAndSlots not only on startup time), copy-paste logic with only flag and doesn't think that this logic became broken (discover task of one type cancel task of another type). So no, it's doesn't make much sense now, but it could make sense if future - I hope somebody will take a look at different types and think why they have been added. But I don't have strong opinion, if you think that it should be fine as boolean flag - I could rework it.

marcosnils:
@Spikhalskiy I understand your opinion and I agree with it to some point. I don't like to add code "just in case something happens in the future" as it may never happen and you have stuff that might confuse someone how's reading the code. 

I prefer to keep thing simple and use a descriptive boolean flag as `discovering` just for `discoverClusterSlots` for the moment. If we need to extend it to use different types we can definitely do it in the future, but for the moment I prefer to keep it simple and narrowed to the desired functionality.

Spikhalskiy:
@marcosnils got it, addressed.

marcosnils:
@Spikhalskiy LGTM. Once another contributor reviews we can mearge. 

@HeartSaVioR ?

HeartSaVioR:
LGTM. Nice finding @Spikhalskiy and thanks for crafting patch.

HeartSaVioR:
Merged into branches master, 2.8, 2.9 respectively!

Spikhalskiy:
@HeartSaVioR @marcosnils Great, thanks for review.

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -79,36 +80,41 @@ public void discoverClusterNodesAndSlots(Jedis jedis) {
   }
 
   public void discoverClusterSlots(Jedis jedis) {
-    w.lock();
+    //If rediscovering is already in process - no need to start one more same rediscovering, just return
+    if (!rediscovering) {
+      w.lock();
+      rediscovering = true;
```
roblg: Late to the party, but a possible perf enhancement here might be to check `rediscovering` again:

```
if (!rediscovering) {
  w.lock();
  if (rediscovering) {
    return;
  }
```

That way, if multiple threads make it past the initial check when under high load and block at lock acquisition, only the first one will do the work.

Spikhalskiy: It shouldn't help. We set it on true in lock and to false before leaving lock, so if we got write lock - it's always false in this var.

marcosnils: I guess it makes sense. Is there a cleaner way to do this instead of repeating the condition twice?

roblg: Ah, right. This _looks_ like a candidate for a double-checked locking pattern, but because we're resetting `rediscovering` it won't work. Darn. It'd be nice to be able to get rid of those extra rediscovers.

(And this, friends, is why I shouldn't read PRs before coffee.)

marcosnils: :+1:

PR #1251, opened 2016-04-05
Cleanup cache state from failed attempts of cluster discovering

Spikhalskiy:
Cleanup cache state from failed attempts of cluster discovering

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisClusterConnectionHandler.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterConnectionHandler.java
@@ -1,5 +1,6 @@
 package redis.clients.jedis;
 
+import java.io.Closeable;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
@@ -10,7 +11,7 @@ import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
 
 import redis.clients.jedis.exceptions.JedisConnectionException;
 
-public abstract class JedisClusterConnectionHandler {
+public abstract class JedisClusterConnectionHandler implements Closeable {
   protected final JedisClusterInfoCache cache;
 
   public JedisClusterConnectionHandler(Set<HostAndPort> nodes,
@@ -71,6 +72,11 @@ public abstract class JedisClusterConnectionHandler {
     } catch (JedisConnectionException e) {
       renewSlotCache();
     }
+  }
+
+  @Override
+  public void close() {
+    cache.reset();
   }
 
   protected List<JedisPool> getShuffledNodesPool() {
--- a/src/main/java/redis/clients/jedis/BinaryJedisCluster.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedisCluster.java
@@ -54,15 +54,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
   @Override
   public void close() {
     if (connectionHandler != null) {
-      for (JedisPool pool : connectionHandler.getNodes().values()) {
-        try {
-          if (pool != null) {
-            pool.destroy();
-          }
-        } catch (Exception e) {
-          // pass
-        }
-      }
+      connectionHandler.close();
     }
   }
 
--- a/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
@@ -16,8 +16,8 @@ import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
 import redis.clients.util.SafeEncoder;
 
 public class JedisClusterInfoCache {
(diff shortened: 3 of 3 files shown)


src/main/java/redis/clients/jedis/JedisClusterConnectionHandler.java
```diff
@@ -34,12 +34,22 @@ public Jedis getConnectionFromNode(HostAndPort node) {
...
-      Jedis jedis = new Jedis(hostAndPort.getHost(), hostAndPort.getPort());
+      Jedis jedis = null;
       try {
+        jedis = new Jedis(hostAndPort.getHost(), hostAndPort.getPort());
         cache.discoverClusterNodesAndSlots(jedis);
         break;
       } catch (JedisConnectionException e) {
-        // try next nodes
+        // close already created resource and try next nodes
+        for (JedisPool pool : cache.getNodes().values()) {
```
marcosnils: Why are you destroying all cluster pools if just one node is incorrect?

marcosnils:
@Spikhalskiy I wouldn't consider a problematic node a node which you can't connect to.

What if the node is down at the moment of initializing JedisCluster and some other cluster node has it on it's `cluster slots`?. It would totally make sense to add it to the cluster nodes list as JedisCluster will try to connect to it whenever it comes back online

src/main/java/redis/clients/jedis/JedisClusterConnectionHandler.java
Spikhalskiy: Imagine situation. We got first node, we discover something and fails in the middle. Now we "try next node" and just make .clear() .clear() on both collection in cache because we will discover everything from this next node. Without releasing allocated resources, connections, etc. So, it's just this releasing.

Spikhalskiy:
@marcosnils No, we will fail on new Jedis(), throw exception and discover just nothing now.

marcosnils:
@Spikhalskiy ?? I've just tried your test with Jedis master branch and it passes. I'm not sure what this is fixing.

Spikhalskiy:
@marcosnils Yes, you are right. ```new Jedis()``` itself doesn't throw exception because doesn't create connection, had different problem. On the same time - cleaning after getting exception is still valid. Now we on "next node" will .clear() possibly non-empty collections without releasing resources. Rework PR without moving ```new Jedis()``` inside try-catch and unit test?

marcosnils:
@Spikhalskiy I see what you're saying but I'm still thinking why we need to destroy pool on initialization. 

If `nodes.clear()` and `slots.clear()` gets called then the garbage collection should take care cleaning up resources and any possible established socket connections. Why do we need to make it explicitly?

Spikhalskiy:
@marcosnils I think just the same reason [why](http://www.informit.com/articles/article.aspx?p=1216151&seqNum=7) we should explicitly close resources and don't rely on finalizers and when it will be called if even called. I don't know, if closing sockets sometimes on gc is fine - no idea why do we have all this BinaryJedis#quit() code.

marcosnils:
@Spikhalskiy seems reasonable. It doesn't hurt to be a bit defensive sometimes. 

LGTM

Thanks for contributing!.

marcosnils:
@HeartSaVioR ping.

Spikhalskiy:
Maybe this invalidation code should be incapsulated in JedisClusterInfoCache method like... "clear" or "invalidate". Because we already do exactly the same in BinaryJedisCluster#close

src/main/java/redis/clients/jedis/JedisClusterConnectionHandler.java
HeartSaVioR: @Spikhalskiy 
I don't think handler is responsible of destroying the pools in JedisClusterInfoCache. If we need to destroy, it would be better to handle in JedisClusterInfoCache, or at least it should be extracted to method of JedisClusterInfoCache.

Ideally we should destroy pool when calling `this.nodes.clear();`. Since we only clear nodes when clearing all states, we can extract this behavior into new method, maybe destroy() or reset().
Definitely it also applies to BinaryJedisCluster. Let cache destroy its pools by itself.

If you think it's beyond the boundary of pull request, we can merge as is and I'll submit new pull request.

HeartSaVioR:
@Spikhalskiy I didn't see your last comment before commenting. What I commented is what you commented. I definitely agree.

Spikhalskiy:
@marcosnils @HeartSaVioR moved to separate method

HeartSaVioR:
Great, LGTM
@marcosnils Please review the change. :)

marcosnils:
even better now. LGTM. 

@Spikhalskiy amazing job!.

Spikhalskiy:
@marcosnils @HeartSaVioR Thank you for review, guys!

HeartSaVioR:
Merged into master, 2.8, 2.9 respectively.
@Spikhalskiy Thanks for the great work!

Issue #1252, opened 2016-04-06
Snowball effect with reconnecting to poor performing node


Spikhalskiy:
We have a problem with JedisPool that aggravates perf issues of redis nodes when this issue only start to appear.

What we have:
- Short timeouts. Like 3ms.
- Significantly loaded redis that sometimes starts to respond slowly with number of connections like... 40000 per node.

If redis is starting to stuck for 4ms, instead of each one read, we start to do
- read
- after timeout and marking Jedis as broken, JedisFactory gently sends quit to Redis in destroyObject
- we establish new connection
- PING-PONG
and only after that we have new Jedis instance for new read, but... actually nothing changed.

So, when our Redis Cluster start to experience some perf issues - we finish it off by invalidating Jedis.

Any thoughts?
One one from me - maybe we could add an ability to pass some type of "InvalidationStrategy" to Jedis? For example, strategy by default will mark as broken and do everything like now and 3rd party can implement it's own strategy, for example, send PING-PONG before quit. "read with timeout - PING-PONG, give it a chance - read" looks better than current mandatory invalidation flow.

I could implement and provide PR for any solution solving or providing possibility to improve current standard flow.

What do you think?

Spikhalskiy:
Final version which is working in our prod and do it fine: https://github.com/Spikhalskiy/jedis/releases/tag/PP-2

It's just master with merged related pull requests.
Issue could be closed after merging PRs to upstream.

marcosnils:
#1249 #1251 #1253 #1256 

@Spikhalskiy amazing contribution. I've had a rough weeks lately. As soon as I have some time I promise to look at those changes.

PR #1253, opened 2016-04-06
Issue #1252: Fix creating a lot of new Jedis instances on unstable cluster, fix slots clearing without filling

Spikhalskiy:
I started to investigate Issue #1252 and PING-PONG counts especially.

The most interesting things I found in stack traces on our runtime:
```
	at redis.clients.jedis.BinaryJedis.ping(BinaryJedis.java:106)
	at redis.clients.jedis.JedisSlotBasedConnectionHandler.getConnection(JedisSlotBasedConnectionHandler.java:41)
	at redis.clients.jedis.JedisSlotBasedConnectionHandler.getConnectionFromSlot(JedisSlotBasedConnectionHandler.java:64)
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:115)
	at redis.clients.jedis.JedisClusterCommand.run(JedisClusterCommand.java:30)
	at redis.clients.jedis.JedisCluster.setex(JedisCluster.java:292)
```

Looks absolutely abnormal - we have nothing for slot and fallback to new Jedis to random node.

Here you can find my changes and fixes for place, that could cause this issue + some refactoring, remove of some duplicate map lookups/keys building.

System behavior before and after applying this changes:
Before
After

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
@@ -66,7 +66,7 @@ public class JedisClusterInfoCache {
           }
 
           HostAndPort targetNode = generateHostAndPort(hostInfos);
-          setNodeIfNotExist(targetNode);
+          setupNodeIfNotExist(targetNode);
           if (i == MASTER_NODE_INDEX) {
             assignSlotsToNode(slotNums, targetNode);
           }
@@ -84,21 +84,24 @@ public class JedisClusterInfoCache {
       rediscovering = true;
 
       try {
-        this.slots.clear();
-
         List<Object> slots = jedis.clusterSlots();
+
+        // We should clear slots after getting result about cluster slots, because if we got exception or timeout in
+        // getting new slots state, which is possible, because it's an external operation - we could got empty slots collection
+        // and ton of new Jedis instances for random nodes, see JedisSlotBasedConnectionHandler#getConnectionFromSlot
+        this.slots.clear();
 
         for (Object slotInfoObj : slots) {
           List<Object> slotInfo = (List<Object>) slotInfoObj;
 
-          if (slotInfo.size() <= 2) {
+          if (slotInfo.size() <= MASTER_NODE_INDEX) {
             continue;
           }
 
           List<Integer> slotNums = getAssignedSlotArray(slotInfo);
 
           // hostInfos
-          List<Object> hostInfos = (List<Object>) slotInfo.get(2);
+          List<Object> hostInfos = (List<Object>) slotInfo.get(MASTER_NODE_INDEX);
           if (hostInfos.isEmpty()) {
             continue;
           }
@@ -106,7 +109,6 @@ public class JedisClusterInfoCache {
           // at this time, we just use master, discard slave information
           HostAndPort targetNode = generateHostAndPort(hostInfos);
 
-          setNodeIfNotExist(targetNode);
           assignSlotsToNode(slotNums, targetNode);
         }
       } finally {
@@ -121,45 +123,51 @@ public class JedisClusterInfoCache {
         ((Long) hostInfos.get(1)).intValue());
   }
 
-  public void setNodeIfNotExist(HostAndPort node) {
+  public JedisPool setupNodeIfNotExist(HostAndPort node) {
     w.lock();
     try {
       String nodeKey = getNodeKey(node);
-      if (nodes.containsKey(nodeKey)) return;
+      JedisPool existingPool = nodes.get(nodeKey);
+      if (existingPool != null) return existingPool;
 
       JedisPool nodePool = new JedisPool(poolConfig, node.getHost(), node.getPort(),
           connectionTimeout, soTimeout, null, 0, null, false, null, null, null);
       nodes.put(nodeKey, nodePool);
+      return nodePool;
     } finally {
       w.unlock();
     }
   }
 
-  public void setNodeIfNotExist(HostAndPort node, boolean ssl) {
+  public JedisPool setupNodeIfNotExist(HostAndPort node, boolean ssl) {
     w.lock();
     try {
       String nodeKey = getNodeKey(node);
-      if (nodes.containsKey(nodeKey)) return;
+      JedisPool existingPool = nodes.get(nodeKey);
+      if (existingPool != null) return existingPool;
 
       JedisPool nodePool = new JedisPool(poolConfig, node.getHost(), node.getPort(),
           connectionTimeout, soTimeout, null, 0, null, ssl, null, null, null);
       nodes.put(nodeKey, nodePool);
+      return nodePool;
     } finally {
       w.unlock();
     }
   }
 
-  public void setNodeIfNotExist(HostAndPort node, boolean ssl, SSLSocketFactory sslSocketFactory,
-      SSLParameters sslParameters, HostnameVerifier hostnameVerifier) {
+  public JedisPool setupNodeIfNotExist(HostAndPort node, boolean ssl, SSLSocketFactory sslSocketFactory,
+                                  SSLParameters sslParameters, HostnameVerifier hostnameVerifier) {
     w.lock();
     try {
       String nodeKey = getNodeKey(node);
-      if (nodes.containsKey(nodeKey)) return;
+      JedisPool existingPool = nodes.get(nodeKey);
+      if (existingPool != null) return existingPool;
 
       JedisPool nodePool = new JedisPool(poolConfig, node.getHost(), node.getPort(),
           connectionTimeout, soTimeout, null, 0, null, ssl, sslSocketFactory, sslParameters,
           hostnameVerifier);
       nodes.put(nodeKey, nodePool);
+      return nodePool;
     } finally {
       w.unlock();
     }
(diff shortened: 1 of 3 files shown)


src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -86,21 +86,24 @@ public void discoverClusterSlots(Jedis jedis) {
       rediscovering = true;
 
       try {
-        this.slots.clear();
-
         List<Object> slots = jedis.clusterSlots();
 
+        // We should clear slots after getting result about cluster slots, because if we got exception or timeout in
```
Spikhalskiy: most important part here

src/main/java/redis/clients/jedis/JedisSlotBasedConnectionHandler.java
```diff
@@ -61,7 +61,14 @@ public Jedis getConnectionFromSlot(int slot) {
       // assignment
       return connectionPool.getResource();
     } else {
-      return getConnection();
+      renewSlotCache(); //It's abnormal situation for cluster mode, that we have just nothing for slot, try to rediscover state
```
Spikhalskiy: And here

Spikhalskiy:
I'm going to put it to prod load today, will let you know if it helps.

src/main/java/redis/clients/jedis/JedisSlotBasedConnectionHandler.java
HeartSaVioR: It seems nice cause renewSlotCache() should succeed to renew slot cache if cluster is not down. Great improvement.

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
HeartSaVioR: At first, when we fix JedisSlotBasedConnectionHandler#getConnectionFromSlot, this comment becomes invalid.

And IMO, clearing slots first or later is up to amount of slot changeset. For example, if most of slots are changed, it would be better to try discovering slots again instead of relying on outdated slot information. If only a little of slots are changed, we should take latter. It would be not a big deal cause we don't do random connect later.

Spikhalskiy: > JedisSlotBasedConnectionHandler#getConnectionFromSlot, this comment becomes invalid.
Not completely.
1) First thread started to discover. Clear nodes. Release lock while going to next node.
2) On this time second thread went to JedisSlotBasedConnectionHandler#getConnectionFromSlot, very possible that we could just use old pool, but we see empty collection and connect to discovering party.
I would prefer to don't publish empty collection while we can leave old one.

> And IMHO, clearing slots first or later is up to amount of slot changeset
I'm not sure if I get you. We clear it completely anyways. I just moved clearing after external code. So, to avoid situation of exception in external code, which leads to publishing empty slots collection.

HeartSaVioR: >> And IMHO, clearing slots first or later is up to amount of slot changeset

> I'm not sure if I got you. We clear it completely anyways. I just moved clearing after external code. So, to avoid situation of exception in external code, which leads to publishing empty slots collection.

Imagine first trial of clusterSlots() failed (so lock is released), and next one calls getConnectionFromSlot(). If we don't clear slots, it will get Jedis instance from cache which could be invalidated (cause we fails to update so doesn't catch up latest information) or not (node for slot is not changed). Former will try to request wrong node, and receive MOVED, which also initiate slot cache update. If we clear cache first, it doesn't need to request but just update slot cache immediately.

Spikhalskiy: Got you. And I'm not sure what is better... It could get right node and don't stuck at all. In worst case - we will get same result with rediscovering call. I would prefer to give a chance and think that more slots is still on place. But if you think that it's better to publish clear state - can revert this part.

HeartSaVioR: Maybe we want to hold a lock during trying all nodes. 
I see what you said, we release the write lock before completing slot cache update so there's race again.

Spikhalskiy: It should be more stable behavior and makes sense. I will take a look how to implement it.

Spikhalskiy:
@HeartSaVioR reworked with one lock for trying all nodes

marcosnils:
I'll check this and #1252 later today.

HeartSaVioR:
@Spikhalskiy Github introduces new feature - 'squash before merging' so all you need to consider is rebasing to catch up with master.
I'll review later today.

Spikhalskiy:
I don't like this feature, because it merges without merge commits :) So, it's impossible to continue branch and merge once more. But you merge in multiple branches - so it's cherry-pick in any case.

Spikhalskiy:
This helped. We get rid of this type of error flow, things became better. But! found one more broken flow :+1:

After merging this PR will start to think what to do with this another flow.
Shortly:
```
	at redis.clients.jedis.BinaryJedis.ping(BinaryJedis.java:106)
	at redis.clients.jedis.JedisSlotBasedConnectionHandler.getConnection(JedisSlotBasedConnectionHandler.java:41)
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:113)
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:131)
	at redis.clients.jedis.JedisClusterCommand.run(JedisClusterCommand.java:30)
	at redis.clients.jedis.JedisCluster.setex(JedisCluster.java:292)
```

We get timeout. After that we decide to try random node 0_o. It's already strange. We don't use random pool, but create new Jedis with PING-PONG 0_o. On this random node we get MOVED and... starting cluster rediscovering. I even don't know which place in this chain to fix firstly. Suggestions?

Spikhalskiy:
Could we finalize it today to make further fixing easier?

HeartSaVioR:
@Spikhalskiy Even better. LGTM.
@marcosnils Could you review new changeset?

HeartSaVioR:
Please note that I didn't look deeply at your last comment. Sorry I don't have time to concentrate. I'll try to have a look and comment later.

Spikhalskiy:
@HeartSaVioR Thank you! Sure, take your time. When I take a look deeply at code and possible solutions I will share my suggestions too in separate PR.

marcosnils:
@Spikhalskiy @HeartSaVioR I'll take care of this today after work.

amazing job!.

Spikhalskiy:
@HeartSaVioR @marcosnils Any updates here?

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -116,50 +116,81 @@ public void discoverClusterSlots(Jedis jedis) {
...
   }
 
+  private void discoverClusterSlots(Jedis jedis) {
+    List<Object> slots = jedis.clusterSlots();
+    this.slots.clear();
+
+    for (Object slotInfoObj : slots) {
+      List<Object> slotInfo = (List<Object>) slotInfoObj;
+
+      if (slotInfo.size() <= MASTER_NODE_INDEX) {
```
xetorthio: Why are we doing this? Masters could not have replicas, right?

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -116,50 +116,81 @@ public void discoverClusterSlots(Jedis jedis) {
...
+        continue;
+      }
+
+      List<Integer> slotNums = getAssignedSlotArray(slotInfo);
+
+      // hostInfos
+      List<Object> hostInfos = (List<Object>) slotInfo.get(MASTER_NODE_INDEX);
+      if (hostInfos.isEmpty()) {
```
xetorthio: I don't think it is possible for redis clusterSlots command not to return the node IP and port, it would be a redis bug

xetorthio:
Checked your code and I understand it, but I have hard time to relate it to the issue because of all the refactoring. Would you mind describing briefly what this code is fixing?

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
Spikhalskiy: It's old code "if (slotInfo.size() <= 2) {", which is absolutely not related to PR. You can ask original author.

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
Spikhalskiy: It's refactored part of old code "if (hostInfos.isEmpty()) {
 -            continue;" which is not related to PR, I just made it more readable.

Spikhalskiy:
@xetorthio For your understanding I separated PR to two commits for review. Main part is here. https://github.com/xetorthio/jedis/pull/1253/commits/a5961d0685ca2f5472da06008501a863f787f6d7 I even underline 2 main places there by comments (https://github.com/xetorthio/jedis/pull/1253#discussion_r58701333 and https://github.com/xetorthio/jedis/pull/1253#discussion_r58701358).
Second commit is refactoring and moving lock to the top level https://github.com/xetorthio/jedis/pull/1253/commits/ec87a52efa87d52817056e40b0e0e1e62267f250 suggested here https://github.com/xetorthio/jedis/pull/1253#discussion_r58717259

xetorthio:
Got you! After your clarifications and checking again the code, ***I think it is OK to merge this***.

I would make further changes down the road since the code (not yours) is hard to understand. Mainly `clusterSlots` command return is extremely ugly, making it super hard to use and even worse to under the code.

Another suggestion for future improvements is to do the cluster slot renewal in a background thread, which should make the locking unnecessary.

Created issues to discuss this: #1346 and #1347

marcosnils:
Downmerged to 2.8 and 2.9 respectively. Thx again everyone involved to make this happen.

PR #1256, opened 2016-04-13
Issue #1252: Random node + rediscovery on connection exception replaced with rediscovery at the end

Spikhalskiy:
Type of address problem with flow "just one timeout -> random node for nothing -> rediscovering" on each timed out request for Issue #1252:

1. Don’t try random node with immediate initiate cluster renewal after that if we got ConnectionTimeout from proper slot jedis. Try proper slot until last attempt, if last attempt failed - just make cluster rediscovery, without trying random nodes which is useless.
2. New special exception for “No reachable nodes”
3. Fixed situation when simple connection exception with tryRandomNode=true proceed as just no reachable nodes
4. Fix rewriting real root cause JedisConnectionException with JedisClusterMaxRedirectionsException

This PR mostly address my comment on another PR https://github.com/xetorthio/jedis/pull/1253#issuecomment-206480244
And should be merged after merging this more critical PR https://github.com/xetorthio/jedis/pull/1253

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisClusterCommand.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterCommand.java
@@ -15,12 +15,12 @@ public abstract class JedisClusterCommand<T> {
   private static final String NO_DISPATCH_MESSAGE = "No way to dispatch this command to Redis Cluster.";
 
   private JedisClusterConnectionHandler connectionHandler;
-  private int redirections;
+  private int maxAttempts;
   private ThreadLocal<Jedis> askConnection = new ThreadLocal<Jedis>();
 
-  public JedisClusterCommand(JedisClusterConnectionHandler connectionHandler, int maxRedirections) {
+  public JedisClusterCommand(JedisClusterConnectionHandler connectionHandler, int maxAttempts) {
     this.connectionHandler = connectionHandler;
-    this.redirections = maxRedirections;
+    this.maxAttempts = maxAttempts;
   }
 
   public abstract T execute(Jedis connection);
@@ -30,7 +30,7 @@ public abstract class JedisClusterCommand<T> {
       throw new JedisClusterException(NO_DISPATCH_MESSAGE);
     }
 
-    return runWithRetries(SafeEncoder.encode(key), this.redirections, false, false);
+    return runWithRetries(SafeEncoder.encode(key), this.maxAttempts, false, false);
   }
 
   public T run(int keyCount, String... keys) {
@@ -51,7 +51,7 @@ public abstract class JedisClusterCommand<T> {
       }
     }
 
-    return runWithRetries(SafeEncoder.encode(keys[0]), this.redirections, false, false);
+    return runWithRetries(SafeEncoder.encode(keys[0]), this.maxAttempts, false, false);
   }
 
   public T runBinary(byte[] key) {
@@ -59,7 +59,7 @@ public abstract class JedisClusterCommand<T> {
       throw new JedisClusterException(NO_DISPATCH_MESSAGE);
     }
 
-    return runWithRetries(key, this.redirections, false, false);
+    return runWithRetries(key, this.maxAttempts, false, false);
   }
 
   public T runBinary(int keyCount, byte[]... keys) {
@@ -80,7 +80,7 @@ public abstract class JedisClusterCommand<T> {
       }
     }
 
-    return runWithRetries(keys[0], this.redirections, false, false);
+    return runWithRetries(keys[0], this.maxAttempts, false, false);
   }
 
   public T runWithAnyNode() {
@@ -95,8 +95,8 @@ public abstract class JedisClusterCommand<T> {
     }
   }
 
-  private T runWithRetries(byte[] key, int redirections, boolean tryRandomNode, boolean asking) {
-    if (redirections <= 0) {
+  private T runWithRetries(byte[] key, int attempts, boolean tryRandomNode, boolean asking) {
+    if (attempts <= 0) {
       throw new JedisClusterMaxRedirectionsException("Too many Cluster redirections?");
     }
 
@@ -128,10 +128,10 @@ public abstract class JedisClusterCommand<T> {
       releaseConnection(connection);
       connection = null;
 
-      if (redirections <= 1) {
+      if (attempts <= 1) {
         //We need this because if node is not reachable anymore - we need to finally initiate slots renewing,
         //or we can stuck with cluster state without one node in opposite case.
-        //But now if maxRedirections = 1 or 2 we will do it too often. For each time-outed request.
+        //But now if maxAttempts = 1 or 2 we will do it too often. For each time-outed request.
         //TODO make tracking of successful/unsuccessful operations for node - do renewing only
         //if there were no successful responses from this node last few seconds
         this.connectionHandler.renewSlotCache();
@@ -140,7 +140,7 @@ public abstract class JedisClusterCommand<T> {
         throw jce;
       }
 
-      return runWithRetries(key, redirections - 1, tryRandomNode, asking);
+      return runWithRetries(key, attempts - 1, tryRandomNode, asking);
     } catch (JedisRedirectionException jre) {
       // if MOVED redirection occurred,
       if (jre instanceof JedisMovedDataException) {
@@ -161,7 +161,7 @@ public abstract class JedisClusterCommand<T> {
         throw new JedisClusterException(jre);
       }
 
-      return runWithRetries(key, redirections - 1, false, asking);
+      return runWithRetries(key, attempts - 1, false, asking);
     } finally {
       releaseConnection(connection);
     }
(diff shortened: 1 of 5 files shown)


src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -117,18 +118,29 @@ private T runWithRetries(byte[] key, int redirections, boolean tryRandomNode, bo
...
       // release current connection before recursion
       releaseConnection(connection);
       connection = null;
 
-      // retry with random connection
-      return runWithRetries(key, redirections - 1, true, asking);
+      if (redirections <= 1) {
+        //We need this because if node is not reachable anymore - we need to finally initiate slots renewing,
+        //or we can stuck with cluster state without one node in opposite case.
+        //But now if maxRedirections = 1 or 2 we will do it too often. For each time-outed request.
```
roblg: > //But now if maxRedirections = 1 or 2 we will do it too often.

This comment worries me a little. The redis cluster my company runs is being used as a cache -- all master nodes, in-memory only. When a node goes down, we just want those requests to fail as fast as possible until we bring it back up or rebalance, so we're currently setting maxRedirects to 1, and setting very low socket connect/read timeouts. With this change, any time a client has a socket timeout, we'll renew the slot cache, which seems suboptimal (particularly because in our case, nothing will have changed).

Spikhalskiy: > With this change, any time a client has a socket timeout, we'll renew the slot cache, which seems suboptimal (particularly because in our case, nothing will have changed).

roblg: I _think_ what's happening to us when we hit this timeouts right now is that we fail with a `JedisConnectException` when `runWithRetries` is called the first time, and then we call `runWithRetries` recursively with `redirections = 0`, which immediately throws a `JedisClusterMaxRedirectionsException`. We'll never get a `JedisMovedDataException`, so we'll never actually try to renew the slot cache. (I think; it's been awhile since I stepped through this code.)

Spikhalskiy: > and then we call runWithRetries recursively with redirections = 0

I see another. We will call recursively with redirections-1 and tryRandom=true. So, next call going to random node, we getting MEVED and making rediscovering EACH timeout. See stacktrace attached to my comment https://github.com/xetorthio/jedis/pull/1253#issuecomment-206480244.

Spikhalskiy: And it's better than we don't do rediscovering at all and just fail. We will never discover slave became master after master fail in your schema.

roblg: :( Ah, I see. Yeah, we were experiencing that too initially, which is ultimately why we set `maxRedirects=1` (so we wouldn't get bounced around to the other nodes, which definitely don't have the data).

Spikhalskiy: Yep, you will not experience it with maxRedirects=1, but for us it's bad solution and what's going on with non values > 1 looks strange)

Do you really want to get rid of cluster rediscovering on timeouts if you have maxRedirects=1?
I could implement solution with tracking success/failures and make rediscovering rarely than 1 per... some time like 5s and only if all requests failed. But it will be more complicated and required idea review firstly.

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -117,18 +118,29 @@ private T runWithRetries(byte[] key, int redirections, boolean tryRandomNode, bo
...
 
       return execute(connection);
-    } catch (JedisConnectionException jce) {
-      if (tryRandomNode) {
-        // maybe all connection is down
-        throw jce;
-      }
 
+    } catch (JedisNoReachableClusterNodeException jnrcne) {
+      //Not sure if JedisNoReachableClusterNodeException exception should extends JedisConnectionException, it's not connection exception
```
marcosnils: It's ok to make it extend JedisConnectionException. You can go ahead and remove the comment.

src/main/java/redis/clients/jedis/JedisClusterCommand.java
roblg: > Do you really want to get rid of cluster rediscovering on timeouts if you have maxRedirects=1?

I'm kind of torn. I don't know what the right thing is here. :(

I like the change to avoid using `tryRandomNode` -- for my use case (and for yours too, it sounds like), it only adds unnecessary hops within the cluster that aren't likely to result in success.

The renewSlotCache change adds unnecessary work for the Jedis clients for my use case, because we don't have any slaves to fail over, BUT it's totally possible my use case is weird, and your proposed solution here makes things better for the majority of users, in which case you should ignore me, and I'll try to figure out a way to make this code pluggable (maybe a RetryHandler, or something?) that could allow more flexibility in how clients fail. In any event, I'm just an enthusiast, and have no say in whether or not this gets merged anyway. :)

Spikhalskiy: I'm totally fine to implement more complex solution in one more PR right after this that will avoid rediscovering often than once per time. And it will be correct and fine for your case. And will not make rediscovering if nodes are just unstable, but available. Just need approving of this idea by @marcosnils and @HeartSaVioR in that case.

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -117,18 +118,29 @@ private T runWithRetries(byte[] key, int redirections, boolean tryRandomNode, bo
...
+      //Remove this if hierarchy reworked and no need to make exception from JedisConnectionException clause
+      throw jnrcne;
+    } catch (JedisConnectionException jce) {
       // release current connection before recursion
       releaseConnection(connection);
       connection = null;
 
-      // retry with random connection
-      return runWithRetries(key, redirections - 1, true, asking);
+      if (redirections <= 1) {
```
marcosnils: @Spikhalskiy what if we change this condition to redirections == 1?. In this case we'll only perform discovery on the last redirection. It will also help with @roblg case that if he configures maxRedirections == 1, he won't get a discovery every on every request until he brings up his instance.

Spikhalskiy: If we change it to 1 - nothing change actually. Yes, it will make rediscovering only after last request and it will not change anything for @roblg case. We can set it to 2. In that case it will make rediscover BEFORE last attempt and solve @roblg case, but it's bad solution. If we have @roblg case but with slaves - failed master could not be caught correctly, because no rediscovering will be proceed.

Spikhalskiy: Now this name is absolutely misleading. It's not number of redirections, it's total number of attempts. So... maxRedirection = 1 means that there are no redirections. Only one attempt.

HeartSaVioR: +1 to rename to maxAttempts.

Spikhalskiy:
@HeartSaVioR @marcosnils any updates here?

src/main/java/redis/clients/jedis/JedisClusterCommand.java
marcosnils: > Now this name is absolutely misleading. It's not number of redirections, it's total number of attempts. So... maxRedirection = 1 means that there are no redirections, only one attempt. I would suggest to rename it to maxAttempts.

This is with the modifications on this PR correct?. Because as JedisCluster is today, if you sex maxRedirections to 1, you can get redirected only once.

marcosnils:
@HeartSaVioR @xetorthio  LGTM!. Please review this whenever you can as it's the last PR before releasing 2.9

src/main/java/redis/clients/jedis/JedisClusterCommand.java
Spikhalskiy: "Because as JedisCluster is today, if you sex maxRedirections to 1, you can get redirected only once." You will be 'directed' only once, but 'redirected' zero times. So, you make one 'attempt'. When I set maxRedirections to 1 I could assume that cluster client will make one request and follow one MOVED if required, when I set it to 0 I could assume that cluster client will make only one main attempt, but it doesn't work this way. (JedisClusterCommand#runWithRetries check ```redirections<=0``` to throw exception) In addition, maxRredirections doesn't include information that this count would be used as a number of attempts in case of any fail. A user could assume that it has only "MOVED" related semantic. There are my thoughts about this name.

marcosnils: @Spikhalskiy I see your point. Another important thing is that the `redirected` variable is decremented both in redirections and timeouts, so as you said it's not reflecting the intent. I believe renaming to maxAttempts and adding proper documentation would be enough

@Spikhalskiy can you please rebase and apply renaming?. I'll push so we can merge this today and release 2.9

Spikhalskiy:
@marcosnils rebased, renaming (maxRedirections -> maxAttempts) done as a separate second commit in this PR.

marcosnils:
@Spikhalskiy thanks a lot.

marcosnils:
I'll go ahead and merge this as it went through a lot of discussion and changes seem to be consistent.

marcosnils:
Downmerged to 2.8 and 2.9 respectively. Thx for the contribution

PR #1261, opened 2016-04-19
Introduce new test: verify delegation via reflection & mock

### LINKED ISSUE #829: Expand functional unit tests to Pipeline, Sharded, Cluster, etc.
We're doing functional tests from only normal Jedis.
Actually other cases call Jedis methods so we assume it wouldn't be a problem.
But since we're human, we can have some mistakes to call methods something else, and it could be serious problem while we don't know about it.
(For example, user calls get, but our code calls delete.)

So we may apply Jedis functional tests to all kind of things, though it should slow down execution of unit tests.

HeartSaVioR:
This pull request is WIP to address #829.

* DelegationVerifyTest checks that method delegates the request to delegator's method with same signature
  * currently it only supports Pipeline
* Whole bunch of parameter handlings are moved to Client class
  * in order to check conveniently, and avoid any possible duplicated works
  * will address this to BinaryJedis, Jedis, JedisCluster, ShardedJedis, etc.
* fix some missed spots while testing DelegationVerifyTest

I'll going to apply test to JedisCluster, and ShardedJedis too, but before that I should find a way to instantiate byte[], String[], byte[][], Map (?) and so on.
Pipeline test passed without addressing those things.

When we can apply this test to JedisCluster and ShardedJedis, I guess we can resolve #829.


src/main/java/redis/clients/jedis/PipelineBase.java
```diff
@@ -882,7 +882,7 @@
 
   @Override
   public Response<Long> zcount(byte[] key, double min, double max) {
-    getClient(key).zcount(key, toByteArray(min), toByteArray(max));
```
HeartSaVioR: toByteArray() doesn't take care about "-inf" and "+inf" so it may be a bug around sorted set with Pipeline.

src/main/java/redis/clients/jedis/PipelineBase.java
```diff
@@ -841,7 +841,7 @@
 
   @Override
   public Response<Long> zadd(byte[] key, double score, byte[] member, ZAddParams params) {
-    getClient(key).zadd(key, score, member);
```
HeartSaVioR: missed spot: params are not passed.

src/main/java/redis/clients/jedis/PipelineBase.java
```diff
@@ -911,7 +911,7 @@
 
   @Override
   public Response<Double> zincrby(byte[] key, double score, byte[] member, ZIncrByParams params) {
-    getClient(key).zincrby(key, score, member);
```
HeartSaVioR: missed spot: params are not passed.

src/main/java/redis/clients/jedis/PipelineBase.java
```diff
@@ -1538,7 +1541,7 @@
 
   @Override
   public Response<Double> geodist(String key, String member1, String member2, GeoUnit unit) {
-    getClient(key).geodist(key, member1, member2);
```
HeartSaVioR: missed spot: `unit` is not passed.

src/test/java/redis/clients/jedis/tests/DelegationVerifyTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+        Method matchingMethod = findMatchingMethod(mockClient.getClass(), method);
+        matchingMethod.invoke(verify(mockClient), parameters);
+        reset(mockClient);
+      }
+    }
+  }
+
+  private boolean needToSkipMethod(Method method) {
+    // FIXME: we should replace vararg to List or Collection or Iterator or sth. to get rid of ambiguity. It's already reported from Scala users.
```
HeartSaVioR: This is another issue I would like to address. We have some methods using vararg which makes ambiguity to Scala.

src/main/java/redis/clients/jedis/BinaryClient.java
```diff
@@ -780,6 +784,13 @@ public void zremrangeByRank(final byte[] key, final long start, final long end)
     sendCommand(ZREMRANGEBYRANK, key, toByteArray(start), toByteArray(end));
   }
 
+  public void zremrangeByScore(final byte[] key, final double start, final double end) {
+    byte[] byteArrayStart = (start == Double.NEGATIVE_INFINITY) ? "-inf".getBytes() : toByteArray(start);
```
HeartSaVioR: I think this kind of conversion is bad cause giving -inf -inf as score range is still valid. It should check both Double.NEGATIVE_INFINITY and Double.POSITIVE_INFINITY.

src/test/java/redis/clients/jedis/tests/DelegationVerifyTest.java
HeartSaVioR: File an issue to #1263.

src/main/java/redis/clients/jedis/BinaryClient.java
HeartSaVioR: File an issue to #1264.

HeartSaVioR:
Also addressed ShardedJedis, and ShardedJedisPipeline.

I couldn't address JedisCluster before finding a way to instantiate byte[], String[], etc since JedisCluster tries to find the key (or keys) to lookup.
I plan to inject custom connection handler so any combinations of keys should work. We just should avoid NPE to continue.

HeartSaVioR:
Rebased against master to apply hotfix #1265.

HeartSaVioR:
Addressed JedisCluster, too.

With this patch I think we can close #829 though there's some limitations on it.

HeartSaVioR:
Buffer overflow again... Do we want to switch off OpenJDK7 testing?

HeartSaVioR:
@marcosnils We seem to miss this. Could you review this?

HeartSaVioR:
@marcosnils @sazzad16 Please review this as I still believe this is nice way to not miss something on parameters and keep consistent. Thanks in advance!

sazzad16:
@HeartSaVioR LGTM!

HeartSaVioR:
@sazzad16 OK. Will record broken case and report once.

HeartSaVioR:
@sazzad16 
Addressed. Please take a look again.

Btw, I had to modify current code to remove calling overloaded method in same class, due to the nature of current test mechanism. Hope this is OK.

sazzad16:
@HeartSaVioR Many unrelated changes have got in. Could you remove those?

sazzad16:
@HeartSaVioR I tried removing params from multiple methods in same class. E.g. removing `unit` from two `PipelineBase.geodist()` methods. But getting report for only one.

HeartSaVioR:
@sazzad16 Nice finding. It occurs when there's matching method but parameters are not we expected. Addressed.

sazzad16:
@HeartSaVioR Great work so far! But two more things:

1. The report gives the name of interfaces. IMO, it would be better off with the name of classes.

2. I have polarized feeling about the changes in BinaryClient, Client and MultiKeyPipelineBase.
    - If we are going to keep the changes, Jedis, BinaryJedis, etc. should also be changed to use these. This would mean code redundancies for `setTimeoutInfinite` and `rollbackTimeout`.
    - Or, we can remove the changes and ignore them in DelegationVerifyTest.

@marcosnils Can you give some input here?

uglide:
Closing as stale

Issue #1264, opened 2016-04-20
conversion of score from double to byte array should consider both "-inf" and "+inf"


HeartSaVioR:
Below logic is what Jedis converts scores typed double to byte arrays. 

```
    byte[] byteArrayMin = (min == Double.NEGATIVE_INFINITY) ? "-inf".getBytes() : toByteArray(min);
    byte[] byteArrayMax = (max == Double.POSITIVE_INFINITY) ? "+inf".getBytes() : toByteArray(max);
```

I think this kind of conversion is bad cause giving "-inf" "-inf" as score range is still valid. 
Conversion of score should check both Double.NEGATIVE_INFINITY and Double.POSITIVE_INFINITY for each score, not handling by context.

id-ilych:
https://github.com/xetorthio/jedis/pull/1271

marcosnils:
Plz check #1273

PR #1271, opened 2016-04-27
Handle infinite doubles in Client (Pipeline, Transaction etc.)

id-ilych:
Client used to encode double values before calling to BinaryClient methods with byte[] arguments thus skipping infinite values handling which is implemented in BinaryClient's methods with double arguments.

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryClient.java
+++ b/src/main/java/redis/clients/jedis/BinaryClient.java
@@ -780,6 +780,13 @@ public class BinaryClient extends Connection {
     sendCommand(ZREMRANGEBYRANK, key, toByteArray(start), toByteArray(end));
   }
 
+  public void zremrangeByScore(final byte[] key, final double start, final double end) {
+    byte[] byteArrayStart = (start == Double.NEGATIVE_INFINITY) ? "-inf".getBytes() : toByteArray(start);
+    byte[] byteArrayEnd = (end == Double.POSITIVE_INFINITY) ? "+inf".getBytes() : toByteArray(end);
+
+    sendCommand(ZREMRANGEBYSCORE, key, byteArrayStart, byteArrayEnd);
+  }
+
   public void zremrangeByScore(final byte[] key, final byte[] start, final byte[] end) {
     sendCommand(ZREMRANGEBYSCORE, key, start, end);
   }
--- a/src/main/java/redis/clients/jedis/Client.java
+++ b/src/main/java/redis/clients/jedis/Client.java
@@ -560,7 +560,7 @@ public class Client extends BinaryClient implements Commands {
 
   @Override
   public void zcount(final String key, final double min, final double max) {
-    zcount(SafeEncoder.encode(key), toByteArray(min), toByteArray(max));
+    zcount(SafeEncoder.encode(key), min, max);
   }
 
   @Override
@@ -570,7 +570,7 @@ public class Client extends BinaryClient implements Commands {
 
   @Override
   public void zrangeByScore(final String key, final double min, final double max) {
-    zrangeByScore(SafeEncoder.encode(key), toByteArray(min), toByteArray(max));
+    zrangeByScore(SafeEncoder.encode(key), min, max);
   }
 
   @Override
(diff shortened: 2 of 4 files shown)


HeartSaVioR:
Hi! Thanks for contributing.
Btw, #1261 should address all the things, so you may want to compare and find missed spots from there.
And you may also know I didn't intend this from #1264.

id-ilych:
@HeartSaVioR unfortunately it's not clear for me how to use your verification. Test seems to passed (w/ lots of "Skipping method due to some issues" messages) even w/o my fixes (though you also added 'zremrangeByScore' overload for `double`). 

I just looked for `toByteArray(final double value)` invocations in Client and avoided all of them reusing `BinaryClient`'s methods with `double` parameters.

id-ilych:
I just added some tests for infinite arguments in pipieline and transaction. But then I tried to reverse my initial changes to verify that tests are failing and something strange happened: 
- (expected) tests are failing on my Windows machine with redis 3.0.501 (https://github.com/MSOpenTech/redis/releases/)
- (what!?!) tests are passing on The Travic CI Linux machine

Can someone check if original problem exists on Linux? My guess is that underlying C stdlib behave differently while parsing infinite numbers. Reference says (http://en.cppreference.com/w/cpp/string/byte/strtod):

> infinity expression. It consists of the following parts:
(optional) plus or minus sign
INF or INFINITY ignoring case

which means that `"-Infinity"` should be as good as `"-inf"`. It's frustraiting...

HeartSaVioR:
@id-ilych 

> I just looked for toByteArray(final double value) invocations in Client and avoided all of them reusing BinaryClient's methods with double parameters.

Yeah, that's what I also addressed via #1261.

> (expected) tests are failing on my Windows machine with redis 3.0.501 (https://github.com/MSOpenTech/redis/releases/)
>(what!?!) tests are passing on The Travic CI Linux machine

I guess tests shouldn't fail since it's still possible to provide double as parameters and why it needs +inf and -inf is just a precision issue.
We can check how command will be sent by attaching idle Redis server for testing by redis-cli, and do `monitor`.
Could you check this to both server (windows folk, original from linux / mac)?

HeartSaVioR:
@id-ilych 
Btw, my test checks that methods listed in API interfaces delegate request to same method with same signature in underlying delegator. So in order to pass my test, methods shouldn't touch the parameter before delegating so I also addressed it. (it's what your pull request is modifying so I said it's duplicated.)

id-ilych:
> I guess tests shouldn't fail since it's still possible to provide double as parameters and why it needs +inf and -inf is just a precision issue or easier to input by hand (cli).

I don't get that part about 'precision issue'. Currently `jedis/master` sends `"-Infinity"` string in `byte[]` form when pipeline/transaction is used for e.g. `zcount`. Never saw redis documentation specifying numbers format except for `-inf`/`+inf` mentioned here http://redis.io/commands/zrangebyscore

>We can check how command will be sent by attaching idle Redis server for testing by redis-cli, and do monitor.
Could you check this to both server (windows folk, original from linux / mac)? 

I need more basic explanation... step-by-step instruction would be great :) My redis knowlege is basic.

I just tested `zcount` behaviour with `redis-cli` on Windows using 3.0.501 (https://github.com/MSOpenTech/redis/releases/):

> 127.0.0.1:6379> zadd foo 2 a
(integer) 1
127.0.0.1:6379> zadd foo 3 b
(integer) 1
127.0.0.1:6379> zadd foo 4 c
(integer) 1
127.0.0.1:6379> zcount foo 2 3
(integer) 2
127.0.0.1:6379> zcount foo 2 +Infinity
(error) ERR min or max is not a float
127.0.0.1:6379> zcount foo -Infinity +Infinity
(error) ERR min or max is not a float
127.0.0.1:6379> zcount foo -Inf +Inf
(integer) 3

And on Linux using 2.8.4 (I don't have Linux machin so used Cloud9 and they have that version installed by default):

> 127.0.0.1:6379> zadd foo 2 a
(integer) 1
127.0.0.1:6379> zadd foo 3 b
(integer) 1
127.0.0.1:6379> zadd foo 4 c
(integer) 0
127.0.0.1:6379> zcount foo 2 3
(integer) 2
127.0.0.1:6379> zcount foo 2 +Infinity
(integer) 3
127.0.0.1:6379> zcount foo -Infinity +Infinity
(integer) 3
127.0.0.1:6379> zcount foo -Inf +Inf
(integer) 3

So it seems that we have to options:

1. `"-Infinity"` and `"+Infinity"` are correct values to send which implies:
  -  conversions to `-inf`/`+inf` in jedis is just traffic optimization (5 bytes less to send)
  - redis server for Windows has a bug

2. Behaviour is undefined/unspecified for `"-Infinity"`/`"+Infinity"` and can change in future which implies that changes from this PR and #1261 are required. (I think this PR is still valuable because it simpler and could be integrated earlier that #1261)

I will file report to redis for Windows so we can see which option is better to take.

id-ilych:
Report is replied (https://github.com/MSOpenTech/redis/issues/455#issuecomment-215351090):
>Currently Windows supports +/-Inf and +/-Infinite. We will add support for +/-Infinity.

The funny thing is that "infinite" spelling doesn't work on Linux :laughing:

HeartSaVioR:
@id-ilych Heh, that's what port is! Porting and maintaining origin should be very hard, especially origin project is still evolving. :)

id-ilych:
@HeartSaVioR fix for windows build was commited today (don't know when it would be included in release).

Any thoughts about PR status? It's still unclear whether 'infinity' is specified or should never be used. Also even in that case it could be an optimization (even though it looks minor).

marcosnils:
@HeartSaVioR @id-ilych check #1273

id-ilych:
@marcosnils Unit tests I wrote have passed on Windows (even w/o mentioned fix for "Infinity")

HeartSaVioR:
@id-ilych 
I feel #1273 and #1261 covers this pull request. Could you take a look and see there's still uncovered spots?
Thanks in advance!

id-ilych:
@HeartSaVioR #1273 seems to cover what I tried to fix.

PR #1278, opened 2016-05-10
Add support for new Redis loadable module system

marcosnils:
@HeartSaVioR @xetorthio ping!.

:bell::bell::bell::bell::bell:


marcosnils:
Plz squash commits before merging!.

fabiopedrosa:
This would indeed be very useful.

src/main/java/redis/clients/jedis/MultiKeyPipelineBase.java
```diff
@@ -8,7 +8,7 @@
 
 public abstract class MultiKeyPipelineBase extends PipelineBase implements
     MultiKeyBinaryRedisPipeline, MultiKeyCommandsPipeline, ClusterPipeline,
-    BinaryScriptingCommandsPipeline, ScriptingCommandsPipeline {
+    BinaryScriptingCommandsPipeline, ScriptingCommandsPipeline, BasicRedisPipeline {
```
HeartSaVioR: We don't need to add this since PipelineBase should implement BasicRedisPipeline.

HeartSaVioR: Oh I see the reason... Is it because methods don't have key to route?
Then I think methods should be treat as multiple key methods. I mean, I think MultiKeyCommandsPipeline is proper place to be placed.

HeartSaVioR:
@marcosnils I fee it's added to wrong interface for pipeline, but otherwise looks good overall.
It also needs to be rebased.

marcosnils:
@HeartSaVioR the interface was already in Jedis, I just realized that it wasn't used anywhere so I added it to `MultiKeyPipelineBase` so we don't forget to implement any method. We can refactor this later. 

Rebased!

HeartSaVioR:
@marcosnils PipelineBase already implements BasicRedisPipeline. It's used for single key commands since those commands are used with ShardedJedis. So, commands which don't have route key have to be placed to multiple keys interface. 'randomKey' is there, too.

marcosnils:
@HeartSaVioR I do't see that `PipelineBase` implements `BasicRedisPipeline`, it implements `BinaryRedisPipeline` which is not the same :tongue:. When I added it in `MultiKeyPipelineBase` I've realized that some methods weren't implemented like `time` which I covered here: 
https://github.com/xetorthio/jedis/pull/1278/files#diff-292c84e057888fbce479de11157d4088R544

HeartSaVioR:
@marcosnils 
Oh sorry I confused BasicRedisPipeline and RedisPipeline. ;) You're right. LGTM.

PR #1301, opened 2016-06-03
Fixing squid S1168 Return an empty collection instead of null Part 2

devFozgul:
 squid:S1168- “ Empty arrays and collections should be returned”. 
This PR will remove 60min TD.
 You can find more information about the issues here: 
 https://dev.eclipse.org/sonar/rules/show/squid:S1168
 Fevzi Ozgul


HeartSaVioR:
I didn't notice this change. If we need to fix the test, this would be the signal of doing wrong.
Let's fix this from #1368

marcosnils:
My bad, let's revert this.

PR #1302, opened 2016-06-03
Fixing squid: S1168 Empty arrays and collections should be returned part 3

devFozgul:
 squid:S1168- “  Empty arrays and collections should be returned”. 
This TD will remove 60min TD.
 You can find more information about the issues here: 
 https://dev.eclipse.org/sonar/rules/show/squid:S1168
Fevzi Ozgul

### DIFF
--- a/src/main/java/redis/clients/jedis/Protocol.java
+++ b/src/main/java/redis/clients/jedis/Protocol.java
@@ -2,6 +2,7 @@ package redis.clients.jedis;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 import redis.clients.jedis.commands.ProtocolCommand;
@@ -168,7 +169,7 @@ public final class Protocol {
   private static byte[] processBulkReply(final RedisInputStream is) {
     final int len = is.readIntCrLf();
     if (len == -1) {
-      return null;
+      return new byte[0];
     }
 
     final byte[] read = new byte[len];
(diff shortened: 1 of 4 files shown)


marcosnils:
Tests are not passing

devFozgul:
Yes, exactly. I checked the logs where the tests failed. This PR required
quite a lot changes to be made in test methods. So I will do it in a more
convenient time. Problem is the assertNull methods for lists. Instead of
assertNull we might be use  Hamcrest empty list checking methods.

marcosnils:
Perfect, let's leave it open until we can get tests to pass. Thx for the contribution!.

devFozgul:
You are welcome. I have a couple of more issues to fix with this
project.However, they are not costly (TD wise) as the ones I have  already
done. You have also another "jedis something"  project  ( I do not remeber
the exact name now ) which I  will be working on in the coming days. I have
not checked it yet whether it has a considerable amount of TD to work on,
we shall see.

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -168,7 +169,7 @@ private static Object process(final RedisInputStream is) {
   private static byte[] processBulkReply(final RedisInputStream is) {
     final int len = is.readIntCrLf();
     if (len == -1) {
-      return null;
+      return new byte[0];
```
sazzad16: [Redis Bulk String](https://redis.io/topics/protocol#resp-bulk-strings)
> The client library API should not return an empty string, but a nil object, when the server replies with a Null Bulk String.

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -194,7 +195,7 @@ private static Long processInteger(final RedisInputStream is) {
   private static List<Object> processMultiBulkReply(final RedisInputStream is) {
     final int num = is.readIntCrLf();
     if (num == -1) {
-      return null;
+      return Collections.emptyList();
```
sazzad16: [Redis Arrays](https://redis.io/topics/protocol#resp-arrays)
> A client library API should return a null object and not an empty Array when Redis replies with a Null Array. This is necessary to distinguish between an empty list and a different condition.

PR #1303, opened 2016-06-03
Fixing squid: S2583 Conditions should not unconditionally evaluate to "TRUE" or to "FALSE"

devFozgul:
 squid:S2583 - “ Conditions should not unconditionally evaluate to ""TRUE"" or to ""FALSE"" ”. 
 You can find more information about the issue here: 
 https://dev.eclipse.org/sonar/rules/show/squid:S2583
Fevzi Ozgul

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisClusterConnectionHandler.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterConnectionHandler.java
@@ -45,9 +45,7 @@ public abstract class JedisClusterConnectionHandler implements Closeable {
       } catch (JedisConnectionException e) {
         // try next nodes
       } finally {
-        if (jedis != null) {
           jedis.close();
-        }
       }
     }
   }
@@ -62,9 +60,7 @@ public abstract class JedisClusterConnectionHandler implements Closeable {
       } catch (JedisConnectionException e) {
         // try next nodes
       } finally {
-        if (jedis != null) {
           jedis.close();
-        }
       }
     }
   }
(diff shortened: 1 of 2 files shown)


src/main/java/redis/clients/jedis/JedisClusterConnectionHandler.java
```diff
@@ -62,9 +60,7 @@ public void renewSlotCache() {
       } catch (JedisConnectionException e) {
         // try next nodes
       } finally {
-        if (jedis != null) {
```
marcosnils: Jedis is initialized as null here: https://github.com/xetorthio/jedis/pull/1303/files#diff-36ffb2068ca7e0edd549920989880b0eR55 if the try bluck throws an exception we need to do this check here.

src/main/java/redis/clients/jedis/JedisSentinelPool.java
```diff
@@ -163,9 +163,7 @@ private HostAndPort initSentinels(Set<String> sentinels, final String masterName
         log.warning("Cannot get master address from sentinel running @ " + hap + ". Reason: " + e
             + ". Trying next one.");
       } finally {
-        if (jedis != null) {
```
marcosnils: Same as https://github.com/xetorthio/jedis/pull/1303/files#r65761426

devFozgul: I will check again but I thought jedis is never null here.

marcosnils: What if `new Jedis()` eventually throws a runtime exception or someone adds something before this line (https://github.com/xetorthio/jedis/pull/1303/files/ae11c9480520fccbb006890323b55076a526dace#diff-e6525ac1a247b59a49332a6584030684R144) that might raise it?

PR #1319, opened 2016-06-15
Using toLowerCase method causes a ERR syntax error if default Locale is Turkish

juliomucor:
Because I dotted and I dottless are different in turkish, e.g. WITHSCORES

http://mattryall.net/blog/2009/02/the-infamous-turkish-locale-bug


marcosnils:
:+1: thx for fixing!.

marcosnils:
Applied to 2.8 and 2.9 respectively. Thx!.

PR #1327, opened 2016-06-25
2.8 - Implemented a specific version of scan methods from Redis Cluster

sheinbergon:
Based on http://www.paluch.biz/blog/162-iterate-over-all-keys-in-a-redis-cluster.html  , 
This implementation supports scanning for "Hash Tag" prefixed based  patterns ( i.e.  {TAG}* ) 

ScanParams were also refactored to allow exposure of params to package classes as well as eliminate possible bugs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.iml
 *.ipr
 *.iws
+nb*
 .project
 .settings/
 .gradle/
--- a/src/main/java/redis/clients/jedis/BinaryJedisCluster.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedisCluster.java
@@ -14,1998 +14,2027 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Pattern;
 
 public class BinaryJedisCluster implements BasicCommands, BinaryJedisClusterCommands,
-    MultiKeyBinaryJedisClusterCommands, JedisClusterBinaryScriptingCommands, Closeable {
+        MultiKeyBinaryJedisClusterCommands, JedisClusterBinaryScriptingCommands , JedisClusterBinaryCustomizedCommands , Closeable {
 
-  public static final short HASHSLOTS = 16384;
-  protected static final int DEFAULT_TIMEOUT = 2000;
-  protected static final int DEFAULT_MAX_REDIRECTIONS = 5;
+    protected static final Pattern SCAN_VALID_MATCH_REGEX_PATTERN = Pattern.compile("\\{[^\\{^\\}]+\\}\\*");
 
-  protected int maxRedirections;
+    public static final short HASHSLOTS = 16384;
+    protected static final int DEFAULT_TIMEOUT = 2000;
+    protected static final int DEFAULT_MAX_REDIRECTIONS = 5;
 
-  protected JedisClusterConnectionHandler connectionHandler;
+    protected int maxRedirections;
 
-  public BinaryJedisCluster(Set<HostAndPort> nodes, int timeout) {
-    this(nodes, timeout, DEFAULT_MAX_REDIRECTIONS, new GenericObjectPoolConfig());
-  }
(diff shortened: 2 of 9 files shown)


.gitignore
```diff
@@ -2,6 +2,7 @@
 *.iml
 *.ipr
 *.iws
+nb*
```
marcosnils: What is this for?

marcosnils:
@sheinbergon thx for implementing this based on @mp911de draft. 

Can you please rebase this against master so we can merge?. We'll take care to down merge to 2.9 and 2.8

src/main/java/redis/clients/jedis/HighLevelBinaryJedisClusterCommands.java
```diff
@@ -0,0 +1,5 @@
+package redis.clients.jedis;
+
+public interface HighLevelBinaryJedisClusterCommands {
```
marcosnils: Why another interface?. This command doesn't seem high level it just checks that the key contains a hash tag. 

:-1:  @HeartSaVioR ?

src/main/java/redis/clients/jedis/HighLevelJedisClusterCommands.java
```diff
@@ -0,0 +1,5 @@
+package redis.clients.jedis;
+
+public interface HighLevelJedisClusterCommands {
```
marcosnils: Same https://github.com/xetorthio/jedis/pull/1327/files#r68688905

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
```diff
@@ -14,14 +14,16 @@
...
 
 public class BinaryJedisCluster implements BasicCommands, BinaryJedisClusterCommands,
-    MultiKeyBinaryJedisClusterCommands, JedisClusterBinaryScriptingCommands, Closeable {
+    MultiKeyBinaryJedisClusterCommands , HighLevelBinaryJedisClusterCommands, JedisClusterBinaryScriptingCommands, Closeable {
 
   public static final short HASHSLOTS = 16384;
   protected static final int DEFAULT_TIMEOUT = 2000;
   protected static final int DEFAULT_MAX_REDIRECTIONS = 5;
-
+  protected static final Pattern SCAN_VALID_MATCH_REGEX_PATTERN = Pattern.compile("\\{[^\\{^\\}]+\\}\\*");
```
sheinbergon: @marcosnils  I only use the regex to validate input since only {TAG}* should be supported in cluster mode . I can disable the limitation though it might create confusion amongst the user as to why this is not working in all scenarios

I will add cluster based scanning implementation that does distributed scanning ( supporting all cases ) in the future but for now only a speciic case is covered

src/main/java/redis/clients/jedis/HighLevelBinaryJedisClusterCommands.java
sheinbergon: @marcosnils Well , it does a bit more than just sending the commands to the cluster as the other implmentations do . I also didn't want to interfere with the other  interface . Let me know what you decide and I'll change it  according to your request

.gitignore
sheinbergon: @marcosnils  I develop on top of netbeans
Didn't want to commit jedis configuration files

sheinbergon:
@marcosnils  sorry for not getting to rebasing on the master , I'll get to it today or tomorrow

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
sheinbergon: @marcosnils  re-reading what you wrote , I see you meant changing the way we check for input matching to be based upon start/end chars. Regex will indeed be slower , I'll fix my code to behave the same as getSlot(String key) method does :smile:

sheinbergon:
@marcosnils  made changes accordiing to you requests/rejects  , now found in #1335.
I'm closing this one

Issue #1333, opened 2016-07-02
Upgrade the JDK versin to 1.7


smadasu:
### Expected behavior

Jedis should run using JDK1.7

### Actual behavior

It still uses 1.6 accroding to pom.xml below
``
                 <plugins>
			<plugin>
				<groupId>org.apache.maven.plugins</groupId>
				<artifactId>maven-compiler-plugin</artifactId>
				<version>3.1</version>
				<configuration>
					<source>1.6</source>
					<target>1.6</target>
				</configuration>
			</plugin>
``
### Steps to reproduce:

Please create a reproducible case of your problem. Make sure
that case repeats consistently and it's not random
1.
2.
3.

#### Java version:
Upgrade to 1.7

marcosnils:
Fixed by #1334

PR #1334, opened 2016-07-02
Upgrade the JDK versin to 1.7 #1333

smadasu:
I have updated the pom.xml to use JDK7. Please review


marcosnils:
This applies to master only.

Issue #1325, opened 2016-06-24
Scan support for JedisClient using tagging


idans-ybrantdigital-com:
Though it makes sense why key scanning is not supported for jedis cluster client
Support for key scanning with tagging should work as depicted here :

http://www.paluch.biz/blog/162-iterate-over-all-keys-in-a-redis-cluster.html

Any chance it might find it's way to the JedisCluster implementation ?

10x

marcosnils:
@idans-ybrantdigital-com patches are always welcome :)

mp911de:
@idans-ybrantdigital-com Maybe you want to take a look [here](https://github.com/mp911de/lettuce/blob/master/src/main/java/com/lambdaworks/redis/cluster/ClusterScanSupport.java) and port it to jedis? Would be a nice addition.

sheinbergon:
@marcosnils  , I created a minimal implementation support only TAG-prefix based scanning , it can be found in the pull request mentioned  #1327  . I interfered as little as possible with the Jedis existing code and made this exetnsion blend as naturally as possible with the rest of your code. 

I tested it and found it to be working as expected :)  Looking forward for your feedback/rejects

@mp911de  That is same great work you've done there. Once someone approves this patch , I'll find the time and port that code as well :)

sheinbergon:
BTW , I made all of the fixes on top of 2.8 as I believe this is considered a rather minor fix . If you'd like me to apply it on some other branch , let me know

HeartSaVioR:
@sheinbergon Hi, all of patch should be against master since all changes must be reflected to master.
Having cluster-wise command is nice, especially if it gets rid of interface difference between Jedis and JedisCluster. 
I'll have a look when I get some time. Thanks!

idans-ybrantdigital-com:
Done by @sheinbergon  , merged by @HeartSaVioR  , will be released in 2.9

Closing

PR #1335, opened 2016-07-02
Implemented a HashTag version of SCAN methods for Redis Cluster

sheinbergon:
Fixing issue #1325 

Here are the is the version of the changes mentioned in #1327  applied against the master + rejects mentioned by @marcosnils fixed

@marcosnils  , looking  forward to your comments/rejects

Closeing #1327

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryJedisCluster.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedisCluster.java
@@ -17,6 +17,7 @@ import java.util.Map;
 import java.util.Set;
 
 import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
+import redis.clients.util.JedisClusterHashTagUtil;
 
 public class BinaryJedisCluster implements BinaryJedisClusterCommands,
     MultiKeyBinaryJedisClusterCommands, JedisClusterBinaryScriptingCommands, Closeable {
@@ -1749,7 +1750,30 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
       }
     }.runBinary(key);
   }
+  
+  @Override
+  public ScanResult<byte[]> scan(final byte[] cursor, final ScanParams params) {
 
+    String matchPattern = null;
+
+    if (params == null || (matchPattern = params.match()) == null || matchPattern.isEmpty()) {
+      throw new IllegalArgumentException(JedisCluster.class.getSimpleName() + " only supports scan requests with non-empty match patterns");
+    }
+
+    if (JedisClusterHashTagUtil.isTaggedMatchPattern(matchPattern)) {
+
+      return new JedisClusterCommand< ScanResult<byte[]>>(connectionHandler,
+              maxRedirections) {
+        @Override
+        public ScanResult<byte[]> execute(Jedis connection) {
+          return connection.scan(cursor, params);
+        }
+      }.runBinary(SafeEncoder.encode(matchPattern));
+    } else {
+      throw new IllegalArgumentException(JedisCluster.class.getSimpleName() + " only supports scan requests with match patterns of the following format : '{<HASH_TAG>}*' where <HASH_TAG> is a tag string of your choice");
+    }
+  }
+  
   @Override
   public ScanResult<Map.Entry<byte[], byte[]>> hscan(final byte[] key, final byte[] cursor) {
     return new JedisClusterCommand<ScanResult<Map.Entry<byte[], byte[]>>>(connectionHandler,
--- a/src/main/java/redis/clients/jedis/BinaryJedisCluster.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedisCluster.java
@@ -1757,7 +1757,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
     String matchPattern = null;
 
     if (params == null || (matchPattern = params.match()) == null || matchPattern.isEmpty()) {
-      throw new IllegalArgumentException(JedisCluster.class.getSimpleName() + " only supports scan requests with non-empty match patterns");
+      throw new IllegalArgumentException(BinaryJedisCluster.class.getSimpleName() + " only supports SCAN commands with non-empty MATCH patterns");
     }
 
     if (JedisClusterHashTagUtil.isClusterCompliantMatchPattern(matchPattern)) {
@@ -1770,7 +1770,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
         }
       }.runBinary(SafeEncoder.encode(matchPattern));
     } else {
-      throw new IllegalArgumentException(JedisCluster.class.getSimpleName() + " only supports scan requests with match patterns of the following format : '{<HASH_TAG>}*' where <HASH_TAG> is a tag string of your choice");
+      throw new IllegalArgumentException(BinaryJedisCluster.class.getSimpleName() + " only supports SCAN commands with MATCH patterns containing hash-tags ( curly-brackets enclosed strings )");
     }
   }
   
--- a/src/main/java/redis/clients/jedis/BinaryJedisCluster.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedisCluster.java
@@ -1760,7 +1760,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
       throw new IllegalArgumentException(JedisCluster.class.getSimpleName() + " only supports scan requests with non-empty match patterns");
     }
 
-    if (JedisClusterHashTagUtil.isTaggedMatchPattern(matchPattern)) {
+    if (JedisClusterHashTagUtil.isClusterCompliantMatchPattern(matchPattern)) {
 
       return new JedisClusterCommand< ScanResult<byte[]>>(connectionHandler,
               maxRedirections) {
--- a/src/main/java/redis/clients/jedis/JedisCluster.java
+++ b/src/main/java/redis/clients/jedis/JedisCluster.java
@@ -18,6 +18,8 @@ import java.util.Set;
 import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
 
 import redis.clients.jedis.params.set.SetParams;
+import redis.clients.util.JedisClusterHashTagUtil;
+import redis.clients.util.SafeEncoder;
 
 public class JedisCluster extends BinaryJedisCluster implements JedisClusterCommands,
     MultiKeyJedisClusterCommands, JedisClusterScriptingCommands {
@@ -1209,6 +1211,29 @@ public class JedisCluster extends BinaryJedisCluster implements JedisClusterComm
     }.run(key);
   }
 
+    @Override
+  public ScanResult<String> scan(final String cursor, final ScanParams params) {
+
+    String matchPattern = null;
+
+    if (params == null || (matchPattern = params.match()) == null || matchPattern.isEmpty()) {
+      throw new IllegalArgumentException(JedisCluster.class.getSimpleName() + " only supports scan requests with non-empty match patterns");
+    }
+
+    if (JedisClusterHashTagUtil.isTaggedMatchPattern(matchPattern)) {
+
+      return new JedisClusterCommand< ScanResult<String>>(connectionHandler,
+              maxRedirections) {
+        @Override
+        public ScanResult<String> execute(Jedis connection) {
+          return connection.scan(cursor, params);
+        }
+      }.runBinary(SafeEncoder.encode(matchPattern));
+    } else {
+      throw new IllegalArgumentException(JedisCluster.class.getSimpleName() + " only supports scan requests with match patterns of the following format : '{<HASH_TAG>}*' where <HASH_TAG> is a tag string of your choice");
+    }
+  }
+  
   @Override
   public ScanResult<Entry<String, String>> hscan(final String key, final String cursor) {
     return new JedisClusterCommand<ScanResult<Entry<String, String>>>(connectionHandler,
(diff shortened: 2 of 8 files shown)


src/main/java/redis/clients/jedis/ScanParams.java
```diff
@@ -1,39 +1,67 @@
...
+
+    String match() {
+        if (params.containsKey(MATCH)) {
+            return new String(params.get(MATCH).array());
+        } else {
+            return null ;
+        }
+    }
+    
+    int count() {
```
HeartSaVioR: It should return null when not set by user.

src/main/java/redis/clients/jedis/ScanParams.java
```diff
@@ -1,39 +1,67 @@
...
-
-  public Collection<byte[]> getParams() {
-    return Collections.unmodifiableCollection(params);
-  }
-}
+
+    private final Map<Keyword, ByteBuffer> params = new EnumMap<Keyword, ByteBuffer>(Keyword.class);
+
+    // Default COUNT parameter value according to Redis documentation - http://redis.io/commands/scan
+    private final static ByteBuffer SCAN_COUNT_DEFAULT = ByteBuffer.wrap(toByteArray(10));
```
HeartSaVioR: That default value should be handled from Redis side, not client side.

sheinbergon: Ok , so what  to return in case not set ? 0 ?

src/main/java/redis/clients/jedis/ScanParams.java
sheinbergon: Are you referring to 'int count()' ?

src/main/java/redis/clients/util/JedisClusterHashTagUtil.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  public static String getHashTag(String key) {
+    return extractHashTag(key, false, false);
+  }
+
+  public static boolean isClusterCompliantMatchPattern(String matchPattern) {
+    String tag = extractHashTag(matchPattern, true, true);
+    if (tag != null && !tag.isEmpty() && matchPattern.startsWith(tag)) {
+      return matchPattern.substring(tag.length()).equals("*");
```
HeartSaVioR: AFAIK, pattern doesn't need to be restricted by "*".
Keys with same hashtag will be placed to same slot, which is only one condition for scan with Redis cluster.

src/main/java/redis/clients/jedis/ScanParams.java
HeartSaVioR: Yes right. I mean it shouldn't have default value. It should not send count parameter instead of sending default value. Default count value for Redis can be changed from Redis side, and we should follow it regardless of Redis versions.

src/main/java/redis/clients/jedis/ScanParams.java
HeartSaVioR: Commented above.

src/main/java/redis/clients/jedis/ScanParams.java
sheinbergon: OK , will fix soon

src/main/java/redis/clients/jedis/ScanParams.java
sheinbergon: OK , will fix soon

src/main/java/redis/clients/util/JedisClusterHashTagUtil.java
sheinbergon: Sorry but that's not the case.
You can test manually and see that running the scan only ( without the '*' globbing ) yields no results . My string match tests behaves as expected and is the right way to scan with Tags within a clustered setup ( Though I think the same goes for standard SCANing )

http://redis.io/commands/scan -   'It is possible to only iterate elements matching a given glob-style pattern'

src/main/java/redis/clients/jedis/ScanParams.java
sheinbergon: @HeartSaVioR  fixed

src/main/java/redis/clients/util/JedisClusterHashTagUtil.java
HeartSaVioR:
```
127.0.0.1:7379> set {hello}world1 1
OK
127.0.0.1:7379> set {hello}world2 1
OK
127.0.0.1:7379> set {hello}world3 1
OK
127.0.0.1:7379> set {hello}world4 1
OK
127.0.0.1:7379> set {hello}man1 1
OK
127.0.0.1:7379> set {hello}man2 1
OK
127.0.0.1:7379> scan 0 MATCH {hello}world*
1) "0"
2) 1) "{hello}world1"
   2) "{hello}world3"
   3) "{hello}world4"
   4) "{hello}world2"
127.0.0.1:7379> scan 0 MATCH {hello}man*
1) "0"
2) 1) "{hello}man1"
   2) "{hello}man2"
127.0.0.1:7379> scan 0 MATCH {hello}man1
1) "0"
2) 1) "{hello}man1"
```

Please refer above result. The thing is, we shouldn't restrict to use only '*'. It's just a one of use case.

src/main/java/redis/clients/jedis/ScanParams.java
```diff
@@ -61,7 +58,7 @@ int count() {
         if (params.containsKey(COUNT)) {
             return params.get(COUNT).getInt();
         } else {
-            return SCAN_COUNT_DEFAULT.getInt();
+            return 0;
```
HeartSaVioR: Let's have consistent between match() and count(). I think it's better to change return type to Integer (to let users know it can be null), and also return null when it's not presented by user.

sheinbergon: @HeartSaVioR  done :+1:

src/main/java/redis/clients/util/JedisClusterHashTagUtil.java
sheinbergon: @HeartSaVioR  I got what you were trying to say now. 

I ran some more tests and noticed that tag doesn't have to be at the start/prefix of the pattern for it to work . In fact , as long as there is a tag inside the match pattern , I extract it and pass it to the Jedis SDK to extract the correct slot and the SCANing will work as usuall . Therefor - as long the match pattern contains a tag , it will be considered as valid 

``
127.0.0.1:5000> SET key1{FF}1yek 46
OK
127.0.0.1:5000> SET key2{FF}2yek 646
OK
127.0.0.1:5000> SET key3{FF}3yek 64364
OK
127.0.0.1:5000> SCAN 0 MATCH k*{FF}* COUNT 2000
1) "0"
2) 1) "key2{FF}2yek"
   2) "key1{FF}1yek"
   3) "key3{FF}3yek"
``

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
```diff
@@ -1749,7 +1750,30 @@ public Double execute(Jedis connection) {
...
+
+      return new JedisClusterCommand< ScanResult<byte[]>>(connectionHandler,
+              maxRedirections) {
+        @Override
+        public ScanResult<byte[]> execute(Jedis connection) {
+          return connection.scan(cursor, params);
+        }
+      }.runBinary(SafeEncoder.encode(matchPattern));
+    } else {
+      throw new IllegalArgumentException(JedisCluster.class.getSimpleName() + " only supports scan requests with match patterns of the following format : '{<HASH_TAG>}*' where <HASH_TAG> is a tag string of your choice");
```
HeartSaVioR: Please update the message that match pattern should have a hashtag but not having other requirements.

src/main/java/redis/clients/jedis/JedisCluster.java
```diff
@@ -1209,6 +1211,29 @@ public Long execute(Jedis connection) {
...
+
+      return new JedisClusterCommand< ScanResult<String>>(connectionHandler,
+              maxRedirections) {
+        @Override
+        public ScanResult<String> execute(Jedis connection) {
+          return connection.scan(cursor, params);
+        }
+      }.runBinary(SafeEncoder.encode(matchPattern));
+    } else {
+      throw new IllegalArgumentException(JedisCluster.class.getSimpleName() + " only supports scan requests with match patterns of the following format : '{<HASH_TAG>}*' where <HASH_TAG> is a tag string of your choice");
```
HeartSaVioR: Same above.

HeartSaVioR:
We're almost done! Personally it may be better to have test cases for new feature, but it's OK if you're not interested to do it.
@marcosnils Please review and comment.

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
```diff
@@ -1749,7 +1750,30 @@ public Double execute(Jedis connection) {
       }
     }.runBinary(key);
   }
+  
+  @Override
+  public ScanResult<byte[]> scan(final byte[] cursor, final ScanParams params) {
+
+    String matchPattern = null;
+
+    if (params == null || (matchPattern = params.match()) == null || matchPattern.isEmpty()) {
```
marcosnils: `matchPattern = params.match()`? Shouldn't we use equality instead?

src/main/java/redis/clients/util/JedisClusterHashTagUtil.java
```diff
@@ -0,0 +1,41 @@
...
+  public static String getHashTag(String key) {
+    return extractHashTag(key, false, false);
+  }
+
+  public static boolean isClusterCompliantMatchPattern(String matchPattern) {
+    String tag = extractHashTag(matchPattern, true, true);
+    return tag != null && !tag.isEmpty() ;
+  }
+
+  private static String extractHashTag(String key, boolean includeBrackets, boolean returnNullWhenTagNotFound) {
```
marcosnils: I don't like having a function with boolean parameters (specially two) as it's super difficult to use it from somewhere else without looking the code. I'll try to think of a better alternative to this.

marcosnils:
PR looks good in general. I've addressed a concern regarding the extractHashTag which I'd like to think before merging.

src/main/java/redis/clients/jedis/JedisCluster.java
sheinbergon: @HeartSaVioR  done

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
sheinbergon: @HeartSaVioR  done , I actually did forget to fix those after disposing of the regex pattern matching

src/main/java/redis/clients/util/JedisClusterHashTagUtil.java
sheinbergon: @marcosnils  though I don't really understand your objection to a method containing 2 boolean arguments ( especially as I made an effort to give them clear names ) , I went ahead and omitted the 'includeBrackets' argument as it wasn't really necessary due to the various changes done as part of the pull request :+1:  :smile_cat:

sheinbergon:
@marcosnils @HeartSaVioR  Thank you for you comments , rejects and observations , I feel as though We've gotten the most out of this feature.

When can I find this released into maven ?  can I see it  inside a minor release for 2.8 ?

Once we're done with it and I'll clear some stuff off my table , I'll implement high level cluster SCAN to run across the cluster :smiling_imp:

HeartSaVioR:
@sheinbergon According to semver, next minor release is 2.9 so we should include new feature to minor release. It should be shipping to 2.9.

src/main/java/redis/clients/util/JedisClusterHashTagUtil.java
marcosnils: @sheinbergon http://martinfowler.com/bliki/FlagArgument.html

sheinbergon:
@HeartSaVioR @marcosnils  so everything is fine now by you guys ?

src/main/java/redis/clients/util/JedisClusterHashTagUtil.java
sheinbergon: @marcosnils I do agree with the above paradigm when it comes to exposing external methods. Notice that I wrapped the private flag method with 2 non-flag methods just as expressed in the artical . When it comes to internal/private methods with a limited use ( especially in static context ) I feel it's a better approach as the alternative would be duplicating the code.

Anyway , can we agree it is sufficient now ?

marcosnils:
LGTM :+1:

sheinbergon:
@marcosnils  @HeartSaVioR  so ... can we change the label for READY FOR MERGE ?

HeartSaVioR:
@sheinbergon 
Yeah sure. I'll take care of it. Please note that we're on distributed time zones, and have a job which should be handled as first priority. :)

HeartSaVioR:
Merged into master and 2.9.
Btw, I had to arrange something to apply this to 2.9. Please have a look at commit https://github.com/xetorthio/jedis/commit/5dd5e0579d51f64fac7a54ef2e7a9cdb03cd1ed8

And since we're treating scan as multi key operation, it should be corrected to master too.

marcosnils:
@HeartSaVioR perfect. You can go ahead and fix directly in master also.

sheinbergon:
@HeartSaVioR  Ok got it . I know and understand we are all very busy , just wanted to make sure this is not left behind . Thank you for your time and effort

HeartSaVioR:
@sheinbergon Yeah we're labeling to not making great efforts to be staled. :) Thanks for your effort!

HeartSaVioR:
@marcosnils Addressed. Btw, I'm still seeing Redis crashing on ModuleTest in local dev, OSX 10.11.

marcosnils:
@HeartSaVioR I don't have OSX so it's hard for me to replicate. :cry:

PR #1336, opened 2016-07-03
Move back interfaces to proper place to respect backward compatibility (2.9)

HeartSaVioR:
2.8.x breaks this, and we're about to rollback.

It shouldn't be backported to master.
If we're planning to release any 2.8.x versions, please merge this to 2.8 branch, too.

@marcosnils Please review and merge when it's OK. Thanks!


marcosnils:
LGTM!.

HeartSaVioR:
Ported back to 2.8 since it's a bug which should be addressed soon.

PR #1339, opened 2016-07-06
Fix for 'ERR max number of clients reached'  MasterListener Thread death

### LINKED ISSUE #1329: JedisSentinelPool: 'ERR max number of clients reached' kills MasterListener
The issue is similar to https://github.com/xetorthio/jedis/issues/1036 
However it can happen in JedisSentinelPool.MasterListener.

Exception in thread "MasterListener-redismaster-[mysecrethost.com:26379]" redis.clients.jedis.exceptions.JedisDataException: ERR max number of clients reached
    at redis.clients.jedis.Protocol.processError(Protocol.java:117)
    at redis.clients.jedis.Protocol.process(Protocol.java:151)
    at redis.clients.jedis.Protocol.read(Protocol.java:205)
    at redis.clients.jedis.Connection.readProtocolWithCheckingBroken(Connection.java:297)
    at redis.clients.jedis.Connection.getRawObjectMultiBulkReply(Connection.java:242)
    at redis.clients.jedis.JedisPubSub.process(JedisPubSub.java:108)
    at redis.clients.jedis.JedisPubSub.proceed(JedisPubSub.java:102)
    at redis.clients.jedis.Jedis.subscribe(Jedis.java:2591)
    at redis.clients.jedis.JedisSentinelPool$MasterListener.run(JedisSentinelPool.java:283)

JedisDataException is not catched and MasterListener thread exit.
Expected: JedisDataException must be catched and listener shall reconnect later.
#### Jedis version:

2.7.3

sheinbergon:
- Fixes #1329 
- Added missing @Override annotations
- Made the logger final static
- As JDK is now 1.7 , Jedis instance is now wrapped in try-with-resources block

### DIFF
--- a/src/main/java/redis/clients/jedis/BuilderFactory.java
+++ b/src/main/java/redis/clients/jedis/BuilderFactory.java
@@ -97,7 +97,7 @@ public final class BuilderFactory {
     @SuppressWarnings("unchecked")
     public Map<String, String> build(Object data) {
       final List<byte[]> flatHash = (List<byte[]>) data;
-      final Map<String, String> hash = new HashMap<String, String>();
+      final Map<String, String> hash = new HashMap<String, String>(flatHash.size()/2, 1);
       final Iterator<byte[]> iterator = flatHash.iterator();
       while (iterator.hasNext()) {
         hash.put(SafeEncoder.encode(iterator.next()), SafeEncoder.encode(iterator.next()));
@@ -117,7 +117,7 @@ public final class BuilderFactory {
     @SuppressWarnings("unchecked")
     public Map<String, String> build(Object data) {
       final List<Object> flatHash = (List<Object>) data;
-      final Map<String, String> hash = new HashMap<String, String>();
+      final Map<String, String> hash = new HashMap<String, String>(flatHash.size()/2, 1);
       final Iterator<Object> iterator = flatHash.iterator();
       while (iterator.hasNext()) {
         hash.put(SafeEncoder.encode((byte[]) iterator.next()),


src/main/java/redis/clients/jedis/JedisSentinelPool.java
```diff
@@ -139,9 +140,7 @@ private HostAndPort initSentinels(Set<String> sentinels, final String masterName
 
       log.fine("Connecting to Sentinel " + hap);
 
-      Jedis jedis = null;
-      try {
-        jedis = new Jedis(hap.getHost(), hap.getPort());
+      try (Jedis jedis = new Jedis(hap.getHost(), hap.getPort())) {
```
marcosnils: This can't be ported to 2.8 as it will have to run in JDK 6

marcosnils:
> As JDK is now 1.7 , Jedis instance is now wrapped in try-with-resources block

master version is JDK 1.7 but keep in mind that this needs to be downmerged to 2.8 and 2.9 as well so it's better to use the old way

marcosnils:
@sheinbergon change seems ok, but can you explain a bit what was happening to understand the scenario?. Do you think it's possible to write a test case for this fix?

src/main/java/redis/clients/jedis/JedisSentinelPool.java
sheinbergon: @marcosnils I have to disagree here . As you already upgraded the JDK to 1.7 in the master branch, I believe that as a principal all new pieces of code written should be employ latest writing techniques , such  as try-with-resources. People looking to resolve a rather minor issue such as the above  should upgrade to newer versions , which is also good practice.

sheinbergon:
@marcosnils  see my comment about jdk 1.7 in the line note

sheinbergon:
@marcosnils  the scenario in general is - every situation that produces an exception which is not JedisConnectionException in the MasterListener Thread causes it to die and the whole functionality of waiting for master changes to stop working in a live production code.  The more specific scenario is due to client connection overload ( too many cilient connect to the sentinel ) and JedisDataException is thrown).

Jedis Exceptions are unchecked by design which is a valid design method just as choosing the other way around . Should they've not been that way , this would've been avoided at code writing time :hammer: 

There-for I don't think this specific scenario has a relevant test case , in a more general notion , we should make sure the MasterListener never dies ( maybe place a watch dog on top of it ) because if it does die production for whatever reason , It's bad

src/main/java/redis/clients/jedis/JedisSentinelPool.java
marcosnils: I understand your point and I believe it's valid. What I'm saying is that makes more difficult to backport stuff to Jedis 2.x as we'll have to do manual merging because compilation will fail in those branches.

marcosnils:
@sheinbergon @HeartSaVioR @xetorthio LGTM. 

Once another contributor approves we can merge.

xetorthio:
Thanks for taking care of this issue. It is important to fix it.

Now even though I like your changes, the only one relevant to fix this issue is catching `JedisException` instead of `JedisConnectionException`. I recommend splitting this PR into 2.

Another thing is the lack of testing. As you said, this is an important issue, and it deserves a test. Regardless of any design decision we made, we should try to find a way to test this. There is already a list of steps to reproduce the issue. That's a good start.

marcosnils:
> Another thing is the lack of testing. As you said, this is an important issue, and it deserves a test. Regardless of any design decision we made, we should try to find a way to test this.

This is a bit tricky to test as we need to do two things which are not easy. 

1 - Force redis to return an error when MasterListener initializes

2 - Found a way to check that the MasterListener is currently down. We don't have API's for that ATM.

src/main/java/redis/clients/jedis/JedisSentinelPool.java
sheinbergon: Again , it's up to you to decide . Not all fixes/features are to be backported IMHO. It's your call . If you it's crucial to backport , I'll dispose of the try-with-resources . 

If not , let's keep it that way , please :)

sheinbergon:
@xetorthio @marcosnils  I'll see what I can do about testing . I'll update here in a few more days

sheinbergon:
@marcosnils  I'll rework on this to include some tests

PR #1343, opened 2016-07-08
Added initial size for HashMaps

harshmoorjani:
To avoid it being resized when Redis Hash has too many keys.

### DIFF
--- a/src/main/java/redis/clients/jedis/BuilderFactory.java
+++ b/src/main/java/redis/clients/jedis/BuilderFactory.java
@@ -97,7 +97,7 @@ public final class BuilderFactory {
     @SuppressWarnings("unchecked")
     public Map<String, String> build(Object data) {
       final List<byte[]> flatHash = (List<byte[]>) data;
-      final Map<String, String> hash = new HashMap<String, String>();
+      final Map<String, String> hash = new HashMap<String, String>(flatHash.size()/2, 1);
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
LGTM. Great finding.

HeartSaVioR:
Yeah I forgot we might release 2.8.x. :)

src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -97,7 +97,7 @@ public String toString() {
     @SuppressWarnings("unchecked")
     public Map<String, String> build(Object data) {
       final List<byte[]> flatHash = (List<byte[]>) data;
-      final Map<String, String> hash = new HashMap<String, String>();
+      final Map<String, String> hash = new HashMap<String, String>(flatHash.size()/2);
```
mp911de: You might want to add load factor `1` as you already know the target size in advance. Load factor `1` prevents rehashing. Same for the change below.

marcosnils:
LGTM after addressing @mp911de comments.

harshmoorjani:
Added load factor `1` as suggested by @mp911de

xetorthio:
Amazing! Good catch! LGTM!

marcosnils:
Merged. Applied to 2.8 and 2.9 respectively.

Issue #1332, opened 2016-07-02
Fix resource leaks and warnings in Junit tests


smadasu:
### Expected behavior

The Redis connections opened in unit tests should be closed properly. There are also lot of warnings in code which can be cleaned up easily
### Actual behavior

The resources are not closed properly and may cause problems in running unit tests. Unnecessary suppresswarnings etc can be cleaned up
### Steps to reproduce:

This is code enhancement
#### Jedis version:

Latest unstable
#### Redis version:

Latest unstable
#### Java version:

1.7

smadasu:
@marcosnils , I cannot use try with resources for java7. Why is Jedis still using 1.6 version of java?

HeartSaVioR:
We discussed about version but seems like we forgot to change. 
@marcosnils @xetorthio I guess we already discussed this - using at least JDK 7. Right?

smadasu:
Here is part of pom.xml


		<plugins>
			<plugin>
				<groupId>org.apache.maven.plugins</groupId>
				<artifactId>maven-compiler-plugin</artifactId>
				<version>3.1</version>
				<configuration>
					<source>**1.6**</source>
					<target>**1.6**</target>
				</configuration>
			</plugin>

This uses 1.6

smadasu:
@marcosnils @HeartSaVioR Do you think it's OK if I change as part of this pull request?

HeartSaVioR:
Sorry I think it should be handled as another commit/issue/pull request for history.

smadasu:
OK. I can create another PR for upgrading to 1.7. How likely is that PR merged into master? Because only   then I can fix this issue.

marcosnils:
fixed by #1334 for master branch.

HeartSaVioR:
Using JDK7 is just a requirement for him to use try-with-finally, and he doesn't submit pull request regarding this. @smadasu Is this correct?

smadasu:
That is correct

HeartSaVioR:
OK. I reopen this.

smadasu:
Thanks

marcosnils:
Oh, sorry I thought you meant change the jdk support only in master. Well, Oracle added an extended support for the jdk 6 until the end of 2018 so I wouldn't recommend to drop support in before that time.
http://www.oracle.com/technetwork/java/eol-135779.html

HeartSaVioR:
Well I think we can't change JDK version even in future if we're concerning Oracle paid support. :)
[...]

PR #1348, opened 2016-07-11
Fix resource leaks and warnings in Junit tests #1332

smadasu:
FIxed resource leak warnings. Please review

### DIFF
--- a/src/test/java/redis/clients/jedis/tests/ConnectionTest.java
+++ b/src/test/java/redis/clients/jedis/tests/ConnectionTest.java
@@ -9,7 +9,6 @@ import org.junit.Test;
 
 import redis.clients.jedis.Connection;
 import redis.clients.jedis.Protocol.Command;
-import redis.clients.jedis.commands.ProtocolCommand;
 import redis.clients.jedis.exceptions.JedisConnectionException;
 
 public class ConnectionTest {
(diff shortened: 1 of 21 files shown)


marcosnils:
@smadasu how / when do you see this warnings?. In your editor?. 

As these are unit tests, it's not super important to do resource cleanup. I'm not sure about this change

@HeartSaVioR @xetorthio ?

xetorthio:
Yeah. They are all on testa, unless I missed something. If that is the case
then I wouldn't worry too much.

smadasu:
Yes. They are all unit tests. It may solve some of the random unit test failures we have seen in the past. But it's up to you guys to merge. This is not a high priority PR.

HeartSaVioR:
@marcosnils @xetorthio 
I like the change though they're just tests. Someone starts from test codes to see the usage patterns, so though they're not critical, but also not bad to change them.

@smadasu 
There're too many inconsistent style issues, whether try after space or not, if after space or not, ( after space or not, and so on. Yes we don't use checkstyle and also don't specify style guideline so you may wonder why I'm saying about style, but even narrowing code to only your changeset, the code style is inconsistent. 

Please have a look at other code blocks and try to keep the code style. We might need to discuss the code style or even set up checkstyle in maven process.

marcosnils:
Closed due to inactivity

PR #1354, opened 2016-07-22
Fix for 'ERR max number of clients reached'  MasterListener Thread death

sheinbergon:
Hi @marcosnils @HeartSaVioR  @xetorthio 

I've created a test case for issue #1329  
Note that this PR doesn't yet include the actual fix ( Change the type of exception  catched to a broader one ) just the test that should fail in the current scenario. I'll add once we confirm the test is valid and up to standards

Haven't gotten round to running it yet though . Let me know what you think

### DIFF
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -202,7 +202,8 @@ public class Connection implements Closeable {
         inputStream = new RedisInputStream(socket.getInputStream());
       } catch (IOException ex) {
         broken = true;
-        throw new JedisConnectionException(ex);
+        throw new JedisConnectionException("Failed connecting to host " 
+            + host + ":" + port, ex);
       }
     }
   }


sheinbergon:
@marcosnils @HeartSaVioR  my tests are failing because config set/get is not supported for redis-sentinel :disappointed:  - http://stackoverflow.com/questions/30376639/redis-sentinel-how-to-get-config-value

I'd like to hear your ideas on how to tackle this (aside from writing command line execution integrations to change values in the config file and restart it). Tnx

sheinbergon:
There is this PR https://github.com/antirez/redis/pull/2777 , but it looks like it never made it to production

sheinbergon:
@marcosnils any response here ?

HeartSaVioR:
@sheinbergon 
Sorry for the late update. Seems like most of changes are indentation mismatch. Could you follow our code style? hbase-formatter.xml is for that, and you can import this to your IDE.
You can even `make format` to reformat whole things but it also change javadoc format and other files so it's only for fixing whole messed things.

I'll take a look at again when formatting issue is gone.

HeartSaVioR:
@sheinbergon 
And if you can't make test because of the lack of Redis, you are not forced to do. I'm OK to drop the test.

sheinbergon:
@HeartSaVioR per your approval , I'll fix this PR to contain only the fix for this scenario

PR #1357, opened 2016-07-25
Avoid a NullPointerException when unsubscribing before subscribing

ekuefler:
In my subscription code, I start up another thread to handle the subscription
before returning the unsubscribe handler. If the caller calls unsubscribe before
the new thread has had a change to call `jedis.subscribe`, I get a
NullPointerException. This makes unsubscribe do nothing in that case. There might
be better ways to handle this situation.


sazzad16:
@HeartSaVioR @marcosnils Close this one? Specially without tests?

sazzad16:
The (Binary)JedisPubSub class(es) are initiated from a SUBSCRIBE command. This is the intended use-case.

Closing due to inactivity and lack of tests.

PR #1361, opened 2016-07-27
eval returns Response<Object>

huguesb:
Eval return value may not be a string and trying to coerce it into one causes a ClassCastException.

This appears to be a regression. Version 2.2.5, from which I am trying to upgrade, did the right thing...

### DIFF
--- a/src/main/java/redis/clients/jedis/PipelineBase.java
+++ b/src/main/java/redis/clients/jedis/PipelineBase.java
@@ -1314,18 +1314,18 @@ public abstract class PipelineBase extends Queable implements BinaryRedisPipelin
     return getResponse(BuilderFactory.DOUBLE);
   }
 
-  public Response<String> eval(String script) {
+  public Response<Object> eval(String script) {
     return this.eval(script, 0);
   }
 
-  public Response<String> eval(String script, List<String> keys, List<String> args) {
+  public Response<Object> eval(String script, List<String> keys, List<String> args) {
     String[] argv = Jedis.getParams(keys, args);
     return this.eval(script, keys.size(), argv);
   }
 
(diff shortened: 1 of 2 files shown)


marcosnils:
LGTM

xetorthio:
This is good, but shouldn't the test actually create a script that returns something other than a string so it makes more sense to return an Object?

marcosnils:
@HeartSaVioR has a PR that will fix all these test cases https://github.com/xetorthio/jedis/pull/749. Eventually we need to review this :)

HeartSaVioR:
AFAIK, we already addressed this but it was only merged into 3.0.0 since it breaks backward compatibility.

HeartSaVioR:
We addressed this at #607. There're similar cases for 3.0.0 so we would want to go ahead of 3.0.0 anyway to resolve such issues.

marcosnils:
@HeartSaVioR too bad we didn't realize earlier so add it in 2.9 in a backward compatible way :(

HeartSaVioR:
@marcosnils Could you elaborate your comment? I don't get it.

marcosnils:
Sure, I'm saying that we could have added the Response<Object> in 2.9 without removing this methods and do it in a backward-compatible manner. If we want to land this before 3.0 we need to release another minor.

HeartSaVioR:
Yes we can do it anyway but I think it will introduce methods with less-intuitive method name. We were not handling #607 to backward compatible way since unlike others it needs to change just return type so we can't rely on method overloading.

Tbdgithub:
Does it fixed in version 2.8.2 ?

sazzad16:
Resolved by #1667

PR #1382, opened 2016-09-11
Disable protected mode for test environment

HeartSaVioR:
* this makes whole tests fail

This resolves the build failure what @gogeof reported.
https://github.com/xetorthio/jedis/issues/1370#issuecomment-246120348

@marcosnils @xetorthio Please review this, and confirm it doesn't fail on Linux.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ STUNNEL_BIN := $(shell which stunnel)
 
 define REDIS1_CONF
 daemonize yes
+protected-mode no
 port 6379
 requirepass foobared
 pidfile /tmp/redis1.pid
@@ -14,6 +15,7 @@ endef
 
 define REDIS2_CONF
 daemonize yes
+protected-mode no
 port 6380
 requirepass foobared
 pidfile /tmp/redis2.pid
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
I don't know when/why it was changed... Only clear thing is that without patch at least OSX build is failing.

marcosnils:
@heartsavior LGTM!. Thx for digging

HeartSaVioR:
@marcosnils Not at all. I'd like to thank @gogeof to report this. I just follow his/her report.

marcosnils:
I still believe it's weird that we needed this for OSX stuff to pass and in linux it worked without problem... But I don't mind that much because it's just for testing.

HeartSaVioR:
Could you test with latest Redis unstable branch now without this patch? Actually I didn't see this before, and saw this after @gogeof reported.

marcosnils:
@HeartSaVioR tests pass in linux without this patch, that's why I believe it's super strange that we need to enable this in OSX. This is a redis configuration which **should** be independent from the underlying OS :confused:

HeartSaVioR:
@marcosnils Yes agreed. Might be better to have a look more and report to Redis.

HeartSaVioR:
Anyway I merged this into master, 2.8, 2.9 branches. I'll try to dig more when I have time.

PR #1383, opened 2016-09-12
support clientName setting for ShardedJedisPool

jiafu1115:
Motivation:

ShardedJedisPool should support setting clientName into Jedis so
that we can trouble shooting easier with client name.

now no client name be showed for client list if we use ShardedJedisPool:
id=18 addr=10.140.201.34:51637 fd=12 name= age=85650 idle=85619 flags=N
db=0 sub=0 psub=0 multi=-1 qbuf=0 qbuf-free=0 obl=0 oll=0 omem=0
events=r cmd=scan

Modifications:

(1) add private member: clientName for ShardedJedisFactory;
(2) add new constructor with clientName for ShardedJedisPool;
(2) When calling makeObject(), it will create connections with all sharded redis servers and set client name for the connections. If exception existed, It will be ignored so that makeObject() can always success.  
note:  
old behavior for ShardedJedisPool is that it won't connect will any sharded  redis server util executing commands; 
new behavior for ShardedJedisPool is that it will connect with all sharded redis server when making object. It is same style with other pools. Though, to avoid one fail connection  cause makeObject() failed. We will ignore the possible connection exception.

Result:

easier to trouble shooting with connection name if using ShardedJedisPool:

id=2 addr=10.224.38.23:33652 fd=5 name=APP_Name age=1350882
idle=1 flags=N db=0 sub=0 psub=0 multi=-1 qbuf=0 qbuf-free=0 obl=0 oll=0
omem=0 events=r cmd=ping


marcosnils:
@jiafu1115 can you please add a test for the functionality you're adding?

jiafu1115:
@marcosnils ok, no problem, Thus, I amn't sure this is the right method to support setting. After all, it change the behavior for makeObject(). What's more, if makeObject() success with one connection fail, The connection won't have time to setClientName again. WDTY for this solution.

I have another solution to add clientName to BinaryClient. WDTY for the two solutions?

jiafu1115:
@marcosnils any update? Thanks

marcosnils:
@jiafu1115 I'm not sure about the overall approach. I don't like polluting the BinaryJedis constructor with extra steps like setting up the client name. If this will be used for pooling, IMHO I beleve it should be kept in the proper PoolFactory initialization function. 

Let's wait to see what @HeartSaVioR or @xetorthio think.

HeartSaVioR:
Agree with @marcosnils. This touches BinaryClient and also BinaryJedis but changeset is only effective for Sharded Jedis which is not good.

jiafu1115:
Thanks for your suggestion. In the first version of this PR. I only change the pool factory with only one file change and little code changes. But I found two disadvantages:  1 the original shard pool factory doesn't open connection when making object .It is perfect design to avoid single connection temp fail influence making object fail

jiafu1115:
If I add set client code into factory as other factorys. The behavior for making objects will be changed to bad design. 2：set client name may should be handled by jedis instead of pool factory. It is similar with select db. Adding set client name and select db to every pool factory is not right style. So I change my solution from changing pool factory  to jedis self.  WDYT

marcosnils:
> The original ShardedJedisFactory doesn't open any connection when making object

I believe this is ok, as `JedisFactory` does this ATM.

> What's more, if I ignore single connection fail or setClientName fail when making object. There is no any chances to set it again if I doesn't change BinaryJedis's code.

I believe this is ok, if you get an exception from Redis while making an object you can throw it to the user check https://github.com/xetorthio/jedis/blob/master/src/main/java/redis/clients/jedis/JedisFactory.java#L100-L123 

> ClientName should be one of client's attributes.

IMHO `select` and `client setname` are different things. `setname` is not an initialization thing (like `select` is), that's why I think it shouldn't be performed when initializing the Jedis object itself.

jiafu1115:
So  do you mean my first version of this PR is ok？

jiafu1115:
@Override
    public PooledObject<ShardedJedis> makeObject() throws Exception {
      ShardedJedis shardedJedis = new ShardedJedis(shards, algo, keyTagPattern);
      for (Jedis jedis : shardedJedis.getAllShards()) {
        try{
          jedis.connect();
          if (clientName != null) {
            jedis.clientSetname(clientName);
          }
        } catch (JedisException je) {
          log.log(Level.SEVERE, "Can't connect to redis at " + jedis.getClient().getHost(), je);
        }
      }

      return new DefaultPooledObject<ShardedJedis>(shardedJedis);
    }

marcosnils:
> So do you mean my first version of this PR is ok？

I believe this is the approach but I haven't looked at the original PR. In the snippet you paste you're swallowing exceptions which IDK if we should do. Can u please revert to the original proposal so we can start from there?

jiafu1115:
@marcosnils I force delete the older version for this pull request. So you can't see the old version. but we can check followed code as example (just change little lines to support without any other file changes):

```
    @Override
    public PooledObject<ShardedJedis> makeObject() throws Exception {
      ShardedJedis shardedJedis = new ShardedJedis(shards, algo, keyTagPattern);
      for (Jedis jedis : shardedJedis.getAllShards()) {
        try{
          jedis.connect();
          if (clientName != null) {
            jedis.clientSetname(clientName);
          }
        } catch (JedisException je) {
          log.log(Level.SEVERE, "Can't connect to redis at " + jedis.getClient().getHost(), je);  //swallow it or throw it?
        }
      }

```
      return new DefaultPooledObject<ShardedJedis>(shardedJedis);
    }

PR #1384, opened 2016-09-14
Support clientName setting for ShardedJedisPool

jiafu1115:
Support clientName setting for ShardedJedisPool

[Motivation]

ShardedJedisPool should support setting clientName into Jedis so
that we can trouble shooting easier with client name.

now no client name be showed for client list if we use ShardedJedisPool:

id=2 addr=10.224.38.23:33652 fd=5 name=  age=1350882
idle=1 flags=N db=0 sub=0 psub=0 multi=-1 qbuf=0 qbuf-free=0 obl=0 oll=0
omem=0 events=r cmd=ping
[Modifications]

(1) add private member: clientName into ShardedJedisFactory;
(2) add new constructor with clientName for ShardedJedisPool;
(3) set client name in ShardedJedisFactory's makeObject() method;

[Result]

easier to trouble shooting with connection name if using
ShardedJedisPool:

id=2 addr=10.224.38.23:33652 fd=5 name=APP_Name age=1350882
idle=1 flags=N db=0 sub=0 psub=0 multi=-1 qbuf=0 qbuf-free=0 obl=0 oll=0
omem=0 events=r cmd=ping


marcosnils:
@jiafu1115 seems code doesn't compile according to travis.

jiafu1115:
@marcosnils  I am fixing it, The error is caused by I edit on line instead of IDE. so you can ignore the error. The basis code is right. So can you help to review it ? Thanks

jiafu1115:
@marcosnils , The compile error had been fixed. You can continue on it now. Thanks

src/main/java/redis/clients/jedis/ShardedJedisPool.java
```diff
@@ -60,17 +72,31 @@ protected void returnResource(final ShardedJedis resource) {
...
+      ShardedJedis shardedJedis = new ShardedJedis(shards, algo, keyTagPattern);
+      for (Jedis jedis : shardedJedis.getAllShards()) {
+        try{
+          jedis.connect();
+          if (clientName != null) {
+            jedis.clientSetname(clientName);
+          }
+        } catch (JedisException je) {
+          log.log(Level.SEVERE, "Can't connect to redis at " + jedis.getClient().getHost(), je);
+          jedis.close();
```
marcosnils: I thought that we wanted to raise the exception here?

src/test/java/redis/clients/jedis/tests/ShardedJedisPoolTest.java
```diff
@@ -54,8 +56,39 @@ public void checkConnections() {
     jedis.close();
     pool.destroy();
   }
-
+  
   @Test
+  public void checkConnectionsWithClientName() {
```
marcosnils: Please format your code, you have a `make format` target in the makefile to do this. Make sure to commit only this files and now all the formatted ones.

src/test/java/redis/clients/jedis/tests/ShardedJedisPoolTest.java
```diff
@@ -54,8 +56,39 @@ public void checkConnections() {
...
+		        String acturalClientName = jedis.clientGetname();
+		        System.out.println("clientName is: " + acturalClientName);
+				assertEquals(clientName, acturalClientName);
+		 	}
+		}
+		
+		assertTrue(isExistConnection);
+		shardedJedis.close();
+	}finally{
+		if(pool!=null)
```
marcosnils: Add {}

src/test/java/redis/clients/jedis/tests/ShardedJedisPoolTest.java
```diff
@@ -54,8 +56,39 @@ public void checkConnections() {
     jedis.close();
     pool.destroy();
   }
-
+
```
marcosnils: Unnecessary line

src/main/java/redis/clients/jedis/ShardedJedisPool.java
jiafu1115: @marcosnils  I doesn't throw it to support this case: one connection fail doesn't cause makeObject can't work.Or one redis server down will cause we can't connet with any  redis server.

src/main/java/redis/clients/jedis/ShardedJedisPool.java
```diff
@@ -60,17 +72,31 @@ protected void returnResource(final ShardedJedis resource) {
...
     }
 
     @Override
     public PooledObject<ShardedJedis> makeObject() throws Exception {
-      ShardedJedis jedis = new ShardedJedis(shards, algo, keyTagPattern);
-      return new DefaultPooledObject<ShardedJedis>(jedis);
+      ShardedJedis shardedJedis = new ShardedJedis(shards, algo, keyTagPattern);
+      for (Jedis jedis : shardedJedis.getAllShards()) {
+        try{
+          jedis.connect();
```
jiafu1115: now I should connect with redis when making object, This behavior is different with original code. So is it ok?

jiafu1115: @marcosnils  now I should connect with redis when making object, This behavior is different with original code. So is it ok?

src/test/java/redis/clients/jedis/tests/ShardedJedisPoolTest.java
jiafu1115: Ok, I will clear up it. Thanks

src/main/java/redis/clients/jedis/ShardedJedisPool.java
jiafu1115: raise exception will cause this case:  one redis server's connection fail will cause we can't work even other redis servers is Ok. Am I right? So I swallow it to make sure makeobject always success.

jiafu1115:
@marcosnils so, just now, after code format and cleanup. There is only one major concern for this PR is exception handle. waiting for your think. 3ks

src/test/java/redis/clients/jedis/tests/ShardedJedisPoolTest.java
jiafu1115: done

src/test/java/redis/clients/jedis/tests/ShardedJedisPoolTest.java
jiafu1115: done

src/test/java/redis/clients/jedis/tests/ShardedJedisPoolTest.java
jiafu1115: done

PR #1386, opened 2016-09-18
JedisCluster should support setting client name.

jiafu1115:
[Motivation]

JedisCluster should support setting client name similar with other types
of Jedis.

[Modifications]

(1) add private member: clientName in JedisClusterInfoCache;
(2) add new constructor with clientName for JedisCluster;

[Result]

JedisCluster support setting client name.

PR #1388, opened 2016-09-19
JedisCluster should support setting client name: Jedis2.9

jiafu1115:
similar with #1386  but code change for 2.9 branch and with some code cleanup


marcosnils:
this PR is not necessary, we'll take care of merging #1386 to the corresponding branches.

Issue #1392, opened 2016-09-19
testInvalidStartNodeNotAdded consistently fails on my local dev.


HeartSaVioR:
### Expected behavior

Tests run: 26, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 25.232 sec - in redis.clients.jedis.tests.JedisClusterTest

### Actual behavior

Tests run: 26, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 24.583 sec <<< FAILURE! - in redis.clients.jedis.tests.JedisClusterTest
testInvalidStartNodeNotAdded(redis.clients.jedis.tests.JedisClusterTest)  Time elapsed: 0.606 sec  <<< ERROR!
redis.clients.jedis.exceptions.JedisConnectionException: java.net.UnknownHostException: not-a-real-host
	at redis.clients.jedis.tests.JedisClusterTest.testInvalidStartNodeNotAdded(JedisClusterTest.java:616)
Caused by: java.net.UnknownHostException: not-a-real-host
	at redis.clients.jedis.tests.JedisClusterTest.testInvalidStartNodeNotAdded(JedisClusterTest.java:616)

### Steps to reproduce:

Run `make stop; make test` to OSX & JDK 1.8 env. Not sure it's failing with Linux.

#### Jedis version:

master, 2.10, 2.9 branches

#### Redis version:

unstable

#### Java version:

Oracle JDK 1.8.0_66 (JDK 1.7 passes)

HeartSaVioR:
```
  private void initializeSlotsCache(Set<HostAndPort> startNodes, GenericObjectPoolConfig poolConfig, String password, String clientName) {
    for (HostAndPort hostAndPort : startNodes) {
      Jedis jedis = new Jedis(hostAndPort.getHost(), hostAndPort.getPort());
      if (password != null) {
        jedis.auth(password);
      }
      if (clientName != null) {
        jedis.clientSetname(clientName);
      }
      try {
        cache.discoverClusterNodesAndSlots(jedis);
        break;
      } catch (JedisConnectionException e) {
        // try next nodes
      } finally {
        if (jedis != null) {
          jedis.close();
        }
      }
    }
  }
```

It makes sense to throw JedisConnectionException when initializing Jedis instance with unknown hostname. Auth failing or clientSetname failing also throw Exception and stop discovering. We need to suppress and try next nodes as same as inner try statement is doing.

marcosnils:
[...]

PR #1402, opened 2016-10-08
improve slots cache initialize‘s retry logic

jiafu1115:
[Motivation]
when jedis.auth(password) throw JedisConnectionException, it won't retry next node. It should be improved.

[Modifications]
move the code about auth/clientSetname into try-catch statements.

[Result]
When jedis fail to setup connection with one node to initial slots cache, it will retry next one.

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisClusterConnectionHandler.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterConnectionHandler.java
@@ -37,13 +37,13 @@ public abstract class JedisClusterConnectionHandler implements Closeable {
   private void initializeSlotsCache(Set<HostAndPort> startNodes, GenericObjectPoolConfig poolConfig, String password, String clientName) {
     for (HostAndPort hostAndPort : startNodes) {
       Jedis jedis = new Jedis(hostAndPort.getHost(), hostAndPort.getPort());
-      if (password != null) {
-        jedis.auth(password);
-      }
-      if (clientName != null) {
-        jedis.clientSetname(clientName);
-      }
       try {
+        if (password != null) {
+          jedis.auth(password);
+        }
+        if (clientName != null) {
+          jedis.clientSetname(clientName);
+        }
         cache.discoverClusterNodesAndSlots(jedis);
         break;
       } catch (JedisConnectionException e) {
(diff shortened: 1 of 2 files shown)


marcosnils:
is this what we really want?. Password management in redis cluster is out of the spec, and it's cluster admin responsibility to make sure that all nodes have the same password. The reason why we make JedisCluster fail fast if a node can't be authenticated is because so that cluster admin realizes about this and fixes the problem. 

@jiafu1115 I'm wondering what's the use-case you're facing that you need the cluster to start in unhealthy state with a node that you can't authenticate to. 

@HeartSaVioR ?

jiafu1115:
Only catch connection exception. auth fail may thrown other exception

jiafu1115:
commented 1 minute ago
Only catch connection exception. auth fail without connection fail may thrown other exception

marcosnils:
@jiafu1115 is this closed intentionally?

jiafu1115:
Sorry . I click wrong button

marcosnils:
@jiafu1115 now I see what you're saying. I guess this is ok. 

LGTM. @HeartSaVioR?

jiafu1115:
@marcosnils  Thanks for your approve and Let's wait @HeartSaVioR 's feedback.

marcosnils:
@jiafu1115 have you checked my last comment regarding the test?. WDYT?

jiafu1115:
I think it should fix the issue #1392. WDTY

marcosnils:
> I think it should fix the issue #1392. WDTY

I forgot to post the review, this is what I'm referring to: https://github.com/xetorthio/jedis/pull/1402#discussion_r82915954

jiafu1115:
@marcosnils Agree with you. I think the original test's goal is the test added by me. So can we just remove the original test and keep my test with test name improved.

marcosnils:
@jiafu1115 It's perfect now. Let's wait for @HeartSaVioR's review as this may close two issues at once.

jiafu1115:
@marcosnils  I often read the issues. I will try to reduce the issues total number due to interesting.

HeartSaVioR:
LGTM, Thanks @jiafu1115 for providing a patch!

HeartSaVioR:
@marcosnils Thanks for putting effort for merging. :)

PR #1407, opened 2016-10-14
2.9 - Support for redis 3.2

LGnord:
Since 3.2, redis provides a new command: reply.
http://redis.io/commands/client-reply

Do you plan to integrate it?

Thanks


marcosnils:
Yup, I'm sure we'll implement it eventually, or you can send a PR to integrate it :)

PR #1414, opened 2016-10-24
Change BinaryJedis bitfield return type to List<Long>

### LINKED ISSUE #1413: BinaryJedis.bitfield should return List<Long> instead of List<byte[]>
### Expected behavior

`BinaryJedis.bitfield` should return `List<Long>` instead of `List<byte[]>`. 
The same command issued via `PiplelineBase` or `JedisCommands` already returns the `Long` variant.

Writing a tests it seems that the return type declaration is wrong and that a `List<Long>` is provided already not matching the return types declaration. This leads to `java.lang.ClassCastException`.

#### Steps to reproduce: 

```java
List<byte[]> results = jedis.bitfield("mykey".getBytes(), new byte[][]{"SET".getBytes(), "i8".getBytes(), "0".getBytes(), "10".getBytes()});
for(byte[] result : results) {  // <-- fails
  System.out.println("result");
}
```

#### Jedis version: 
2.9.0

marcosnils:
Fixes #1413

### DIFF
--- a/src/test/java/redis/clients/jedis/tests/commands/BitCommandsTest.java
+++ b/src/test/java/redis/clients/jedis/tests/commands/BitCommandsTest.java
@@ -194,12 +194,12 @@ public class BitCommandsTest extends JedisCommandTestBase {
 
   @Test
   public void testBinaryBitfield() {
-    List<byte[]> responses = jedis.bitfield(SafeEncoder.encode("mykey"), SafeEncoder.encode("INCRBY"),
-        SafeEncoder.encode("i5"), SafeEncoder.encode("100"), SafeEncoder.encode("1"),
-        SafeEncoder.encode("GET"), SafeEncoder.encode("u4"), SafeEncoder.encode("0")
+    List<Long> responses = jedis.bitfield(SafeEncoder.encode("mykey"), SafeEncoder.encode("INCRBY"),
+            SafeEncoder.encode("i5"), SafeEncoder.encode("100"), SafeEncoder.encode("1"),
+            SafeEncoder.encode("GET"), SafeEncoder.encode("u4"), SafeEncoder.encode("0")
     );
-    assertEquals(1L, responses.get(0));
-    assertEquals(0L, responses.get(1));
+    assertEquals(1L, responses.get(0).longValue());
+    assertEquals(0L, responses.get(1).longValue());
   }
 
 }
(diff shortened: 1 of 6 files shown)


marcosnils:
@HeartSaVioR mind to review plz?

src/test/java/redis/clients/jedis/tests/commands/BitCommandsTest.java
```diff
@@ -194,12 +194,10 @@ public void testBitfield() {
 
   @Test
   public void testBinaryBitfield() {
-    List<byte[]> responses = jedis.bitfield(SafeEncoder.encode("mykey"), SafeEncoder.encode("INCRBY"),
-        SafeEncoder.encode("i5"), SafeEncoder.encode("100"), SafeEncoder.encode("1"),
-        SafeEncoder.encode("GET"), SafeEncoder.encode("u4"), SafeEncoder.encode("0")
-    );
-    assertEquals(1L, responses.get(0));
-    assertEquals(0L, responses.get(1));
+    List<Long> responses = jedis.bitfield("mykey".getBytes(), "INCRBY".getBytes(), "i5".getBytes(),
```
HeartSaVioR: How about keep using SafeEncoder.encode()?

HeartSaVioR:
LGTM except minor comment.

src/test/java/redis/clients/jedis/tests/commands/BitCommandsTest.java
HeartSaVioR: I'd like to have a consistent way to convert string to binary, and if my memory is right, someone complained that this way doesn't work on several charsets.

marcosnils: ok, let's keep it for consistency

marcosnils:
@HeartSaVioR updated!

HeartSaVioR:
Seems still same. Might be a bug on Github?

marcosnils:
> Seems still same. Might be a bug on Github?

my bad :sweat_smile:. Change should be OK now

HeartSaVioR:
Heh great :) LGTM.
@christophstrobl Could you check this again? This is ready to merge but I'd like to check this again so that we don't make a mistake again.

christophstrobl:
thanks for taking care so quickly! LGTM.

HeartSaVioR:
Merged into master, 2.10, 2.9 respectively!

PR #1416, opened 2016-10-28
concept of smartEval

smagellan:
I suppose this PR is related with https://github.com/xetorthio/jedis/issues/1232. Please review


sazzad16:
@smagellan Thanks for the PR. But adding _smart_ methods won't keep Jedis light/small which it wants to be. Closing.

PR #1420, opened 2016-11-02
Give the host and port for a Redis server in exceptions if the

manolama:
connection fails.


HeartSaVioR:
LGTM too. More informative. Thanks @manolama for the change.

HeartSaVioR:
Merged into master, 2.10, 2.9 branches.

Issue #1428, opened 2016-11-15
Missing overload for SET without NX/XX


danielbcorreia:
### Expected behavior

I was expecting to have this kind of usage with the SET command:

```jedis.set(key, value, "PX", time);```

### Actual behavior

The only option currently is to use this:

```jedis.set(key, value, "NX", "PX", time);```

Which is not ideal since we must specify NX/XX, then disabling full key override.

I can submit a pull request adding the required overloads (4 of them), but I've seen another approach merged into master in 2015 that was never released (https://github.com/xetorthio/jedis/pull/878) that should be better in terms of usage, although it changes a lot of things from what I've seen.

Let me know how you want to proceed, since we can't use Jedis as it is (we'll need to fork it internally just to fix this if we can't get it to a Jedis release).

#### Jedis version:

2.9.0

#### Redis version:

3.2

#### Java version:

8

danielbcorreia:
Anyone? Should I add those overloads in a pull request or should we follow another approach?

marcosnils:
@HeartSaVioR I remember having a discussion about this in the past. I've tried searching the thread but I coudln't find it.

HeartSaVioR:
@marcosnils Sorry I don't remember it. I'm OK to have it for 2.10.0, but ideally we would be better to discuss about releasing 3.0.0 seriously.
cc. to @xetorthio

danielbcorreia:
Hi everyone, a few days ago I created pull request #1433 to address this for 2.10. Can we try to release it with this version?

sazzad16:
Closed by #1433

PR #1433, opened 2016-11-21
Add support for SET without specifying NX/XX flag

danielbcorreia:
Relates to #1428

This is a temporary solution so Jedis can support SET without specifying an NX/XX flag, but a more permanent solution is already somewhere on the 3.0 branch line.


sazzad16:
LGTM!

sazzad16:
@danielbcorreia I know this has been a long time. But could you be able to rebase this?

PR #1443, opened 2016-12-12
Call renew slots before final retry

### LINKED ISSUE #1439: JedisCluster throws JedisConnectionException when a cluster master killed
### Expected behavior

I have a Redis Cluster with 4 master nodes and 4 slave nodes. Each master has 1 slave. I use JedisCluster keep writing data to Redis Cluster. When I killed a redis master, the slave switch to master as expected, and I thought JedisCluster would reconnect to the new master.

### Actual behavior

Actually I find JedisCluster throw JedisConnectionException.
If I want my program could tolerate 1 master node failure, I should catch the JedisConnectionException, try to create a new JedisCluster by calling new JedisCluster(nodes), and write data again. But this way isn't what I expected, or Jedis guides described.

I tried this test with Jedis 2.9.0 and 2.8.2, all happed as above. But when I run this test with Jedis 2.8.1, it works as I expected.

According to Jedis 2.9.0 sources, and code changes in Github, I thought the 131-141 lines of JedisClusterCommand.java may have a miss. As described in Issue #1252, the codes could be adding some more logic as below:

131      if (attempts == 1) {
132        //We need this because if node is not reachable anymore - we need to finally initiate slots renewing,
133        //or we can stuck with cluster state without one node in opposite case.
134        //But now if maxAttempts = 1 or 2 we will do it too often. For each time-outed request.
135        //TODO make tracking of successful/unsuccessful operations for node - do renewing only
136        //if there were no successful responses from this node last few seconds
137        this.connectionHandler.renewSlotCache();
+++      }
138
+++      if (attempts < 1) {
139        //no more redirections left, throw original exception, not JedisClusterMaxRedirectionsException, because it's not MOVED situation
140        throw jce;
141      }

or change logic as below:

131      if (attempts == 1) {
132        //We need this because if node is not reachable anymore - we need to finally initiate slots renewing,
133        //or we can stuck with cluster state without one node in opposite case.
134        //But now if maxAttempts = 1 or 2 we will do it too often. For each time-outed request.
135        //TODO make tracking of successful/unsuccessful operations for node - do renewing only
136        //if there were no successful responses from this node last few seconds
137        //this.connectionHandler.renewSlotCache();
+++      tryRandomNode = true;
+++      }
138
+++      if (attempts < 1) {
139        //no more redirections left, throw original exception, not JedisClusterMaxRedirectionsException, because it's not MOVED situation
140        throw jce;
141      }

### Steps to reproduce:

1. Running a Redis Cluster which contains master nodes and slave nodes.
2. Keeping writing data to cluster by using JedisCluster. 
3. Killing one master, and the slave switching to master.
4. The test throwing a JedisConnectionException. 

Redis run as a cluster with 4 master and 4 slave.
And using JedisCluster with default timeout and retry times.

#### Jedis version:

Jedis 2.9.0, 2.8.2 

#### Redis version:

Redis 3.2.1

#### Java version:

Java 8u92

ljrmorgan:
We lost one of our cluster's master nodes and started seeing a lot of `JedisConnectionException`s being thrown. The cluster itself was working, but Jedis seemed to be hanging on to the old state. Searching in the issues we found #1439 which seems to point out the issue: the call to `cluster slots` to refresh the state of the cluster is called _after_ the final retry, so has no effect.

With this change we call `renewSlotCache()` _before_ the final retry. As before `renewSlotCache()` will never be called if `maxAttempts` is two, since that would mean calling `cluster slots` after a single failure.

I've added a failing (well, now passing!) unit test to validate the fix. I'm using Mockito in that test which isn't being used elsewhere in the project, so I added it as a separate commit to make it easier to remove if you didn't want to add a dependency just for this bug fix.

Fixes #1439

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -60,6 +60,13 @@
 			<scope>test</scope>
 		</dependency>
 		<dependency>
+			<groupId>org.mockito</groupId>
+			<artifactId>mockito-core</artifactId>
+			<version>2.3.0</version>
+			<type>jar</type>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
 			<groupId>org.apache.commons</groupId>
 			<artifactId>commons-pool2</artifactId>
 			<version>2.4.2</version>
--- a/src/main/java/redis/clients/jedis/JedisClusterCommand.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterCommand.java
@@ -128,15 +128,15 @@ public abstract class JedisClusterCommand<T> {
       releaseConnection(connection);
       connection = null;
 
-      if (attempts <= 1) {
-        //We need this because if node is not reachable anymore - we need to finally initiate slots renewing,
-        //or we can stuck with cluster state without one node in opposite case.
-        //But now if maxAttempts = 1 or 2 we will do it too often. For each time-outed request.
+      int attemptsRemaining = attempts - 1;
+      if (attemptsRemaining == 1 && maxAttempts > 2) {
+        // May be that the node is not reachable anymore, try calling cluster slots before our final retry
+        // Only do this if maxAttempts > 2 to avoid calling cluster slots on any connection error
         //TODO make tracking of successful/unsuccessful operations for node - do renewing only
         //if there were no successful responses from this node last few seconds
         this.connectionHandler.renewSlotCache();
-
-        //no more redirections left, throw original exception, not JedisClusterMaxRedirectionsException, because it's not MOVED situation
+      } else if (attemptsRemaining <= 0) {
+        // No more redirections left, throw original exception, not JedisClusterMaxRedirectionsException, because it's not MOVED situation
         throw jce;
       }
 
--- a/src/test/java/redis/clients/jedis/JedisClusterCommandTest.java
+++ b/src/test/java/redis/clients/jedis/JedisClusterCommandTest.java
@@ -1,0 +1,53 @@
+package redis.clients.jedis;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.InOrder;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.junit.MockitoJUnitRunner;
+import redis.clients.jedis.exceptions.JedisConnectionException;
+
+import static org.mockito.ArgumentMatchers.anyInt;
+import static org.mockito.Mockito.when;
+
+/**
+ * @author Louis Morgan
+ */
+@RunWith(MockitoJUnitRunner.class)
+public class JedisClusterCommandTest {
+
+  @Mock
+  private Jedis connection;
+
+  @Mock
+  private JedisClusterConnectionHandler connectionHandler;
+
+  @Test
+  public void onConnectionExceptionClusterRenewsSlotCacheAndThenRetries() {
+    when(connectionHandler.getConnectionFromSlot(anyInt())).thenReturn(connection);
+    when(connection.get("foo")).thenThrow(new JedisConnectionException("Command failed"));
+
+    try {
+      new JedisClusterCommand<String>(connectionHandler, 3) {
+        @Override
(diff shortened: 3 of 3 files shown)


marcosnils:
@ljrmorgan thx for the change. It might take some time until I can review @HeartSaVioR @Spikhalskiy can you help with this please?

Spikhalskiy:
> the call to cluster slots to refresh the state of the cluster is called after the final retry, so has no effect.

I'm not sure it's the reason. Ok, we initiate cluster renew after first bunch of requests and this requests would fail, but we schedule this renew and in some time new requests should get a fresh state. So, original code shouldn't stick in old state infinitely.

> if maxAttempts is two, since that would mean calling cluster slots after a single failure.

I don't like this change. This would mean that we will actually never schedule slots renew with maxAttempts=2. And only with this value. We would schedule it with other values. So, as a result we get pretty different Jedis behaviour when maxAttempts=2 and maxAttempts>2. Could be very misleading and unexpectedly.

ljrmorgan:
Thanks for looking at this @HeartSaVioR and @Spikhalskiy, and apologies for the late response.

I've been looking through our splunk logs from when we hit this, and I think my understanding of the issue was a bit off before.

We saw several hundred JedisConnectionExceptions each time a master node was lost, during a fairly quiet time for us traffic wise. We would see these exceptions over a period of a minute or two after each master loss, so it does _seem_ like Jedis recovers. I say seem there, because we're using Jedis via Spring Data Redis, and I'm not sure if the exceptions stop because Jedis recovers itself, or if Spring Data Redis is doing something to recover. I'm not particularly familiar with the internals of Spring Data Redis unfortunately.

I think there might still be some value in this change. In particular, calls that would otherwise fail with a JedisConnectionException could instead recover, which would have avoided the few hundred errors that we saw on each master loss. That feels more robust to me, and we'd call `cluster slots` anyway, so we aren't making additional Redis calls. Perhaps just letting the call fail is a less surprising thing to do though?

I agree that the way I'm handling the `maxAttempts=2` case is incorrect, I hadn't appreciated that Jedis was recovering by calling refresh after the final attempt, and I think I misunderstood the comment about calling `cluster slots` too aggressively. I'd propose:
- Calling `cluster slots` _before_ the final retry if maxAttempts > 2
- Calling `cluster slots` _after_ the final retry otherwise to avoid calling it after one connection error if maxAttempts == 2.

Does that seem like a reasonable approach? Please let me know and I can make the corresponding change to my PR.

Thanks!

ljrmorgan:
Incidentally, immediately after this we also saw a *lot* of calls to Spring Data Redis' `RedisCache::get()` which never returned. We had to kill the machines before we could investigate further unfortunately, and I haven't been able to reproduce it since. I'm not sure if that was caused/aggravated by these exceptions. Looking through your issues it sounds like it could be #1158 that caused this, I'll try setting `setMaxWaitMillis` on our `JedisPoolConfig`, which we're currently just default constructing. This subsequent issue was the real show stopper for us, rather than the JedisConnectionExceptions themselves.

marcosnils:
notifying @mp911de as he has experience in spring-data-redis

ljrmorgan:
@marcosnils thanks, I'll try to reproduce on my end and isolate the hang, that should help determine if it's the Jedis issue I linked to or a Spring Data Redis issue (or an issue in our code)

mp911de:
Spring Data Redis isn't doing much in terms of recovery. Certain commands perform a topology lookup but the majority of commands uses simply `JedisCluster`.

ljrmorgan:
@marcosnils @mp911de thanks guys, the hang seems to be #1158, so not a Spring Data Redis issue. I'll add a comment and stack trace to that issue, I don't want to derail this issue too much.

I still think calling `cluster slots` _before_ the final retry rather than after is worth doing to avoid some `JedisConnectionException`s, what do you think?

sazzad16:
Since #2358, there will be more `renewSlotCache()` calls internally. IMO, we can close this in favor of that.

PR #1447, opened 2016-12-27
enhance of restore command

zts1993:
1. restore  command support REPLACE arg
2. Add cluster dump restore support

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryClient.java
+++ b/src/main/java/redis/clients/jedis/BinaryClient.java
@@ -1083,8 +1083,20 @@ public class BinaryClient extends Connection {
     sendCommand(DUMP, key);
   }
 
-  public void restore(final byte[] key, final int ttl, final byte[] serializedValue) {
-    sendCommand(RESTORE, key, toByteArray(ttl), serializedValue);
+  public void restore(final byte[] key, final int ttl, final byte[] serializedValue, final String... replaceArgs) {
+
+    boolean replace = false;
+    for (String replaceArg : replaceArgs) {
+      if (replaceArg.equalsIgnoreCase(Protocol.DATA_REPLACE)) {
+        replace = true;
+      }
+    }
+
+    if (replace) {
+      sendCommand(RESTORE, key, toByteArray(ttl), serializedValue, SafeEncoder.encode(Protocol.DATA_REPLACE));
+    } else {
+      sendCommand(RESTORE, key, toByteArray(ttl), serializedValue);
+    }
   }
 
   public void pexpire(final byte[] key, final long milliseconds) {
--- a/src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
+++ b/src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
@@ -14,9 +14,11 @@ import java.util.Set;
 
 import org.junit.Test;
 
+import redis.clients.jedis.Protocol;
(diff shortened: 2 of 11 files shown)


src/main/java/redis/clients/jedis/BinaryClient.java
```diff
@@ -1083,8 +1083,20 @@ public void dump(final byte[] key) {
     sendCommand(DUMP, key);
   }
 
-  public void restore(final byte[] key, final int ttl, final byte[] serializedValue) {
-    sendCommand(RESTORE, key, toByteArray(ttl), serializedValue);
+  public void restore(final byte[] key, final int ttl, final byte[] serializedValue, final String... replaceArgs) {
```
marcosnils: Hmmm I don't like adding varargs just to make the replace parameter optional, it makes the interface less clear. I believe we should add a new `restore` method that makes this explicit.

src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
```diff
@@ -488,6 +490,23 @@ public void dumpAndRestore() {
     assertTrue(jedis.exists("foo2"));
   }
 
+  @Test(expected = JedisBusyException.class)
```
HeartSaVioR: Can we get the exception consistently? Or does it rely on timing?

src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
```diff
@@ -488,6 +490,23 @@ public void dumpAndRestore() {
...
+    jedis.restore("foo1", 0, sv);
+    assertTrue(jedis.exists("foo1"));
+  }
+
+  @Test
+  public void dumpAndRestoreReplace() {
+    jedis.set("foo1", "bar1");
+    byte[] sv = jedis.dump("foo1");
+    assertTrue(jedis.exists("foo1"));
+    jedis.restore("foo1", 0, sv, Protocol.DATA_REPLACE);
```
HeartSaVioR: I'm not in favor of exposing Protocol to user. It should be a param class which might be similar to SetParam.

https://github.com/xetorthio/jedis/blob/master/src/main/java/redis/clients/jedis/params/set/SetParams.java

src/main/java/redis/clients/jedis/BinaryClient.java
HeartSaVioR: Yes and I think introducing param class like SetParam should be better.

src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
zts1993: i had considered about Params but , it looks like Params contains Key-value , replace is the only arg , i am not sure where the "REPLACE" should be placed.

src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
zts1993: JedisBusyException is more accurate but using JedisException is ok .
no relations to time.

marcosnils: If we already have JedisBusyException I guess it's best to expect that one instead of a generic one

zts1993:
Hi , all. i start a new pull request at #1451 . 
Closing now

Issue #968, opened 2015-04-29
How to know when scan() is done?


veselov:
Hi.

From implementation, and Redis docs, it looks like that to know if scan() is finished, one needs to compare the latest cursor to literal `0`, right?

If so, may be ScanResult can have a isLast() method that would do that?

marcosnils:
@veselov nice observation. What about renaming the method to something like `next` so the interface can be used like the `Iterator` one. 
i.e:

```java
while (scanResult.next()) {
  //Do something
}
```

marcosnils:
@veselov maybe we can make ScanResult implement `Iterator`? so it can be used in `foreach` loops also.

marcosnils:
Hmmm thinking about this implementing `Iterator` wouldn't be the best thing because for the `next` method we'd have to call Jedis and do some exception handling, etc. I believe we can just provide a `hasNext()` method as @veselov originally proposed. 

@HeartSaVioR @veselov @xetorthio what do u think?

HeartSaVioR:
For supporting various versions of Redis, Jedis didn't evaluate return value to determine sth. .
I'm not 100% sure that antirez never changes representation of "end of scan".
If we're sure, we can just go on. Actually it's better to have it.

veselov:
I am all for making ScanResult return an iterator, and even stream for 1.8, as a utility mechanism. The caveat will be that the next(), or the stream producer may throw an exception. Which, as long as the implementation is ready for, can be handled. On the low level, just having a hasMore()/scanFinished() method is sufficient to prevent the Jedis user to having to have Redis knowledge on how to do determine if the scan is finished.

allanwax:
When the scan returns an index of `0` it is done

HeartSaVioR:
We know about that, but if Jedis has to know about that, and when antirez changes value for end of scan, next Jedis version couldn't be compatible with previous version. We were trying our best to avoid it.

veselov:
I agree with @HeartSaVioR point, but wouldn't that be tantamount to changing the protocol? Changing `cursor==0` condition is no different than changing any other contract Redis has, but Jedis does depend on other contracts.

HeartSaVioR:
http://redis.io/commands/ttl
You can see that return values are changed once. (Added -2)
I don't think it shouldn't be occurred again, but who knows.

corydolphin:
Hi all, awesome discussion here! I think there are some great ideas here.

[...]

Issue #1458, opened 2017-01-06
IMPROVEMENT: add more prominent warning to keys() methods


kutzi:
On Jedis/MultiKeyCommands there's a long Javadoc describing the performance issues of KEYS.
But as obviously most devs don't really read the Javadoc - and then we end up with keys() calls in production code - could you please add a more prominent warning to it?

The only thing which I actually thing could work, would be .@Deprecated, though

marcosnils:
@kutzi thx for the suggestion. I don't believe **@Deprecated** is an option as it doesn't reflect the reality which is that KEYS is still a valid and long standing command. 

I might be wrong, but I believe this is just a matter of understanding how redis works rather than improving documentation from the redis side. I don't think there's much we can do from our end as KEYS vs SCAN depends entirely on the use case

@HeartSaVioR ?

kutzi:
Totally agree that it is a valid command. However no one should ever be using it IMHO.
I also find @Deprecated a perfect solution, but it's the only one I know of which would give an automated warning when using the method.

marcosnils:
> However no one should ever be using it IMHO.

I respectfully disagree. Even the redis official doc (https://redis.io/commands/KEYS) says the it should be used with extremely precaution, for debbuging or changing keypsace layout (there's no other command that does this).

I still believe that Jedis users should understand what the base redis commands do, what are trade-offs of using them and when to do so.

kutzi:
https://redis.io/commands/KEYS
"Don't use KEYS in your regular application code. "

Yes, there might be edge cases when you need to use it - even in Java code and not just with the CLI.

So, IMO just put a description to @deprecated saying why it is isn't actually deprecated, but should be used with extreme care and e.g. linking to official documentation.

kutzi:
If there would be any other way than @deprecated I would perfectly agree to use that one instead

marcosnils:
> So, IMO just put a description to @deprecated saying why it is isn't actually deprecated, but should be used with extreme care and e.g. linking to official documentation.

This will definitely not happen, deprecated has a totally different meaning from an API perspective and will bring more confusion than clarification.

> If there would be any other way than @deprecated I would perfectly agree to use that one instead

Me too, but as there's no other way I don't see there's much to be done besides adding a regular javadoc which is already there.

kutzi:
>  deprecated has a totally different meaning from an API perspective
agree

>  and will bring more confusion than clarification.
[...]

PR #1459, opened 2017-01-06
#1458 add a more prominent warning for keys()

kutzi:
Also added some doc for scan() so that the proposed alternative is more accessible

### DIFF
--- a/src/main/java/redis/clients/jedis/ScanResult.java
+++ b/src/main/java/redis/clients/jedis/ScanResult.java
@@ -25,6 +25,15 @@ public class ScanResult<T> {
     return SafeEncoder.encode(cursor);
   }
 
+  /**
+   * Is the iteration complete. I.e. was the complete dataset scanned.
+   *
+   * @return true if the iteration is complete
+   */
+  public boolean isCompleteIteration() {
+    return ScanParams.SCAN_POINTER_START.equals(getCursor());
+  }
+
   public byte[] getCursorAsBytes() {
     return cursor;
   }
--- a/src/main/java/redis/clients/jedis/ScanResult.java
+++ b/src/main/java/redis/clients/jedis/ScanResult.java
@@ -17,6 +17,10 @@ public class ScanResult<T> {
     this.results = results;
   }
 
+  /**
+   * Returns the new value of the cursor
+   * @return the new cursor value. {@link ScanParams#SCAN_POINTER_START} when a complete iteration has finished
+   */
   public String getCursor() {
     return SafeEncoder.encode(cursor);
   }
@@ -25,6 +29,10 @@ public class ScanResult<T> {
     return cursor;
   }
 
+  /**
+   * The scan results from the current call.
+   * @return the scan results
+   */
   public List<T> getResult() {
     return results;
   }
--- a/src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
+++ b/src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
@@ -621,14 +621,19 @@ public class AllKindOfValuesCommandsTest extends JedisCommandTestBase {
       jedis.set("a" + i, "a" + i);
     }
 
-    ScanResult<String> result = scanCompletely();
+    ScanResult<String> result = jedis.scan(SCAN_POINTER_START);
+    // note: in theory Redis would be allowed to already return all results on the 1st scan,
+    // but in practice this never happens for data sets greater than a few tens
+    // see: https://redis.io/commands/scan#number-of-elements-returned-at-every-scan-call
+    assertFalse(result.isCompleteIteration());
+
+    result = scanCompletely(result.getCursor());
 
     assertNotNull(result);
     assertTrue(result.isCompleteIteration());
   }
 
-  private ScanResult<String> scanCompletely() {
-    String cursor = SCAN_POINTER_START;
+  private ScanResult<String> scanCompletely(String cursor) {
     ScanResult<String> scanResult;
     do {
       scanResult = jedis.scan(cursor);
--- a/src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
+++ b/src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
@@ -4,6 +4,7 @@ import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertNotNull;
 import static redis.clients.jedis.ScanParams.SCAN_POINTER_START;
 import static redis.clients.jedis.ScanParams.SCAN_POINTER_START_BINARY;
 import static redis.clients.jedis.params.set.SetParams.setParams;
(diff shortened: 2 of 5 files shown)


marcosnils:
> f you don't like that method than I can revert it.

Which method?.

kutzi:
The one in this commit: https://github.com/xetorthio/jedis/pull/1459/commits/5d6e9f4984822d648e75f2fd0972fc6c5bf3c3bd

kutzi:
BTW: I looked into extending the tests, but then I found a bit strange what they're asserting already.
For example:
https://github.com/xetorthio/jedis/blob/master/src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java#L554
Why would it assert that the cursor is '0' after the 1st scan() call already?
That's against the guarantees (or better the lack of that) of SCAN:
https://redis.io/commands/scan#number-of-elements-returned-at-every-scan-call

src/main/java/redis/clients/jedis/ScanResult.java
```diff
@@ -17,14 +17,31 @@ public ScanResult(byte[] cursor, List<T> results) {
...
   public String getCursor() {
     return SafeEncoder.encode(cursor);
   }
 
+  /**
+   * Is the iteration complete. I.e. was the complete dataset scanned.
+   *
+   * @return true if the iteration is complete
+   */
+  public boolean isCompleteIteration() {
```
marcosnils: This needs a test

marcosnils:
> Why would it assert that the cursor is '0' after the 1st scan() call already?

this asserts that the cursor is being set properly in the ScanResult object.

If you do `scan 0`, which indicates a new scan, the current cursor should always be 0 until you do a next `scan` with the next cursor

kutzi:
No scan should return the new position of the cursor.
Which just happens to be 0 in that case, because the data set is very small

src/main/java/redis/clients/jedis/ScanResult.java
kutzi: Sure, if you tell me how to run tests

marcosnils:
> No scan should return the new position of the cursor.
> Which just happens to be 0 in that case, because the data set is very small

That's what I meant. For this specific case as the default COUNT is 10 and we're only inserting two keys, the cursor will be always 0. That's how we test the we're populating the cursor correctly as it's deterministic

src/main/java/redis/clients/jedis/ScanResult.java
marcosnils: Install latest `redis` version from `unstable` branch, `make` and just run `make test` from the project root

kutzi:
For all practical cases that may work, but it's against the guarantees.
"SCAN family functions do not guarantee that the number of elements returned per call are in a given range. The commands are also allowed to return zero elements,"
That's all what I said

kutzi:
And to illustrate my point: I think unit tests should assert that the guarantees are kept and *not* that some by-chance practical results

kutzi:
BTW: count is just a *hint* and does not guarantee that that number of elements are returned.
Read https://redis.io/commands/scan#the-count-option

src/main/java/redis/clients/jedis/ScanResult.java
kutzi: There's no way to just run and debug a single test?

marcosnils:
> And to illustrate my point: I think unit tests should assert that the guarantees are kept and not that some by-chance practical results

The purpose of Jedis tests is not to test `redis` guarantees, that's why `redis `has it's own tests. We only test that Jedis is parsing the redis protocol correctly and that we populate our objects with the proper values. 

You're right that it **might** happen that either the count or the cursor doesn't return as expected due to SCAN guarantees. However, we haven't had any errors with SCAN tests so far. 

Having said that, if you think that the test can be improved, we'd like to hear your opinions about it can be improved. Please open an issue to discuss about it if you have any ideas.

src/main/java/redis/clients/jedis/ScanResult.java
marcosnils: Yes, run `make start` and then you can debug your test using your IDE debugger

kutzi: Thanks

kutzi: No matter if I do make, make test or make start, I get:

```
~/work/jedis $ make
rm -vf /tmp/redis_cluster_node*.conf 2>/dev/null
rm dump.rdb appendonly.aof - 2>/dev/null
make: [cleanup] Fehler 1 (ignoriert)
echo "$REDIS1_CONF" | redis-server -

*** FATAL CONFIG FILE ERROR ***
Reading the configuration file, at line 2
>>> 'protected-mode no'
Bad directive or wrong number of arguments
make: *** [start] Fehler 1
```

marcosnils: Did you install latest redis version form unstable ranch?

kutzi:
> The purpose of Jedis tests is not to test redis guarantees

BTW: that was not what I was saying/demanding, but let's not stress this issue anymore.
I'll open another PR with some suggested changes

src/main/java/redis/clients/jedis/ScanResult.java
kutzi: Yes, I can successfully connect to it with redis-cli

kutzi: I mean I downloaded the tar from here https://redis.io/download and build it, if that's what you mean with 'install from unstable branch'

marcosnils:
> I mean I downloaded the tar from here https://redis.io/download and build it, if that's what you mean with 'install from unstable branch'

src/main/java/redis/clients/jedis/ScanResult.java
marcosnils: You need to clone redis from the official repo https://github.com/antirez/redis run `make` and `make install` to install the latest version.

kutzi: I did this now, but it makes no difference

marcosnils: what does `redis-server --version` print?

kutzi: Wait.
Sorry, there's still another redis on the path

marcosnils: Just what I thought.

kutzi: :-)

kutzi:
There you are

src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
```diff
@@ -615,6 +616,29 @@ public void scanCount() {
   }
 
   @Test
+  public void scanIsCompleteIteration() {
+    for (int i = 0; i < 100; i++) {
+      jedis.set("a" + i, "a" + i);
+    }
+
+    ScanResult<String> result = scanCompletely();
```
marcosnils: WDYT if we add an assertion that `result.isCompleteIteration()` is false before the scan?. Otherwise, if `isCompleteIteration()` returns a hardcoded `true` this test wouldn't catch it.

kutzi: Well, there we are again at the problem, that *in theory* I cannot test this reliably ;-)
But nevertheless I added it

marcosnils:
LGTM, once another contributor reviews we can merge

HeartSaVioR:
@kutzi @marcosnils 
Sorry for late reviewing. Missed this PR. LGTM. Thanks for your contributions.

marcosnils:
merged to master and all corresponding branches. Thx @kutzi for the work!

sazzad16:
Backport commits:
2.10: https://github.com/xetorthio/jedis/commit/4fb6a77b4f802d7740aa6de217ace2409b33affb
2.9: https://github.com/xetorthio/jedis/commit/6d8211bc47dcadf37fa418798fabf89da5d6680f

sazzad16:
Resolves #968

sazzad16:
fixes #968

sazzad16:
Resolves #968

Issue #1463, opened 2017-01-21
IMPROVMENT: Jedis should distingusih between exhausted pool to other failures


shmulik-ubimo:
### Expected behavior

In continuance to issue #1236, I would expected Jedis to let me know that it failed to retrieve a connection from the inner pool, since the pool is exhausted by throwing an exception that indicates rather then a general one.   

### Actual behavior

When Jedis fails to retrieve a connection from the pool, it throws a `JedisException`, which is too general and can be thrown for a failure in the implemented `activateObject()` or `validateObject()` method.  

### Steps to reproduce:
1. Create a `GenericObjectPoolConfig` object with `setMaxTotal(1)` and `setBlockWhenExhausted(false)`.
2. Create a `JedisPool` with the configuration from step (1).
3. Try to retrieve a connection from the pool by `getResource()`.
4. Try to retrieve another connection and expect a `JedisException`. 

#### Jedis version:
3.0.0-SNAPSHOT
#### Redis version:
3.0.7
#### Java version:
1.8

marcosnils:
Let's continue the discussion in #1464

PR #1464, opened 2017-01-21
Throws JedisExhaustedPoolException when the inner connection pool is …

shmulik-ubimo:
…exhausted

Implements #1463, by introducing the new `JedisExhaustedPoolException`

### DIFF
--- a/src/main/java/redis/clients/jedis/exceptions/JedisExhaustedPoolException.java
+++ b/src/main/java/redis/clients/jedis/exceptions/JedisExhaustedPoolException.java
@@ -1,8 +1,8 @@
 package redis.clients.jedis.exceptions;
 
 /**
- * This exception will be thrown when the Jedis client won't be able to retreive a connection from
- * the pool, since all the connections are being used (aka. an "exhausted" pool).
+ * This exception will be thrown when the Jedis client isn't able to retrieve a connection from
+ * the pool, since all the connections are being used (a.k.a. an "exhausted" pool).
  */
 public class JedisExhaustedPoolException extends JedisException {
 
--- a/src/main/java/redis/clients/jedis/exceptions/JedisExhaustedPoolException.java
+++ b/src/main/java/redis/clients/jedis/exceptions/JedisExhaustedPoolException.java
@@ -1,0 +1,20 @@
+package redis.clients.jedis.exceptions;
+
+/**
+ * This exception will be thrown when the Jedis client won't be able to retreive a connection from
+ * the pool, since all the connections are being used (aka. an "exhausted" pool).
+ */
+public class JedisExhaustedPoolException extends JedisException {
+
+  public JedisExhaustedPoolException(String message) {
+    super(message);
+  }
+
+  public JedisExhaustedPoolException(Throwable e) {
+    super(e);
+  }
+
+  public JedisExhaustedPoolException(String message, Throwable cause) {
+    super(message, cause);
+  }
+}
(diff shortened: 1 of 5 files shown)


src/main/java/redis/clients/util/Pool.java
```diff
@@ -48,6 +49,11 @@ public T getResource() {
     try {
       return internalPool.borrowObject();
     } catch (NoSuchElementException nse) {
+      if (null == nse.getCause()) { // The exception was caused by an exhausted pool
```
marcosnils: This same code snippet could be validated by the program catching the exception. I don't really see a benefit by throwing this exception. Another thing is:, are you sure that every time getCasue returns null is because the pool is exhausted?. How do you know this?

shmulik-ubimo: The client can use the same code snippet by himself but first, the client should be aware of it by digging up to the inner use of `commons-pool2` and the way it distinguish between the two scenarios. 
I bump into it by accident, while examining a Jedis's monitoring failure in my own code and going through `commons-pool2` implementation.
I think that since Jedis's code catches and rethrows the customized exception, it should be specific on the cause and shouldn't be assume that the client aware of its inner use of third parties.
I'll add another unit test for the other scenario, if this still relevant.

marcosnils: @shmulik-ubimo got it. If you've been looking into the commos-pool code then it's fine. I believe it might be better then to add this logic to JedisPool instead of just JedisCluster. What do you think?. This way if someone is using `JedisPool` it will get this new exception

shmulik-ubimo: I meant to add it to `Pool.java` where `GenericObjectPool<T> internalPool` is declared (and from which `JedisPool` inherits)

marcosnils: > I meant to add it to Pool.java where GenericObjectPool<T> internalPool is declared (and from which JedisPool inherits)

That sounds perfect. Mind changing this PR?

HeartSaVioR: +1 to add this to Pool.

HeartSaVioR:
LGTM

HeartSaVioR:
Since this is not a bugfix I just set milestone to next minor. If someone wants to address this to 2.9.1 that would be also OK.

marcosnils:
@shmulik-ubimo this looks amazing!. Mind preserving the formatting?. Otherwise `make format` will reformat and we'll end in an endless formatting loop :tongue:

src/main/java/redis/clients/jedis/exceptions/JedisExhaustedPoolException.java
```diff
@@ -0,0 +1,20 @@
+package redis.clients.jedis.exceptions;
+
+/**
+ * This exception will be thrown when the Jedis client won't be able to retreive a connection from
```
avram: Small spelling/grammar nits:
"won't" --> "isn't" (or "wasn't")
"retreive" --> "retrieve"
"aka." --> "a.k.a."

shmulik-ubimo: Thanks!

shmulik-ubimo:
@marcosnils I ran `make format` on my branch and it changed 49 files, which only one of them is relevant to this PR. I've only fixed the single file formatting issue.

marcosnils:
@shmulik-ubimo  can you rebase with master so tests pass please?

shmulik-ubimo:
@marcosnils I've rebased it, the build is still failing.

marcosnils:
@shmulik-ubimo thx for rebasing. Build si failing due to travis issues. Just re ran and hopefully it'll fix it.

marcosnils:
merged to master and 2.10 accordingly. Thx for contributing @shmulik-ubimo!.

shmulik-ubimo:
@marcosnils Thanks!

Issue #1340, opened 2016-07-06
Feature proposal : Integrate a more advanced logging framework


sheinbergon:
Though I understand the considerations of keeping Jedis lean & mean and the overall aim of not binding to specific logging frameworks in 3rd party libraries in the likes of Jedis, I find the JUL interface to be rather crude / obsolete .

Plus , it pales in comparison to other logging solutions performance wise : 

http://blog.takipi.com/the-logging-olympics-a-race-between-todays-top-5-logging-frameworks/
https://logging.apache.org/log4j/2.x/performance.html

I have very good experience with Log4J2 in other projects I use it.

@marcosnils @HeartSaVioR  was this ever discussed  ?

xetorthio:
Never discussed before. I wonder what other big java projects are using...

marcosnils:
It has been discussed indeed. https://github.com/xetorthio/jedis/issues/1219

HeartSaVioR:
Yeah discussed before. Note that we're only using log to JedisSentinelPool.

sheinbergon:
@HeartSaVioR  @marcosnils  I've gone over #1219 thoroughly , the arguments there indeed make sense ( and not surprising at all ) . 

If I were to implement an internal logger factory as @mp911de mentioned (and portrayed here  https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/internal/logging/InternalLoggerFactory.java#L51 )   but make it a bit smartter ( like support placeholders in a unified fashion )  -

Is that something you'd be inclined to merge into jedis ? it won't effect the dependency graph and would provide  a nicer logging factory ? 

Hoping for a yes :) let me know  :)

marcosnils:
@sheinbergon @HeartSaVioR I do agree that as Jedis / Redis is becoming more complex (with cluster and modules) it'd be nice to add logging in order to print some debug information when needed. @sheinbergon if dependency graph is kept small and adding logging doesn't require too much effort, I believe that we can include it and use it when needed.

sheinbergon:
@marcosnils  that's great . Let's leave this issue open , I'll work on it as soon as I clear some other stuff from the table

r0main:
Hi,

Which logging interface do you want to use ? 

[...]

PR #1467, opened 2017-02-10
Replace JUL by slf4j (#1340)

r0main:
A pull request to replace JUL by slf4j.

Used log4j2 with slf4j bridge for testing.

Just had a little trouble to replace  "logp" of JUL. It seems there is no equivalent in SLF4j. Did my best to simulate the same log with SLF4j.

I didn't manage to launch all unit tests. Only a fraction works on my environment. But I have the same number of failure before and after. I was able to saw logging redirect through log4j2 during unit testing. All seems OK, but please launch unit tests to be sure before merging.

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisSentinelPool.java
+++ b/src/main/java/redis/clients/jedis/JedisSentinelPool.java
@@ -5,10 +5,10 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.logging.Level;
-import java.util.logging.Logger;
 
 import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import redis.clients.jedis.exceptions.JedisConnectionException;
 import redis.clients.jedis.exceptions.JedisException;
@@ -28,7 +28,7 @@ public class JedisSentinelPool extends JedisPoolAbstract {
 
   protected Set<MasterListener> masterListeners = new HashSet<MasterListener>();
 
-  protected Logger log = Logger.getLogger(getClass().getName());
+  protected Logger log = LoggerFactory.getLogger(getClass().getName());
 
   private volatile JedisFactory factory;
   private volatile HostAndPort currentHostMaster;
@@ -137,7 +137,7 @@ public class JedisSentinelPool extends JedisPoolAbstract {
     for (String sentinel : sentinels) {
       final HostAndPort hap = HostAndPort.parseString(sentinel);
 
-      log.fine("Connecting to Sentinel " + hap);
+      log.debug("Connecting to Sentinel {}", hap);
 
       Jedis jedis = null;
       try {
@@ -149,19 +149,19 @@ public class JedisSentinelPool extends JedisPoolAbstract {
         sentinelAvailable = true;
 
         if (masterAddr == null || masterAddr.size() != 2) {
-          log.warning("Can not get master addr, master name: " + masterName + ". Sentinel: " + hap
-              + ".");
+          log.warn("Can not get master addr, master name: {}. Sentinel: {}", masterName, hap);
           continue;
         }
 
         master = toHostAndPort(masterAddr);
-        log.fine("Found Redis master at " + master);
+        log.debug("Found Redis master at {}", master);
         break;
       } catch (JedisException e) {
         // resolves #1036, it should handle JedisException there's another chance
         // of raising JedisDataException
-        log.warning("Cannot get master address from sentinel running @ " + hap + ". Reason: " + e
-            + ". Trying next one.");
+        log.warn(
+          "Cannot get master address from sentinel running @ {}. Reason: {}. Trying next one.", hap,
+          e.toString());
       } finally {
         if (jedis != null) {
           jedis.close();
@@ -278,7 +278,7 @@ public class JedisSentinelPool extends JedisPoolAbstract {
           j.subscribe(new JedisPubSub() {
             @Override
             public void onMessage(String channel, String message) {
-              log.fine("Sentinel " + host + ":" + port + " published: " + message + ".");
+              log.debug("Sentinel {}:{} published: {}.", host, port, message);
 
               String[] switchMasterMsg = message.split(" ");
 
@@ -287,13 +287,15 @@ public class JedisSentinelPool extends JedisPoolAbstract {
                 if (masterName.equals(switchMasterMsg[0])) {
                   initPool(toHostAndPort(Arrays.asList(switchMasterMsg[3], switchMasterMsg[4])));
                 } else {
-                  log.fine("Ignoring message on +switch-master for master name "
-                      + switchMasterMsg[0] + ", our master name is " + masterName);
+                  log.debug(
+                    "Ignoring message on +switch-master for master name {}, our master name is {}",
+                    switchMasterMsg[0], masterName);
                 }
 
               } else {
-                log.severe("Invalid message received on Sentinel " + host + ":" + port
-                    + " on channel +switch-master: " + message);
+                log.error(
+                  "Invalid message received on Sentinel {}:{} on channel +switch-master: {}", host,
+                  port, message);
               }
             }
           }, "+switch-master");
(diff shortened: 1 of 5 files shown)


marcosnils:
LGTM @sheinbergon @HeartSaVioR WDYT?

marcosnils:
seems like redis unstable is not compiling that's why tests failed. I'll re-run when it gets fixed.

HeartSaVioR:
I know slf4j is used widely so application which depends on Jedis tends to depend on slf4j too. 
But for other cases, could we leave some notes that Jedis depends on slf4j and application should include logging framework and plugin? 
Maybe we just need to describe that we just added slf4j and don't add any plugins for logging framework, given that most users are already familiar with slf4j.

HeartSaVioR:
And since we're adding dependency, we need to decide which version lines we're OK to apply: minor or bugfix.
@marcosnils WDYT?

sheinbergon:
@r0main  @marcosnils  LGTM. 
SLF4ĵ is easily bridgble with other logging frameworks(such as log4j2) so it's good.

However, I dont think we should include log4j2 tests because they are way too specific and do not actuslly relate to the slf4j integration itself

sheinbergon:
@r0main going over the commits, I noticed stuff like "stringifying" exceptions [```e.toString()```] . Maybe things were like this in their JUL version, but I think it's a good idea to do it properly. Also - combining bind params with exceptions in a manner which I'm not sure behaves properly( it's better to check and be on the safe side). @marcosnils IMO it'd be a good idea to relate to these issues before proceefing

r0main:
@sheinbergon I had the same thoughts than you about toString on exceptions... I chose to stick to what was done before, but maybe it' is not the best approach ? I did not want to modify the behaviour in this pull request, just replace JUL by SLF4J. I felt it was not my choice to make.

 - What are your doubts with parametrized messages and exception ? You can do this since SLF4J 1.6.0 (May 2010) : https://www.slf4j.org/faq.html#paramException And I know it works with log4j2. Don't know about Logback.

 - What other framework than log4j2 do you want to use for testing ? Logback ?

marcosnils:
@r0main  can you rebase with master so tests pass please?

r0main:
Tests are fine on Java 7 but not on Java 8 with only one test failing : 
```
redis.clients.jedis.tests.JedisClusterTest
testReturnConnectionOnRedirection(redis.clients.jedis.tests.JedisClusterTest)  Time elapsed: 0.639 sec  <<< ERROR!
java.lang.Exception: Unexpected exception, expected<redis.clients.jedis.exceptions.JedisClusterMaxRedirectionsException> but was<redis.clients.jedis.exceptions.JedisException>
	at redis.clients.jedis.tests.JedisClusterTest.testReturnConnectionOnRedirection(JedisClusterTest.java:592)
Caused by: java.lang.IllegalStateException: Invalidated object not currently part of this pool
	at redis.clients.jedis.tests.JedisClusterTest.testReturnConnectionOnRedirection(JedisClusterTest.java:592)
```
Do you have an idea why is it failing just on Java 8 ? I still can't run the tests on my machine, sorry... But I re-read my commit. I don't see how I could have created that regression. Didn't change any try/catch clause.

Otherwise you can see log are formatted correctly during test event with parametrized message with exception : 
```
[ERROR] 2017-02-17T15:30:09,062 [main]                                  r.c.j.HostAndPort - redis.clients.jedis.HostAndPort.getLocalHostQuietly : cant resolve localhost address
java.net.UnknownHostException: jedis: jedis: Name or service not known
	at java.net.InetAddress.getLocalHost(InetAddress.java:1505) ~[?:1.8.0_111]
	at redis.clients.jedis.HostAndPort.getLocalHostQuietly(HostAndPort.java:106) [classes/:?]
	at redis.clients.jedis.HostAndPort.<clinit>(HostAndPort.java:13) [classes/:?] 
...
```

marcosnils:
@r0main tests sometimes fail due to machine slowness. Just re-ran them and everything looks good now.

r0main:
OK cool ! Tell me if you need anything else on the subject.

sheinbergon:
@r0main disregard the whole parameterized exception, it's my bad ( guess my memory of the matter was wrong). In regards to stringfying exceptions - I still think this should part of the PR.

We can always make another one though , so up to you. Thanks again

r0main:
@sheinbergon As the stringify happen on a log.warn. Do you want me to attach the stacktrace with the log or just the class (calling e.getClass().getName() instead of e.toString()) ? 

Maybe that why the developer stringified the exception in the first place. To not include the stacktrace on a warn message. I'm not sure the stacktrace is relevant in this scenario. But I'm not sure of the opposite either :)

What do you think ?

marcosnils:
> However, I dont think we should include log4j2 tests because they are way too specific and do not actuslly relate to the slf4j integration itself

@sheinbergon any extra thoughts about this comment before merging?

marcosnils:
@HeartSaVioR :bell:

src/main/java/redis/clients/jedis/JedisSentinelPool.java
```diff
@@ -287,29 +287,31 @@ public void onMessage(String channel, String message) {
...
               }
             }
           }, "+switch-master");
 
         } catch (JedisException e) {
 
           if (running.get()) {
-            log.log(Level.SEVERE, "Lost connection to Sentinel at " + host + ":" + port
-                + ". Sleeping 5000ms and retrying.", e);
+            log.error("Lost connection to Sentinel at {}:{}. Sleeping 5000ms and retrying.", host,
```
HeartSaVioR: I guess `e` is treated to one of Object varargs, not treated to Throwable object. @r0main Could you check it?
If I'm right, you need to use traditional string concatenation or String.format() in this case.

HeartSaVioR:
Found a missing spot. Others are OK.

src/main/java/redis/clients/jedis/JedisSentinelPool.java
sheinbergon: That was my initial concern ( and a recollection of how log4j2 cannot mix up varags and exceptions together without sacrificing the stacktrace formatting.

sheinbergon:
@marcosnils  https://github.com/xetorthio/jedis/pull/1467#issuecomment-281181481 I just think tests should pass without failing using the default NO-OP backend. But that's just my POV. It's not crucial

src/main/java/redis/clients/jedis/JedisSentinelPool.java
HeartSaVioR: Yes, it's not `syntactically` possible in method signature. It will require more method call or other trick even if library provides such feature.

mballoni: If I understood it right you are talking about the throwable parameter, right?
SLF4J preserves the stacktrace whenever the last parameter is the throwable:
https://www.slf4j.org/faq.html#paramException

HeartSaVioR: Great. Smart enough.

marcosnils:
> I just think tests should pass without failing using the default NO-OP backend

makes sense. @r0main is there a reason to include lo4j2 during tests?. I see that test outputs are filled with logs now. I'm not sure if logging during tests is what we want.

mballoni:
Usually I prefer logback.
However for tests purpose, as stated by @r0main , it is just a matter of readability in case of test errors so whatever implementation chosen will fit the purpose.

marcosnils:
> If you really don't want to see logs during tests I can use the NO-OP.

+1 on removing logj2 dependency and just use no-op for tests. In any case, logging in test wasn't used before

marcosnils:
I'll merge this as it is whenever I have some time as changing log4j2 to logback doesn't really make a difference.
Thx everyone for your contribution.

r0main:
> I'll merge this as it is whenever I have some time as changing log4j2 to logback doesn't really make a difference.

Do you want me to correct the pull request ? I tested logback-classic:1.2.1 on my machine earlier today to see if it behave correctly. I just have to push if you want. Do you want me to make a new commit or rebase + force push ?

marcosnils:
@r0main it's not necessary. I don't see any difference in using log4j2 or logback for testing purposes. It's fine as it is for me.

HeartSaVioR:
OK great. LGTM

HeartSaVioR:
Merged into master and 2.10 branches. Thanks everyone!

PR #1468, opened 2017-02-17
HostAndPort.convertHost() to resolve localhost in a lazy manner

### LINKED ISSUE #1424: Jedis should not require localhost to be resolved so it can run clean in serverless env
### Expected behavior
Jedis should not require localhost to be resolved so it can run clean in serverless env - aws lambda. 

### Actual behavior
When running jedis on aws lambda, it tries to resolve localhost which is not available in serverless env resulting in a "SEVERE" log output.

Write here what happens instead ...

    15:02:45 SEVERE: cant resolve localhost address
    15:02:45 java.net.UnknownHostException: ip-10-29-121-116: ip-10-29-121-116: unknown error
    15:02:45 at java.net.InetAddress.getLocalHost(InetAddress.java:1505)
    15:02:45 at redis.clients.jedis.HostAndPort.getLocalHostQuietly(HostAndPort.java:105)

### Steps to reproduce:
Use jedis client in aws lambda to connect to aws redis service endpoint
    public class RedisClientDao {

    private static JedisPool pool = null;

    static {
        String endpoint = Preconditions.checkNotNull((String)Config.get("CACHE_ENDPOINT"));
        JedisPoolConfig poolConfig = new JedisPoolConfig();
        poolConfig.setJmxEnabled(false);
        pool = new JedisPool(poolConfig, endpoint);

    }

    public static Jedis getRedisClient() {
        return pool.getResource();
    }
    }

jedis 2.9.0 running on java 1.8 running on aws lambda connecting to redis 3.2.4

#### Jedis version:
2.9
#### Redis version:
3.2.4
#### Java version:
1.8

jkrahul:
fixes #1424

### DIFF
--- a/src/main/java/redis/clients/jedis/HostAndPort.java
+++ b/src/main/java/redis/clients/jedis/HostAndPort.java
@@ -10,7 +10,7 @@ public class HostAndPort implements Serializable {
   private static final long serialVersionUID = -519876229978427751L;
 
   protected static Logger log = LoggerFactory.getLogger(HostAndPort.class.getName());
-  public static final String LOCALHOST_STR = getLocalHostQuietly();
+  public static String localhost;
 
 
   private String host;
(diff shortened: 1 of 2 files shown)


jkrahul:
There is some issue with redis dependencies in Travis. Tests passed in my local with `redis unstable branch`

marcosnils:
LGTM @HeartSaVioR ?

marcosnils:
@jkrahul can you rebase with master so tests pass please?

src/main/java/redis/clients/jedis/HostAndPort.java
```diff
@@ -90,15 +90,50 @@ public static HostAndPort parseString(String from){
...
+        return host;
+    } catch (Exception e) {
+      // Not a valid IP address
+      log.logp(Level.WARNING, HostAndPort.class.getName(), "convertHost", "'" + host + "' is not a valid IP address",
+              e);
       return host;
     }
   }
 
+  public static void setLocalhost(String localhost) {
```
HeartSaVioR: Are there any actual usages for `setLocalhost()`? `setLocalhost()` and `getLocalhost()` can make race condition. 

Btw, `getLocalhost()` itself can make race condition but it might be OK given that `getLocalHostQuietly()` always returns same value.

HeartSaVioR:
@jkrahul I've done reviewing and left comment.

src/main/java/redis/clients/jedis/HostAndPort.java
allanwax: @marcosnils @jkrahul The only use I can think of is a potential one.  If the programmer wants to override the natural local host value, say with the `ENDPOINT` value, then they can make this call change it.  This would be done before actually creating a Jedis instance.

HeartSaVioR: @jkrahul @allanwax 
If we want to keep this as it is, let's leave a javadoc comment for this method describing that this method is not thread-safe so it shouldn't be used with getLocalhost() concurrently.

HeartSaVioR:
@jkrahul Please rebase and address review comment. Thanks in advance!

src/main/java/redis/clients/jedis/HostAndPort.java
allanwax: Wrapping the operations inside `setLocalhost` and `getLocalhost` will make it safe and is less heavy weight than `synchronized`.

jkrahul:
@marcosnils @HeartSaVioR @allanwax review comments have been addressed. Please check

marcosnils:
I believe we're good now. @allanwax @HeartSaVioR ?

src/main/java/redis/clients/jedis/HostAndPort.java
allanwax: @marcosnills @jkrahul I was imagining the wrapping being a ReentrantLock.  I'm good .

HeartSaVioR:
@marcosnils Yes LGTM.

HeartSaVioR:
Merged to master and 2.10 respectively. @jkrahul Thanks for contribution!

PR #1476, opened 2017-03-04
Better data validation for Transaction server-side errors ( and more )

### LINKED ISSUE #1473: Writing to read-only slave fails silently inside of MULTI block
### Expected behavior

When writing to a read-only slave node outside of a transaction throws a `JedisDataException`.  I would expect the same behavior inside of a transaction, as seen in the command-line redis client:

```
127.0.0.1:7379> multi
OK
127.0.0.1:7379> set test test
(error) READONLY You can't write against a read only slave.
127.0.0.1:7379> exec
(empty list or set)
```

### Actual behavior

Writing to a read-only slave as part of a transaction fails silently.

### Steps to reproduce:

This fails to write anything without producing any errors:

```
Jedis j = new Jedis("read-only-slave");

Transaction t = j.multi();
t.set("test", "test");
List<Object> results = t.exec();
```

sheinbergon:
Fixes #1473 
This fix should ensure 'null' response don't pass as valid

### DIFF
--- a/src/main/java/redis/clients/jedis/Queable.java
+++ b/src/main/java/redis/clients/jedis/Queable.java
@@ -10,9 +10,14 @@ public class Queable {
     pipelinedResponses.clear();
   }
 
+  /* 
+    Certain server-side failure scenarios create "null" response-data
+    Scenarios we should fend against.
+    TODO - improve "data" Object validation
+  */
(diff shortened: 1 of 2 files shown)


sheinbergon:
@marcosnils so some "pipelining" events are expected to contain

PR #1484, opened 2017-03-15
Psync support

Aegeaner:
Support redis partial sync since version 2.8 .


sazzad16:
@Aegeaner Appreciate your effort. But, PSYNC is for master-slave intercommunication. I don't think we need this in Jedis.

@HeartSaVioR @marcosnils Closing.

PR #1496, opened 2017-04-01
Prevent ShardedJedis from closing healthy connections

thirstycrow:
Problem

The current behavior of ShardedJedisPool is destroying the ShardedJedis instance once it is detected invalid. It impacts the performance when a shard is broken, because connections to the healthy shards will also be closed, and the client will have to communicate with the healthy shards with short lived connections.

Solution

Instead of abandon the ShardedJedis instance, we can just replace the broken connections.

### DIFF
--- a/src/main/java/redis/clients/jedis/Jedis.java
+++ b/src/main/java/redis/clients/jedis/Jedis.java
@@ -35,6 +35,17 @@ import redis.clients.util.Slowlog;
 public class Jedis extends BinaryJedis implements JedisCommands, MultiKeyCommands,
     AdvancedJedisCommands, ScriptingCommands, BasicCommands, ClusterCommands, SentinelCommands, ModuleCommands {
 
+  public static void destroy(Jedis jedis) {
+    try {
+      try {
+        jedis.quit();
+      } catch(Exception e) {
+      }
+      jedis.disconnect();
+    } catch(Exception e) {
+    }
+  }
+	
   protected JedisPoolAbstract dataSource = null;
 
   public Jedis() {
--- a/src/main/java/redis/clients/jedis/ShardedJedis.java
+++ b/src/main/java/redis/clients/jedis/ShardedJedis.java
@@ -785,21 +785,14 @@ public class ShardedJedis extends BinaryShardedJedis implements JedisCommands, C
   @Override
   public void close() {
     if (dataSource != null) {
-      boolean broken = false;
-
-      for (Jedis jedis : getAllShards()) {
+      for (JedisShardInfo shardInfo: getDistinctShardInfo()) {
+        Jedis jedis = getShard(shardInfo);
         if (jedis.getClient().isBroken()) {
-          broken = true;
-          break;
+          replaceShard(shardInfo);
+          Jedis.destroy(jedis);
         }
       }
-
-      if (broken) {
-        dataSource.returnBrokenResource(this);
-      } else {
-        dataSource.returnResource(this);
-      }
-
+      dataSource.returnResource(this);
     } else {
       disconnect();
     }
--- a/src/main/java/redis/clients/jedis/ShardedJedisPool.java
+++ b/src/main/java/redis/clients/jedis/ShardedJedisPool.java
@@ -77,32 +77,27 @@ public class ShardedJedisPool extends Pool<ShardedJedis> {
     public void destroyObject(PooledObject<ShardedJedis> pooledShardedJedis) throws Exception {
       final ShardedJedis shardedJedis = pooledShardedJedis.getObject();
       for (Jedis jedis : shardedJedis.getAllShards()) {
-        try {
-          try {
-            jedis.quit();
-          } catch (Exception e) {
-
-          }
-          jedis.disconnect();
-        } catch (Exception e) {
-
-        }
+        Jedis.destroy(jedis);
       }
     }
 
     @Override
     public boolean validateObject(PooledObject<ShardedJedis> pooledShardedJedis) {
-      try {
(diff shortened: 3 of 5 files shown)


marcosnils:
@thirstycrow thx for the contribution. Any chance you can add a test for this?

thirstycrow:
@marcosnils Finally, I added some test cases. It took a while to find a way to get broken jedis clients.

src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -35,6 +35,17 @@
 public class Jedis extends BinaryJedis implements JedisCommands, MultiKeyCommands,
     AdvancedJedisCommands, ScriptingCommands, BasicCommands, ClusterCommands, SentinelCommands, ModuleCommands {
 
+  public static void destroy(Jedis jedis) {
```
HeartSaVioR: I think it can be a method for object, not for class.

src/main/java/redis/clients/jedis/ShardedJedisPool.java
```diff
@@ -77,32 +77,27 @@ public ShardedJedisFactory(List<JedisShardInfo> shards, Hashing algo, Pattern ke
...
+      for (JedisShardInfo shardInfo : jedis.getDistinctShardInfo()) {
+        Jedis shard = jedis.getShard(shardInfo);
+        boolean valid = true;
+        try {
+          valid = shard.ping().equals("PONG");
+        } catch(Exception ex) {
+          valid = false;
+        }
+        if (!valid) {
+          jedis.replaceShard(shardInfo);
```
HeartSaVioR: Newly created Jedis instance may fail to ping, or it even throws any Exception while creating Jedis which `validateObject()` doesn't want to.
I think we should take care of that (drop the whole instance and return false), since `validateObject()` should ensure the instance is good.

src/test/java/redis/clients/jedis/tests/ShardedJedisPoolTest.java
```diff
@@ -103,6 +105,53 @@ public void checkPoolRepairedWhenJedisIsBroken() {
...
+    JedisShardInfo shardInfo0 = shards.get(0);
+    JedisShardInfo shardInfo1 = shards.get(1);
+    ShardedJedis jedis1 = pool.getResource();
+    Jedis shard10 = jedis1.getShard(shardInfo0);
+    Jedis shard11 = jedis1.getShard(shardInfo1);
+    breakClient(shard10);
+    assertTrue(shard10.getClient().isBroken());
+    jedis1.close();
+
+    ShardedJedis jedis2 = pool.getResource();
```
HeartSaVioR: I'm not sure this is guaranteed by Apache Commons Pool. Maybe need to change the pool config to have size 1 to ensure this.

src/test/java/redis/clients/jedis/tests/ShardedJedisPoolTest.java
```diff
@@ -103,6 +105,53 @@ public void checkPoolRepairedWhenJedisIsBroken() {
...
+    ShardedJedisPool pool = new ShardedJedisPool(config, withBrokenShards);
+    JedisShardInfo shardInfo0 = withBrokenShards.get(0);
+    JedisShardInfo shardInfo1 = withBrokenShards.get(1);
+    ShardedJedis jedis1 = pool.getResource();
+    Jedis shard10 = jedis1.getShard(shardInfo0);
+    Jedis shard11 = jedis1.getShard(shardInfo1);
+    jedis1.close();
+
+    ShardedJedis jedis2 = pool.getResource();
+    assertEquals(jedis1, jedis2);
```
HeartSaVioR: Same here: I'm not sure this is guaranteed by Apache Commons Pool. Maybe need to change the pool config to have size 1 to ensure this.

src/main/java/redis/clients/jedis/ShardedJedis.java
```diff
@@ -785,21 +785,14 @@ public Long bitpos(String key, boolean value, BitPosParams params) {
...
-      boolean broken = false;
-
-      for (Jedis jedis : getAllShards()) {
+      for (JedisShardInfo shardInfo: getDistinctShardInfo()) {
+        Jedis jedis = getShard(shardInfo);
         if (jedis.getClient().isBroken()) {
-          broken = true;
-          break;
+          replaceShard(shardInfo);
+          Jedis.destroy(jedis);
```
marcosnils: Do we really need to force `destroy` if the connection is broken?. I think the `quit` call will always fail in the `destroy` function if the connection is broken.

src/main/java/redis/clients/jedis/Jedis.java
marcosnils: I don't really see the need of this function. If the connection is broken, it's almost sure that `quit` will always fail.

src/test/java/redis/clients/jedis/tests/ShardedJedisPoolTest.java
marcosnils: ^ same

src/main/java/redis/clients/jedis/Jedis.java
thirstycrow: What if `quit` timeout? Will jedis close the connection on timeout?

sazzad16:
@HeartSaVioR @marcosnils Close due to inactivity?

marcosnils:
@thirstycrow u still want to push this forward?

thirstycrow:
Sorry for late reply. I close this one because it only solved part of the problem we were facing. In the end, we had to wrote a new implementation to improve performance and stability for interacting with a client sharded redis cluster.

Issue #1512, opened 2017-05-17
Performance decrease a lot during redis cluster fail over period 


Neway6655:
### Expected behavior
I have a Redis Cluster with 5 master nodes and 5 slave nodes, each master has one slave. 
I have running a stress test by calling jedis's hmget method, then I killed one master, so redis cluster would fail over, and my expectation is the performance might be decreased a little during fail over period, since some keys might hit the failing master node. And it will recover after redis cluster finish master/slave switch.

### Actual behavior
But actually, performance was decreased a lot during the fail over period, about 75 percentage, the performance was down to 1/4 as before. And after redis cluster recovered, the performance recovered as well.

### Steps to reproduce:
1. start redis cluster.
2. start a stress test.
3. kill one redis master node.

#### Jedis version:
2.8.2
#### Redis version:
3.0.5
#### Java version:
JDK 1.7u21

Neway6655:
I found jedis will try to renewSlotCache when JedisConnectionException happened, while redis cluster fail over may need some time (several seconds or minutes), code snippet as below:

```
   catch (JedisConnectionException jce) {
      // release current connection before recursion
      releaseConnection(connection);
      connection = null;

      if (attempts <= 1) {
        //We need this because if node is not reachable anymore - we need to finally initiate slots renewing,
        //or we can stuck with cluster state without one node in opposite case.
        //But now if maxAttempts = 1 or 2 we will do it too often. For each time-outed request.
        //TODO make tracking of successful/unsuccessful operations for node - do renewing only
        //if there were no successful responses from this node last few seconds
        this.connectionHandler.renewSlotCache();

        //no more redirections left, throw original exception, not JedisClusterMaxRedirectionsException, because it's not MOVED situation
        throw jce;
      }

      return runWithRetries(key, attempts - 1, tryRandomNode, asking);
    }
```

so during the fail over, requests hit at the failing node will fail into JedisConnectionException, then renewSlotCache will be called continuously, this may decrease the performance a lot since renewSlotCache will cause write lock called, which will lock the read lock in getConnectionFromSlot method.

marcosnils:
@Neway6655 good catch. Is there something that you propose to overcome this situation?. We haven't been actively developing Jedis for while, so we really depend on our users to help us with their problems and to propose a possible solution. 

Also @HeartSaVioR or @Spikhalskiy  might have a word regarding this.

kenny319:
[...]

PR #1516, opened 2017-05-31
Performance decrease a lot during redis cluster fail over period #1512

kenny319:
The fix aims to reduce the impaction to the normal nodes when the failover is in progress. The core principle is to narrow the use of the write lock in JedisClusterInfoCache.

1. Make sure only one thread can retrieve the new slots by using another lock
2. Compare the new slots with old slots and replace the old slots with new slots when they are different only avoiding unnecessary occupation of the write lock.


kenny319:
The tests on the 2.8 seem like not stable very much

marcosnils:
@kenny319 yes, sometimes tests fail because of travis. Is this PR ready to be reviewed?. I've seen that you made several changes. 

On the other hand, how's the performance on fail-over with this applied?

kenny319:
Let me create a new PR and add the testing result.

PR #1521, opened 2017-06-12
Performance decrease a lot during redis cluster fail over period #1512

kenny319:
The fix of issue #1512  aims to reduce the impaction to the normal nodes when the failover is in progress. The core principle is to narrow the use of the write lock in JedisClusterInfoCache.

1. Make sure only one thread can retrieve the new slots by using another lock
2. Compare the new slots with old slots and replace the old slots when they are different only avoiding unnecessary occupation of the write lock.

Actually the improved ratio of performance depends on the ratio of the cost spent on accessing redis to the cost spent on the rest compution of a service. After applying the patch the performance is proved a lot, in our test case we have a web service which accesses a redis cluster with 5 master/slave pairs, the QPS increases 50% after using the patch. 

redis.clients.jedis.JedisClusterInfoCache.setupNodeIfNotExist(HostAndPort) method uses the write lock as well when the slots migration is in progress, but in our test case the performance just degrades 7%, so we do not change it since the migration is in full control by ourselves.

Please help review the PR.

Thanks

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
@@ -5,9 +5,12 @@ import redis.clients.util.SafeEncoder;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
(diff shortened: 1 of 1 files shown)


marcosnils:
@kenny319 thx for taking the time to do all this hard work. I'll try to give it a try as soon as I get some time. @HeartSaVioR @Spikhalskiy would love your opinion if you have some time.

kenny319:
@marcosnils Is there any update on this review? We really hope to use the official release, if the PR could be merged, instead of a customized version made by ourselves.

marcosnils:
@kenny319 sorry, I haven't had any time to dedicate to this. I'll try to check it during the weekend.

HeartSaVioR:
I'll also try to spend time to review it. Sorry I visit to late on this.

HeartSaVioR:
@kenny319 Before that, could we have PR for master instead of 2.8? Porting back the patch will be handled by collaborators.

sazzad16:
@HeartSaVioR @marcosnils Close? PR is for a stale branch and without tests.

marcosnils:
@sazzad16. I wouldn't close this as @kenny319 says this patch brought some performance improvements to his use case which me be others as well

@kenny319, is there a chance you can provide some code as a benchmark to validate your results?. We don't really have a lot of bandwidth to replicate the scenario ourselves.

sazzad16:
@marcosnils I was asking because I've seen some other PRs got closed just because of not being for master branch. Also we need some tests/benchmarks to verify improvement.

marcosnils:
> Also we need some tests/benchmarks to verify improvement.

That's what I requested. Let's leave this open as it might help others until we see if @kenny319 can provide some benchmarks.

sazzad16:
Closing, as it is created against branch `2.8`.

PR #1531, opened 2017-06-18
Add missing method on RedisPipeline and BinaryRedisPipeline.

PhantomThief:


marcosnils:
@PhantomThief thx for the PR. Now that we're doing this, maybe it's a good idea to add "@override" to the implementation functions?

Issue #1527, opened 2017-06-16
redis.clients.jedis.BinaryJedisCluster : Request to add feature to get reference of ConnectionHandler from JedisClusterConnection.


saurabhygk:
Before starting thing, I want to clear that this not an issue but DETAILED request for new feature
### Expected behavior

I am expecting reference of Connection Handler to get native connection Jedis by implementing some method in either redis.clients.jedis.BinaryJedisCluster or redis.clients.jedis.JedisCluster classes.

### Actual behavior

In actual behaviour, We want to implement pipeline in redis 3.2 cluster but spring-data-redis JedisClusterConnection does not supports the pipeline on cluster. So we planned to create our own by getting native jedis connection from JedisPool. So for that, in current version JedisCluster class do not have any feature which will return us the reference of connection handler. so for that we require to overload BinaryJedisCluster class in our application with same package name and provide some method which returns connectionHandler reference object. But this is not good choice in the scenario if Jedis client version changed in future and some changes implemented in BinaryJedisCluster class then we need to again overload that class and provide that method. Also this approach/solution gives the issue of class loading. Sometime ClassLoader loads the Jedis BinaryJedisCluster class instead of our application overloaded class.

### Steps to reproduce:
This is not issue but DETAILED request. To implement pipeline in our application we need connection handler to get native connection from JedisPool. So for that we need to following steps.
1.  We need to overload BinaryJedisCluster class with package name redis.clients.jedis in our application. We need to implemented following method
`
	public JedisSlotBasedConnectionHandler getConnectionHandler() {
		return (JedisSlotBasedConnectionHandler) this.connectionHandler;
        }
`

2. After that in code where we want Jedis native connection, we have call by following way

// Get Jedis Cluster Connection
``
JedisClusterConnection jedisClusterConnection = (JedisClusterConnection) 
stringRedisTemplate.getConnectionFactory().getClusterConnection();
``
// Get Jedis Cluster reference from Native Connection.
``
JedisCluster jedisCluster = jedisClusterConnection.getNativeConnection();
``
// This method call is new method which we have implemented to get Connection Handler reference from overloaded method, because Jedis library not provides it.
``
JedisSlotBasedConnectionHandler jedisClusterConnectionHandler = jedisCluster.getConnectionHandler();
``

 -----------------------------OR -----------------------------
Change the visibility of getConnectionFromSlot(int slot) into JedisCluster and expose method into BinaryJedisCluster class same as getClusterNodes() method

`
[...]

PR #1532, opened 2017-06-21
Pull request to change visibility of getConnectionFromSlot method

saurabhygk:
For issue number: https://github.com/xetorthio/jedis/issues/1527


marcosnils:
LGTM!. @HeartSaVioR @allanwax @xetorthio ?

allanwax:
ok

saurabhygk:
@HeartSaVioR @xetorthio I am sorry for trouble. Please review changes and help me out.

Thank you @marcosnils and @allanwax

allanwax:
@HeartSaVioR @xetorthio @saurabhygk Since I'm not a git person (I'm still using svn very productively) I'm not used to using git tools to do various things.  I did check out the master to see what the current definition of the method is.  While there is potential for bad things to happen since you can now get directly at the connection, the ability to get the connection allows things like pipelining to happen, with the caveats that anything can happen between getting the connection and actually executing the action(s).  So I think the benefits outweigh the risks.  If issue come up in the future you can still do it the hard way if you really want to.

saurabhygk:
@allanwx Thank you for your comment. I am sorry but can you please suggest me what kind of risk will be there with this method. By using this method we are taking the connection and perform the action with that connection only and its user/developer responsibility how to be behave with connection. Also one more thing in single redis connection Jedis supports pipeline then what's the issue with redis cluster. I am curious to know that why Jedis does not support pipeline on cluster.

saurabhygk:
@allanwax sorry for typo in name

allanwax:
@saurabhygk The risks are (a) the connection for the slot becomes invalid between the time you fetch it and use it (failover, network problems, etc.), (b) bad code: somebody gets the connection and does something with it that they shouldn't.  (a) can happen at any time so you need to code defensively. (b) is less likely but if someone writes bad code it can take down the whole program with errors and be hard to find.

But as I said, I think the benefits outweigh the risks.

HeartSaVioR:
@saurabhygk 
I intentionally didn't implement JedisCluster with Pipeline because it is really hard to guarantee safeness. Please refer below discussion from Redis group: 
https://groups.google.com/d/msg/redis-db/4I0ELYnf3bk/Lrctk0ULm6AJ

HeartSaVioR:
Sorry maybe commenting to #1527 would be right. Anyway the code change looks fine.

allanwax:
@saurabhygk @HeartSaVioR I absolutely agree that trying things like pipeline on a cluster is inherently unsafe because of all the previously mentioned reasons.  If, however, you are tolerant of failure and bad data there may be some benefit.

I've written and published to the group several times my class called ReadOnlySlave which uses some of the ideas of bypassing the normal cluster logic and treating the slaves as individual servers.  It works and allows some parallel read operations (but not pipelined) on the cluster. **But** it also is monitoring slot movement and server failovers and tries to recover from them, with the ultimate fallback to do a regular read to the cluster.  The data I'm using does not have to be perfect and in my case will self-heal over time.

In other circumstances there is no tolerance for bad data and pipelining on the cluster will always fail eventually no matter how clever your program is.

So, getting the connection from the slot -- go ahead, but know that you can (and will) shoot yourself in the foot if you don't play by the rules.

HeartSaVioR:
@saurabhygk Merged into master and 2.10 branches respectively.

marcosnils:
Thx heart for taking care

saurabhygk:
@HeartSaVioR Thank you very much for handling Pull Request. I will be continue my self to contribute in Jedis.

PR #1542, opened 2017-07-05
Place the connection of the socket in the BinaryClient constructor

iaiengxia:
After testing, when Socket to establish a connection, Socket state will no longer change, unless it calls the appropriate method, if the network causes the Socket connection is disconnected or the server is automatically disconnected, Socket state will not change, So the following code can not detect whether the connection is available
`socket != null && socket.isBound() && !socket.isClosed() && socket.isConnected()
        && !socket.isInputShutdown() && !socket.isOutputShutdown()`
That is, every time you send a command to detect the availability of the connection is not necessary, so intend to remove the detection code, and connect the Socket code to the construction method.

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryClient.java
+++ b/src/main/java/redis/clients/jedis/BinaryClient.java
@@ -50,24 +50,29 @@ public class BinaryClient extends Connection {
 
   public BinaryClient() {
     super();
+    connect();
   }
 
   public BinaryClient(final String host) {
     super(host);
+    connect();
   }
 
   public BinaryClient(final String host, final int port) {
     super(host, port);
+    connect();
   }
 
   public BinaryClient(final String host, final int port, final boolean ssl) {
     super(host, port, ssl);
+    connect();
   }
 
   public BinaryClient(final String host, final int port, final boolean ssl,
       final SSLSocketFactory sslSocketFactory, final SSLParameters sslParameters,
       final HostnameVerifier hostnameVerifier) {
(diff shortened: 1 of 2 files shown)


marcosnils:
@iaiengxia IIUC according to your statement if we change the `isConnected`  logic to `socket != null && true;` then it'll be the same right?. As the socket state doesn't change per your observations.

iaiengxia:
Very sorry, now see this reply, you are right. Also in the actual development, I encountered a Socket disconnected situation, I think it should be added with broken cable reconnection function, do you think it is necessary

marcosnils:
@iaiengxia well, I've tried changing the `isConnected` function and then some tests failed. So seems like the logic there is actually doing something. Have you tried this?

iaiengxia:
@marcosnils Can you look at the reason for the failure of the test, is it because the modified isConnected function

iaiengxia:
@marcosnils On the Socket state, I was tested. So I do not think there is a need to verify the availability of the socket every time I send a message (unless the heartbeat is sent, otherwise it is not verified), so I moved the Socket connection to the BinaryClient constructor.

HeartSaVioR:
If my understanding is right, the purpose of `isConnected()` is performing **lazy-connection**. So isConnected() can't determine disconnection or so (unless doing heartbeat, as you said earlier), but I guess it can determine at least socket is initialized. If then the code is implemented properly.

HeartSaVioR:
Maybe we could take another approach like flagging if you observe that `isConnected()` is a bottleneck.

sazzad16:
Breaks several tests. Closing.

PR #1550, opened 2017-07-18
Add SSL support to JedisCluster

joroda:
These changes add SSL support to JedisCluster. Tests have also been added. I've tried to solve issue of not being able to get SSL ports from a Redis server by introducing a configuration setting where users can specify a mapping between the ports reported by Redis and the actual SSL ports that should be used. This approach seemed the most flexible.


marcosnils:
hey @joroda this is really amazing. I'll try to review whenever I have some time.

damiri-ts:
@joroda I've been looking at the code to try to understand it. I will outline my analysis of the code here just to make sure I understand it correctly.

- The client maintains a local map of standard ports to TLS ports for each node in the cluster
- When the client connects to a single node in the cluster, the node returns the hosts and standard ports of all nodes in the cluster to the client
- The client uses the information from the server and it local map of ports to determine the TLS port to use when connecting to the other nodes in the cluster.

It seems that one limitation of this solution, is that the client needs to know all hosts, ports, and TLS ports in the cluster ahead of time. This means that if another node is added to the cluster, the client will not be immediately aware of this.

Did I understand this correctly?

joroda:
@damiri-ts Your understanding is correct. I'm not sure that there is a way to avoid the client needing to know about all ports/TLS ports ahead of time. However, instead of using a map for these ports, it may be more flexible for the user to provide a Function<int, int> that does the mapping between normal and TLS ports.

In the case that the relationship between port and TLS port was easily predictable, the user could provide something like (int p) -> p + 1000 to adapt to nodes being added without having to change code.

Please let me know what you think about this.

damiri-ts:
@joroda Providing a port mapping function is a clever idea. I believe that this would be the best solution, short of modifications to the server, because it would allow clients to have immediate knowledge of additional nodes on the cluster, provided some convention for calculating TLS ports from non-TLS ports.

Thank you for your contribution.

joroda:
@damiri-ts Sounds great. I've just pushed a change to implement this. In order to do this in a way supported by Java 7, I've added the JedisClusterPortMap inteface instead of using Java 8's lambda function feature. Here is the interface:

```Java
public interface JedisClusterPortMap {
  int GetSSLPort(int port);
}
```

Users will have to provide an implementation of the interface which could look something like this:

```Java
JedisClusterPortMap portMap = new JedisClusterPortMap() {
    public int GetSSLPort(int port) {
      return port + 1000;
    }
};
```

Please let me know what you think of this.

damiri-ts:
@joroda It looks good to me. I think it's the most flexible and practical solution without server-side changes.

joroda:
@damiri-ts @marcosnils Let me know if this needs anything else before it can be merged.

HeartSaVioR:
I'll have a look when I have some time to catch up. Sorry to visit too late.

joroda:
@HeartSaVioR Just wanted to check in to see if you've had the chance to take a look at this.

HeartSaVioR:
@marcosnils @sazzad16 Please review this. Thanks in advance.

mp911de:
Looks pretty decent, it would also make sense to provide an address mapper if the IP address is different than the one reported by `CLUSTER INFO`/`CLUSTER SLOTS` (e.g. running the SSL tunnel on a different machine).

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -33,19 +33,32 @@
...
+
   private static final int MASTER_NODE_INDEX = 2;
 
   public JedisClusterInfoCache(final GenericObjectPoolConfig poolConfig, int timeout) {
-    this(poolConfig, timeout, timeout, null, null);
+    this(poolConfig, timeout, timeout, null, null, false, null, null, null, null);
   }
 
   public JedisClusterInfoCache(final GenericObjectPoolConfig poolConfig,
-      final int connectionTimeout, final int soTimeout, final String password, final String clientName) {
```
sazzad16: This breaks backward compatibility. Keep this constructor and add another one. It is possible to call the new constructor from here.

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -146,53 +159,27 @@ private void discoverClusterSlots(Jedis jedis) {
...
-      JedisPool nodePool = new JedisPool(poolConfig, node.getHost(), node.getPort(),
-          connectionTimeout, soTimeout, password, 0, clientName, false, null, null, null);
-      nodes.put(nodeKey, nodePool);
-      return nodePool;
-    } finally {
-      w.unlock();
-    }
-  }
-
-  public JedisPool setupNodeIfNotExist(HostAndPort node, boolean ssl) {
```
sazzad16: This breaks backward compatibility. But I don't see any usage either. I guess we should have separate PRs to remove/deprecate this.

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -146,53 +159,27 @@ private void discoverClusterSlots(Jedis jedis) {
...
+      Integer mappedPort = portMap.getSSLPort(port);
+      if (mappedPort != null) {
+        port = mappedPort;
+      }
     }
+    return new HostAndPort(host, port);
   }
 
-  public JedisPool setupNodeIfNotExist(HostAndPort node, boolean ssl, SSLSocketFactory sslSocketFactory,
-                                  SSLParameters sslParameters, HostnameVerifier hostnameVerifier) {
```
sazzad16: This breaks backward compatibility. But I don't see any usage either. I guess we should have separate PRs to remove/deprecate this.

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -33,19 +33,32 @@
...
+  private SSLSocketFactory sslSocketFactory;
+  private SSLParameters sslParameters;
+  private HostnameVerifier hostnameVerifier;
+  private JedisClusterPortMap portMap;
+
   private static final int MASTER_NODE_INDEX = 2;
 
   public JedisClusterInfoCache(final GenericObjectPoolConfig poolConfig, int timeout) {
-    this(poolConfig, timeout, timeout, null, null);
+    this(poolConfig, timeout, timeout, null, null, false, null, null, null, null);
```
sazzad16: This won't not be required if following constructor is kept.

joroda:
@sazzad16 I've added the requested changes and attempted to resolve merge conflicts.

src/test/java/redis/clients/jedis/tests/SSLJedisClusterTest.java
```diff
@@ -0,0 +1,262 @@
...
+      jc2 = new JedisCluster(new HostAndPort("127.0.0.1", 8379), DEFAULT_TIMEOUT, DEFAULT_TIMEOUT,
+          DEFAULT_REDIRECTIONS, "cluster", null, DEFAULT_CONFIG, true, 
+          null, null, hostnameVerifier, portMap);
+      Assert.fail("The code did not throw the expected NullPointerException.");
+    } catch (NullPointerException e) {
+      // Null pointer exception occurs from closing Jedis object that did not connect due to custom
+      // hostname validation. When closing this Jedis object, the RedisOutputStream in the underlying 
+      // Connection object is null and causes this exception
+    }
+    if (jc2 != null) {
```
sazzad16: If `jc2 != null`, execution will not reach this point because of previous `Assert.fail()`. Put this block in finally and remove following `Assert.fail()`.

src/test/java/redis/clients/jedis/tests/SSLJedisClusterTest.java
```diff
@@ -0,0 +1,262 @@
...
+          SSLException.class, e.getCause().getClass());
+      Assert.assertEquals("Unexpected second inner exception.",
+          SSLException.class, e.getCause().getCause().getClass());
+      Assert.assertEquals("Unexpected third inner exception",
+          RuntimeException.class, e.getCause().getCause().getCause().getClass());
+      Assert.assertEquals("Unexpected fourth inner exception.",
+          InvalidAlgorithmParameterException.class, e.getCause().getCause().getCause().getCause().getClass());
+    }
+    
+    if (jc != null) {
```
sazzad16: If `jc != null`, execution will not reach this point because of previous `Assert.fail()`. Put this block in finally and remove following `Assert.fail()`.

joroda:
@sazzad16 Thanks for reviewing. I've made the requested changes. I haven't looked into the test that is failing but it is also failing intermittently when testing locally as well.

src/test/java/redis/clients/jedis/tests/SSLJedisClusterTest.java
```diff
@@ -0,0 +1,262 @@
...
+          DEFAULT_REDIRECTIONS, "cluster", null, DEFAULT_CONFIG, true, 
+          null, sslParameters, null, portMap);
+      Assert.fail("The code did not throw the expected JedisConnectionException.");
+    } catch (JedisConnectionException e) {
+      Assert.assertEquals(SSLException.class, e.getCause().getClass());
+      Assert.assertEquals(SSLHandshakeException.class, e.getCause().getCause().getClass());
+      Assert.assertEquals(CertificateException.class, e.getCause().getCause().getCause().getClass());
+    }
+    
+    if (jc != null) {
```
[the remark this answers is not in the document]
joroda: I decided to go with `connectByIpAddressFailsWithSSLParameters` since in this case, there is not really a _right_ `SSLParameters` that will work.

sazzad16:
@joroda Good work! I tried but couldn't see travis log because of `There was an error while trying to fetch the log.`

LGTM if it passes the tests.

sazzad16:
@HeartSaVioR It's been some time you've LGTMed. In case you want to review again!

@marcosnils ?

joroda:
@sazzad16 Now that I think about it more, it's probably a good idea to incorporate the suggestion made by @mp911de now rather than later. It makes sense that in most cases the SSL tunnel will be run on a different machine than the redis server. It will also be much easier to add this functionality now without breaking changes than it would be later.

src/test/java/redis/clients/jedis/tests/SSLJedisTest.java
```diff
@@ -247,7 +247,7 @@ public void connectWithShardInfoAndEmptyTrustStore() throws Exception {
    * Creates an SSLSocketFactory that trusts all certificates in
    * truststore.jceks.
    */
-  private static SSLSocketFactory createTrustStoreSslSocketFactory() throws Exception {
+  public static SSLSocketFactory createTrustStoreSslSocketFactory() throws Exception {
```
sazzad16: Remove public.

src/test/java/redis/clients/jedis/tests/SSLJedisTest.java
```diff
@@ -271,7 +271,7 @@ private static SSLSocketFactory createTrustStoreSslSocketFactory() throws Except
    * Creates an SSLSocketFactory with a trust manager that does not trust any
    * certificates.
    */
-  private static SSLSocketFactory createTrustNoOneSslSocketFactory() throws Exception {
+  public static SSLSocketFactory createTrustNoOneSslSocketFactory() throws Exception {
```
sazzad16: Remove public.

src/test/java/redis/clients/jedis/tests/SSLJedisTest.java
```diff
@@ -297,7 +297,7 @@ public void checkServerTrusted(X509Certificate[] chain, String authType) {
    * for production.
    * 
    */
-  private static class BasicHostnameVerifier implements HostnameVerifier {
+  public static class BasicHostnameVerifier implements HostnameVerifier {
```
sazzad16: Remove public.

sazzad16:
@joroda Great! Thanks!

xusun-CMU:
Hi, 

Is there any timeline for merging this change? 

Thanks,
Edward

rameshmaddi:
Please merge these changes,  I am waiting for these changes to be merged.

sgopan:
Any idea when these changes will be will merged. We badly need these changes.

akhiljain1986:
Is there any ETA to merge these changes ?

johnynek:
Maintainers: is there any update? We are considering running our own build of jedis to get this feature. Of course, a release would be easier for us.

Is there anything blocking? It has a "ready to merge" tag, so maybe everything is ready to go?

sazzad16:
@johnynek There isn't any approximate date of release. So please use your own build for the time being. Thanks!

marcosnils:
Downmerged

sazzad16:
@marcosnils Thank you for merging these PRs :+1: . I was planning to do this weekend :)

marcosnils:
Np. Trying to understand why tests do not pass in travis but they do in my local machine :confused:

sazzad16:
Hmm, that's odd. Because it [did pass](https://travis-ci.org/xetorthio/jedis/builds/332864632) during merge request.

sazzad16:
@marcosnils Please revert this from 2.9.

marcosnils:
Done.

marcosnils:
I'll revert this PR because of two reasons:

1. Tests don't pas
2. There are some things in the design which I don't like and I'd like to change.

@joroda can you please open the PR again?

joroda:
The tests were passing after my latest change to this PR, but that was a while ago so I wonder if other changes to the master branch since then have broken something. I'm not sure I'll have time to make additional changes, but please build off of my contribution if it is helpful. Otherwise, I will probably have time to work on this in a few months.

marcosnils:
@joroda Apologize for the inconveniences and we really appreciate your effort. Would you mind re-opening the PR as-is so we can have a fresh start?. Thx!.

joquijada:
Is this solution already part of a Maven release of the jedis artifact? I see [here](https://search.maven.org/search?q=g:redis.clients) it still lists 2.9.0, which is what I'm currently using and also phasing the SSL issue while connecting to AWS Redis Cluster.

marcosnils:
Nope 2.10.0 hasn't been released yet.

analytically:
Doesn't seem part of 2.10.0 or 3.0.0.

PR #1553, opened 2017-07-24
add redis4.0 moduleExecute

jingbanghu:
if there is something wrong,please tell me

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -53,11 +53,9 @@
 
 	<dependencies>
 		<dependency>
-			<groupId>junit</groupId>
-			<artifactId>junit</artifactId>
-			<version>4.11</version>
-			<type>jar</type>
-			<scope>test</scope>
+			<groupId>org.slf4j</groupId>
+			<artifactId>slf4j-api</artifactId>
+			<version>1.7.22</version>
 		</dependency>
 		<dependency>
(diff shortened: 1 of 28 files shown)


pom.xml
```diff
@@ -168,29 +168,29 @@
...
-				    <execution>
-				      <id>bundle-manifest</id>
-				      <phase>process-classes</phase>
-				      <goals>    
-				        <goal>manifest</goal>
-				      </goals>   
-				    </execution>
-				</executions>		
-			</plugin>
+<!-- 			<plugin> -->
```
marcosnils: Hmm is there a reason why this is commented?

src/main/java/redis/clients/jedis/BinaryClient.java
```diff
@@ -1,1303 +1,1312 @@
-package redis.clients.jedis;
```
marcosnils: This whole class changed?

src/main/java/redis/clients/jedis/Client.java
```diff
@@ -1,1220 +1,1230 @@
-package redis.clients.jedis;
```
marcosnils: This whole class changed?

jingbanghu:
i will change it later

PR #1556, opened 2017-08-07
Check jedis isConnected when destory it in ShardedJedisPool

wantedlzx:
In some cases, a jedis object may not have connected, but need to be destroyed in Sharded env.

For example, three redis node(r1, r2, r3), r1 is out of network, when we make a ShardedJedis object, and use it to find a key that located in r1, we will get a connectionTimeout Exception, and the jedis of r1 will be marked as broken, but other two jedis object(r2, r3) has not call connect() method. In this case, r1, r2, r3 will first set up TCP connection, then send quit, and close TCP connection next, but we dont need do this!

So I add a check like JedisFactory.

### DIFF
--- a/src/main/java/redis/clients/jedis/ShardedJedisPool.java
+++ b/src/main/java/redis/clients/jedis/ShardedJedisPool.java
@@ -77,15 +77,17 @@ public class ShardedJedisPool extends Pool<ShardedJedis> {
     public void destroyObject(PooledObject<ShardedJedis> pooledShardedJedis) throws Exception {
       final ShardedJedis shardedJedis = pooledShardedJedis.getObject();
       for (Jedis jedis : shardedJedis.getAllShards()) {
-        try {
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
LGTM. @marcosnils WDYT?

marcosnils:
LGTM!>

marcosnils:
Merged, to master and up to 2.9!. Thx @wantedlzx for the contribution!.

jaikit:
@wantedlzx @marcosnils In your use case, let's assume connection to r2 was made. Now when jedis of r1 is marked as broken, is there any specific reason all shards are disconnected instead of just the broken jedis shard ? By sending QUIT to all shards and disconnecting, it leads to churn of connections in Shard env

jaikit:
BTW - thanks for adding fix to check isConnected . It is much better than earlier version !

PR #1562, opened 2017-08-17
Returns the pool according to the slot

dingli112:
Returns the pool according to the slot

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ tags
 .idea
 *.aof
 *.rdb
+*DS_Store
+/.gitignore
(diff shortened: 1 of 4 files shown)


marcosnils:
Hey, thx for the change. Can you please revert the formatting changes?. It makes the PR difficult to follow.  Thx. 

Additionally, I've seen that you didn't add tests for the method you're adding. We require that so we make sure that your code is actually working.

HeartSaVioR:
This patch reminds me to accept #1532 and now I'm thinking it was my mistake. I don't want to open the chance to mess JedisCluster instance.
Could we expose HostAndPort instead of Jedis/JedisPool so that users can just create the new connection and no more having a chance to mess the internal?

marcosnils:
> Could we expose HostAndPort instead of Jedis/JedisPool so that users can just create the new connection and no more having a chance to mess the internal?

Yeah, I think this would be the best approach. Exposing the connection could lead to multiple bugs because users are not handling things as needed. :+1: on exposing HostAndPort and make users create the connection themselves. @HeartSaVioR shall we close this?

HeartSaVioR:
Yes. Closing.

PR #1566, opened 2017-08-18
active refresh currentHostMaster

w775450:
Here's a problem,A master-slave switch occurred during the time of the client network failure,unable to receive subscription message,then currentHostMaster point to the slave.

Added code for active refresh

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisSentinelPool.java
+++ b/src/main/java/redis/clients/jedis/JedisSentinelPool.java
@@ -274,6 +274,16 @@ public class JedisSentinelPool extends JedisPoolAbstract {
           if (!running.get()) {
             break;
           }
+          
+          /*
+           * Added code for active refresh
+           */
+          List<String> masterAddr = j.sentinelGetMasterAddrByName(masterName);  
+          if (masterAddr == null || masterAddr.size() != 2) {
+            log.warn("Can not get master addr, master name: {}. Sentinel: {}：{}.",masterName,host,port);
+          }else{
+              initPool(toHostAndPort(masterAddr)); 
+          }
 
           j.subscribe(new JedisPubSub() {
             @Override


marcosnils:
@HeartSaVioR would know. :)

HeartSaVioR:
Sorry it needs to take time to reload the context. Active refresh should help on that scenario, but just curious it is sufficient to only placing it before subscription.

HeartSaVioR:
@w775450 Could you please refine the PR's title and also commit message? Thanks in advance.

w775450:
It's not a question of the reload time.When the network is disconnected for a long time，the socket communication  will be broken and JedisConnectionException will be thrown, in this senario  it can not received the subscription message . so it needs a manually active refresh

HeartSaVioR:
> Sorry it needs to take time to reload the context.

@w775450 Sorry above statement was my joke. :) I have been out of Jedis for months and just get back so I need some time to warm up myself to play as Jedis collaborator.

w775450:
@HeartSaVioR :)

HeartSaVioR:
LGTM. @marcosnils If you are not familiar with this, I'll just treat it as enough reviewed. Please let me know if it's the case. If not please review.

marcosnils:
merged to master and corresponding branches.

gkorland:
See #1910 it seems like adding initPool() call on the MasterListener run() can cause concurrency issues with the Main thread running constructor & then getResource()

gkorland:
The simple solution I see here is to add some critical section on initPool() and make sure no two threads are executing it in parallel.
@marcosnils @HeartSaVioR ?

vattezhang:
@w775450 , I'm a little confused, the code you add is just run at the first time of masterlistner. If the network failed after the masterListner, how can these code reinit the pool?

CrazyCafeBaby:
> @w775450 , I'm a little confused, the code you add is just run at the first time of masterlistner. If the network failed after the masterListner, how can these code reinit the pool?

because the masterListener while run until be shuntdown

`      running.set(true);

      while (running.get()) {

        j = new Jedis(host, port);

        try {
          // double check that it is not being shutdown
          if (!running.get()) {
            break;
          }
`

Issue #1263, opened 2016-04-20
Resolve ambiguity between single argument and varargs by replacing varargs to Collection<>


HeartSaVioR:
Relevant issue: #566 

Due to ambiguity issue Scala users should use workaround like...
```
val stringArray = Array[String]("mystring")
j.del(stringArray:_*)
```

In order to avoid this issue, as I stated from #566, we should define two methods as...

del(String key);
del(String firstKey, String secondKey, String... remainKeys);

which I really don't like it.

Spring Data Redis uses Collection<> to avoid ambiguity. Some of the method definitions are...

https://github.com/spring-projects/spring-data-redis/blob/master/src/main/java/org/springframework/data/redis/core/ScanOptions.java
https://github.com/spring-projects/spring-data-redis/blob/master/src/main/java/org/springframework/data/redis/core/ZSetOperations.java

We should use Collection<> at least ambiguity methods, but it hurts consistency.
So I suggest we replace all methods using varargs to Collection<> to make consistency.

It changes Jedis APIs completely so this should be applied to only Jedis 3.0.0.

If we really want to apply workaround like `k1, k2, keys` I'm not sure that it breaks backward compatibility. It anyway could break `binary` backward compatibility.

mp911de:
This question kept me busy for a while. I identified about 50 methods which accept varargs and are candidates for some sort of Collection API interface (e.g. `Iterable<T>`). Varargs is syntax sugar as the compiler creates the bytecode for an array. Using the Collection API requires the user to create always a `Set` or a `List` if the method accepts only a Collection interface. Working in parallel, providing varargs and Collection API interfaces together isn't a good idea. This pattern leads to more confusion and a higher risk of bugs as in certain cases the compiler just accepts arguments but does not call the method you'd expect at first sight.

```java
public class MyTest {

    void del(Object... objects) {
        System.out.println("array");
    }

    void del(List<Object> objects) {
        System.out.println("iterable");
    }

    public static void main(String[] args) {

        MyTest mytest = new MyTest();
        mytest.del(Arrays.asList("a"));
    }
}
```
Output: `array`

```java
public class MyTest {

    void del(String... objects) {
        System.out.println("array");
    }

    void del(List<String> objects) {
        System.out.println("iterable");
    }

    public static void main(String[] args) {

        MyTest mytest = new MyTest();
        mytest.del(Arrays.asList("a"));
    }
}
```

Output: `iterable`


[...]

PR #1574, opened 2017-08-24
Removed exists(key) method from Commands interface

sazzad16:
Removed exists(key) methods. These can be served by existing exists(array_of_keys) methods.

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryClient.java
+++ b/src/main/java/redis/clients/jedis/BinaryClient.java
@@ -137,10 +137,6 @@ public class BinaryClient extends Connection {
     sendCommand(EXISTS, keys);
   }
 
-  public void exists(final byte[] key) {
-    sendCommand(EXISTS, key);
-  }
-
   public void del(final byte[]... keys) {
     sendCommand(DEL, keys);
   }
--- a/src/main/java/redis/clients/jedis/Client.java
+++ b/src/main/java/redis/clients/jedis/Client.java
@@ -60,16 +60,12 @@ public class Client extends BinaryClient implements Commands {
   }
 
   @Override
-  public void exists(final String key) {
-    exists(SafeEncoder.encode(key));
-  }
-
-  @Override
   public void exists(final String... keys) {
     final byte[][] bkeys = SafeEncoder.encodeMany(keys);
     exists(bkeys);
   }
 
+  @Override
   public void del(final String... keys) {
     final byte[][] bkeys = new byte[keys.length][];
     for (int i = 0; i < keys.length; i++) {
--- a/src/main/java/redis/clients/jedis/commands/Commands.java
+++ b/src/main/java/redis/clients/jedis/commands/Commands.java
@@ -19,8 +19,6 @@ public interface Commands {
 
   void get(final String key);
 
-  void exists(final String key);
-
   void exists(final String... keys);
 
   void del(final String... keys);


marcosnils:
LGTM!. As this breaks backwards compatibility, it needs to go in 3.0. Additionally we need to open a PR against 2.10 and deprecate the method there.

marcosnils:
@HeartSaVioR ?

sazzad16:
@marcosnils How about removing `del`(key) methods? These can also be served by existing del(array_of_keys) methods.

sazzad16:
@marcosnils 

If we are going to break backward compatibility anyway, can we remove `exists(key)` methods from `JedisCommands` interface family? The issue is that `exists(key)` methods return Boolean while `exists(keys...)` methods return Long. But an argument can be made that Redis itself returns Long. Also a similar command `del` returns Long in both cases.

What do you think?

marcosnils:
@sazzad16 agree, I vote to also remove the `boolean exist(key)` method from all interfaces so we're consistent with the `Redis` interface and all the other Jedis methods.

@HeartSaVioR ?

sazzad16:
@marcosnils 

Similar to `EXISTS`, there are other commands like `DEL`, `PFCOUNT` (also upcoming `TOUCH`, `UNLINK`) where method(key) and method(array_of_keys) both are available. In these cases, return types are same. Should we keep only method(array_of_keys) option for these commands? WDYT?

Furthermore, for `SCRIPT EXISTS` command there are `Boolean scriptExists(sha1)` and `List<Boolean> scriptExists(array_sha1)`. What about these? Redis does _Array reply_ for these cases.

HeartSaVioR:
We still need to have two different overloading methods to support ShardJedis and also JedisCluster (single key / multiple keys). While I'm all for dropping support of ShardJedis (I've been proposing that) maybe we could reconsider unifying Jedis and JedisCluster interfaces (I made it separated, but I guess it would be easier to use if they're unified.)

sazzad16:
@HeartSaVioR 

:+1: 

> unifying Jedis and JedisCluster interfaces

Probably `move` is the only method which shouldn't be in JedisCluster but should be in Jedis. How it can be approached?

HeartSaVioR:
Maybe need to discuss, but my thought is that just throwing UnsupportedOperationException in these cases. (Jedis can support but JedisCluster can't support)

At first I don't want to do it so made them separate (individual interfaces), but after the change I couldn't make Jedis/JedisCluster transparent because they're using individual interfaces.

So even though throwing UnsupportedOperationException feels odd, it looks like worth to do that, and we could leave warning message in javadoc.

HeartSaVioR:
So in overall, I don't think we could remove method with singular parameter, and then the problem (ambiguous method signatures) still exists. (#1263)

sazzad16:
@HeartSaVioR Should I close this PR or keep it open for discussions?

sazzad16:
It seems we can't remove `exists(key)` from JedisCommands interface family right now. But, IMHO, we can at least remove this option from `Commands` interface. Firstly, this doesn't stop any support. Secondly, `exists` will be similar to `del` at in Commands and related interfaces. (There is only one option for `del` in Commands interface.)

HeartSaVioR:
Yes LGTM for such boundary.

HeartSaVioR:
Merged into master.

sazzad16:
@marcosnils Created new PR #1597 against 2.10

PR #1575, opened 2017-08-24
Added 'psetex' command in binary and cluster interfaces

sazzad16:
Added 'psetex' command in BinaryJedisCommands, BinaryJedisClusterCommands, JedisClusterCommands interfaces.

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryShardedJedis.java
+++ b/src/main/java/redis/clients/jedis/BinaryShardedJedis.java
@@ -138,6 +138,12 @@ public class BinaryShardedJedis extends Sharded<Jedis, JedisShardInfo> implement
   }
 
   @Override
+  public String psetex(byte[] key, long milliseconds, byte[] value) {
+    Jedis j = getShard(key);
+    return j.psetex(key, milliseconds, value);
+  }
+
+  @Override
   public Long decrBy(byte[] key, long integer) {
     Jedis j = getShard(key);
     return j.decrBy(key, integer);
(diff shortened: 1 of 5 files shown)


src/main/java/redis/clients/jedis/BinaryJedisCluster.java
```diff
@@ -270,21 +270,21 @@ public Long execute(Jedis connection) {
   }
 
   @Override
-  public String psetex(final byte[] key, final long milliseconds, final byte[] value) {
```
marcosnils: I don't understand the changes in this file. Seems like you're just renaming things but nothing is actually being added / removed?. What's the purpose of the changes here?

sazzad16: I have placed `psetex` after `setex`. Yes, this change does not have any implementation value.

But in other interfaces and classes (except this one) `setex` places earlier than `psetex`. I thought it to be a good idea to maintain it here too.

I definitely can change this if you oppose this idea.

HeartSaVioR: I'm OK to arrange the sequence of methods, more clearly, IMHO ideally we should.

HeartSaVioR:
LGTM

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
sazzad16: It seems code re-arranging makes merging harder. So undoing those changes.

HeartSaVioR:
I think reverting the change would make @marcosnils happier. I'll just change the status to READY_TO_MERGE.

sazzad16:
Merged into master, 2.10.

Issue #1412, opened 2016-10-20
Jedis support TOUCH Command 


darionyaphet:
### Expected behavior

Jedis support TOUCH Command . 
### Actual behavior

I couldn't find `touch` at Jedis class .

marcosnils:
@darionyaphet probably because the command is not implemented. What do you think about crafting a PR if you need it?

darionyaphet:
@marcosnils I will try to implement this command :)

marcosnils:
@darionyaphet amazing!. Let me know if you need some help or run into any issues!.

cobbzilla:
👍 I too would love to see the `touch` command implemented

marcosnils:
@cobbzilla PR's are welcome!.

cobbzilla:
@marcosnils Could you point me to a similarly-simply command already in the codebase that I could use as an example? Not promising a PR but if you can guide me just a bit I might be able to... :)

marcosnils:
@cobbzilla if you check this project PR's you can see many commands implementations. Just look for the word "command" in the PR list.

sazzad16:
PR #1579

PR #1579, opened 2017-08-26
Implemented TOUCH command

sazzad16:


src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
```diff
@@ -356,6 +356,46 @@ public void ttl() {
   }
 
   @Test
+  public void touch() {
```
marcosnils: What about adding an `OBJECT IDLETIME <key>` call in some of the examples to validate that Jedis is actually running the `touch` command?. This way we can verify that the access time of the specified key(s) was actually changed.

HeartSaVioR:
LGTM. @marcosnils

sazzad16:
#1412

marcosnils:
Merged to master and 2.10 respectively. Thx @sazzad16 for your contributions!.

PR #1582, opened 2017-08-28
Altered returning primitive 'long' to 'Long' in pfcount methods

sazzad16:


HeartSaVioR:
`pfcount` guarantees to return 0 even if the key doesn't exist. Do you see the cases which needs `null`?

sazzad16:
No, I do not see such needs. This is just to keep the continuity with entire codebase.

sazzad16:
Redis rarely sends `null` as an Integer reply. Still all the methods, except pfcount, have non-primitive Long as return type. If there is no such need, these should be changed altogether.

PR #1584, opened 2017-08-29
rename DEFAULT_MAX_REDIRECYIONS to DEFAULT_MAX_ATTEMPTS

wantedlzx:
Find that maxRedirection has been rename to maxAttempts in JedisCluster.

Should we also rename DEFAULT_MAX_REDIRECYIONS  to DEFAULT_MAX_ATTEMPTS.


HeartSaVioR:
LGTM. @marcosnils WDYT?

HeartSaVioR:
@wantedlzx Sorry I didn't aware this patch is against 2.8. We don't treat such kind of PR unless it is only applied to such version. Could you make this patch against master?

wantedlzx:
hi, @HeartSaVioR 

Thanks for review this patch, I have made a patch against to master (#1595)

Issue #1565, opened 2017-08-18
The performance decreases a lot when one of the client sharded nodes is down.


kenny319:
Given 5 sharded redis nodes scenario , once one of them is down and when a request is going to access this node the SharedJedis instance will be closed, even if other 4 node's connections are able to provide service and they are destroyed as well, if other requests want to access the online nodes the connections have to be recreated. 

The expected behavior is the throughput goes to the normal nodes will not be impacted.

But the actual behavior is even the connections to the normal nodes are closed and reopened repeatedly, that will impact the performance a lot.

PR #1587, opened 2017-08-30
The performance decreases a lot when one of the client sharded nodes is down.

wantedlzx:
Hi, this patch is for issue #1565.

I made a test on my laptop: 
Use ten threads to call get(xx) method all the time with Round Robin for three servers

--> Three normal Redis servers
       1. without patch & with patch :  QPS ≈ 24000

--> Two normal servers and one abnormal server
      1. without patch : QPS = 6423
      2. with patch      : QPS = 13380

It seems works well due to that it not need to close all the Jedis object since it maybe isConnected.

But I don't know whether this change I made is reasonable. SO, can you give some suggestion for this patch?


HeartSaVioR:
We don't take PR for specific version unless there's merge conflict. Please craft a PR against master. Thanks!

PR #1596, opened 2017-09-11
Altered usage of runBinary with run in JedisCluster

sazzad16:
`JedisClusterCommand.runBinary` is used in only BinaryJedisCluster and `JedisClusterCommand.run` is used only in JedisCluster. The only exception is altered in this PR.


sazzad16:

PR #1600, opened 2017-09-11
Removed checkIsInMultiOrPipeline

sazzad16:
Calling `checkIsInMultiOrPipeline` should be the responsibility of public methods which would actually be called by users. Thing is, it is done properly in those methods. As a result, calling this again in private methods leads `checkIsInMultiOrPipeline` to being called twice for same operation.


HeartSaVioR:
LGTM. @marcosnils Please review.

sazzad16:

PR #1605, opened 2017-09-13
Add ability to reset password in JedisFactory at runtime

yapei123:
* add a new constructor in JedisPool
 * add a new method in JedisFactory
 * change JedisFactory to public

PR #1606, opened 2017-09-13
Add ability to reset password in JedisFactory

yapei123:
* Change JedisFactory to public
* add a new method to reset password

### DIFF
--- a/src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
+++ b/src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
@@ -17,11 +17,12 @@ import org.junit.Test;
 
 import redis.clients.jedis.HostAndPort;
 import redis.clients.jedis.Jedis;
+import redis.clients.jedis.JedisFactory;
 import redis.clients.jedis.JedisPool;
 import redis.clients.jedis.JedisPoolConfig;
 import redis.clients.jedis.Transaction;
 import redis.clients.jedis.exceptions.InvalidURIException;
-import redis.clients.jedis.exceptions.JedisException;
+import redis.clients.jedis.exceptions.JedisConnectionException;
 import redis.clients.jedis.exceptions.JedisExhaustedPoolException;
 
 public class JedisPoolTest {
@@ -381,4 +382,40 @@ public class JedisPoolTest {
     return clientList.split("\n").length;
   }
 
+  @Test
+  public void testResetInvalidPassword() {
+    JedisFactory factory = new JedisFactory(hnp.getHost(), hnp.getPort(), 2000, 2000, "foobared", 0, "my_shiny_client_name");
+    
+
+    try(JedisPool pool = new JedisPool(new JedisPoolConfig(), factory);
+        Jedis obj1 = pool.getResource();){
+      obj1.set("foo", "bar");
+      assertEquals("bar", obj1.get("foo"));
+      assertEquals(1, pool.getNumActive());
+      
+      factory.setPassword("wrong password");
+      try (Jedis obj2 = pool.getResource()) {
+        fail("Should not get resource from pool");
+      } catch (JedisConnectionException e) {}
+    }
+  }
+  
+  @Test
+  public void testResetValidPassword() {
+    JedisFactory factory = new JedisFactory(hnp.getHost(), hnp.getPort(), 2000, 2000, "bad password", 0, "my_shiny_client_name");
+    JedisPool pool = new JedisPool(new JedisPoolConfig(), factory);
+    Jedis obj = null;
+    try {
+      pool.getResource();
+      fail("Could not get resource from pool");
+    } catch (JedisConnectionException e) {
+      factory.setPassword("foobared");
+      obj = pool.getResource();
+      obj.set("foo", "bar");
+      assertEquals("bar", obj.get("foo"));
+    } finally {
+      obj.close();
+      pool.close(); 
+    }
+  }
 }
--- a/src/test/java/redis/clients/jedis/tests/JedisSentinelPoolTest.java
+++ b/src/test/java/redis/clients/jedis/tests/JedisSentinelPoolTest.java
@@ -2,6 +2,7 @@ package redis.clients.jedis.tests;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
 
 import java.util.HashSet;
 import java.util.Set;
@@ -12,6 +13,8 @@ import org.junit.Test;
 
 import redis.clients.jedis.HostAndPort;
 import redis.clients.jedis.Jedis;
+import redis.clients.jedis.JedisFactory;
+import redis.clients.jedis.JedisPoolConfig;
 import redis.clients.jedis.JedisSentinelPool;
 import redis.clients.jedis.Transaction;
 import redis.clients.jedis.exceptions.JedisConnectionException;
@@ -160,6 +163,42 @@ public class JedisSentinelPoolTest {
     }
 
     assertTrue(pool.isClosed());
+  }
+  
+  @Test
+  public void testResetInvalidPassword() {
+    JedisFactory factory = new JedisFactory(null, 0, 2000, 2000, "foobared", 0, "my_shiny_client_name");
+    
+    try(JedisSentinelPool pool = new JedisSentinelPool(MASTER_NAME, sentinels, new JedisPoolConfig(), factory);
+        Jedis obj1 = pool.getResource(); ) {
+      obj1.set("foo", "bar");
+      assertEquals("bar", obj1.get("foo"));
+      assertEquals(1, pool.getNumActive());
+      
+      factory.setPassword("wrong password");
+      try (Jedis obj2 = pool.getResource();) {
+        fail("Should not get resource from pool");
+      }catch (JedisConnectionException e) {}
+    }
+  }
+ 
+  @Test
+  public void testResetValidPassword() {
+    JedisFactory factory = new JedisFactory(null, 0, 2000, 2000, "wrong password", 0, "my_shiny_client_name");
+    JedisSentinelPool pool = new JedisSentinelPool(MASTER_NAME, sentinels, new JedisPoolConfig(), factory);
+    Jedis obj = null;
+    try {
+      pool.getResource();
+      fail("Could not get resource from pool");
+    } catch (JedisConnectionException e) {
+      factory.setPassword("foobared");
+      obj = pool.getResource();
+      obj.set("foo", "bar");
+      assertEquals("bar", obj.get("foo"));
+    } finally {
+      obj.close();
+      pool.close();
+    }
   }
 
   private void forceFailover(JedisSentinelPool pool) throws InterruptedException {
--- a/src/main/java/redis/clients/jedis/JedisFactory.java
+++ b/src/main/java/redis/clients/jedis/JedisFactory.java
@@ -18,11 +18,11 @@ import redis.clients.util.JedisURIHelper;
 /**
  * PoolableObjectFactory custom impl.
  */
-class JedisFactory implements PooledObjectFactory<Jedis> {
+public class JedisFactory implements PooledObjectFactory<Jedis> {
   private final AtomicReference<HostAndPort> hostAndPort = new AtomicReference<HostAndPort>();
+  private final AtomicReference<String> password = new AtomicReference<String>();
   private final int connectionTimeout;
   private final int soTimeout;
-  private final String password;
   private final int database;
   private final String clientName;
   private final boolean ssl;
@@ -30,7 +30,7 @@ class JedisFactory implements PooledObjectFactory<Jedis> {
   private final SSLParameters sslParameters;
   private final HostnameVerifier hostnameVerifier;
 
-  JedisFactory(final String host, final int port, final int connectionTimeout,
+  public JedisFactory(final String host, final int port, final int connectionTimeout,
       final int soTimeout, final String password, final int database, final String clientName) {
     this(host, port, connectionTimeout, soTimeout, password, database, clientName,
         false, null, null, null);
@@ -43,7 +43,7 @@ class JedisFactory implements PooledObjectFactory<Jedis> {
     this.hostAndPort.set(new HostAndPort(host, port));
     this.connectionTimeout = connectionTimeout;
     this.soTimeout = soTimeout;
-    this.password = password;
+    this.password.set(password);
     this.database = database;
     this.clientName = clientName;
     this.ssl = ssl;
@@ -68,7 +68,7 @@ class JedisFactory implements PooledObjectFactory<Jedis> {
     this.hostAndPort.set(new HostAndPort(uri.getHost(), uri.getPort()));
     this.connectionTimeout = connectionTimeout;
     this.soTimeout = soTimeout;
-    this.password = JedisURIHelper.getPassword(uri);
+    this.password.set(JedisURIHelper.getPassword(uri));
     this.database = JedisURIHelper.getDBIndex(uri);
     this.clientName = clientName;
     this.ssl = JedisURIHelper.isRedisSSLScheme(uri);
@@ -79,6 +79,10 @@ class JedisFactory implements PooledObjectFactory<Jedis> {
 
   public void setHostAndPort(final HostAndPort hostAndPort) {
     this.hostAndPort.set(hostAndPort);
+  }
+
+  public void setPassword(final String password) {
+    this.password.set(password);
   }
 
   @Override
@@ -115,8 +119,8 @@ class JedisFactory implements PooledObjectFactory<Jedis> {
 
     try {
       jedis.connect();
-      if (password != null) {
-        jedis.auth(password);
+      if (password.get() != null) {
+        jedis.auth(password.get());
       }
       if (database != 0) {
         jedis.select(database);
--- a/src/main/java/redis/clients/jedis/JedisPool.java
+++ b/src/main/java/redis/clients/jedis/JedisPool.java
@@ -6,6 +6,7 @@ import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLParameters;
 import javax.net.ssl.SSLSocketFactory;
 
+import org.apache.commons.pool2.PooledObjectFactory;
 import org.apache.commons.pool2.impl.GenericObjectPool;
 import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
 
@@ -227,6 +228,10 @@ public class JedisPool extends JedisPoolAbstract {
       final SSLParameters sslParameters, final HostnameVerifier hostnameVerifier) {
     super(poolConfig, new JedisFactory(uri, connectionTimeout, soTimeout, null, sslSocketFactory,
         sslParameters, hostnameVerifier));
+  }
+  
+  public JedisPool(GenericObjectPoolConfig poolConfig, PooledObjectFactory<Jedis> factory) {
+    super(poolConfig, factory);
   }
 
   @Override
--- a/src/main/java/redis/clients/jedis/JedisSentinelPool.java
+++ b/src/main/java/redis/clients/jedis/JedisSentinelPool.java
@@ -95,6 +95,15 @@ public class JedisSentinelPool extends JedisPoolAbstract {
     initPool(master);
   }
 
+  public JedisSentinelPool(String masterName, Set<String> sentinels, final GenericObjectPoolConfig poolConfig, final JedisFactory factory) {
+    this.poolConfig = poolConfig;
+    this.factory = factory;
+
+    HostAndPort master = initSentinels(sentinels, masterName);
+    initPool(poolConfig, factory);
+    initPool(master);
+  }
+
   @Override
   public void destroy() {
     for (MasterListener m : masterListeners) {


marcosnils:
dupe of #1605 ?

yapei123:
@marcosnils I removed that branch since of some messed up. I will close #1605. I just saw your comments.

yapei123:
@marcosnils  I would love to add it to JedisCluster and add test. Can I have some helps from your or other member on how to run the tests? I've tried to run 'make start', 'make test' but failed.

marcosnils:
@yapei123 sure, please share your erros and I'll try to help you to get the tests passed.

yapei123:
@marcosnils 
$ make start
make: *** [stunnel] Error 1
$ make test
gcc -shared -o /tmp/testmodule.so -fPIC src/test/resources/testmodule.c
make: *** [stunnel] Error 1

Is there anything I need to do before running 'make'? I've tried to make my local redis running, but 'Make' still doesn't work.

yapei123:
@marcosnils  I did minor modification to the Makefile to make it run. (remove the line STUNNEL_BIN, and change the redis-git path.

I can run 'make start/stop', but 'make test' failed with the below info:
---------------------------------------------------------------------------------------------
...
echo "$REDIS_SENTINEL4" > /tmp/sentinel4.conf && redis-server /tmp/sentinel4.conf --sentinel
echo "$REDIS_CLUSTER_NODE1_CONF" | redis-server -
echo "$REDIS_CLUSTER_NODE2_CONF" | redis-server -
echo "$REDIS_CLUSTER_NODE3_CONF" | redis-server -
echo "$REDIS_CLUSTER_NODE4_CONF" | redis-server -
echo "$REDIS_CLUSTER_NODE5_CONF" | redis-server -
sleep 2
mvn -Dtest=!SSL*, clean compile test
/bin/sh: mvn: command not found
make: *** [test] Error 127
--------------------------------------------------------------------------------------------------

Thanks.
Yan

marcosnils:
@yapei123 do you have maven installed?. Seems like the makefile is complaining that maven is not installed. Which linux distribution are you using?

yapei123:
@marcosnils  Thank you. I fixed the mvn problem. 'make test' is successful now.

yapei123:
@marcosnils  I got the below error randomly:

`java.lang.Exception: Unexpected exception, expected<redis.clients.jedis.exceptions.JedisConnectionException> but was<redis.clients.jedis.exceptions.JedisClusterMaxRedirectionsException>
	at org.junit.internal.runners.statements.ExpectException.evaluate(ExpectException.java:28)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)

Caused by: redis.clients.jedis.exceptions.JedisClusterMaxRedirectionsException: Too many Cluster redirections?
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:100)
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:164)
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:164)
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:164)
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:164)
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:164)
	at redis.clients.jedis.JedisClusterCommand.run(JedisClusterCommand.java:33)
	at redis.clients.jedis.JedisCluster.set(JedisCluster.java:128)
	at redis.clients.jedis.tests.JedisClusterTest.testResetInvalidPassword(JedisClusterTest.java:654)`

yapei123:
@marcosnils  I've run JedisClusterTest.java only from my IDE several times, all tests passed all the time.  JedisClusterTest randomly failed when executing 'make test'. What might be the problem?

marcosnils:
@yapei123 maybe some weird race condition or something. If tests pass on you IDE then it's ok. Is this PR ready to be reviewed?. Can you please squash the commits?

yapei123:
@marcosnils  yes, it's ready for reviewing. what do you mean by squashing the commits?

marcosnils:
As you can see this PR has 15 commits. It's simpler to review if you just squash them all together and you just send few of them. Check https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

yapei123:
@marcosnils  I had hard time to squash my commits into one. I might have to clone a new branch to my local and copy the changed codes there. is it ok if I create a new pull request?

yapei123:
@marcosnils it's hard to squash to only one commit. Except that, it's ready to code review now. Thanks for your helps.

yapei123:
@marcosnils just realized the JedisCluster might not be able to add the feature of 'reset password at runtime' since the JedisFactory needs to be created for each node and put into a List and it's hard to get hold of the JedsFactory for each node.

yapei123:
@marcosnils  the JedisTest.java is always passing in my local.

yapei123:
ready for code reviewing. Thanks!

chrisribble:
It would be really cool to get this merged so that an application can update the Redis password at runtime. @marcosnils is there anything holding up merging this?

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
```diff
@@ -380,5 +381,34 @@ public void testCloseConnectionOnMakeObject() {
...
+    obj2.set("foo2", "bar2");
+    pool.close();
+  }
+  
+  @Test
+  public void testResetValidPassword() {
+    JedisFactory factory = new JedisFactory(hnp.getHost(), hnp.getPort(), 2000, 2000, "bad password", 0, "my_shiny_client_name", false, null, null, null);
+    JedisPool pool = new JedisPool(new JedisPoolConfig(), factory);
+    try {
+      pool.getResource();
```
HeartSaVioR: add `fail()` below `pool.getResource()`

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
```diff
@@ -380,5 +381,34 @@ public void testCloseConnectionOnMakeObject() {
...
+  @Test(expected = JedisConnectionException.class)
+  public void testResetInvalidPassword() {
+    JedisFactory factory = new JedisFactory(hnp.getHost(), hnp.getPort(), 2000, 2000, "foobared", 0, "my_shiny_client_name", false, null, null, null);
+    JedisPool pool = new JedisPool(new JedisPoolConfig(), factory);
+    Jedis obj1 = pool.getResource();
+    obj1.set("foo", "bar");
+    assertEquals("bar", obj1.get("foo"));
+    assertEquals(1, pool.getNumActive());
+    
+    factory.setPassword("wrong password");
```
HeartSaVioR: Let's wrap with try-catch and remove expected statement. The test will pass even JedisConnectionException is thrown from obj1.

src/test/java/redis/clients/jedis/tests/JedisSentinelPoolTest.java
```diff
@@ -161,6 +163,36 @@ public void customClientName() {
...
+  @Test(expected = JedisConnectionException.class)
+  public void testResetInvalidPassword() {
+    JedisFactory factory = new JedisFactory(null, 0, 2000, 2000, "foobared", 0, "my_shiny_client_name", false, null, null, null);
+    JedisSentinelPool pool = new JedisSentinelPool(MASTER_NAME, sentinels, new JedisPoolConfig(), factory);
+    Jedis obj1 = pool.getResource();
+    obj1.set("foo", "bar");
+    assertEquals("bar", obj1.get("foo"));
+    assertEquals(1, pool.getNumActive());
+    
+    factory.setPassword("wrong password");
```
HeartSaVioR: Let's wrap with try-catch and remove expected statement. The test will pass even JedisConnectionException is thrown from obj1.

src/test/java/redis/clients/jedis/tests/JedisSentinelPoolTest.java
```diff
@@ -161,6 +163,36 @@ public void customClientName() {
...
+    obj2.set("foo2", "bar2");
+    pool.close();
+  }
+  
+  @Test
+  public void testResetValidPassword() {
+    JedisFactory factory = new JedisFactory(null, 0, 2000, 2000, "wrong password", 0, "my_shiny_client_name", false, null, null, null);
+    JedisSentinelPool pool = new JedisSentinelPool(MASTER_NAME, sentinels, new JedisPoolConfig(), factory);
+    try {
+      pool.getResource();
```
HeartSaVioR: add `fail()` below `pool.getResource()`

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
yapei123: done

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
yapei123: done

src/test/java/redis/clients/jedis/tests/JedisSentinelPoolTest.java
yapei123: done

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
yapei123: done

src/test/java/redis/clients/jedis/tests/JedisSentinelPoolTest.java
yapei123: done

src/test/java/redis/clients/jedis/tests/JedisSentinelPoolTest.java
yapei123: done

yapei123:
@HeartSaVioR @marcosnils I've pushed the codes based on @HeartSaVioR's comments. Please let me know if there is anything else needed.

HeartSaVioR:
@sazzad16 @marcosnils Please review this.

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
```diff
@@ -381,4 +382,39 @@ private int getClientCount(final String clientList) {
     return clientList.split("\n").length;
   }
 
+  @Test
+  public void testResetInvalidPassword() {
+    JedisFactory factory = new JedisFactory(hnp.getHost(), hnp.getPort(), 2000, 2000, "foobared", 0, "my_shiny_client_name", false, null, null, null);
+    JedisPool pool = new JedisPool(new JedisPoolConfig(), factory);
+    Jedis obj1 = pool.getResource();
```
sazzad16: `obj1` and all other internally created `Jedis` objects are never closed.

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
```diff
@@ -381,4 +382,39 @@ private int getClientCount(final String clientList) {
...
+    obj1.set("foo", "bar");
+    assertEquals("bar", obj1.get("foo"));
+    assertEquals(1, pool.getNumActive());
+    
+    factory.setPassword("wrong password");
+    boolean fail = false;
+    try {
+      pool.getResource();
+    } catch (JedisConnectionException e) {
+      pool.close();
```
sazzad16: All `pool.close();` within `catch` block should be in `finally` block.

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
```diff
@@ -381,4 +382,39 @@ private int getClientCount(final String clientList) {
...
+    JedisPool pool = new JedisPool(new JedisPoolConfig(), factory);
+    Jedis obj1 = pool.getResource();
+    obj1.set("foo", "bar");
+    assertEquals("bar", obj1.get("foo"));
+    assertEquals(1, pool.getNumActive());
+    
+    factory.setPassword("wrong password");
+    boolean fail = false;
+    try {
+      pool.getResource();
```
sazzad16: Instead of variable `fail`, method `fail()` can be used here as well as in other cases.

sazzad16:
@yapei123 Left some comments. I didn't repeat same comment. So each comment is applicable in multiple places.

sazzad16:
@HeartSaVioR Do we really need this?

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
yapei123: done

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
yapei123: done

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
yapei123: done

yapei123:
@sazzad16 I've addressed all your comments. Got a conflict and don't have write access to resolve it.

sazzad16:
@yapei123 To resolve conflict, you can **fetch** master branch of this repo and **rebase** your working branch with that one.

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
```diff
@@ -381,4 +382,41 @@ private int getClientCount(final String clientList) {
...
+    assertEquals("bar", obj1.get("foo"));
+    assertEquals(1, pool.getNumActive());
+    
+    factory.setPassword("wrong password");
+    try {
+      pool.getResource();
+      fail("Could not get resource from pool");
+    } finally {
+      obj1.close();
+      pool.destroy();
```
sazzad16: `pool.close()` would be better here.

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
```diff
@@ -381,4 +382,41 @@ private int getClientCount(final String clientList) {
...
 
+  @Test(expected = JedisConnectionException.class)
+  public void testResetInvalidPassword() {
+    JedisFactory factory = new JedisFactory(hnp.getHost(), hnp.getPort(), 2000, 2000, "foobared", 0, "my_shiny_client_name", false, null, null, null);
+    JedisPool pool = new JedisPool(new JedisPoolConfig(), factory);
+    Jedis obj1 = pool.getResource();
+    obj1.set("foo", "bar");
+    assertEquals("bar", obj1.get("foo"));
+    assertEquals(1, pool.getNumActive());
+
```
sazzad16: Move `obj1.close()` here. There's nothing related to `obj1` in try.

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
```diff
@@ -381,4 +382,41 @@ private int getClientCount(final String clientList) {
...
+      obj1.close();
+      pool.destroy();
+    }
+  }
+  
+  @Test
+  public void testResetValidPassword() {
+    JedisFactory factory = new JedisFactory(hnp.getHost(), hnp.getPort(), 2000, 2000, "bad password", 0, "my_shiny_client_name", false, null, null, null);
+    JedisPool pool = new JedisPool(new JedisPoolConfig(), factory);
+    Jedis obj = null;
```
sazzad16: If operation fails in try, `obj` will be null in finally and will get NPE. Declare directly in catch.

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
```diff
@@ -381,4 +382,41 @@ private int getClientCount(final String clientList) {
...
+    try {
+      pool.getResource();
+      fail("Could not get resource from pool");
+    } catch (JedisConnectionException e) {
+      factory.setPassword("foobared");
+      obj = pool.getResource();
```
sazzad16: Declare `obj` here.

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
```diff
@@ -381,4 +382,41 @@ private int getClientCount(final String clientList) {
...
+      obj.set("foo", "bar");
+      assertEquals("bar", obj.get("foo"));
```
sazzad16: Close `obj` after this.

src/test/java/redis/clients/jedis/tests/JedisSentinelPoolTest.java
```diff
@@ -2,6 +2,7 @@
```
sazzad16: Most of my comments in `JedisPoolTest.java` are also applicable in this class. So use those suggestions :)

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
```diff
@@ -381,4 +382,41 @@ private int getClientCount(final String clientList) {
...
   }
 
+  @Test(expected = JedisConnectionException.class)
+  public void testResetInvalidPassword() {
+    JedisFactory factory = new JedisFactory(hnp.getHost(), hnp.getPort(), 2000, 2000, "foobared", 0, "my_shiny_client_name", false, null, null, null);
+    JedisPool pool = new JedisPool(new JedisPoolConfig(), factory);
+    Jedis obj1 = pool.getResource();
+    obj1.set("foo", "bar");
+    assertEquals("bar", obj1.get("foo"));
+    assertEquals(1, pool.getNumActive());
```
sazzad16: @HeartSaVioR @marcosnils Should we use it?

@yapei123 Wait for what others say.

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
```diff
@@ -381,4 +382,41 @@ private int getClientCount(final String clientList) {
     return clientList.split("\n").length;
   }
 
+  @Test(expected = JedisConnectionException.class)
```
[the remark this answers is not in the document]
yapei123: done

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
yapei123: done

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
yapei123: done

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
yapei123: done

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
yapei123: done

src/test/java/redis/clients/jedis/tests/JedisSentinelPoolTest.java
yapei123: done

marcosnils:
@yapei123 PR still has conflicts. Have you rebased?

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
yapei123: @sazzad16 if I move obj1.close() here, after calling factory.setPassword("wrong password"), the pool.getResource() will not make a new object for Jedis. The old Jedis which has the correct password got returned.

yapei123: @sazzad16 I changed to use Jedis.quit and it works well.

marcosnils:
@yapei123 thx for rebasing. once @sazzad16 approves we can merge.

yapei123:
@marcosnils Thanks. There is no conflicts now. But The Travis CI build failed on a method I didn't touch. In my local, it runs good. Please let me know what I should do.

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
```diff
@@ -381,4 +382,41 @@ private int getClientCount(final String clientList) {
...
 
+  @Test
+  public void testResetInvalidPassword() {
+    JedisFactory factory = new JedisFactory(hnp.getHost(), hnp.getPort(), 2000, 2000, "foobared", 0, "my_shiny_client_name", false, null, null, null);
+    JedisPool pool = new JedisPool(new JedisPoolConfig(), factory);
+    Jedis obj1 = pool.getResource();
+    obj1.set("foo", "bar");
+    assertEquals("bar", obj1.get("foo"));
+    assertEquals(1, pool.getNumActive());
+    obj1.quit();
```
sazzad16: Use `close()`

src/test/java/redis/clients/jedis/tests/JedisSentinelPoolTest.java
```diff
@@ -161,6 +164,44 @@ public void customClientName() {
...
+  
+  @Test
+  public void testResetInvalidPassword() {
+    JedisFactory factory = new JedisFactory(null, 0, 2000, 2000, "foobared", 0, "my_shiny_client_name", false, null, null, null);
+    JedisSentinelPool pool = new JedisSentinelPool(MASTER_NAME, sentinels, new JedisPoolConfig(), factory);
+    Jedis obj1 = pool.getResource();
+    obj1.set("foo", "bar");
+    assertEquals("bar", obj1.get("foo"));
+    assertEquals(1, pool.getNumActive());
+    obj1.quit();
```
sazzad16: Use `close()`

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
```diff
@@ -381,4 +382,41 @@ private int getClientCount(final String clientList) {
...
+    
+    factory.setPassword("wrong password");
+    try {
+      pool.getResource();
+     fail("Expecting JedisConnectionException");
+    } catch (JedisConnectionException e) {
+    } finally {
```
sazzad16: Also need a `catch (Exception)` here and other similar places. Sorry, didn't notice earlier.

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
```diff
@@ -381,4 +382,41 @@ private int getClientCount(final String clientList) {
...
+    Jedis obj1 = pool.getResource();
+    obj1.set("foo", "bar");
+    assertEquals("bar", obj1.get("foo"));
+    assertEquals(1, pool.getNumActive());
+    obj1.quit();
+    
+    factory.setPassword("wrong password");
+    try {
+      pool.getResource();
+     fail("Expecting JedisConnectionException");
```
sazzad16: @yapei123 What happened to your old message?

@marcosnils @HeartSaVioR I like message like `Should not get resource from pool` better. WDYT?

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
yapei123: When using close(),  the factory.setPassword() doesn't take any effect in this case since  makeObject() of JedisFactory.java doesn't get invoked so the same Jedis object returned with the old password.

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
yapei123: I can combine them together?

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
yapei123: done

src/test/java/redis/clients/jedis/tests/JedisSentinelPoolTest.java
yapei123: the same reason

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
sazzad16: Then remove `quit()`

sazzad16:
@HeartSaVioR @marcosnils The code looks very good now. I am just confused about one thing. Jedis.close() in between Pool.getResource() doesn't work. As long as my confusion is not resolved, it's you guys' call. I won't mind if you decide to go ahead.

yapei123:
@sazzad16 I moved the obj1.close() to inside finally(), as long as it is called after the second pool.getResource(), it will work.

yapei123:
@sazzad16 Thank you for all your comments.

sazzad16:
@yapei123 That `obj1.close();`[[link 1](https://github.com/xetorthio/jedis/pull/1606/files#diff-987dacebf0c53007c8da06018f771f39R402)][[link 2](https://github.com/xetorthio/jedis/pull/1606/files#diff-987dacebf0c53007c8da06018f771f39R402)] aren't working before try. IMO, users are likely to call close after some operations and get another resource for some other operations.

chrisribble:
@sazzad16: Would you agree that it is okay for existing connections in the pool to not have the password changed since those existing connections should have already been authorized with the previous password and should continue to function until they are closed?

@yapei123 and I were discussing this just now and agree the test needs to use try-with-resources and we think it would be okay to close the JedisPool before changing the password and then to create a new JedisPool with the same JedisFactory and verify that the password is the new password. This should simulate the scenario where connections were broken/closed and the new password is used for new connections.

The alternative would be to force the connections in the pool to be disconnected (or to otherwise remove them from the pool). I can think of a few ways to do that:

1. Maintain a separate Redis server instance in the tests just for this scenario so that we can shut it down (to break the connections), then start it back up.

2. Manually force the JedisPool to flush the connections in the test after we change the password; looking at JedisPool and Pool<T>, I don't really see a way to do that.

yapei123:
Hi @sazzad16, I have changed the test method according to @chrisribble by using try-with-resources in the first time, then create a new Pool after factory changing password, and it's working.

yapei123:
@sazzad16 the build failed due to unrecognized JedisFacotory's constructor which is weird. I have never changed it's constructors. All tests past in my local.

sazzad16:
@chrisribble @yapei123 

> Would you agree that it is okay for existing connections in the pool to not have the password changed since those existing connections should have already been authorized with the previous password and should continue to function until they are closed?

Yes, but getting new resource after resetting password should be possible whether previous resource is closed or not. Do you agree on this?

> ... we think it would be okay to close the JedisPool before changing the password and then to create a new JedisPool with the same JedisFactory ...

Closing previous Pool and creating new one is possible right now, even without this PR. What would be the purpose of this PR then? Isn't it to avoid creating new pool?

chrisribble:
@sazzad16 I agree that it should be possible to get a new resource from the same pool after resetting the password, but I think the issue @yapei123 is running into is that once you successfully acquire a resource it is sitting in the pool and subsequent calls to getResource return that existing resource, so it is difficult to prove that setting the wrong password has caused the behavior to change.

Currently the tests is doing this:
1. Use the right password to construct the pool
2. Get a resource from the pool
3. Change to the wrong password
4. Try to prove that getting a resource fails now since the password is wrong

4 doesn't work because the item is already in the pool; perhaps if the test was in the reverse order it would demonstrate the behavior better?

1. Use the wrong password to construct the pool
2. Try to prove that getting a resource fails
3. Change to the right password
4. Try to prove that getting a resource succeeds

What do you think?

sazzad16:
@chrisribble IMO, both scenarios should be successful. Choosing only one doesn't sound right to me.

Waiting for your response on my query about creating new pool.

yapei123:
@chrisribble @sazzad16  There are two tests: one test reset password from bad to good, another from good to bad. The first one has no issue.

chrisribble:
@sazzad16 Yes, the purpose of this PR is to avoid creating a new pool when you want to change the password that is used for *new* connections (resources).

The purpose of the PR is *not* to force the existing resources to be closed/released/etc when the password is changed. 

Instead of creating a new pool in the test, what if we request a second resource while holding the first resource? That will ensure that the pool has to create a new connection with the bad password.

Sequence of actions:
1. Create pool with good password
2. Request resource
3. Pool uses good password to make connection
4. Pool adds resource to pool and hands out to caller
5. Change password to bad password
6. Without closing first resource (and thus avoiding release back to pool), request another resource
7. Pool uses bad password to attempt to make connection
8. Pool throws exception and caller verifies thrown

How does that sound?

sazzad16:
@chrisribble 
> Without closing first resource (and thus avoiding release back to pool), request another resource

I'm strongly against this idea that already requested resources would _enforced to be unclosed_ for a valid operation in Pool.getResource(). I hope you'd understand my viewpoint.

PS: This is my own view. Other collaborators may agree with you. We can wait for them to respond.

chrisribble:
@sazzad16 I guess I don't understand completely because it's not that the underlying code has to enforce that the requested resource is unclosed; it's just that there's no straightforward way to validate that the password has been changed without keeping the previously-requested resource unclosed because there is no external way for the caller to tell the pool to evict a closed resource.

Without the closed (released) resource being evicted from the pool, we will get it again the next time we call getResource. Do you have any other ideas on a safe way to evict the resource from the pool without resorting to clearing the entire pool?

I think maybe it's a mistake to be writing a test for this behavior using JedisPool and that the test should be directly validating the behavior on JedisFactory. What do you think?

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
```diff
@@ -381,4 +382,40 @@ private int getClientCount(final String clientList) {
...
+    }
+  }
+  
+  @Test
+  public void testResetValidPassword() {
+    JedisFactory factory = new JedisFactory(hnp.getHost(), hnp.getPort(), 2000, 2000, "bad password", 0, "my_shiny_client_name");
+    JedisPool pool = new JedisPool(new JedisPoolConfig(), factory);
+    Jedis obj = null;
+    try {
+      pool.getResource();
```
sazzad16: Get this resource and close it properly, just to avoid any possible connection leak.

src/test/java/redis/clients/jedis/tests/JedisSentinelPoolTest.java
```diff
@@ -161,6 +164,42 @@ public void customClientName() {
...
+    }
+  }
+ 
+  @Test
+  public void testResetValidPassword() {
+    JedisFactory factory = new JedisFactory(null, 0, 2000, 2000, "wrong password", 0, "my_shiny_client_name");
+    JedisSentinelPool pool = new JedisSentinelPool(MASTER_NAME, sentinels, new JedisPoolConfig(), factory);
+    Jedis obj = null;
+    try {
+      pool.getResource();
```
sazzad16: Get this resource and close it properly, just to avoid any possible connection leak. try-with-resources can be used for convenience.

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
```diff
@@ -381,4 +382,40 @@ private int getClientCount(final String clientList) {
...
+      fail("Could not get resource from pool");
+    } catch (JedisConnectionException e) {
+      factory.setPassword("foobared");
+      obj = pool.getResource();
```
sazzad16: Close `obj` within this catch block. Optionally, you can _declare_ `obj` (aka `Jedis obj = pool.getResource();`) here with/without try-with-resources.

src/test/java/redis/clients/jedis/tests/JedisSentinelPoolTest.java
```diff
@@ -161,6 +164,42 @@ public void customClientName() {
...
+      fail("Could not get resource from pool");
+    } catch (JedisConnectionException e) {
+      factory.setPassword("foobared");
+      obj = pool.getResource();
```
sazzad16: Close `obj` within this catch block. Optionally, you can _declare_ `obj` (aka `Jedis obj = pool.getResource();`) here with/without try-with-resources.

PR #1609, opened 2017-09-20
Moved scan command into MultiKey... interface

sazzad16:


marcosnils:
hey @sazzad16 . Thx for your contributions!. I guess this PR has changes that shouldn't be in here right?

Like some guideline removas and copyright changes.

sazzad16:
@marcosnils I needed to create a build with some pending changes. This is happened when I was trying to merge into common place.

I will fix these as soon as I can :)

HeartSaVioR:
LGTM. TBH, I guess it breaks backward compatibility, but I would rather not thinking someone uses JedisCluster with interface.
@mp911de Could you please check there's any issues on Spring Data Redis if we adopt the change?
@marcosnils Please review this as well.

HeartSaVioR:
This is actually a bugfix, not improvement. So I just marked target version as bugfix version. Please let me know if someone doesn't agree.

mp911de:
Thanks for the ping. This change doesn't break Spring Data Redis as we don't use `SCAN` in cluster yet and we directly use `JedisCluster`/`BinaryJedisCluster`.

For anybody using the interfaces directly, it's a breaking change which shouldn't happen in a patch release at all. Wouldn't it be possible to deprecate `scan(…)` in `BinaryJedisClusterCommands` with a hint (`@see …`) so you can remove the method in 3.0?

marcosnils:
> This is actually a bugfix, not improvement. So I just marked target version as bugfix version. Please let me know if someone doesn't agree.

But it still breaks backwards compatibility. I'd vote for marking it as *deprecated* and push the change to 3.0

@HeartSaVioR @mp911de @sazzad16 ?

mp911de:
Introducing the method on `MultiKeyBinaryJedisClusterCommands` and deprecating the old one doesn't for 2.10 would be a nice move and allow users to adapt the new code before 3.0.

sazzad16:
@marcosnils @mp911de agreed!

sazzad16:
PR for 2.10 #1668

sazzad16:
Merged into master.

PR #1610, opened 2017-09-20
ScanParam `match` binary version

sazzad16:


src/main/java/redis/clients/util/JedisClusterHashTagUtil.java
```diff
@@ -14,6 +14,10 @@ public static String getHashTag(String key) {
     return extractHashTag(key, true);
   }
 
+  public static boolean isClusterCompliantMatchPattern(byte[] matchPattern) {
```
HeartSaVioR: Looks like this is unrelated to the actual changeset. Minimizing changeset helps us avoiding merge conflict.

sazzad16:
Merged into master, 2.10.

PR #1612, opened 2017-09-20
ClusterCommands interface is implemented in BinaryJedis

sazzad16:
`ClusterCommands` methods are accessible from only `Jedis` but not `BinaryJedis`. This change would allow us to access those methods from both classes.

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryJedis.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedis.java
@@ -34,7 +34,8 @@ import redis.clients.util.JedisURIHelper;
 import redis.clients.util.SafeEncoder;
 
 public class BinaryJedis implements BasicCommands, BinaryJedisCommands, MultiKeyBinaryCommands,
-    AdvancedBinaryJedisCommands, BinaryScriptingCommands, Closeable {
+    AdvancedBinaryJedisCommands, BinaryScriptingCommands, ClusterCommands, Closeable {
+
   protected Client client = null;
   protected Transaction transaction = null;
   protected Pipeline pipeline = null;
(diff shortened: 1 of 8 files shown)


src/main/java/redis/clients/jedis/BinaryJedisCluster.java
```diff
@@ -17,11 +18,14 @@
 import java.util.Set;
 
 import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
-import redis.clients.util.JedisClusterHashTagUtil;
 
 public class BinaryJedisCluster implements BinaryJedisClusterCommands,
     MultiKeyBinaryJedisClusterCommands, JedisClusterBinaryScriptingCommands, Closeable {
 
+  public static enum Reset {
```
HeartSaVioR: This change will trigger backward incompatibility since it is exposed to `clusterReset`.

src/main/java/redis/clients/jedis/BinaryJedis.java
```diff
@@ -3776,4 +3777,151 @@ public Long hstrlen(byte[] key, byte[] field) {
...
+
+  @Override
+  public String clusterMeet(final String ip, final int port) {
+    checkIsInMultiOrPipeline();
+    client.clusterMeet(ip, port);
+    return client.getStatusCodeReply();
+  }
+
+  @Override
+  public String clusterReset(final BinaryJedisCluster.Reset resetType) {
```
HeartSaVioR: This change will trigger backward incompatibility.

sazzad16:
@HeartSaVioR What if we move `Reset` back in JedisCluster? This won't look good but won't break backward compatibility either.

PR #1615, opened 2017-10-02
Code re-usage in JedisCluster constructors

sazzad16:
JedisCluster(HostAndPort node, ...) constructors should just be convenient options for JedisCluster(Set<HostAndPort> nodes, ...) constructors.
This way, passing default values resides in one place instead of multiple for same code flow.


HeartSaVioR:
LGTM. @marcosnils Please review.

HeartSaVioR:
I'll just change the label since @sazzad16 is a collaborator.

sazzad16:

PR #1616, opened 2017-10-02
removing squid:s2925

alainlompo:
Removing sonar squid:s2925 violation, using [org.awaitability](https://github.com/awaitility/awaitility/wiki/Usage#fork-destination-box)

See also: [Sonar Squid:s2925, thread.sleep should not be used in tests](https://sonarcloud.io/organizations/default/rules#rule_key=squid%3AS2925)

### DIFF
--- a/src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
+++ b/src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
@@ -1,6 +1,5 @@
 package redis.clients.jedis.tests.commands;
 
-
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
@@ -10,19 +9,16 @@ import static redis.clients.jedis.ScanParams.SCAN_POINTER_START;
 import static redis.clients.jedis.ScanParams.SCAN_POINTER_START_BINARY;
 import static redis.clients.jedis.params.set.SetParams.setParams;
 
-
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
-import java.util.concurrent.Callable;
-import java.util.concurrent.TimeUnit;
 
 import org.junit.Test;
-import static org.awaitility.Awaitility.*;
 
 import redis.clients.jedis.Protocol.Keyword;
 import redis.clients.jedis.ScanParams;
 import redis.clients.jedis.ScanResult;
+import redis.clients.jedis.tests.utils.AwaitilityUtils;
 import redis.clients.util.SafeEncoder;
 
 public class AllKindOfValuesCommandsTest extends JedisCommandTestBase {
@@ -360,23 +356,6 @@ public class AllKindOfValuesCommandsTest extends JedisCommandTestBase {
 
   }
 
-  private Callable<Boolean> jedisTargetIsIdle(final String target) {
-    return new Callable<Boolean>() {
-      @Override
-      public Boolean call() throws Exception {
-        return jedis.objectIdletime(target) > 0;
-      }
-    };
-  }
-
-  private Callable<Boolean> jedisTargetIsIdle(final byte[] target) {
-    return new Callable<Boolean>() {
-      @Override
-      public Boolean call() throws Exception {
-        return jedis.objectIdletime(target) > 0;
-      }
-    };
-  }
   @Test
   public void touch() throws Exception {
     final String foo1 = "foo1";
@@ -385,7 +364,7 @@ public class AllKindOfValuesCommandsTest extends JedisCommandTestBase {
 
     jedis.set(foo1, "bar1");
 
-    await().atMost(2, TimeUnit.SECONDS).until(jedisTargetIsIdle(foo1));
+    AwaitilityUtils.waitFor(2000);
 
     assertTrue(jedis.objectIdletime("foo1") > 0);
 
@@ -409,7 +388,7 @@ public class AllKindOfValuesCommandsTest extends JedisCommandTestBase {
 
     jedis.set(bfoo1, bbar1);
 
-    await().atMost(2, TimeUnit.SECONDS).until(jedisTargetIsIdle(bfoo1));
+    AwaitilityUtils.waitFor(2000);
 
     assertTrue(jedis.objectIdletime(bfoo1) > 0);
 
--- a/src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
+++ b/src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
@@ -1,5 +1,6 @@
 package redis.clients.jedis.tests.commands;
 
+
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
@@ -9,11 +10,15 @@ import static redis.clients.jedis.ScanParams.SCAN_POINTER_START;
 import static redis.clients.jedis.ScanParams.SCAN_POINTER_START_BINARY;
 import static redis.clients.jedis.params.set.SetParams.setParams;
 
+
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
 
 import org.junit.Test;
+import static org.awaitility.Awaitility.*;
 
 import redis.clients.jedis.Protocol.Keyword;
 import redis.clients.jedis.ScanParams;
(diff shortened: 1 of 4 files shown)


HeartSaVioR:
LGTM. @marcosnils Please review this as well.

alainlompo:
@HeartSaVioR any nnews about the review from @marcosnils? Can someone else reviews if he is not yet available? Thanks

HeartSaVioR:
@sazzad16 Could you please review this as well?

sazzad16:
@HeartSaVioR I purposely avoided this as I have not enough idea :(

src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
```diff
@@ -355,14 +360,33 @@ public void ttl() {
...
   public void touch() throws Exception {
-    long reply = jedis.touch("foo1", "foo2", "foo3");
+    final String foo1 = "foo1";
+    long reply = jedis.touch(foo1, "foo2", "foo3");
     assertEquals(0, reply);
 
-    jedis.set("foo1", "bar1");
+    jedis.set(foo1, "bar1");
+
+    await().atMost(2, TimeUnit.SECONDS).until(jedisTargetIsIdle(foo1));
```
sazzad16: The idea here is to wait for at least 1 second. 2000ms is used to be safe. Is at most 2 seconds safe here?

src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
```diff
@@ -355,14 +360,33 @@ public void ttl() {
...
 
-    Thread.sleep(2000);
     assertTrue(jedis.objectIdletime("foo1") > 0);
```
sazzad16: `jedis.objectIdletime("foo1") > 0` is already called in `...until(jedisTargetIsIdle(foo1))`. Why use it again?

src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
```diff
@@ -385,7 +409,8 @@ public void touch() throws Exception {
 
     jedis.set(bfoo1, bbar1);
 
-    Thread.sleep(2000);
+    await().atMost(2, TimeUnit.SECONDS).until(jedisTargetIsIdle(bfoo1));
```
sazzad16: Similar concern as earlier.

src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
```diff
@@ -385,7 +409,8 @@ public void touch() throws Exception {
...
+
     assertTrue(jedis.objectIdletime(bfoo1) > 0);
```
sazzad16: Similar concern as earlier.

src/test/java/redis/clients/jedis/tests/commands/ClusterBinaryJedisCommandsTest.java
```diff
@@ -186,14 +189,18 @@ private static String getNodeId(String infoOutput) {
   }
 
   private void waitForClusterReady() throws InterruptedException {
-    boolean clusterOk = false;
-    while (!clusterOk) {
```
sazzad16: This loop used to check a condition after each 50ms. Don't we need repeated checking anymore?

src/test/java/redis/clients/jedis/tests/commands/ClusterBinaryJedisCommandsTest.java
```diff
@@ -186,14 +189,18 @@ private static String getNodeId(String infoOutput) {
...
-      if (node1.clusterInfo().split("\n")[0].contains("ok")
-          && node2.clusterInfo().split("\n")[0].contains("ok")
-          && node3.clusterInfo().split("\n")[0].contains("ok")) {
-        clusterOk = true;
+    await().atMost(50, TimeUnit.SECONDS).until(clusterIsReady());
```
sazzad16: 50 seconds???

alainlompo:
@sazzad16 thank you for the review.

**There are about 15 Thread.sleep(). Only 3 of them are addressed in this PR and I am concerned about all of those**.

Actually there are much more squid in which I am interested than the s2925. But it is a significant amount of effort to fix. Here are a mention of a various types of squids that I found in the code

- squid:S3066
- squid:S2153
- squid:s2293
- squid:s1192
- squid:s00107

And many more. I was quite surprise by the important amount of these types of violations in the code, which led me to think that it was maybe not of a big interest. Fixing a few of these and creating a pull request seemed a good way to see if there was any interest in fixing the squid violations on the project. So now I can see there is some interest in it and I can proceed further.

Regarding your remarks on the 3 squid:s2925 that I suggested, I took your remarks into account and changed the approach. In my new approach I am replacing strictly the Thread.sleep statements, nothing less, nothing more, and I use

`await().atLeast(duration, ...).until(timeIsElapsed(timeToElapse...)`

Please take a look at the newly pushed code. If it is OK, just write back and I will push the 12 other squid:s2925 fixes before the PR can be approved. After that I can proceed one squid type at a time and create a PR for each of them until it is all processed.

I also refactored the waiting code into an Util class

HeartSaVioR:
@alainlompo 
Please don't address squad issues unless you understand how the relevant code works. For example, we have been receiving PRs regarding squad fix, and already denied the patch about "empty list instead of null" because Redis guide clearly says the client must return null instead of empty list due to classify twos.

HeartSaVioR:
@sazzad16 Thanks for the detailed review. Your review comments are great I missed these things somehow.

alainlompo:
@HeartSaVioR , thank you for this information. Such cases should be normally occuring only in a few cases, right? It should be great if a comment could be added for such pieces of code so as to explain why the squid could not be fixed, maybe even an annotation, something like:

```
@IgnoreSquid("squid:s1168", "Using empty list instead of null is not possible for this method because Redis guide clearly says the client must return null instead of empty list due to classify twos.")
public T targetMethod(....)
```

That way this information which is important can be part of our code base documentation and would be also easily visibile to anyone who tries an attempt to fix such issue.

I also think that the case of using Thread.sleep in test code does not fall into the above situation, unless you say otherwise. It would be great if I could get a new feedback from you or @sazzad16 about the update on that fix, so I can fix the 12 other cases (when the feedback is a GO)

sazzad16:
@alainlompo 

First, to answer your question:
> Such cases should be normally occuring only in a few cases, right?

No. It's very much normal for Jedis. So, even comments or annotations are not feasible options. Because, then, we'd have to add those on almost all methods!

Second, your PR is looks much better now and I truly appreciate your effort. But I'm concerned about your _understanding_ at some places. E.g. to work in `AllKindOfValuesCommandsTest.touch()` someone should at least know about Redis `TOUCH` and `OBJECT IDLETIME` commands. Because it would be vital to get the significance of 2000ms with `> 0` checking.

So, as @HeartSaVioR already said:
> Please don't address squad issues unless you understand how the relevant code works.

HeartSaVioR:
Personally I'm also not supportive to squad because it is not integrated to Jedis codebase. We don't use squad but we need to leave annotation for squad: what a counterintuitive!
If we really want to introduce something like checkstyle to make our code much safer, it should be integrated to build process, and then we could have exceptions on rules.

alainlompo:
@HeartSaVioR @sazzad16 thanks for your review and great comments. I guess the squids fixing is here a very sensitive thing and require deeper knowledge of the project. Maybe can you let me know more about your current improvement roadmap.

marcosnils:
going back to this PR, is it really necessary?. I even see that even awaitility uses `Thread.sleep` in their tests. (https://github.com/awaitility/awaitility/blob/70a97670ac54ddcf686db5b9d8fe337a52742d01/awaitility/src/test/java/org/awaitility/WaitForAtomicBooleanTest.java). Does make sense what we're trying to "fix" here?

AFAIK nobody had errors with Jedis tests.

alainlompo:
@marcosnils, thank you, for your illustration, they are testing awaitility, they have to use something else to be able to do that. Awaitility is precisely used as a convenient replacement to these `Thread.sleep`. If they would then use it that's a contradictory behaviour worth stopping their project.

Quoting from [sonarcloud.io](https://sonarcloud.io/organizations/default/rules#rule_key=squid%3AS2925)

> Using Thread.sleep in a test is just generally a bad idea. It creates brittle tests that can fail unpredictably depending on environment ("Passes on my machine!") or load. Don't rely on timing (use mocks) or use libraries such as **Awaitility** for asynchroneous testing.

But I agree that if there is not need to fix these squids, the pull request should be closed.

sazzad16:
Closing this for now.

PR #1617, opened 2017-10-03
Removed null checking where Redis does not return null

sazzad16:
Removed null checking (before returning emptySet) where Redis does not return null.


HeartSaVioR:
The commands you are addressing will never at least return null array (even non-exist key), right? If you meant this, LGTM.
@marcosnils Please review this as well.

sazzad16:
@HeartSaVioR Yes. For those calls Redis just returns 0-sized array.

sazzad16:

PR #1618, opened 2017-10-03
Null response of Redis is kept unaltered

sazzad16:
If Redis's response is null, user should get that. Changing it to empty collection hides the significance of null.

### DIFF
--- a/src/main/java/redis/clients/jedis/BuilderFactory.java
+++ b/src/main/java/redis/clients/jedis/BuilderFactory.java
@@ -1,7 +1,6 @@
 package redis.clients.jedis;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -73,7 +72,7 @@ public final class BuilderFactory {
     @SuppressWarnings("unchecked")
     public List<String> build(Object data) {
       if (null == data) {
-        return Collections.emptyList();
+        return null;
       }
       List<byte[]> l = (List<byte[]>) data;
       final ArrayList<String> result = new ArrayList<String>(l.size());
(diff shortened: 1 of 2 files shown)


HeartSaVioR:
This breaks backward compatibility without being notice to the end users. Could we find the affecting places and comment to javadoc if possible?

sazzad16:
@HeartSaVioR IMHO, this is a fix for bugs introduced in #1299 . Similar bugs were introduced in #1301 (from same author) and were reverted.

sazzad16:
There are a large number of affecting places. Should we really have to comment all those in javadoc, if we consider these as bugfix?

marcosnils:
> If Redis's response is null, user should get that. Changing it to empty collection hides the significance of null.

For what I've seen redis commands that usually return a list, there's no `null` value in there. The `redis` server returns `empty list of set` which is what I thought we wanted to represent with the change in #1299.

I.E, if you do `lrange test 0 1` over an unexisting list, you'll get an empty list and not a `nil` type. That's why I think #1299 reflects this. @sazzad16 @HeartSaVioR  Is there a case you see where this is not the intention?

sazzad16:
@marcosnils `BLPOP` and `BRPOP` both return `nil` when no element could be popped and the timeout expired.

HeartSaVioR:
@marcosnils Redis protocol spec describes that null array shouldn't be interpreted as empty list so that end users don't confuse.
https://redis.io/topics/protocol
```
A client library API should return a null object and not an empty Array when Redis replies with a Null Array. This is necessary to distinguish between an empty list and a different condition (for instance the timeout condition of the BLPOP command).
```

@sazzad16 I agree. Let's just fix it without addressing javadoc.

sazzad16:

scadgek:
Hi, it looks like this is still an issue in the 2.9.0 lib published to Maven. Is it planned to publish 2.9.x with a fix? We need both TLS support and this regression fixed.

PR #1619, opened 2017-10-03
Added missing flush in getRawObjectMultiBulkReply

sazzad16:
`flush()` was missing in `getRawObjectMultiBulkReply()`. It is added there. Also `getObjectMultiBulkReply()` just calls `getRawObjectMultiBulkReply()`. So `flush()` is removed from `getObjectMultiBulkReply()` as it is already added in `getRawObjectMultiBulkReply()`.

### DIFF
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -273,11 +273,11 @@ public class Connection implements Closeable {
 
   @SuppressWarnings("unchecked")
   public List<Object> getRawObjectMultiBulkReply() {
+    flush();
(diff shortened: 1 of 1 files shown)


HeartSaVioR:
LGTM. @marcosnils Please review this as well.

sazzad16:
@HeartSaVioR My bad. Please check again.

sazzad16:

werkt:
FWIW - this change means that there is no way to safely synchronize subscription clients. When any attempt to synchronously send a message to subscribe can be interleaved with a flush resulting from a message having been received, protocol desynchronization can occur.

PR #1621, opened 2017-10-04
Using available BinaryClient methods for zrange operations from BinaryJedis

sazzad16:
1. Using available `BinaryClient` methods for zrange operations from `BinaryJedis`. Currently, in `BinaryJedis`, those calls are forwarded to similar public methods. But other operations in `BinaryJedis` and even same operations in `Jedis` call client methods directly.

2. Removed `BinaryClient` methods having both `byte[]` and `String` as params. This does not go along with current project design. Moreover, these methods are not used anywhere in this project.

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryClient.java
+++ b/src/main/java/redis/clients/jedis/BinaryClient.java
@@ -617,7 +617,6 @@ public class BinaryClient extends Connection {
   }
 
   public void zcount(final byte[] key, final double min, final double max) {
-
     sendCommand(ZCOUNT, key, toByteArray(min), toByteArray(max));
   }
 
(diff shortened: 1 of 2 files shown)


HeartSaVioR:
@sazzad16 Looks like the changeset only contains 2. Which point are you addressing 1.?

sazzad16:
@HeartSaVioR Thank you for reviewing. Please check now.

sazzad16:
@HeartSaVioR I was thinking about splitting this PR. `1` is for 2.9.1/2.10.0. `2` is for 3.0.0 with deprecation in 2.10.0. That's why I removed label and milestone. Now WDYT?

PS: Don't get confused with my similar comment in another PR :)

sazzad16:
This PR is separated into #1675 and #1676

PR #1623, opened 2017-10-07
Used assertNull(value) instead of assertEquals(null, value)

sazzad16:


HeartSaVioR:
LGTM. Since this is just a small fix I'll just handle it by myself.

sazzad16:

PR #1626, opened 2017-10-10
Removed touch(key) method from BinaryClient

sazzad16:
Removed touch(key) method which can be served by touch(array_of_keys) method.

Similar to #1574 

Note: `touch(String key)` is already non-existent in Client and Commands.


sazzad16:
`TOUCH` is not released yet. So this should be easy to remove without deprecation.

marcosnils:
LGTM!.

PR #1627, opened 2017-10-10
Proper usage of 'import static' for Command(s)

sazzad16:


HeartSaVioR:
LGTM. Since this is just a small fix I'll just handle it by myself.

sazzad16:
Merged into master.

Backport not possible due to conflict because of `import static redis.clients.jedis.Protocol.Keyword.*;` in respective branches.

PR #1638, opened 2017-10-22
Reduced encoding while creating Tuple and kept only one Set<Tuple> builder

sazzad16:
1. In following block of code
```
result.add(new Tuple(SafeEncoder.encode(iterator.next()), Double.valueOf(SafeEncoder
    .encode(iterator.next()))));
```
there are actually 2 conversions happening. First a byte[] to String within the block.
And then String to byte[] conversion in Tuple constructor.

Changing the block to
```
result.add(new Tuple(iterator.next(), Double.valueOf(SafeEncoder.encode(iterator.next()))));
```
reduces to 0 conversions.

2. As Tuple supports both byte[] and String at once, there is no need to have two separate
TUPLE_ZSET and TUPLE_ZSET_BINARY.

### DIFF
--- a/src/main/java/redis/clients/jedis/BuilderFactory.java
+++ b/src/main/java/redis/clients/jedis/BuilderFactory.java
@@ -266,8 +266,7 @@ public final class BuilderFactory {
       final Set<Tuple> result = new LinkedHashSet<Tuple>(l.size()/2, 1);
       Iterator<byte[]> iterator = l.iterator();
       while (iterator.hasNext()) {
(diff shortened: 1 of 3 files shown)


sazzad16:
@HeartSaVioR I was thinking about splitting `1` and `2`. `1` is for 2.9.1. But 2 is for 3.0.0 with deprecation in 2.10.0. WDYT?

sazzad16:
This PR is separated into #1677 and #1678

HeartSaVioR:
Makes sense. Let's close this and handle this from other PRs.

PR #1640, opened 2017-10-22
Update readme - added maven central and license badges

pushkyn:
I think it's good idea - add some useful badges


marcosnils:
LGTM. I'll go ahead and merge this

PR #1641, opened 2017-10-22
 Modified Tuple.equals()

### LINKED ISSUE #1533: Java equals/hashcode contract broken on Tuple class
### Expected behavior
Tuple hashCode() and equals() implementation should obey the Java contract for these methods.
Per the Java Object hashcode() javadoc:
"If two objects are equal according to the equals(Object) method, then calling the hashCode method on each of the two objects must produce the same integer result."

### Actual behavior
Tuple hashCode() implementation considers the score in its calculation, whereas equals() does not. As a result, you can have 2 Tuples that are equal, with differing hashCodes.

### Steps to reproduce:
This code shows the problem. The 3rd printout should be FALSE, not TRUE.

        Tuple tuple1 = new Tuple("user1", 10.0);
        Tuple tuple2 = new Tuple("user1", 5.0);

        System.out.println(tuple1.hashCode());
        System.out.println(tuple2.hashCode());

        System.out.println(tuple1.equals(tuple2));
    }

#### Jedis version:
2.9

#### Redis version:
N/A

#### Java version:
1.8

sazzad16:
Based on following question and answer:
Q: https://github.com/xetorthio/jedis/issues/1533#issuecomment-324300040 by @sazzad16
A: https://github.com/xetorthio/jedis/issues/1533#issuecomment-338466018 by @HeartSaVioR 

Resolves #1533

### DIFF
--- a/src/main/java/redis/clients/jedis/Tuple.java
+++ b/src/main/java/redis/clients/jedis/Tuple.java
@@ -1,6 +1,7 @@
 package redis.clients.jedis;
 
 import java.util.Arrays;
+import java.util.Objects;
 
 import redis.clients.util.ByteArrayComparator;
 import redis.clients.util.SafeEncoder;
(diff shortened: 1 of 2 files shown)


marcosnils:
LGTM!. @HeartSaVioR ?

sazzad16:

Issue #1123, opened 2015-10-01
NumberFormatException on Infinite Sorted Set Score


efmr:
Redis supports Infinite values on Sorted Sets by setting them to "inf" or "-inf". This is an issue for Jedis since any method that retrieves the score from the sorted set will throw an NumberFormatException since `Double.valueOf("inf")` is not a valid input string for conversion.

I suggest matching `Double.POSITIVE_INFINITY` (and negative) when the input string matches "inf" or"-inf".

Is there a workaround this issue?

Following is an example code that generates this exception.

``` java
import org.junit.Assert;
import org.junit.Test;
import redis.clients.jedis.Jedis;
import redis.clients.jedis.ZParams;

public class JedisResponseTest {
    @Test
    public void infiniteSortedSetScore() {
        Jedis jedis = new Jedis("localhost");
        /*
        Setting up a sorted set member score to be equal to inf since
          .zadd("foo", Double.POSITIVE_INFINITY, "bar");
        OR
          .zadd("foo", "inf", "bar");
        are not implemented
        */
        jedis.zadd("foo", Double.MAX_VALUE, "bar");
        ZParams params = new ZParams();
        params.weightsByDouble(2d);
        jedis.zinterstore("foobar", params, "foo");
        //should convert redis response "inf" to Double.POSITIVE_INFINITY
        // but throws numberFormatException instead
        Double result = jedis.zscore("foobar", "bar");
        Assert.assertTrue(result != null && result.equals(Double.POSITIVE_INFINITY));
        /*
        java.lang.NumberFormatException: For input string: "inf"
            at sun.misc.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:1250)
            at java.lang.Double.valueOf(Double.java:504)
            at java.lang.Double.<init>(Double.java:597)
            at redis.clients.jedis.Jedis.zscore(Jedis.java:1474)
            at JedisResponseTest.infiniteSortedSetScore(JedisResponseTest.java:22)
        (...)
        */
    }
}
```

HeartSaVioR:
Maybe we need to upmerge #463 and merge.

sazzad16:
Resolved by #1642

PR #1642, opened 2017-10-22
Implemented ability to read infinity

sazzad16:
- Implemented ability to read infinity.
- Replaced Double conversions with infinity reading.
- Added test case.
- Removed Jedis.getTupledSet() and replaced it with BinaryJedis.get~~Binary~~TupledSet().

Closes #463 #1123

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryJedis.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedis.java
@@ -1748,20 +1748,14 @@ public class BinaryJedis implements BasicCommands, BinaryJedisCommands, MultiKey
   public Double zincrby(final byte[] key, final double increment, final byte[] member) {
     checkIsInMultiOrPipeline();
     client.zincrby(key, increment, member);
-    String newscore = client.getBulkReply();
-    return Double.valueOf(newscore);
+    return BuilderFactory.DOUBLE.build(client.getOne());
   }
 
   @Override
   public Double zincrby(final byte[] key, final double increment, final byte[] member, final ZIncrByParams params) {
     checkIsInMultiOrPipeline();
     client.zincrby(key, increment, member, params);
-    String newscore = client.getBulkReply();
-
-    // with nx / xx options it could return null now
-    if (newscore == null) return null;
-
-    return Double.valueOf(newscore);
+    return BuilderFactory.DOUBLE.build(client.getOne());
   }
 
   /**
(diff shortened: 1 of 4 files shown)


marcosnils:
LGTM!. @HeartSaVioR ?

src/main/java/redis/clients/jedis/BinaryJedis.java
```diff
@@ -2507,18 +2501,19 @@ public Long zcount(final byte[] key, final byte[] min, final byte[] max) {
       final int offset, final int count) {
     checkIsInMultiOrPipeline();
     client.zrangeByScoreWithScores(key, min, max, offset, count);
-    return getBinaryTupledSet();
+    return getTupledSet();
   }
 
-  private Set<Tuple> getBinaryTupledSet() {
+  protected Set<Tuple> getTupledSet() {
+    checkIsInMultiOrPipeline();
```
HeartSaVioR: Do we need to check this again while reading response from Redis?

src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -16,8 +16,15 @@
...
     @Override
     public Double build(Object data) {
-      String asString = STRING.build(data);
-      return asString == null ? null : Double.valueOf(asString);
+      String string = STRING.build(data);
+      if (string == null) return null;
+      try {
+        return Double.valueOf(string);
+      } catch (NumberFormatException e) {
+        if (string.equals("inf")) return Double.POSITIVE_INFINITY;
```
HeartSaVioR: `inf` -> `+inf`

https://redis.io/commands/zrangebyscore

```
Exclusive intervals and infinity

min and max can be -inf and +inf, so that you are not required to know the highest or lowest score in the sorted set to get all elements from or up to a certain score.
```

src/main/java/redis/clients/jedis/BinaryJedis.java
sazzad16: No, we don't. Nice finding @HeartSaVioR . It's happened because PR is crafted before merging #1600 .

src/main/java/redis/clients/jedis/BuilderFactory.java
sazzad16: First of all, the one you've mentioned is about request, not response. This PR is about reading Redis' responses.

After using several commands in several ways, I have found that Redis accepts both `+inf` and `inf` as positive infinity in requests. But, in responses, it only sends `inf` as positive infinity. In this PR, I have already added a test where Redis returns positive infinity. Please check that. It definity returned `inf`, otherwise the test wouldn't have passed.

Lastly, if you still want, I can add another condition for `+inf`.

sazzad16:
@HeartSaVioR @marcosnils rebased. PTAL again!

src/main/java/redis/clients/jedis/BuilderFactory.java
HeartSaVioR: Let's add `+inf` as it doesn't hurt and guard against Redis bugfix or regression or so.

sazzad16: Done!

marcosnils:
LGTM. Make sure to squash the changes when merging plz.

sazzad16:

PR #1646, opened 2017-10-28
ListPosition is renamed and moved out from BinaryClient

sazzad16:


sazzad16:
@HeartSaVioR I agree at _master only_. Please check #1674 for 2.10.

sazzad16:
Merged into master.

PR #1647, opened 2017-10-28
ClusterReset is renamed and moved out from JedisCluster

sazzad16:


HeartSaVioR:
Ah you may want to also resolve conflict too.

sazzad16:
@HeartSaVioR I do agree! Please check #1673 for 2.10.

sazzad16:
Merged into master.

PR #1655, opened 2017-11-06
Add ScanCursor ulility for easier iteration using SCAN-like commands

antonio-tomac:
It is known that using Jedis.keys("*") is bad, especially if key-space is big.
So, alternative is to use Jedis.scan() to avoid many problems with keys command.
But, using scan commands is somewhat cumbersome because we need to deal with cursors which can look ugly and obfuscate application logic with this technical detail.

I've implemented a class `ScanCursor` for easier/nicer iteration using SCAN, SSCAN, ZSCAN and HSCAN command.
Example usage is:
```
Jedis jedis = ...;
long total = 0;
for (String key : ScanCursor.builder(jedis).pattern("prefix*").scan()) {
    //do something with each matched key
    total += key.length();
}
```

### DIFF
--- a/src/main/java/redis/clients/util/ScanCursor.java
+++ b/src/main/java/redis/clients/util/ScanCursor.java
@@ -1,0 +1,405 @@
+package redis.clients.util;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Iterator;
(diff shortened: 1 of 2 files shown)


uglide:

PR #1661, opened 2017-11-19
Use int instead of short

sazzad16:


marcosnils:
@sazzad16 why `int` if short is big enough for the amount of possible slots?

sazzad16:
@marcosnils How about [Why does the Java API use int instead of short or byte?](https://stackoverflow.com/questions/27122610/why-does-the-java-api-use-int-instead-of-short-or-byte)

HeartSaVioR:
I'll wait for @marcosnils to approve as well, given that he participated to review.

marcosnils:
LGTM

sazzad16:
Merged into master, 2.10.

PR #1663, opened 2017-11-21
Move Jedis initialization in try block

sazzad16:
There is a null checking in `finally` which is meaningless if object creation is outside `try`-block.


marcosnils:
LGTM!

sazzad16:

PR #1665, opened 2017-11-23
JedisPoolConfig should use softMinEvictableIdleTimeMillis instead of minEvictableIdleTimeMillis

wantedlzx:
The JedisPoolConfig should have the "softMinEvictableIdleTimeMillis" (get and set) as the GenericObjectPool, to avoid destroying idle connections below the "minIdle" value. In this case, connections are dropped and immediately recreated with no reason. The "minEvictableIdleTimeMillis" should not be considered because there are ways to validate idle connections before they are used.

In other case, considered JVM heap, if youngGC happened frequently, maybe the idle connection will promote to old gen. If use minEvictableIdleTimeMillis, old gen will grow slowly.


sazzad16:
@wantedlzx Would it make any difference when `minIdle` equals 0 (default value)?

sazzad16:
@wantedlzx Would it make any difference when `minIdle` equals 0?

sazzad16:
@marcosnils @HeartSaVioR Close?

sazzad16:
Closed due to inactivity.

PR #1666, opened 2017-11-23
bitOpNot should throw exception for multiple source keys

pkashokkumar:
1. Modified bitOpNot to throw Data exception if there are multiple source keys
2. Wrote a new Unit test to verify that this exception is thrown

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryClient.java
+++ b/src/main/java/redis/clients/jedis/BinaryClient.java
@@ -24,6 +24,7 @@ import javax.net.ssl.SSLSocketFactory;
 
 import redis.clients.jedis.Protocol.Command;
 import redis.clients.jedis.Protocol.Keyword;
+import redis.clients.jedis.exceptions.JedisDataException;
 import redis.clients.jedis.params.geo.GeoRadiusParam;
 import redis.clients.jedis.params.set.SetParams;
 import redis.clients.jedis.params.sortedset.ZAddParams;
@@ -1077,7 +1078,7 @@ public class BinaryClient extends Connection {
       break;
     case NOT:
       kw = Keyword.NOT;
-      len = Math.min(1, len);
+      if (len > 1) throw new JedisDataException("BITOP NOT must be called with a single source key");
       break;
     }
 
--- a/src/test/java/redis/clients/jedis/tests/commands/BitCommandsTest.java
+++ b/src/test/java/redis/clients/jedis/tests/commands/BitCommandsTest.java
@@ -9,6 +9,7 @@ import org.junit.Test;
 import redis.clients.jedis.BitOP;
 import redis.clients.jedis.BitPosParams;
 import redis.clients.jedis.Protocol;
+import redis.clients.jedis.exceptions.JedisDataException;
 import redis.clients.util.SafeEncoder;
 
 import java.util.List;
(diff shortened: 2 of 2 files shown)


sazzad16:
@pkashokkumar If an exception is to be thrown before calling Redis, then why allow the user to call with multiple source keys at the first place?

pkashokkumar:
public void bitop(final BitOP op, final String destKey, final String... srcKeys) {}

The bitop method supports all the four BITOP operations namely AND, OR, XOR, NOT.
All the operations except NOT accept multiple source keys. So a user is allowed to pass multiple source keys to this method as long as the operation is not NOT. 

Hence, the only option would be to throw exception and fail the method if user enters multiple source keys for NOT operation, right?

This is similar to how the BITOP command works in redis-cli too. The help and hints shows support for multiple source keys, but it throws an error if multiple source keys are input for a NOT operation.

sazzad16:
@pkashokkumar 
> This is similar to how the BITOP command works in redis-cli too.

No, it's not similar to redis-cli. redis-cli actually calls Redis, gets error message and shows that.

marcosnils:
> No, it's not similar to redis-cli. redis-cli actually calls Redis, gets error message and shows that.

If that is the case then we shouldn't change anything from this side as well.

sazzad16:
> If that is the case then we shouldn't change anything from this side as well.

@marcosnils Do you mean, we shouldn't change anything compared to what redis-cli does?

pkashokkumar:
The current behavior of bitop method in jedis is different than the BITOP redis-cli command in the following way:
redis-cli bitop command when called with multiple source keys throws error:
> BITOP NOT destkey srckey1 srckey2
> (error) ERR BITOP NOT must be called with a single source key.

but jedis.bitop() when called with same arguments ignores srckey2 and passes only the first source key(srckey1) to Redis and hence the method succeeds.

I think this should be fixed to have a consistent behavior vis-a-vis redis-cli.
In my opinion, following are the two ways to fix it:

1. As shown in my pull request, jedis can proactively check if there are multiple source keys and throw exception so that the method fails. This looks similar to redis-cli but like @sazzad16 mentioned the difference is that unlike redis-cli, we are not letting the command go to Redis and fail from there.

2. Comment out the "len = Math.min(1, len);" line from the bitops method in BinaryClient.
This will result in all the source keys getting passed to Redis and it will throw back an error.
This should make it similar to the redis-cli behavior and also lets Redis trigger the error.

Sample execution result with Option 2:
java -jar target/redis-app-1.0-SNAPSHOT.jar 
Exception in thread "main" redis.clients.jedis.exceptions.JedisDataException: ERR BITOP NOT must be called with a single source key.
	at redis.clients.jedis.Protocol.processError(Protocol.java:131)
	at redis.clients.jedis.Protocol.process(Protocol.java:165)
	at redis.clients.jedis.Protocol.read(Protocol.java:219)
	at redis.clients.jedis.Connection.readProtocolWithCheckingBroken(Connection.java:310)
	at redis.clients.jedis.Connection.getIntegerReply(Connection.java:261)
	at redis.clients.jedis.Jedis.bitop(Jedis.java:2994)
	at org.ak.pocs.RedisApp.bitOpNot(RedisApp.java:63)
	at org.ak.pocs.RedisApp.main(RedisApp.java:27)

So @marcosnils , @sazzad16 I think we should implement Option 2 in jedis.

sazzad16:
Resolved in #1679

PR #1667, opened 2017-11-23
EVAL, EVALSHA should return Object

### LINKED ISSUE #1493: evalsha(String sha1, int numKeys, String... args) returns unexpected result.
When using lua scripts redis can return different types, and I was expecting that 
evalsha will return response that holds returned type that can be correctly parsed or cast to wanted type.
So I was expecting that  `Response<String> evalsha(String sha1, int numKeys, String... args) ` will return correct type only sa string representation, but it returned correct type and cast to `byte[]` failed in 
`BuilderFactory.STRING` factory. 

I think that 
```java
public Response<String> evalsha(String sha1, int numKeys, String... args) {
    getClient(sha1).evalsha(sha1, numKeys, args);
    return getResponse(BuilderFactory.STRING);
  }
```
should be:
```java
public Response<Object> evalsha(String sha1, int numKeys, String... args) {
    getClient(sha1).evalsha(sha1, numKeys, args);
    return getResponse(BuilderFactory.EVAL_BINARY_RESULT);
  }
```

### LINKED ISSUE #1504: ClassCastException while trying to get eval response of Lua script that returns a list
### Expected behavior

Original REDIS-Lua script that returns and clears a list when the fourth element is added: 

`eval "local res;if (redis.call('RPUSH','dalist',ARGV[1])==4) then res=redis.call('LRANGE','dalist',0,3);redis.call('DEL','dalist');return res; else return {} end" 1 argname 9999
`
Jedis code that invokes the previous script inside a transaction:

```
Jedis jedis = JedisUtil.getPool().getResource();
jedis.watch("dalist");
Transaction tx = jedis.multi();
Response<Long> r=tx.rpush("dalist","2322");
tx.eval("local res;if (redis.call('RPUSH','dalist',ARGV[1])==4) then res=redis.call('LRANGE','dalist',0,3);redis.call('DEL','dalist');return res; else return {} end",0,"99999");
        
List<Object> rs=tx.exec();
```

Expected: an empty array/4-elements array

### Actual behavior

Exception in thread "main" java.lang.ClassCastException: java.util.ArrayList cannot be cast to [B
	at redis.clients.jedis.BuilderFactory$5.build(BuilderFactory.java:58)
	at redis.clients.jedis.BuilderFactory$5.build(BuilderFactory.java:56)
	at redis.clients.jedis.Response.build(Response.java:61)
	at redis.clients.jedis.Response.get(Response.java:37)
	at redis.clients.jedis.Transaction.exec(Transaction.java:54)

When exec is invoked:
List<Object> rs=tx.exec();

### Steps to reproduce:

* Just start a Redis instance and run the code above.

redis.pool.maxActive=1024
redis.pool.maxIdle=200
redis.pool.maxWait=1000
redis.pool.testOnBorrow=true
redis.pool.testOnReturn=true
redis.ip=127.0.0.1
redis.port=6379

sazzad16:
This is fixed in master. But there are several issues #1361 #1493 #1504 regarding this. I understand that this breaks backward compatibility. IMHO, this should also be considered as BUG. So I think, we should include this in at least 2.10. May be in 2.9 too?

closes #1361, closes #1493, closes #1504

### DIFF
--- a/src/main/java/redis/clients/jedis/PipelineBase.java
+++ b/src/main/java/redis/clients/jedis/PipelineBase.java
@@ -1522,32 +1522,32 @@ public abstract class PipelineBase extends Queable implements BinaryRedisPipelin
     return getResponse(BuilderFactory.DOUBLE);
   }
 
-  public Response<String> eval(String script) {
+  public Response<Object> eval(String script) {
     return this.eval(script, 0);
   }
 
-  public Response<String> eval(String script, List<String> keys, List<String> args) {
+  public Response<Object> eval(String script, List<String> keys, List<String> args) {
     String[] argv = Jedis.getParams(keys, args);
     return this.eval(script, keys.size(), argv);
   }
 
-  public Response<String> eval(String script, int numKeys, String... args) {
+  public Response<Object> eval(String script, int numKeys, String... args) {
     getClient(script).eval(script, numKeys, args);
-    return getResponse(BuilderFactory.STRING);
+    return getResponse(BuilderFactory.EVAL_RESULT);
   }
 
-  public Response<String> evalsha(String script) {
+  public Response<Object> evalsha(String script) {
     return this.evalsha(script, 0);
   }
 
-  public Response<String> evalsha(String sha1, List<String> keys, List<String> args) {
+  public Response<Object> evalsha(String sha1, List<String> keys, List<String> args) {
     String[] argv = Jedis.getParams(keys, args);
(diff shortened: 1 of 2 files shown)


marcosnils:
LGTM!

sazzad16:
@marcosnils @HeartSaVioR How about including in 2.9?

sazzad16:
@marcosnils @HeartSaVioR How about including in 2.9 considering it as bugfix?

PR #1673, opened 2017-11-26
ClusterReset is renamed and moved out from JedisCluster with previous implementation deprecated

sazzad16:
#1647 for 2.10


sazzad16:
Merged into 2.10.

PR #1674, opened 2017-11-26
ListPosition is renamed and moved out from BinaryClient with previous implementation deprecated

sazzad16:
#1646 for 2.10


marcosnils:
LGTM!.

sazzad16:
Merged into 2.10.

PR #1675, opened 2017-11-27
Using available BinaryClient methods for zrange operations from BinaryJedis

sazzad16:
Currently, in BinaryJedis, those calls are forwarded to similar public methods. But in Jedis, similar operations are done other way.

Only zremrangeByScore was not in BinaryJedis and it is added.

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryClient.java
+++ b/src/main/java/redis/clients/jedis/BinaryClient.java
@@ -763,6 +763,10 @@ public class BinaryClient extends Connection {
     sendCommand(ZREMRANGEBYRANK, key, toByteArray(start), toByteArray(stop));
   }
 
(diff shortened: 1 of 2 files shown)


HeartSaVioR:
@sazzad16 
Btw you can open the build result to see why it fails, and rebuild if it is intermittent. Unless there's specific reason (there's known issue in Redis unstable branch, master branch already broken, etc.) we should verify the build success.

sazzad16:
@HeartSaVioR Thx!
Actually, I looked for rebuild option but didn't find it. So I thought that there isn't any. After you said, I looked again, found sign in option, signed in, looked again and found restart job option.

marcosnils:
LGTM!.

sazzad16:
Merged into master, 2.10.

PR #1677, opened 2017-11-27
Reduced encoding while creating Tuple

sazzad16:
In following code:
```
new Tuple(SafeEncoder.encode(iterator.next())
```
there are actually 2 conversions happening. First a byte[] to String. And then String to byte[] in Tuple constructor.

Changing the code to:
```
new Tuple(iterator.next())
```
reduces the number of conversions to 0.

Also, used available DOUBLE.build().


sazzad16:

PR #1678, opened 2017-11-27
Kept only one Set<Tuple> builder

sazzad16:
As Tuple supports both byte[] and String at once, there is no need to have two separate TUPLE_ZSET and TUPLE_ZSET_BINARY.


sazzad16:
Merged into master.

PR #1679, opened 2017-11-27
Modify BITOP NOT to fail with multiple sources keys

### LINKED ISSUE #1635: BITOP NOT unexpected behavior
### Expected behavior

`BITOP` `NOT` with multiple source keys should either throw an exception or not be allowed at all.

Following is the response from Redis in such case:
```
ERR ERR BITOP NOT must be called with a single source key.
```

### Actual behavior

It does a successfull operation with only first key.

### Steps to reproduce:
```
new Jedis().bitop(BitOP.NOT, "a", "b", "c", "d");
```

#### Jedis version:
2.9.0
#### Redis version:
4.0.1

sazzad16:
This fixes #1635 in a backward compatible way.


sazzad16:

PR #1681, opened 2017-11-27
Re-order if-else of keys and scan in JedisCluster(s)

sazzad16:


sazzad16:
Merged master, 2.10, 2.9.

PR #1682, opened 2017-11-27
else after return

sazzad16:
- Remove else after return in `GEORADIUS_WITH_PARAMS_RESULT`
- Remove else after return in `MODULE_LIST`
- Stop creating ArrayList with default size when empty is required


sazzad16:

PR #1684, opened 2017-11-27
Remove usages of toByteArray from PipelineBase

sazzad16:
**Dependant on #1675**


sazzad16:
Merged into master, 2.10.

PR #1685, opened 2017-11-29
Update commons-pool2 to version 2.4.3

sazzad16:
https://issues.apache.org/jira/browse/POOL-303 is released and should fix #1158

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -60,7 +60,7 @@
 		<dependency>
 			<groupId>org.apache.commons</groupId>
 			<artifactId>commons-pool2</artifactId>
-			<version>2.4.2</version>
+			<version>2.4.3</version>
 			<type>jar</type>
 			<scope>compile</scope>
 		</dependency>


sazzad16:
@HeartSaVioR @marcosnils PTAL
PS: I'm not sure about milestone.

HeartSaVioR:
I'd rather avoiding dependency upgrade in bugfix version, but if it is necessary for fixing the bug, I'm OK with that.

sazzad16:

sunsuk7tp:
@sazzad16 @HeartSaVioR 
Although I've already used Jedis2.9 with commons-pool 2.4.3 for standalone master/slave Redis in production environment, 
I met another kind of connection stuck issue due to regression by POOL-303. 
When the master node is down and many app. threads try `JedisPool#getResource`, 
each thread waits in `GenericObjectPool#borrowObject` forever until pool is free. 
There commons-pool's `maxWaitMillis` or Jedis's `(connection/so)timeout` have no effect. 
That's why last invalid `getResource` takes too long time to abort the process. 

I reported on commons-pool side at [POOL-347](https://issues.apache.org/jira/browse/POOL-347).
Anyway, commons-pool 2.4.3 may relieve #1158 certainly but it makes behavior worse in master down. 
I recommend to rollback this commit or wait for fixed 2.6.x for safe Jedis 2.9.1 release.

PR #1686, opened 2017-12-01
Add dump and restore into JedisCluster

sunsuk7tp:
Here let me add missing [DUMP](https://redis.io/commands/dump) and [RESTORE](https://redis.io/commands/restore) for `(Binary)JedisCluster` also.

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryJedis.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedis.java
@@ -3430,12 +3430,14 @@ public class BinaryJedis implements BasicCommands, BinaryJedisCommands, MultiKey
     return client.getIntegerReply();
   }
 
+  @Override
   public byte[] dump(final byte[] key) {
     checkIsInMultiOrPipeline();
     client.dump(key);
     return client.getBinaryBulkReply();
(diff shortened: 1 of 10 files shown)


sunsuk7tp:
As well as existing `(Binary)Jedis`'s methods, I dare not to add methods in interfaces for now.
But I suppose they should be defined to `(Binary)JedisClusterCommands` interfaces.

sazzad16:
@sunsuk7tp IMO, you should add methods in interfaces. Not only in cluster interfaces, but also in other interfaces, if you can.

sazzad16:
@HeartSaVioR @marcosnils Would we still keep return of dump and argument of restore as `byte[]`?

sunsuk7tp:
@sazzad16 Added into all interfaces.

sazzad16:
LGTM! @HeartSaVioR @marcosnils

marcosnils:
LGTM!

sazzad16:
Merged into master, 2.10.

sazzad16:
Merged into master, 2.10. Thanks @sunsuk7tp for contribution!

PR #1689, opened 2017-12-04
Move package util into jedis

sazzad16:
In `util` package:
- Some classes contain the name `Jedis`
- Some classes are dependant on elements in package `jedis` (e.g. `exceptions`)

IMHO, both of these should be avoided. A simple solution for this is to move entire `util` into `jedis`.

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryJedis.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedis.java
@@ -28,8 +28,8 @@ import redis.clients.jedis.params.GeoRadiusParam;
 import redis.clients.jedis.params.SetParams;
 import redis.clients.jedis.params.ZAddParams;
 import redis.clients.jedis.params.ZIncrByParams;
-import redis.clients.util.JedisByteHashMap;
-import redis.clients.util.JedisURIHelper;
+import redis.clients.jedis.util.JedisByteHashMap;
+import redis.clients.jedis.util.JedisURIHelper;
 
 public class BinaryJedis implements BasicCommands, BinaryJedisCommands, MultiKeyBinaryCommands,
     AdvancedBinaryJedisCommands, BinaryScriptingCommands, Closeable {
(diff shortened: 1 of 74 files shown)


HeartSaVioR:
Since it is clear backward incompatibility, I'd like to have @marcosnils also taking a look.

sazzad16:
@marcosnils ?

sazzad16:
@marcosnils PING

marcosnils:
This is already marked as "READY TO MERGE". LGTM!

sazzad16:
@marcosnils I just marked `READY TO MERGE` after waiting two months for your opinion :)

sazzad16:
@marcosnils I just marked `READY TO MERGE` after waiting two months for your opinion :)
Wanted to make sure you don't oppose the idea.

sazzad16:
Merged into master.

gkorland:
i think we need to add "proxy" files on the original packages to prevent backward issues

gkorland:
```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'redisConnectionFactory' defined in class path resource [org/springframework/boot/autoconfigure/data/redis/JedisConnectionConfiguration.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.data.redis.connection.jedis.JedisConnectionFactory]: Factory method 'redisConnectionFactory' threw exception; nested exception is java.lang.NoClassDefFoundError: redis/clients/util/SafeEncoder
```

sazzad16:
@gkorland 
As long as 2.x is being used, there shouldn't a problem. Besides, I can see spring-data-redis is officially using [jedis-2.9.0](https://github.com/spring-projects/spring-data-redis/blob/1887aa6eefa17aabc799a5850b79c71ba1c7eac2/pom.xml#L25). They'd also be able to use 2.10.x easily.

However, it requires only 7 changes in 6 files to make spring-data-redis be compatible with jedis-3.x. This should be a simple task. Isn't it? :)

PR #1693, opened 2017-12-09
Remove setupNodeIfNotExist for ssl

sazzad16:
There are no usage of these methods. From #1550, it is evident that these are not handy even for ssl operations.


sazzad16:
@HeartSaVioR @marcosnils

sazzad16:
Closed by #1705

PR #1694, opened 2017-12-09
Deprecate setupNodeIfNotExist for ssl

sazzad16:
Deprecation patch for #1693


sazzad16:
@HeartSaVioR @marcosnils

PR #1695, opened 2017-12-09
Modify ConnectionHandlingCommandsTest

sazzad16:
* Remove extending JedisCommandTestBase. This saves unnecessary setUp, tearDown.

* <strike>Remove reading from getRedisServers.</strike> Provide HostAndPort for Jedis and BinaryJedis. Only host (evidently `localhost`) is read which doesn't make complete sense. Either both host and port or none should be read.

### DIFF
--- a/src/test/java/redis/clients/jedis/tests/commands/ConnectionHandlingCommandsTest.java
+++ b/src/test/java/redis/clients/jedis/tests/commands/ConnectionHandlingCommandsTest.java
@@ -5,20 +5,19 @@ import static org.junit.Assert.assertEquals;
 import org.junit.Test;
 
 import redis.clients.jedis.BinaryJedis;
-import redis.clients.jedis.HostAndPort;
-import redis.clients.jedis.tests.HostAndPortUtil;
+import redis.clients.jedis.Jedis;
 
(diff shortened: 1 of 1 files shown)


sazzad16:
@HeartSaVioR @marcosnils

src/test/java/redis/clients/jedis/tests/commands/ConnectionHandlingCommandsTest.java
```diff
@@ -5,20 +5,19 @@
 import org.junit.Test;
 
 import redis.clients.jedis.BinaryJedis;
-import redis.clients.jedis.HostAndPort;
-import redis.clients.jedis.tests.HostAndPortUtil;
+import redis.clients.jedis.Jedis;
 
-public class ConnectionHandlingCommandsTest extends JedisCommandTestBase {
-  protected static HostAndPort hnp = HostAndPortUtil.getRedisServers().get(0);
```
HeartSaVioR: While taking only host is not valid, I think we still need to rely on this. Could we provide host and port for Jedis and BinaryJedis instance from this?

sazzad16: addressed

sazzad16:

PR #1703, opened 2017-12-09
Refactor imports

sazzad16:


marcosnils:
LGTM!.

marcosnils:
@sazzad16 is there a reason this can't go into 2.9.1?

sazzad16:
@marcosnils I was concerned about how much I'll be able to backport to 2.9, let alone 2.10. However, I can give that a try.

Changing milestone.

HeartSaVioR:
I guess it doesn't bring any bugfix and also any improvement, so I don't mind even if it will be merged in only 3.0.0. Backporting is already a huge pain: Let's avoid unnecessary hurdle.

sazzad16:

PR #1707, opened 2017-12-12
add socks5 proxy with  userpass auth

foxundermoon:
add proxy feature for connection
```java
  URI jedisUri=new URI("redis://:redispassword@10.10.x.x:6379?proxy=socks5://user:password@proxyhost:1080");
        Jedis jedis = new Jedis(jedisUri);
        String rst= jedis.set("__test_set_over_proxy","aha");

```


sureshkandimalla:
Does this works now, i don't see these classes or proxy setting for client in code?

PR #1712, opened 2017-12-17
Close jedis quietly in iteration

sunsuk7tp:
`jedis.close` can also throw `JedisException` when it cannot return resource to the pool. 
It causes the followings:

- Break `JedisClusterConnectionHandler#getConnection` on the way
    - If all jedis are unavailable, `JedisNoReachableClusterNodeException` should be thrown as spec. 
- Break `JedisClusterCommand#runWithRetries` on the way
    - because retry for-loop does not care `JedisException` 

In order to fix these issues, this PR fixed to just ignore JedisException in `close()`.


sunsuk7tp:
Rebased to fix conflict with #1727

uglide:

PR #1716, opened 2017-12-18
Generic Support for RedisModules

diegopacheco:
@xetorthio 

I'm working on custom redis module. Would like to add generic commands capability to Jedis. So basically you can pass a generic command name and var-args of arguments and them get the result to a specific type like Long, List<Object>, String.

Cheers,
Diego Pacheco

### DIFF
--- a/src/main/java/redis/clients/jedis/Client.java
+++ b/src/main/java/redis/clients/jedis/Client.java
@@ -3,7 +3,6 @@ package redis.clients.jedis;
 import static redis.clients.jedis.Protocol.toByteArray;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


marcosnils:
Hi @diegopacheco thx for taking the time to make a PR. Before moving forward, I'd suggest opening an issue as a "design discussion" to propose an alternative interface to the current implementation https://github.com/xetorthio/jedis/blob/master/src/test/java/redis/clients/jedis/tests/ModuleTest.java#L33-L47 and highlight your use-case about why this interface would be useful. 

Remember that Jedis is a general purpose redis library for Java. So any changes that we make -specially on the interfaces- should be as generic as possible so anyone can benefit.

marcosnils:
In addition to that, try searching in the current issues/pr as some discussions around modules are already on-going. I.E: https://github.com/xetorthio/jedis/issues/1713

diegopacheco:
@marcosnils 

Thanks for the feedback. I can do my use case with: 

https://github.com/xetorthio/jedis/blob/master/src/test/java/redis/clients/jedis/tests/ModuleTest.java#L33-L47

Cheers,
Diego Pacheco

PR #1718, opened 2017-12-19
Reduce sleep time to resolve test failing issue

sazzad16:
Recent PRs are failing after these lines.

### DIFF
--- a/src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
+++ b/src/test/java/redis/clients/jedis/tests/commands/AllKindOfValuesCommandsTest.java
@@ -403,7 +403,7 @@ public class AllKindOfValuesCommandsTest extends JedisCommandTestBase {
 
     jedis.set("foo1", "bar1");
 
-    Thread.sleep(2000);
+    Thread.sleep(1100); // little over 1 sec
     assertTrue(jedis.objectIdletime("foo1") > 0);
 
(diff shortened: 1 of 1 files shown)


sazzad16:
@HeartSaVioR @marcosnils I'm applying myself. Hope this is okay.

sazzad16:
@HeartSaVioR @marcosnils

marcosnils:
@sazzad16  wondering why's the reason that reducing the sleep makes it work.

sazzad16:
@marcosnils Your guess is as good as mine.
I played with different combinations of Jedis timeout, sleep time and Redis timeout. This turned out to be the minimal change that worked every the time in travis.

marcosnils:
If this makes tests more reliable then LGTM!

sazzad16:
@marcosnils I won't say this is a _reliable_ fix but more of a _quick_ fix.

Travis builds are failing for all the PRs. I thought it'd be a good idea to have a quick fix. We may create an issue to dig into this.

sazzad16:
From December 12th, 2017, Travis jobs are running in **Trusty** environment. This could be the reason. PRs before this date were passing (1706 and earlier) but PRs after this date are failing (1710 and later).

@marcosnils @HeartSaVioR

marcosnils:
@sazzad16 got it. Let's go ahead and merge this until we find a better fix.

sazzad16:
Merged into master, 2.10.

PR #1726, opened 2018-01-06
Ignore non-connected shards when closing ShardedJedis

### LINKED ISSUE #1558: Should check whether jedis is connecteded or not when destory it in ShardedJedisPool
Suppose exists three redis node(r1, r2, r3), r1 is out of network.

When we make a ShardedJedis object, and use it to find a key that located in r1, we will get a connectionTimeout Exception, and the jedis of r1 will be marked as broken, but other two jedis object(r2, r3) has not call connect() method. In this case, r1, r2, r3 will first set up TCP connection, then send quit, and close TCP connection next, but we dont need do this!

Maybe we should check it before destory it. And only when a jedis's isConnected is true, we destory it.

sazzad16:
Closes #1558


marcosnils:
LGTM!

sazzad16:

PR #1727, opened 2018-01-07
Stop closing Jedis object from caller

sazzad16:
When there is no exception from getResource, only inner Jedis objects are closed. But if there is exception in very first iteration of loop, Jedis passed by caller gets closed.
This change closes only the inner objects.


marcosnils:
LGTM!. Wondering how you spotted this :)

sazzad16:

Issue #1183, opened 2015-12-17
Response exception after discarded a transaction


yaxing:
Sample code:

```
Jedis j = new Jedis("host", port);
Transaction t = j.multi();
Response<String> resp = t.get("key");
t.discard();
String data = resp.get();
```

Such exception got thrown at the last line: 

```
redis.clients.jedis.exceptions.JedisDataException: Please close pipeline or multi block before calling this method.
```

I see that `generateResponse` is not called in [`discard()`](https://github.com/xetorthio/jedis/blob/master/src/main/java/redis/clients/jedis/Transaction.java#L78-L84). 

Is this expected? The confusing part is that it's the same exception where you forgot to call `exec()` or even the same when you forgot to sync a pipeline, thus making it hard to handle for a user. Is there any way to handle it better, maybe throwing a different exception saying the transaction got discarded?

marcosnils:
@yaxing thanks for reporting the issue. I'll check this out during the weekend. 

Any suggestions about how you'd expect this to work as a Jedis user?

yaxing:
@marcosnils thanks. What I think might work better would be a different type of exception for such cases. like TransactionDiscarded, PipelineClosed etc.

marcosnils:
I thought about this and I believe the best approach would be to change the type of the exception to `JedisException` instead of `JedisDataException` and to improve the logging message. The reason for this are:

1) Currently `JedisDataException` is used mainly when Redis returns an error about the operation. The pipeline/transaction validation we're doing it's something from the Jedis side, so it doesn't make sense to return this type of exception

2) What would be the benefit of returning a custom exception on this case?. You usually return a custom exception when you want your code to do some custom exception handling. In the case of Pipeline / Transaction being used incorrectly (as in the example of this issue) there's nothing to recover from because it's an incorrect use and Jedis users would need to change to code to fix this.

3) I see that `build` method in `Response` class can also returns a `JedisDataException` which also leads to confusion about what to do in terms of the state of the transaction/pipeline vs what redis returned. I guess this reinforces the fact that we should use a different exception (`JedisException`) when pipline / transaction is being used incorrectly.

@HeartSaVioR @xetorthio @yaxing what do you think?

HeartSaVioR:
@marcosnils 
Sorry to be late.
I also think we shouldn't use JedisDataException which is not ERR from Redis side.
Could you address this? Thanks!

marcosnils:
[...]

PR #1733, opened 2018-01-09
Introduce JedisBatchOperationException

sazzad16:
- Pipeline and Transaction related exceptions, which do not come from Redis, are changed to throw JedisBatchOperationException.
- SafeEncoder is changed to throw JedisException.

Inspired by https://github.com/redis/jedis/issues/1183#issuecomment-166799828 by @marcosnils and https://github.com/xetorthio/jedis/issues/1183#issuecomment-169230449 by @HeartSaVioR


sazzad16:
@HeartSaVioR @marcosnils ?

sazzad16:
@marcosnils @gkorland Please review this before releasing 3.0.0. This PR will remove wrongful usages of `JedisDataException`.

src/main/java/redis/clients/jedis/exceptions/JedisBatchOperationException.java
```diff
@@ -0,0 +1,10 @@
+package redis.clients.jedis.exceptions;
+
+public class JedisBatchOperationException extends JedisException {
```
gkorland: Can you make it extend "JedisDataException" this way it won't make a backward issue?

sazzad16: @gkorland It was done deliberately.

[According to marcosnils](https://github.com/xetorthio/jedis/issues/1183#issuecomment-166799828) @marcosnils 
> I thought about this and I believe the best approach would be to change the type of the exception to `JedisException` instead of `JedisDataException` and to improve the logging message. ... Currently `JedisDataException` is used mainly when Redis returns an error about the operation. The pipeline/transaction validation we're doing it's something from the Jedis side, so it doesn't make sense to return this type of exception. ...

[According to HeartSaVioR](https://github.com/xetorthio/jedis/issues/1183#issuecomment-169230449) @HeartSaVioR 
> ... I also think we shouldn't use JedisDataException which is not ERR from Redis side. ...

I agree with both of them. And that's why I crafted this PR.

sazzad16: @gkorland Please check #2360

yangbodong22011: i agree @gkorland advice, we should be compatible with `JedisDataException`, because maybe user code has `catch JedisDataException`.

So, the inheritance path is ： JedisBatchOperationException extends JedisDataException extends JedisException.

sazzad16: @yangbodong22011 WDYT about #2360 ?

PR #1738, opened 2018-01-13
Closed Jedis should not be operable without password

sazzad16:
when operating on a password protected Redis.


sazzad16:
@HeartSaVioR @marcosnils ?

sazzad16:
@marcosnils I don't actually remember! I probably saw a comment regarding that why a closed Jedis object is operable. I know it's the design to keep it that way. But thought it to be a good idea to make at least password protected Jedis inoperable. I still think so.

If you don't agree, please feel free to remove milestone and label or let me know.

sazzad16:
Related https://github.com/redis/jedis/pull/2368#issuecomment-790262400

PR #1739, opened 2018-01-13
Reset timeout to initial value

sazzad16:
In two test methods, Redis config is changed but is not reset to initial value. Because of this, `redis.clients.jedis.exceptions.JedisConnectionException: Unexpected end of stream.` occurs time to time in tests. This PR will resolve this issue.

### DIFF
--- a/src/test/java/redis/clients/jedis/tests/JedisTest.java
+++ b/src/test/java/redis/clients/jedis/tests/JedisTest.java
@@ -53,21 +53,46 @@ public class JedisTest extends JedisCommandTestBase {
     jedis.get("foo");
   }
 
-  @Test(expected = JedisConnectionException.class)
+  @Test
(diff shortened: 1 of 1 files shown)


sazzad16:
@HeartSaVioR @marcosnils Please review this. Once timeout is set to 1 second, tests with 1+ seconds of sleep are tend to fail. Resetting timeout should resolve this issue.

sazzad16:
@marcosnils @gkorland Please review this.

sazzad16:
@gkorland Just being lazy, but I think we can use `try-finally` in entire project in separate PR.

marcosnils:
> Just being lazy, but I think we can use `try-finally` in entire project in separate PR.

LGTM!. 

Good catch @sazzad16

PR #1741, opened 2018-01-14
Remove inner packages of params

sazzad16:
- It doesn't seem necessary to have inner packages, containing 1 (may be 2) classes, inside params package. So those classes are moved in params package.
- Constructors of Params classes are made public. There isn't seem any extra benefit of keeping those private.


sazzad16:
Merged into master.

Issue #1745, opened 2018-01-15
Type safe keys


OndraZizka:
### Current behavior

Currently, you can do:

    jedis.hset("foo", "bar");
    jedis.get("foo");

That is easy to overlook on review (nothing hints the key is for a hash) and will end with a runtime exception.

### Suggested behavior

    public String hget(final String key, final String field) { }

    public class HashKey {
           public HashKey(String key) { ... }
    }
    public default String hget(final HashKey hkey, final String field) {
           return hkey.getKey();
    }

This would prevent such mistakes:

    public static final HashKey KEY_FOO = new HashKey("foo");
    
    jedis.key(KEY_FOO);  // Compilation error
    jedis.hkey(KEY_FOO); // Nicely works with code-completion as a bonus.

#### Jedis version:
2.9.0

sazzad16:
@OndraZizka This alters user's data which we don't want.

OndraZizka:
@sazzad16  What do you mean by that?

OndraZizka:
Note for myself:
 (.*) h(.*)\(String key(.*)\);
default $1 h$2(HashKey key$3) { return h$2(key.getKey()$3); }

OndraZizka:
I am still not sure what @sazzad16  meant by "alters user's data".

sazzad16:
@OndraZizka Before your post was edited, it seemed to me you were talking about adding a prefix to user provided key which would alter (modify) user provided key.

However, about your current logic:
What if wrong type of key is created as below?
```
public static final HashKey KEY_FOO = new HashKey("foo");
public static final HashKey KEY_BOO = new StringKey("foo");

jedis.key(KEY_BOO);  // No compilation error, but runtime error
jedis.hkey(KEY_FOO);
```

Users have to know about the key type before creating the object. If a user can create an object which is safe to use, IMHO, s/he is able to do safe operations as well.

OndraZizka:
What do you mean by `jedis.key(KEY_BOO)`? There's no such method: http://static.javadoc.io/redis.clients/jedis/2.9.0/redis/clients/jedis/Jedis.html

OndraZizka:
I assume you mean get/hget.
In which case, that's exactly the point of the change: To reduce the places where you can make errors to the definition of constants. These constants can be maintained by an app architect. And then other developers who use them would not be able to use wrong Jedis method with given key constant.

Ofcourse, someone could create their own constant. But why would they do it?

sazzad16:
[...]

PR #1746, opened 2018-01-15
#1745 type safe keys - demo

OndraZizka:
Here is a demonstration of what #1745 could look like.
This is Java 8 style for brevity, I can rewrite for Java 7.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -101,8 +101,8 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.1</version>
 				<configuration>
-					<source>1.7</source>
-					<target>1.7</target>
+					<source>1.8</source>
+					<target>1.8</target>
 				</configuration>
 			</plugin>
 			<plugin>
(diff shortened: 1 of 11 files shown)


pom.xml
```diff
@@ -101,8 +101,8 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.1</version>
 				<configuration>
-					<source>1.7</source>
-					<target>1.7</target>
+					<source>1.8</source>
+					<target>1.8</target>
```
OndraZizka: Just for demo purposes.

src/main/java/redis/clients/jedis/keys/JedisCommandsTypeSafe.java
```diff
@@ -0,0 +1,303 @@
...
+import java.util.Map;
+import java.util.Set;
+
+public interface JedisCommandsTypeSafe extends JedisCommands {
+
+    default String set(SetKey key, String value) { return set(key.getKey(), value); }
+
+    default String set(SetKey key, String value, SetParams params) { return set(key.getKey(), value, params); }
+
+    String get(String key);
```
OndraZizka: I'm letting these in to track which are left to add.

sazzad16:
@HeartSaVioR @marcosnils Close?

PR #1756, opened 2018-01-24
Rename the argument key to sampleKey

sazzad16:
where it is not sent to Redis as an argument.

Based on https://github.com/xetorthio/jedis/issues/1257#issuecomment-211627200


sazzad16:
Merged into master, 2.10.

PR #1759, opened 2018-01-29
Stop modifying global config

sazzad16:
because it can effect tests which will be executed afterwards.


sazzad16:

PR #1761, opened 2018-02-02
Small fixes in the Jedis and BinaryJedis classes

pechenoha:


src/main/java/redis/clients/jedis/BinaryJedis.java
```diff
@@ -252,11 +252,11 @@ public String quit() {
...
 
   /**
-   * Test if the specified keys exist. The command returns the number of keys existed Time
-   * complexity: O(N)
+   * Test if the specified keys exist. The command returns the number of keys exist.
+   * Time complexity: O(N)
    * @param keys
-   * @return Integer reply, specifically: an integer greater than 0 if one or more keys existed 0 if
-   *         none of the specified keys existed
+   * @return Integer reply, specifically: an integer greater than 0 if one or more keys exist
```
sazzad16: Add a full stop at the end.

src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -180,11 +180,11 @@ public String get(final String key) {
...
 
   /**
-   * Test if the specified key exists. The command returns the number of keys existed Time
-   * complexity: O(N)
+   * Test if the specified keys exist. The command returns the number of keys exist.
+   * Time complexity: O(N)
    * @param keys
-   * @return Integer reply, specifically: an integer greater than 0 if one or more keys were removed
-   *         0 if none of the specified key existed
+   * @return Integer reply, specifically: an integer greater than 0 if one or more keys exist
```
sazzad16: Add a full stop at the end.

src/main/java/redis/clients/jedis/BinaryJedis.java
pechenoha: I suppose the full stop should be at the end of the next line (258), but the comma - at the end of this line (258), right?

src/main/java/redis/clients/jedis/Jedis.java
pechenoha: The same question as above.

src/main/java/redis/clients/jedis/BinaryJedis.java
sazzad16: What we need is a proper punctuation after 258's `exist`. So I suggested a full stop. But your suggestion also works.

sazzad16:
LGTM!

sazzad16:

PR #1763, opened 2018-02-06
Calculate cluster slot only once for same key

sazzad16:
even if re-attempt happens.

Also, necessary variables are made to be final.


sazzad16:
@HeartSaVioR @marcosnils

sazzad16:

PR #1770, opened 2018-02-11
Modify cluster related exceptions

sazzad16:
- Add JedisClusterOperationException. When JedisClusterException is not because of a direct response from Redis (aka JedisDataException), this JedisClusterOperationException is to be used instead.

- JedisClusterMaxRedirectionsException is renamed to JedisClusterMaxAttemptsException. Because this is not thrown after limited redirections but limited attempts. Also, it is now a subclass of JedisClusterOperationException instead of JedisDataException.


sazzad16:
@HeartSaVioR @marcosnils

sazzad16:
Merged into master.

PR #1771, opened 2018-02-11
Stop throwing some exceptions in JedisClusterCommand

sazzad16:
In JedisClusterCommand, stop throwing custom exceptions when key is null or keys array is null or empty. This is not what is practised in other Jedis classes.

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisClusterCommand.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterCommand.java
@@ -11,8 +11,6 @@ import redis.clients.jedis.util.JedisClusterCRC16;
 
 public abstract class JedisClusterCommand<T> {
 
-  private static final String NO_DISPATCH_MESSAGE = "No way to dispatch this command to Redis Cluster.";
-
   private final JedisClusterConnectionHandler connectionHandler;
   private final int maxAttempts;
   private final ThreadLocal<Jedis> askConnection = new ThreadLocal<Jedis>();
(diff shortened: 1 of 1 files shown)


sazzad16:
@HeartSaVioR @marcosnils

marcosnils:
@smadasu by applying this change, this will cause a `NullPointerException` everytime you set a key to null from here: https://github.com/xetorthio/jedis/blob/master/src/main/java/redis/clients/jedis/util/JedisClusterHashTagUtil.java?utf8=%E2%9C%93#L27

In other Jedis classes ,we also check this at a different level: 

https://github.com/xetorthio/jedis/blob/master/src/main/java/redis/clients/jedis/util/SafeEncoder.java?utf8=%E2%9C%93#L28

I think it's OK to leave it like this.

sazzad16:
@marcosnils Please check #1774. It is opposite approach of this PR.

jiafu1115:
@sazzad16  @marcosnils  this PR is similar with #1774 which has more better implement. so can we close this one and keep #1774 as 3.0.0's milestone?  it is small improvement instead of key feature.
WDTY?

sazzad16:
@jiafu1115 If you think the implementation of #1774 is good, you can LGTM that :)

jiafu1115:
@sazzad16 yes. this PR loss some implement which is in #1774 . so suggest close this PR and continue #1774 , BTW, can you help to review one tiny change PR #1814 , Thanks

marcosnils:
Shall we close this in favor of #1774 ?

PR #1772, opened 2018-02-11
Stop returning JedisConnectionException at the last attempt of recursion

sazzad16:
Currently, if JedisConnectionException occurs at the last attempt of recursion, it is returned ignoring all the exceptions occured in previous attempts.

But returning MaxRedirectionsException as usual seems more logical at this spot.


sazzad16:
@HeartSaVioR @marcosnils

sazzad16:

PR #1773, opened 2018-02-11
Ditch localhost resolution

sazzad16:
The duty of setting/providing hosts properly can be left for the users. Moreover, altering values provided by the users or Redis may not be a good idea as because this may lead into some system dependant issues.

Note: Default host is set to `127.0.0.1`, similar to [redis-cli](https://github.com/antirez/redis/blob/5.0.5/src/redis-cli.c#L6994).

### DIFF
--- a/src/main/java/redis/clients/jedis/HostAndPort.java
+++ b/src/main/java/redis/clients/jedis/HostAndPort.java
@@ -1,20 +1,12 @@
 package redis.clients.jedis;
 
 import java.io.Serializable;
-import java.net.InetAddress;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 public class HostAndPort implements Serializable {
   private static final long serialVersionUID = -519876229978427751L;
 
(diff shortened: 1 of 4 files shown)


sazzad16:
@HeartSaVioR @marcosnils

marcosnils:
I'd like to see what @allanwax thinks about this.

allanwax:
@sazzad16 @marcosnils I have not looked at the code change for this.

If this is related to setting the localhost, the original purpose of the change was to stop the newbie issue with people switching from single Redis instances to cluster and specifying 'localhost' as the host name.  This doesn't work in a cluster.

Someone in the last year produced code which I liked where the default is the ip address (in my original code change) of the localhost but you can change it to whatever you want before creating the cluster.  This allows for Amazon host IDs.

So it's fine to allow people to set whatever they want, but an additional check is needed in a cluster to flag that using 'localhost' in a cluster won't work.

And if this not the topic, please inform me

sazzad16:
Closed by https://github.com/redis/jedis/commit/858c8051301854e8343eef66aa68c6b9dab1c328

PR #1774, opened 2018-02-13
JedisClusterCRC16.getSlot() wraps NullPointerException

sazzad16:
- This is similar to SafeEncoder.encode(String).
- This helps JedisClusterCommand specially in case of multi key operations.

Also includes,
- Move CRC16 tests in one place.
- Fix a wrong usage of DEFAULT_REDIRECTIONS.


sazzad16:
@marcosnils @HeartSaVioR

marcosnils:
LGTM. Can be merged after rebased.

sazzad16:
@marcosnils 3.0.0 is the one to go for. Because

1. Previously, a null key would cause `NullPointerException`. What if people have implementation catching this one?

2. `JedisClusterOperationException` is only available in 3.0.0.

marcosnils:
> Previously, a null key would cause NullPointerException. What if people have implementation catching this one?

This is a bug and shouldn't be thrown. 



> JedisClusterOperationException is only available in 3.0.0.

We can't do anything about this. Let's schedule it for 3.0 then :)

sazzad16:
Merged into master.

PR #1775, opened 2018-02-14
Unset data source while closing pool resource

sazzad16:
This prevents exception being thrown when pooled resource is being closed multiple times.

Some of probable causes of those exceptions:
- `java.lang.IllegalStateException: Object has already been returned to this pool or is invalid`
- `java.lang.IllegalStateException: Invalidated object not currently part of this pool`

### DIFF
--- a/src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
+++ b/src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
@@ -361,6 +361,31 @@ public class JedisPoolTest {
   }
 
   @Test
+  public void closeResourceTwice() {
+    JedisPool pool = new JedisPool(new JedisPoolConfig(), hnp.getHost(), hnp.getPort(), 2000);
+    Jedis j = pool.getResource();
+    j.auth("foobared");
+    j.ping();
+    j.close();
+    j.close();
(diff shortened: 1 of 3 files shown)


sazzad16:
@marcosnils @HeartSaVioR

src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
```diff
@@ -360,6 +360,31 @@ public void testAddObject() {
 
   }
 
+  @Test
```
bsnyder788: 👍

marcosnils:
@smadasu should this be released in 2.9.1?. What if someone was catching exceptions and doing something when this happened?. Seems like a backwards incompatible change. How about releasing it in 3.0?

sazzad16:
@marcosnils I saw a couple of issues where it turned out that the reason of the exception was that the user was closing Jedis twice, unknowingly! I considered, throwing exception while closing a already closed Jedis, as a bug. So as a bugfix, I set this PR for 2.9.1.

marcosnils:
Ok, I guess it makes sense.

sazzad16:

PR #1783, opened 2018-03-07
catch JedisDataException during pipeline sync

QooYoung:
The method client.getOne() should be called enough times even if JedisDataException is thrown, otherwise the next commands sent to server via the same connection may be crashed due to the remaining responses in its receive buffer.

### DIFF
--- a/src/main/java/redis/clients/jedis/ShardedJedisPipeline.java
+++ b/src/main/java/redis/clients/jedis/ShardedJedisPipeline.java
@@ -5,6 +5,8 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Queue;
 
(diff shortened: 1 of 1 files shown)


sazzad16:
@QooYoung Change looks good. But it would be much more helpful with proper tests. Would you be able to add some tests?

yangbodong22011:
@sazzad16 Does this PR can slove this question https://github.com/redis/jedis/issues/2488#issue-839666322 which you metioned.

sazzad16:
@yangbodong22011 No.

> an issue related to concerned class

I meant that this PR approaches an issue related to `ShardedJedisPipeline` class. And that issue is not even listed in my comment. This signifies how much flaws ShardedJedisPipeline class contains.

PR #1801, opened 2018-04-13
Fix: wrong list commands test case name

tramchamploo:
Fix wrong names for `lset` and `lindex` test cases.


sazzad16:
@tramchamploo Nice catch and thanks for your contribution!

sazzad16:

PR #1809, opened 2018-05-06
Don't auth if password is empty as REDIS doesn't support it

### LINKED ISSUE #1808: If this.password is empty, it should not be auth
https://github.com/xetorthio/jedis/blob/df040537f47c8bf51ed66d16007c2a3f9f1bb802/src/main/java/redis/clients/jedis/JedisFactory.java#L118

marcosnils:
Fixes #1808


marcosnils:
Just double checked and Redis actually **allows** you to start a server with an empty password and authenticate against it. 

```
redis-server - <<< "requirepass \"\""
redis-cli
127.0.0.1:6379> auth ""
OK
127.0.0.1:6379> 
```

PR #1814, opened 2018-05-15
log the error for redirection of request

jiafu1115:
current, when moved/asked happen (such as fail over happen), no any message hint the event. 
so suggest log it as warn to show the event.

test result example:
WARNING: should request to : 127.0.0.1:7380

BTW： the tests are not stable.


sazzad16:
@jiafu1115 In case of cluster reconfiguration, node repair like scenarios user can be bombarded with this kind of logs. These scenarios may be rare but expected. So, IMHO, we should not log these exceptions.

sazzad16:
@marcosnils WDYT?

jiafu1115:
node repair should only happen in control limit time period. it shouldn't take too much time. Take normal case, for example, I take manual fail over and no any message for the application to know the event. so I raise this PR.

sazzad16:
@marcosnils @HeartSaVioR Close?

jiafu1115:
@sazzad16 change to fine from warning to let user have more choice to decide if track the change.

jiafu1115:
will open new one due to the PR is to older branch.

Issue #1817, opened 2018-05-30
jedis not support rename-command


fuyuanpai:
For safety reason, we have rename some dangerous command like flushdb, eval in case any malicious client call this command.
But jedis can not call the rename command, will jedis add support for rename command?

sazzad16:
@fuyuanpai You are very much welcome to craft PR.

gkorland:
@sazzad16 please review my PR

sazzad16:
@gkorland Which PR? If you mean #1822 , I am actually waiting for opinions from someone else.

@fuyuanpai Does #1822 fulfills your requirement?

fuyuanpai:
Thanks for @gkorland 
@sazzad16  #1822 meets my requirements

sazzad16:
@gkorland Please see my review comments.

sazzad16:
Rename command is now deprecated https://github.com/redis/redis/commit/7604ab7118d1154e9120ea41a88d9c214f2202c3 and #1822 is closed.

PR #1787, opened 2018-03-15
 Fix #1780 to replace Jedis, JedisPool, JedisCluster constructor

warrenzhu25:

PR #1822, opened 2018-06-18
Support rename-command (#1817)

gkorland:


gkorland:
It doesn't seem like the test failure is related to the PR

src/test/java/redis/clients/jedis/tests/commands/RenameProtocolCommandsTest.java
```diff
@@ -0,0 +1,51 @@
...
+	public void renameCommand() {
+		jedis.set("mykey", "hello world");
+		String value = jedis.get("mykey");
+		assertEquals("hello world", value);
+	}
+
+	@Test(expected = JedisDataException.class)
+	public void disabledCommand() {
+		jedis.set("mykey", "hello world");
+		jedis.del("mykey");
```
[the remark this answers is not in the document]
sazzad16: You can also user exists or get at the end to make it more tangible.

src/test/java/redis/clients/jedis/tests/commands/RenameProtocolCommandsTest.java
```diff
@@ -0,0 +1,59 @@
...
+    String value = jedis.get("mykey");
+    assertEquals("hello world", value);
+  }
+
+  @Test
+  public void disabledCommand() {
+    jedis.set("mykey", "hello world2");
+    try {
+      jedis.del("mykey");
+      fail();
```
sazzad16: Add a message.

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -259,16 +259,20 @@ public static Object read(final RedisInputStream is) {
...
     Command() {
-      raw = SafeEncoder.encode(this.name());
+    	rename(this.name());
     }
 
     @Override
     public byte[] getRaw() {
       return raw;
     }
+
```
sazzad16: Remove spaces to make an empty line.

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -259,16 +259,20 @@ public static Object read(final RedisInputStream is) {
...
+    	rename(this.name());
     }
 
     @Override
     public byte[] getRaw() {
       return raw;
     }
+    
+    public void rename(final String newName) {
+    	raw = SafeEncoder.encode(newName);
```
sazzad16: Replace tab character.

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -259,16 +259,20 @@ public static Object read(final RedisInputStream is) {
     PFADD, PFCOUNT, PFMERGE, READONLY, GEOADD, GEODIST, GEOHASH, GEOPOS, GEORADIUS,
     GEORADIUSBYMEMBER, MODULE, BITFIELD, HSTRLEN, TOUCH, SWAPDB;
 
-    private final byte[] raw;
+    private byte[] raw;
 
     Command() {
-      raw = SafeEncoder.encode(this.name());
+    	rename(this.name());
```
sazzad16: Replace tab character.

src/test/java/redis/clients/jedis/tests/commands/RenameProtocolCommandsTest.java
```diff
@@ -0,0 +1,59 @@
...
+    String value = jedis.get("mykey");
+    assertEquals("hello world", value);
+  }
+
+  @Test
+  public void disabledCommand() {
+    jedis.set("mykey", "hello world2");
+    try {
+      jedis.del("mykey");
+      fail("JedisDataException expected");
```
sazzad16: Message can be more self explanatory, something like `DEL command should be unknown`

sazzad16:
LGTM!

@marcosnils @HeartSaVioR

marcosnils:
What if we also add a `rename` method to the `Jedis` class which takes a `Command` and a `String` as a rename?. The reason why I'm suggesting this is because from the API perspective Jedis users will expect to see the `rename` method. This is how @mp911de did it in lettuce also (https://github.com/lettuce-io/lettuce-core/blob/dde37483a31215ebf17f7141fea6454971ea9a08/src/test/java/io/lettuce/core/commands/KeyCommandTest.java#L212-L214)

Thoughts?

sazzad16:
@marcosnils Your provided link is about [**rename**](https://redis.io/commands/rename) command which is already implemented in `Jedis`

https://github.com/xetorthio/jedis/blob/d1fb17ce3021f8bac135aef3fd306d62c877c13c/src/main/java/redis/clients/jedis/Jedis.java#L304-L308

But this PR is about supporting [**rename-command**](https://redis.io/topics/security#disabling-of-specific-commands) of _Redis Security_.

marcosnils:
@smadasu you're right. How about if we add the `renameCommand` or something similar to the Jedis API. 
I'm not entirely sure if doing `Protocol.<CMD>.rename("something")`  is the cleanest way from the API perspective.

sazzad16:
@marcosnils 

> How about if we add the renameCommand or something similar to the Jedis API.

1. What we have in Jedis/BinaryJedis/JedisCluster etc are [Redis commands](https://redis.io/commands/). But `rename-command` is not a Redis command.

2. Unlike Redis commands, rename-command is done by configuration file.

> I'm not entirely sure if doing Protocol.<CMD>.rename("something") is the cleanest way from the API perspective.

I'm also concerned about that. But I couldn't find another way, let alone a better one, to support this feature.

I can see only two options regarding this. Either use an available solution or leave that feature out from supporting.

marcosnils:
> ename-command shouldn't be part of the API since as @sazzad16 mentioned a command is renamed using the Redis configuration file. Meaning, a user shouldn't really rename methods in runtime.

Right. Still the `Protocol.<CMD>.rename("something")` doesn't convince me. Can we try to look for an alternative for this?. Maybe check how / if other libraries are doing it?

gkorland:
@marcosnils [node_redis](https://github.com/NodeRedis/node_redis) is doing it on client initialization

`
redis.createClient([options])
`

`
*rename_commands* -
Passing an object with renamed commands to use instead of the original functions. For example, if you renamed the command KEYS to "DO-NOT-USE" then the rename_commands object would be: { KEYS : "DO-NOT-USE" } . 
`

gkorland:
Same with .NET client [StackExchange.Redis](https://stackexchange.github.io/StackExchange.Redis/Configuration)


```
var commands = new Dictionary<string,string> {
        { "info", null }, // disabled
        { "select", "use" }, // renamed to SQL equivalent for some reason
};

var options = new ConfigurationOptions {
    // ...
    CommandMap = CommandMap.Create(commands),
    // ...
}
```

gkorland:
@marcosnils I don't think our current design fit adding it constructor settings

marcosnils:
Maybe we can resurrect https://github.com/xetorthio/jedis/pull/1787 and add it there? @gkorland WDYT?

sazzad16:
@marcosnils Don't know if it's relevant, but a note: currently, there is a way to alter localhost in Jedis which is done without constructor.

https://github.com/xetorthio/jedis/blob/d1fb17ce3021f8bac135aef3fd306d62c877c13c/src/main/java/redis/clients/jedis/HostAndPort.java#L120-L124

gkorland:
@marcosnils I don't think #1787 will help here unless we also moving from Enums and make a bigger redesign. It's probably cleaner to do it as part of the constructor but are we ready to make this refactor?

As @sazzad16 mentioned I think this pattern is already being used in the product for another rare use case so I think it can be a good compromise.

marcosnils:
> As @sazzad16 mentioned I think this pattern is already being used in the product for another rare use case so I think it can be a good compromise.

I'm not against this change, I'm just trying to evaluate all other alternatives as I'm not 100% convinced about this approach. As you're saying, the `HostAndPort` case is *rare*. In this case, the `rename` feature is something common to Redis / Jedis and there might be some other things like this in the future. 

I guess we could move forward with this approach and refactor it later in the future, but taking the opportunity that we'll be probably releasing 3.0 next, I'd like to take advantage of the interface breaking changes and do it the right way.

marcosnils:
Additionally, I just thought about the current use-case:

What if you have two Jedis instances pointing to different redis DB's that are configured differently?. This PR basically forces that all redis instances to have the exact same rename configuration or it won't be possible to use Jedis at all. I believe this limitation is non-trivial and now that I think about it I'm almost sure that this is definitely not right.

gkorland:
@marcosnils when do you expect 3.0 will be released? Will it include #1787 ?

sazzad16:
@gkorland Want to go ahead with this one?

marcosnils:
Shall we close this?. We agreed that the implementation is not the desired one.

jaipreet-chhatwal:
@gkorland  Have we finalized a version in which jedis support for rename-command will be rolled out.

gkorland:
@jaipreet-chhatwal no, we didn't get to an agreement how it should be done

jaipreet-chhatwal:
@gkorland  As a part of security assessment, many financial institutions have mandated to rename redis commands in Sentinel and Redis server. As expected post changes not able to access commands via Jedis.
I think this makes this feature much more mandatory and is a make-break scenario for using Jedis because of this concern.

Having said, will you be able to review my changes to enable this feature.

mp911de:
> As a part of security assessment, many financial institutions have mandated to rename redis commands in Sentinel and Redis server. As expected post changes not able to access commands 

Why should a client library compensate for something that is pointless in the first place? The whole point of renaming commands is making sure they do not get invoked. Putting renamed commands back into place circumvents the renaming effort.

If you need to invoke a renamed command for whatever reason, then I suggest using the `execute()` API that accepts a command name.

gkorland:
@mp911de, @jaipreet-chhatwal comment sounds like a reasonable reason especially with the last noise around Redis "security" issues

gkorland:
@marcosnils @sazzad16 should we keep my original naive solution?
Should not impact any future solution we might implement in the future on the `Jedis` level.

It feels like since most of the applications (EJB, Classloader, ApplicationContext or OSGI) usually connect to a single Redis it should be good enough, while not impacting the performance.

Makefile
```diff
@@ -91,10 +91,14 @@ define REDIS8_CONF
 daemonize yes
 protected-mode no
 port 6386
+requirepass foobared
```
sazzad16: Test requires instance at 6386 port to be password-less. Either create a new instance or create a new one.

sazzad16:
@gkorland I vote to include this feature. IMO, adding solution for one configuration is far better than not having solution for multiple configuration.

And I'm okay with your implementation. Not to mention, we've waited a year for an alternate solution.

Makefile
gkorland: "a new instance or create a new one"? what is the different between "new instance" and "new one"?

sazzad16: Oops!

"Either use password-less instance or create a new one."

jaipreet-chhatwal:
@gkorland @sazzad16  :  When we configure Redis with Sentinel for HA, arent we suppoosed to configure the same Redis password in every instance. Isn't this expected already that we need to have same configuration across nodes. 
So i am not able to understand the problem with 
"**adding solution for one configuration is far better than not having solution for multiple configuration." since we are already** "
Please correct me if i am missing a very basic point and not on same page with you guys.

Also, can we conclude on the version in which this rename-command feature will be supported.

sazzad16:
@jaipreet-chhatwal 

> Isn't this expected already that we need to have same configuration across nodes.

**Yes and No.**

Within same sentinel, it is expected that the nodes within that sentinel are of same configuration. For example, they are expected to have same password.

But Jedis is not limited to connect with only one Redis instance/sentinel/cluster. It allows you to connect with many of them at once. So you can set up two different sentinels (two different group of nodes, say A and B) and connect with them simultaneously within same application using Jedis. Those two set of sentinel nodes are not required to have same configuration. For example, both set of nodes A and B are not required to have same password, even though all nodes of sentinel A are expected to have same password as well as all nodes of sentinel B are expected to have same password.

I hope this clears your confusion about my statement.

sazzad16:
@jaipreet-chhatwal 

> can we conclude on the version in which this rename-command feature will be supported.

Not yet.

gkorland:
@sazzad16 if we agree this solution is better than nothing, let's mark it as 3.2.0?

sazzad16:
@gkorland sure thing!

@marcosnils It has been a year. Humbly speaking, we have waited enough for an alternate solution.

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -261,16 +261,20 @@ public static Object read(final RedisInputStream is) {
...
+      rename(this.name());
     }
 
     @Override
     public byte[] getRaw() {
       return raw;
     }
+
+    public void rename(final String newName) {
+      raw = SafeEncoder.encode(newName);
```
sazzad16: Couldn't this be an issue in multithreading? Synchronized/Lock may be required here.

gkorland: You're right, but volatile should do, I'm trying to find a way to avoid the performance penalty on each getRaw() call

src/test/java/redis/clients/jedis/tests/commands/RenameProtocolCommandsTest.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  @Test
+  public void disabledCommand() {
+    jedis.set("mykey", "hello world2");
+    try {
+      jedis.del("mykey");
+      fail("'DEL' command should be unknown");
+    } catch (JedisDataException expected) {
+      assertEquals("ERR unknown command 'DEL'", expected.getMessage());
```
yangbodong22011, line 54:
```suggestion
            Assert.assertTrue(expected.getMessage().contains("ERR unknown command `DEL`"));
```

src/main/java/redis/clients/jedis/Protocol.java
mina-asham: * +1 to volatile (or AtomicReference), at the moment it's not though, so we need to update that to either volatile or AtomicReference.
* We should make this fail (and have tests for) if `newName` is:
  * null
  * empty
  * an existing command name

mina-asham (changes requested):

sazzad16:
@gkorland Command renaming is deprecated in Redis. See https://github.com/redis/redis/commit/7604ab7118d1154e9120ea41a88d9c214f2202c3
Should we close this PR? Or, keep this PR open for discussion?

gkorland:
Lets close

Issue #1819, opened 2018-06-05
JedisSentinelPool is not thread safe. It can create sentinel connection leak


khemkasudeep:
### Expected behavior

JedisSentinelPool() should create connection to sentinels in MasterListener thread.
Then pool.close() should always close that connection

JedisSentinelPool pool = new JedisSentinelPool()
pool.close()

### Actual behavior

There is a race around condition where connection is not closed some times.

### Steps to reproduce:

for(i=0; i<10000;i++){
    JedisSentinelPool pool = new JedisSentinelPool();
    pool.close();
}

check number of client connections for your sentinel hosts.
It should shootup.

alternatively, check how many threads are waiting on MasterListener as below -
jstack pid | grep "MasterListener-mastername-" | wc -l

### Reason:

This is a race around condition in  [JedisSentinelPool](https://github.com/xetorthio/jedis/blob/master/src/main/java/redis/clients/jedis/JedisSentinelPool.java#L241) class.

This can happen if the code execution happens in the order mentioned below. 

JedisSentinelPool pool = new JedisSentinelPool();    -->  MasterListener().start()  -->  MasterListener.run --> j = new Jedis(host, port);  --> if (!running.get()){}   --> pool.close(); -> MasterListener. shutdown -> j.disconnect();  -->  j.subscribe();

Basically here creation of MasterListener and its shutdown are interleaved. If we create  MasterListener and say the thread is running but has not subscribed for messages with sentinel yet and in the meanwhile if someone one calls shutdown, caller would expect that MasterListener thread should close sentinel connections and eventually stop running. 

But instead after shutdown has executed, MasterListener can go and call subscribe which will open a connection to sentinel and will keep listening on it for ever.

gkorland:
See https://github.com/xetorthio/jedis/issues/1910

khemkasudeep:
hi @gkorland  1910 is not same as this issue. this issue is only related to shutdown of JedisSentinelPool

akshay12390:
@khemkasudeep we also faced same issue in our application recently. We are using version 2.9.0.  Is the fix for this issue available in later versions now?

khemkasudeep:
@akshay12390  This fix is not merged. I didn't received response on this thread. 
However I have tested this change and it works perfectly.

khemkasudeep:
@akshay12390  can you describe the problem you faced?

akshay12390:
[...]

PR #1825, opened 2018-06-20
Make JedisSentinelPool MasterListener shutdown thread safe. (#1819)

khemkasudeep:
https://github.com/xetorthio/jedis/issues/1819

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisSentinelPool.java
+++ b/src/main/java/redis/clients/jedis/JedisSentinelPool.java
@@ -5,6 +5,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.ReentrantLock;
 
 import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
 import org.slf4j.Logger;
@@ -245,7 +246,8 @@ public class JedisSentinelPool extends JedisPoolAbstract {
     protected int port;
     protected long subscribeRetryWaitTimeMillis = 5000;
     protected volatile Jedis j;
-    protected AtomicBoolean running = new AtomicBoolean(false);
+    protected AtomicBoolean isShutdown = new AtomicBoolean(false);
+    protected final ReentrantLock lock = new ReentrantLock();
 
     protected MasterListener() {
     }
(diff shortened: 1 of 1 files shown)


src/main/java/redis/clients/jedis/JedisSentinelPool.java
```diff
@@ -315,7 +317,12 @@ public void onMessage(String channel, String message) {
 
         } catch (JedisException e) {
 
-          if (running.get()) {
+          try {
+            lock.unlock();
```
gkorland: In this case we'll always get to:
line 341: log.debug("lock is not held by MasterListener");

src/main/java/redis/clients/jedis/JedisSentinelPool.java
```diff
@@ -315,7 +317,12 @@ public void onMessage(String channel, String message) {
...
+          }catch (IllegalMonitorStateException ex){
+            log.debug("lock is not held by MasterListener");
+          }
+          if (!isShutdown.get()) {
```
gkorland: Why do you need to check this? We had the lock until line 321 so obviously the isShutdown was false. And anyway by the the time you check it on line 325 to the log on line 326 it can be set to true.

src/main/java/redis/clients/jedis/JedisSentinelPool.java
```diff
@@ -315,7 +317,12 @@ public void onMessage(String channel, String message) {
 
         } catch (JedisException e) {
 
-          if (running.get()) {
+          try {
+            lock.unlock();
+          }catch (IllegalMonitorStateException ex){
```
gkorland: How can you get IllegalMonitorStateException? The lock is to be held by the current Thread at this stage.

khemkasudeep: @gkorland Ideally the lock won't be held till this stage. It should be released at line 290.
We want to release the lock before starting to read from sentinel socket which is blocking call. 

This is necessary to be able to shutdown JedisSentinelPool or else if someone tries to shutdown, it will block on MasterListener thread, possibly for ever.

src/main/java/redis/clients/jedis/JedisSentinelPool.java
khemkasudeep: I think we need to try release lock at both the places.
Here at line 321, we need to release because after this we sleep for 5 seconds. I don't want anyone trying to shutdown, wait for it. 

The reason we need to shutdown in finally (line 339) is if there is some unchecked exception which we are not catching, it can result in lock being never released. In that case anyone calling shutdown will be blocked for ever

khemkasudeep:
Hi @sazzad16 
Will you be able to look at it.

uglide:

PR #1829, opened 2018-07-17
Replace long else-if with switch-case

gkorland:


sazzad16:
LGTM!

@marcosnils

marcosnils:
LGTM!.

PR #1831, opened 2018-07-19
Update some comment in Jedis.java

wxisme:
In Java, 1 and true, 0 and false are not equivalent. And in the comments of other methods whose return value is Boolean, use true and false.

### DIFF
--- a/src/main/java/redis/clients/jedis/Jedis.java
+++ b/src/main/java/redis/clients/jedis/Jedis.java
@@ -452,7 +452,7 @@ public class Jedis extends BinaryJedis implements JedisCommands, MultiKeyCommand
   }
 
   /**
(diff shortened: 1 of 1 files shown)


gkorland:
LGTM!

sazzad16:
@wxisme Similar change is also applicable for `exists` method. Moreover same issues can be found in `BinaryJedis`. Would you care to fix those?

wxisme:
@sazzad16 Yes, I have committed it.

wxisme:
@sazzad16 I have fixed `dont`. thanks

sazzad16:
LGTM!

@marcosnils @HeartSaVioR

sazzad16:
@marcosnils @HeartSaVioR As these are minor changes in comments, I've set READY TO MERGE directly.

PR #1834, opened 2018-07-26
cleanup jedis client move get and ping from sendCommand to specific

gkorland:
remove from jedis.java direct calls to client.sendCommand for GET and PING and move these methods to use the standard way like the rest of the commands


sazzad16:
LGTM!

@marcosnils @HeartSaVioR

marcosnils:
LGTM!.

PR #1836, opened 2018-07-31
Fixed some package importing

wxisme:
I checked the code files and fixed some package importing.


marcosnils:
LGTM!. The thing about this PR's is that are super hard to downmerge. I'll set it as a 3.0 milestone if you agree @sazzad16

sazzad16:
@marcosnils agreed!

PR #1838, opened 2018-08-01
Add volatile for localhost in HostAndPort.java

wxisme:
The method `getLocalhost()` in HostAndPort.java uses a double check lock, but `localhost` is not a `volatile` variable.


wxisme:
There is a problem with this PR, I will resubmit it.

PR #1843, opened 2018-08-10
CLIENT PAUSE command implementation

wxisme:
CLIENT PAUSE command implementation and test case.


wxisme:
I'll re-open the PR as it contains changes which are out of the scope.

PR #1847, opened 2018-08-13
CLIENT PAUSE command implementation

wxisme:
CLIENT PAUSE command implementation and test case.


sazzad16:
Merged into master, 2.10.

PR #1857, opened 2018-08-20
replace slots cache atomically

kutzi:
See https://github.com/xetorthio/jedis/issues/1855


kutzi:
I just realised that this won't fix the most common source of this error (which I have observed): directly after application startup when the slots cache hasn't been initialised, yet

uglide:

PR #1862, opened 2018-09-02
initial commit for gradle support

smadasu:
Adding gradle support along with gradle wrapper


marcosnils:
Hey @smadasu thx for the contribution. I'm curious about the necessity to add gradle builds to Jedis. 

Additionally, seems like gradle-wrapper.jar is missing in order for this to work, right?

smadasu:
Yeah. That's true. gradle wrapper jar is missing. Let me send it through. Sorry

smadasu:
The necessity to add gradle is that some developers may wish to use gradle in stead of maven.

smadasu:
I just realized the gradle test will have to start redis instances. Because of this the tests might fail. WIll have to get back on that

marcosnils:
> I just realized the gradle test will have to start redis instances. Because of this the tests might fail. WIll have to get back on that

I guess you can do `make start && gradle test`. 

> The necessity to add gradle is that some developers may wish to use gradle instead of maven.

I haven't seen many (or any?) issues about gradle support requests. I usually try to avoid adding new stuff unless it's required. In this particular situation given the low activity on the project, I'm just wondering how useful this would be if it's you who it'll be using it mainly.

smadasu:
Well. OK. If there is not much activity We don't have to do this.

marcosnils:
> Well. OK. If there is not much activity We don't have to do this.

Don't get me wrong, I do appreciate the effort, it's just that maybe it's not necessary if nobody besides you will be using it.

smadasu:
That's fine. I am OK with it. I don't to push changes in if nobody else is using them.

Issue #1863, opened 2018-09-03
JDK8 discussion


smadasu:
Switching the version to 1.8 is trivial in pom.xml. What else is preventing us to do that?

marcosnils:
I guess nothing. Wanna give it a shot.

smadasu:
Sure.

Martian101:
I encountered this error when using `public String set(final String key, final String value, final SetParams params)` method in JDK 1.8.0_111:
```java
java.lang.ClassCastException: java.lang.Integer cannot be cast to [C

	at com.gexin.clients.jedis.params.set.SetParams.getByteParams(SetParams.java:75)
	at com.gexin.clients.jedis.BinaryClient.set(BinaryClient.java:106)
	at com.gexin.clients.jedis.Client.set(Client.java:53)
	at com.gexin.clients.jedis.Jedis.set(Jedis.java:136)
```
It seems that a type inference error occured when the type parameter is used in the return type:
```java
  public <T> T getParam(String name) {
    if (params == null) return null;

    return (T) params.get(name);
  }
```
I can ensure this issue after some test both in JDK 1.7 and 1.8:
``` java
    @Test
    public void testT() {
        System.out.println(String.valueOf(getParam()));
    }

    public static <T> T getParam() {
        return (T) "AAA";
    }
```

JensenZz:
@Martian101   maybe you can find your answer      https://stackoverflow.com/questions/23155907/classcast-error-java-7-vs-java-8

sazzad16:
Switching to JDK1.8 will be required if we want to upgrade Commons Pool to version `2.7.0`.

References:
- #2039 
- #2040

sazzad16:
Resolved by #2396

PR #1864, opened 2018-09-09
Upgrade JDK to 8 refer #1863

smadasu:


smadasu:
This build fails and succeeds intermittently on my local sand box. I don't know why

src/main/java/redis/clients/jedis/params/SetParams.java
```diff
@@ -71,11 +71,13 @@ public SetParams xx() {
 
     if (contains(EX)) {
       byteParams.add(SafeEncoder.encode(EX));
-      byteParams.add(SafeEncoder.encode(String.valueOf(getParam(EX))));
+      final Object param = getParam(EX);
+      byteParams.add(SafeEncoder.encode(String.valueOf(param)));
```
gkorland: How are these changes related to JDK 8?

sazzad16:
@gkorland About https://github.com/xetorthio/jedis/pull/1864#pullrequestreview-173921091

Those changes are related to JDK 8. The [job log](https://travis-ci.org/xetorthio/jedis/jobs/426423621) contains related exceptions. Here https://stackoverflow.com/questions/23155907/classcast-error-java-7-vs-java-8 are some details about the case.

However, IMHO, we should come up with a generic solution for these situations. Otherwise, there would be redundant codes over and over again.

@marcosnils ping!

src/main/java/redis/clients/jedis/params/SetParams.java
sazzad16: Resolved in #2019

sazzad16:
Added same milestone as #2040

sazzad16:
Resolved by #2396

PR #1875, opened 2018-10-11
Fix JedisClusterInfoCache

thirstycrow:
JedisCluster gets into an unrecoverable state,  if it failed to get cluster nodes from any of the start nodes at initialization time. Because the start nodes were 'forgotten' after initialization, JedisClusterInfoCache keeps trying to renew the slot map, but no node is available, as a result JedisNoReachableClusterNodeException is kept being thrown until the application restarts.


sazzad16:
Resolved by #2643

Issue #1876, opened 2018-10-13
Useless retry when jedicluster configured password and try to connect to a node without password


StillerRen:
### Expected behavior
do special handling of JedisDataException: "ERR Client sent AUTH, but no password is set"，do not retry to connect to the node with <attemps> times.
This exception means that no normal connection can be established no matter how many times you try again.

### Actual behavior
when I create a instance of JedisCluster **with password**, and try to connect to a node **without password**, it would report the exception JedisDataException: "ERR Client sent AUTH, but no password is set",  and then the exception is packaged as JedisConnectionException: "Could not get a resource from the pool", JedisClusterCommand would catch this exception and retry <attempts> times.

      // release current connection before recursion
      releaseConnection(connection);
      connection = null;
      if (attempts <= 1) {
        //We need this because if node is not reachable anymore - we need to finally initiate slots renewing,
        //or we can stuck with cluster state without one node in opposite case.
        //But now if maxAttempts = 1 or 2 we will do it too often. For each time-outed request.
        //TODO make tracking of successful/unsuccessful operations for node - do renewing only
        //if there were no successful responses from this node last few seconds
        this.connectionHandler.renewSlotCache();

        //no more redirections left, throw original exception, not JedisClusterMaxRedirectionsException, because it's not MOVED situation
        throw jce;
      }

      return runWithRetries(key, attempts - 1, tryRandomNode, asking);
    }

   at the last time, it would call **renewSlotCache()** to renew the slot message.

Default retries is 5 times, and it would call "Cluster" command, all of these would spend a lot of time.

when some nodes in cluster have password and the other don't, when I use the JedisCluster configured password to send a request to the node without password, it would cause client latency increasing sharply.

I think we need to do special handling of JedisDataException: "ERR Client sent AUTH, but no password is set", and report the exception directly.

### Steps to reproduce:

Please create a reproducible case of your problem. Make sure
that case repeats consistently and it's not random
1. create a cluster that some nodes in cluster have password and the other don't
2. create a JedisCluster instance with password
3. send request to the cluster

#### Jedis version:
2.9.0
#### Redis version:
4.0.11
#### Java version:
java8

gkorland:
@StillerRen sounds like a nice improvement, do you care to submit a PR?

sazzad16:
[...]

PR #1899, opened 2018-11-22
Special handling for jedis exception

StillerRen:
special handing for the issue: [#1876](https://github.com/xetorthio/jedis/issues/1876)


gkorland:
It seems like a backward issue

marcosnils:
NACK!. 

Why are we wrapping the exception into a JedisConnectionException?. Shouldn't just we  let it fail and let the caller catch the corresponding exception?. We could also apply it for 3.0 as it breaks backwards compatibility.
@smadasu @gkorland ?.

PR #1902, opened 2018-11-23
Bug fix in JedisCluster scan

### LINKED ISSUE #1901: Scan not supporting hash tag
https://github.com/xetorthio/jedis/blob/53194dd03bf397211566ba48a60e9f1e549c85ae/src/main/java/redis/clients/jedis/JedisCluster.java#L1337

Shouldn't there be a NOT (!) sign for this? Looks like this is saying if the matchPattern is cluster compliant, then throw error, whereas it should be otherway.

I'm currently trying to scan for keys with match pattern of "crossing:{123}-*" (without double quotes), and it's throwing the exception from line 1338 of JedisCluster.

sazzad16:
Closes #1901

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisCluster.java
+++ b/src/main/java/redis/clients/jedis/JedisCluster.java
@@ -1334,7 +1334,7 @@ public class JedisCluster extends BinaryJedisCluster implements JedisClusterComm
           + " only supports SCAN commands with non-empty MATCH patterns");
     }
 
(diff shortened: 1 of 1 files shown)


marcosnils:
LGTM!>

sazzad16:

marcosnils:
Just realize this doesn't have a test. Shouldn't we add one?

marcosnils:
Yes, I've seen those but maybe in this case it's work to add a JedisCluster test specifically to test that condition.

PR #1909, opened 2018-12-02
add publish plugins

gkorland:


gkorland:
@marcosnils any objection? I'm used this pom.xml to deploy all the versions lately

PR #1911, opened 2018-12-10
fix #1910 protect initPool from being called in parallel

### LINKED ISSUE #1910: Race condition in JedisSentinelPool.initPool() causes JedisConnectionException on getResource()
JedisSentinelPool.initPool() is called concurrently by multiple threads during statup: The thread creating `JedisSentinelPool` and `MasterListener`. The consequence is that calls to `getResource()` fail with an exception.

Reproduction of the issue requires just the right timing so calling the offending code in a `while(true)` loop allows reproduction instead of single-shot execution.

Jedis version: 3.0.0

```
redis.clients.jedis.exceptions.JedisConnectionException: Could not get a resource from the pool

	at redis.clients.jedis.util.Pool.getResource(Pool.java:59)
	at redis.clients.jedis.JedisSentinelPool.getResource(JedisSentinelPool.java:209)
	at org.springframework.data.redis.connection.jedis.Reproducer.reproduceSentinelPoolRaceCondition(Reproducer.java:43)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)
Caused by: java.lang.NullPointerException
	at redis.clients.jedis.util.Pool.getResource(Pool.java:50)
	... 24 more
```

Reproducer

```java  
public class Reproducer {

  @Test
  public void reproduceSentinelPoolRaceCondition() {

    while (true) {

      GenericObjectPoolConfig poolConfig = new JedisPoolConfig();
      JedisSentinelPool jedisSentinelPool = new JedisSentinelPool("mymaster",
          new HashSet<>(Arrays.asList("127.0.0.1:26379", "127.0.0.1:26380")), 
          poolConfig, 1000, 1000, null, 0,
          "jedis-client");

      jedisSentinelPool.getResource().close();

      jedisSentinelPool.destroy();
    }
  }
}
```

gkorland:


sazzad16:
@gkorland 
1. Did you run your patch against the _Reproducer_ provided in #1910?
2. Does your test without the patch reproduce #1910's issue?

mp911de:
I think there are two issues:

1. Order of invocation
2. Synchronization/visibility issues

* `initPool(…)` in `JedisSentinelPool`  should be called before starting any threads
* The pool should not be set by any thread but only-re-initialized to avoid visibility issues.

gkorland:
> * Did you run your patch against the _Reproducer_ provided in #1910?
@sazzad16 yes, and I even added similar test

> * Does your test without the patch reproduce #1910's issue?
Yes

sazzad16:
LGTM!

gkorland:
@mp911de
> 1. Order of invocation
You are right, I guess a cleaner solution should be released later on, I think we should release a quick fix for 3.0.1 and then work on a cleaner solution for 3.1.0

> 2. Synchronization/visibility issues
See commit ed2ed51

gkorland:
@sazzad16 no, the former used this as the object to lock on which means it "exposed" the lock object while the later is using a private internal object

gkorland:
@marcosnils @sazzad16 it seems like this bug is too critical and a fix should be released as soon as possible

sazzad16:
@gkorland okay. go ahead

PR #1916, opened 2018-12-17
Skip nodes which lacks of slots during slot cache initialization

yu74n:
This PR adds a strict checking for getting slot info from the node without slots.
In master promotion, for example, a user might leave accidentally a old master/slave node after initialized in a nodes list of a cluster.
When a nodes list of a cluster includes a node without slots, skip the node and continue finding other valid nodes and use it.


sazzad16:
During a master down scenario, this will not allow to create a JedisCluster object; (in case of Jedis 3.x, an operable JedisCluster object).

A proper solution to #2503 may help in this regard.

PR #1917, opened 2018-12-17
update supported versions

gkorland:


gkorland:
@sazzad16 comment added

PR #1918, opened 2018-12-17
Move dataSource reset before connection returned

gkorland:
Move dataSource reset before connection returned to pool.
If datasource reset after the connection returned to pool, it might reset the datasource after it was allocated to another thread


kutzi:
In the absence of any memory barrier on the `dataSource` field this fixes IMHO nothing.
See https://github.com/xetorthio/jedis/commit/02f2cc5cce2f44efeaaafe351a2facf66988ddbf#r31795161

PR #1923, opened 2019-01-05
Throw an exception when trying to read from a broken connection

mina-asham:
- When attempting to read from a broken connection, Jedis could potentially try to read some corrupt data and interpret it according to the Redis protocol, this makes it throw an exception instead of trying that.
- https://github.com/xetorthio/jedis/issues/1747 describes that possible misinterpretations of the protocol can lead to trying to create very large arrays
- Also refactor clientPause test, it was testing that clients fail, CLIENT PAUSE command pauses all clients, so they should be expected to be delayed rather than failed


mina-asham:
ping, it's quite a small diff...

mina-asham:
Another ping... @marcosnils @gkorland

marcosnils:
LGTM. @smadasu I've changed it to 3.0.2 as it's not adding anything new features.

sazzad16:
@marcosnils Sure!

BTW, It's @sazzad16, not @smadasu ;)

suryavtkt:
what does it mean broken connection, what are the possibilities to broke a connection

PR #1925, opened 2019-01-07
Update versions of dependencies

sazzad16:


sazzad16:
@marcosnils @gkorland I think we should update our dependencies. I'm not 100% about `commons-pool2` though.

gkorland:
@sazzad16 I think we were waiting for commons-pool2 2.6.1?

sazzad16:
@gkorland Yes, we are waiting for commons-pool2 2.6.1. That's I've expressed my concern over it. However, we're waiting for quite some time and 2.6.0 is the latest one we can use right now.

Keeping commons-pool2 aside, I think we should consider updating other dependencies. Our log dependencies are old and JUnit is very old.

gkorland:
@sazzad16 the only thing is that I don't think we should have it in 3.0.2 but in 3.1.0

sazzad16:
@gkorland Okay. done.

marcosnils:
LGTM!.

PR #1928, opened 2019-01-09
Fixed the logic that prevents all cluster slot requests sent to specific instances

petpetg:
Fixed the logic that prevents all cluster slot requests sent to specific instances when jedis is initialized on many processes(such as storm with a lot of works).
It can cause high CPU and the impact of many slow queries.

Please review.
thanks.


petpetg:
When the amount of requests is large, it will result in high CPU and a lot of slow queries.

petpetg:
When a storm application with 200 workers is started, the CPU utilization of one instance exceeds 20%, while that of the other instances is only 2%. At the same time, there will be a lot of requests blocked and slow queries.

petpetg:
Hi @sazzad16 
    This changed was fixed the logic that prevents all cluster slot requests sent to specific instances when jedis is initialized on many processes(such as storm with a lot of works). It can cause high CPU and the impact of many slow queries.
    When a storm application with 200 workers is started, the CPU utilization of one instance exceeds 20%, while that of the other instances is only 2%. At the same time, there will be a lot of requests blocked and slow queries.

Please review.
thanks.

sazzad16:
@petpetg I understand and I already reviewed. Please address what was said in the review. Thx!

petpetg:
@sazzad16  Done. Thx!

dengliming (dismissed):

src/main/java/redis/clients/jedis/JedisClusterConnectionHandler.java
```diff
@@ -1,6 +1,9 @@
 package redis.clients.jedis;
 
 import java.io.Closeable;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
```
sazzad16, line 6:
```suggestion
```

src/main/java/redis/clients/jedis/JedisClusterConnectionHandler.java
```diff
@@ -117,8 +120,10 @@ public Jedis getConnectionFromNode(HostAndPort node) {
   }
 
   private void initializeSlotsCache(Set<HostAndPort> startNodes, JedisClientConfig clientConfig) {
+    List<HostAndPort> startNodeList = new ArrayList<>(startNodes);
```
sazzad16, line 123:
```suggestion
    ArrayList<HostAndPort> startNodeList = new ArrayList<>(startNodes);
```

sazzad16 (approved):

PR #1930, opened 2019-01-09
[Feature] ROLE command implementation

aeter:
Description:
* implements the ROLE Redis command, https://redis.io/commands/role

Testing:
* ran ```TEST=ControlCommandsTest make test```, observed 0 failures.


gkorland:
@aeter thanks for the PR, don't you think we should also add this API to the none binary API?

aeter:
@gkorland - a good point! I've updated the code, rebuilt redis locally and reran ```TEST=ControlCommandsTest make test```, with 0 test failures.

I've just noticed there has been an older PR, of about 5 years ago (https://github.com/xetorthio/jedis/pull/664) as well as a related issue in the tracker (https://github.com/xetorthio/jedis/issues/1124). Having this in mind, feel free to handle (close/merge/etc.) the current PR as you prefer.

aeter:
My code after a few improvements turns out to be very similar to the older PR code (https://github.com/xetorthio/jedis/pull/664) by @HeartSaVioR, so I'll just close my PR for now, it makes no sense to have 2 PRs open with similar code.

gkorland:
@sazzad16 should we reopen this PR since https://github.com/xetorthio/jedis/pull/664 seems a bit too old now

sazzad16:
@gkorland ok. but we should return something better than `List<Object>`.

gkorland:
Ya, I thought about it too, but the problem is that the returned value changes according to the role

yangbodong22011:
@sazzad16 @gkorland Regarding the return value of the `ROLE` command, I thought of the following methods:

- Return a class named `RoleResponse` and `masterRoleResponse`, `slaveRoleResponse`, `sentinelRoleResponse` inherits `RoleResponse`, the disadvantage of this is that the user needs to use the `instanceof` keyword to determine and convert to the correct subclass.
- Return `List<Object>`, p.s. [Lettuce](https://github.com/lettuce-io/lettuce-core/blob/main/src/main/java/io/lettuce/core/api/sync/BaseRedisCommands.java#L89) also returns ` List<Object>`.

I support returning to `List<Object>`, if someone votes for `RoleResponse` or other methods, we can continue the discussion.

Issue #1545, opened 2017-07-10
Common interface for binary clustered and non-clustered redis needed


greenrd:
In 2.9.0, RedisCommands is implemented by both Jedis and JedisCluster, but does not include the binary versions of the commands, and there appears to be no common interface which includes binary versions of the commands shared between Jedis and JedisCluster.

sazzad16:
@greenrd There is no `RedisCommands`. Could you please clarify which interface you're talking about?

greenrd:
@sazzad16 Sorry, it was a typo. I meant `JedisCommands`.

HeartSaVioR:
We've discussed about this before while adding interface for JedisCluster. 
The issue is that JedisCluster doesn't support all the commands. If we apply origin interfaces to (Binary)JedisCluster, some methods in JedisCluster should be implemented as throwing UnsupportedException. 

While in discussion we thought it is bad, so we adopt separate interfaces in JedisCluster, but I have been thinking I might be wrong. Unifying interface might have more worth than avoiding UnsupportedException. I would like to hear opinions again.

@marcosnils @sazzad16 @greenrd @mp911de (as you thumb up :) )

sazzad16:
👍

mp911de:
Having a common interface with commands shared between Redis Cluster and Redis Standalone (Master/Slave) is beneficial as it does no further distinction in client code. Right now, we need to operate either on `Jedis` or `JedisCluster` and make that distinction in the code.

I propose to design an interface that:

1. Summarizes all data-type specific command interfaces (keys, string, lists, Lua, …).
2. Does not expose cluster commands (I made that mistake with `RedisClusterCommands` which is also used for Redis Standalone in Lettuce).
3. Does not expose transactional commands or standalone commands (like `MULTI` and `SELECT`).
4. Has no state (i.e. not extending `Closeable` and such, the API is intended for command usage, not lifecycle control).

You end up with an interface (well, two, one for byte arrays and one for strings) that exposes just the intersection of commands supported by Redis Standalone and Cluster. That interface can be the new API implemented by `Jedis` and `JedisCluster` and should be promoted in code samples for broad adoption. Not sure about transactions and pipelining right now, that can be the next step.

Does this make sense?

enryold:
Hi, I've just created a fast wrapper class for JedisPool/JedisCluster as workaround 
Maybe it could be useful for somebody. 

https://gist.github.com/enryold/76d5cc9f909a6ed96fffe13c6842785c

[...]

PR #1932, opened 2019-01-11
Introduce a common parent interface for JedisClusterCommands and JedisCommands

ptahchiev:
Unify the common methods from JedisClusterCommands and JedisCommands in a common parent interface to avoid duplication.


src/main/java/redis/clients/jedis/commands/JCommands.java
```diff
@@ -0,0 +1,327 @@
...
+
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * @author Petar Tahchiev
+ * @since 2.0.1
+ */
+public interface JCommands {
```
gkorland: I think we need a better naming

gkorland:
in general it LGTM, but
1. I think we need a better name than JCommands
2. @marcosnils was there any logic behind this duplication ?

src/main/java/redis/clients/jedis/commands/JCommands.java
ptahchiev: Agree

ptahchiev: I have renamed `JCommands` to `JedisBaseCommands`

gkorland:
LGTM!

marcosnils:
LGTM!. I've assigned to 4.0.0 as it breaks backwards compatibility.

gkorland:
@marcosnils  how does it break?

marcosnils:
@gkorland for projects that were using JedisCommands or JedisClusterCommands the code won't compile anymore because methods were moved to a new interface. We're breaking the API for external consumers

gkorland:
@marcosnils But that is the all idea, JedisCommands and JedisClusterCommands interfaces extend the new JedisBaseCommands interface, is should still compile.

ptahchiev:
@marcosnils why would it not compile when the existing interfaces extend from the new interface?

ptahchiev:
This PR is marked as ready to merge, but since 2 months nothing has happened. And now it is unmergeable, because both interfaces added new methods. And guys what: `JedisClusterCommands` methods use `Long` while `JedisCommands` use `long`. This wouldn't have happened if both methods were in the same parent interface like I suggested. 
@gkorland @marcosnils  please provide some constructive feedback on why you don't want this merged or how to improve it so I can get it merged.

gkorland:
@ptahchiev you're right, I don't see a reason not to merge it to 3.1, but @marcosnils had some concerns.
BTW, I tried to merge/fix the branch using GH merge GUI but it seems like I broke your branch, can you please review it.

ptahchiev:
@gkorland Like I said now some of the methods have diverged. Two methods have the same name and parameters, but one returns `java.lang.Long` and the other one returns `long`. I can rewrite my PR but now it will  bigger. Can you please advise me which return signature to use - `java.lang.Long` or `long` is correct?

sazzad16:
@gkorland I do not support to merge this in 3.x.

If you really want to merge this now, I'd suggest to merge this ONLY in master (which will act as 4.0). Then we'd have to manage 4.0/master and 3.x separately.

src/main/java/redis/clients/jedis/commands/JedisBaseCommands.java
```diff
@@ -0,0 +1,327 @@
...
+import redis.clients.jedis.params.SetParams;
+import redis.clients.jedis.params.ZAddParams;
+import redis.clients.jedis.params.ZIncrByParams;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * @author Petar Tahchiev
```
sazzad16: Please remove author tag.

src/main/java/redis/clients/jedis/commands/JedisBaseCommands.java
```diff
@@ -0,0 +1,327 @@
...
+ * @since 2.0.1
```
sazzad16: ???

sazzad16:
@ptahchiev @gkorland @marcosnils 

I'm a bit concerned about the name `JedisBaseCommands`. What are trying to mean by the word `Base`? Furthermore, this may create confusion between `JedisBaseCommands` and `BasicCommands` which contain separate methods.

**So, I would strongly suggest a name change.**

Just dropping a suggestion: may be something like, `JedisCommonCommands`?

ptahchiev:
@sazzad16 I have no problem with a name change. However, I do believe that this should be merged in the 3.x branch as it is not a breaking change and is in fact a blocker to other open-source projects (jedis-quartz, spring-boot with jedis-quartz, etc). @sazzad16  is the name change your only objection?

sazzad16:
@ptahchiev You've asked a simple but complicated question! Before answering anything, I purposefully avoided myself from participating here. because I don't exactly support this PR with the idea of introducing a new common parent interface.

I lean more towards **unifying** the concerned interfaces. This idea was expressed by @HeartSaVioR earlier in #1545.<sup>[1]</sup>

[1]: https://github.com/xetorthio/jedis/issues/1545#issuecomment-347687167

ptahchiev:
Hi @sazzad16 I am totally OK with unifying both the interfaces into one. Like I said I raised this issue (https://github.com/xetorthio/jedis/issues/1931) because I cannot upgrade the quartz-redis-jobstore to the jedis 3.0 and I am really blocked. I would really like to see a decision being made and some progress on this issue.

sazzad16:
@ptahchiev Thank you for understanding!

And, for this approach, you've already got two collaborators participating which is enough! I just spoke out when I felt to, for the format of the project.

src/main/java/redis/clients/jedis/commands/JedisClusterCommands.java
```diff
@@ -19,7 +19,10 @@
 import java.util.Map;
 import java.util.Set;
 
-public interface JedisClusterCommands {
+public interface JedisClusterCommands extends JedisBaseCommands {
+  
+  Long waitReplicas(final String key, final int replicas, final long timeout);
+  
   String set(String key, String value);
```
sazzad16: `JedisClusterCommands` shouldn't contain the methods from `JedisBaseCommands`.

kamyuentse:
ping

thattolleyguy:
What's the status of this issue? I'd be more than happy to help complete it. We upgraded to jedis 3.0 as part of upgrading to spring boot 2 but are now missing commands because the JedisClusterCommands and JedisCommands are not in sync.

gkorland:
@thattolleyguy can you please be more specific about the missing commands? these are probably commands that are not supported by the cluster API

thattolleyguy:
@gkorland These are the methods missing from `JedisCluster`:

- `hincrByFloat(String key, String field, double value)`
- `move(String key, int dbIndex)`
- `bitpos(String key, boolean value)`
- `bitpos(String key, boolean value, BitPosParams params)`
- `hscan(String key, String cursor, ScanParams params)`
- `sscan(String key, String cursor, ScanParams params)`
- `zscan(String key, String cursor, ScanParams params)`

All of them except `move` and `bitpos` are supported in the `BinaryJedisCluster` so as far as I can tell they should be supported in the `JedisCluster` as well. Maybe this is a separate issue but I thought it was related as they seem to be common commands between `Jedis` and `JedisCluster`.

Should the `move` and `bitpos` commands be supported in `JedisCluster`? I'm pretty new to redis so I don't know if they do or don't work in a cluster.

sazzad16:
Resolved by #2533

PR #1935, opened 2019-01-17
Remove reset after connection was returned to pool

### LINKED ISSUE #1920: JedisPool exhausted in Jedis 2.10.0
Jedis.java

    @Override
    public void close() {
        if (dataSource != null) {
            if (client.isBroken()) {
                this.dataSource.returnBrokenResource(this);
            } else {
                this.dataSource.
                        returnResource(this);
            }
            this.dataSource = null;  // <-- This line 
        } else {
            super.close();
        }
    }

-----

JedisSentinelPool.java
       
    @Override
    public Jedis getResource() {
        while (true) {
            Jedis jedis = super.getResource();
            jedis.setDataSource(this);  // <-- This line 

            // get a reference because it can change concurrently
            final HostAndPort master = currentHostMaster;
            final HostAndPort connection = new HostAndPort(jedis.getClient().getHost(), jedis.getClient()
                    .getPort());

            if (master.equals(connection)) {
                // connected to the correct master
                return jedis;
            } else {
                returnBrokenResource(jedis);
            }
        }
    }

In the case of concurrency

1. Thread A return an object but not run to "this.dataSource = null" yet

2. Thread B borrow an object and set dataSource to this;

3. And Then Thread A run this.dataSource = null;

4. Finally Thread A will never returnResource because dataSource is null

#### Jedis version:
2.10.0

gkorland:
Fix #1920 (#1918)
Fix bad backport of commit

### DIFF
--- a/src/main/java/redis/clients/jedis/Jedis.java
+++ b/src/main/java/redis/clients/jedis/Jedis.java
@@ -3636,7 +3636,6 @@ public class Jedis extends BinaryJedis implements JedisCommands, MultiKeyCommand
       } else {
         pool.returnResource(this);
       }
-      this.dataSource = null;
     } else {
       super.close();
     }


gkorland:
@sazzad16 you can't in 2.10 is was 
```
  protected Pool<Jedis> dataSource = null;
```
And there was no JedisPoolAbstract

sazzad16:
@gkorland oh, ok!

LGTM!

sazzad16:
Merged into 2.10.

PR #1942, opened 2019-01-27
Add Unix domain socket support

mina-asham:
Fixes issue https://github.com/xetorthio/jedis/issues/492, and revives pull request https://github.com/xetorthio/jedis/pull/1132

The code is mostly based on the previous pull request with one main modification: exposing the unix domain socket file in the constructor. This makes the interface more clear from a Jedis client point of view.

Note: Tests were already failing due to latest Redis unstable version, but the test for this commit is working fine, and all tests pass on Redis 5.0 branch.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,6 +48,7 @@
 		<redis-hosts>localhost:6379,localhost:6380,localhost:6381,localhost:6382,localhost:6383,localhost:6384,localhost:6385</redis-hosts>
 		<sentinel-hosts>localhost:26379,localhost:26380,localhost:26381</sentinel-hosts>
 		<cluster-hosts>localhost:7379,localhost:7380,localhost:7381,localhost:7382,localhost:7383,localhost:7384,localhost:7385</cluster-hosts>
+		<uds-hosts>/tmp/redis_6379.sock</uds-hosts>
     	<github.global.server>github</github.global.server>
 	</properties>
 
@@ -63,6 +64,11 @@
 			<version>2.6.0</version>
 			<type>jar</type>
 			<scope>compile</scope>
+		</dependency>
+		<dependency>
+			<groupId>com.kohlschutter.junixsocket</groupId>
+			<artifactId>junixsocket-native-common</artifactId>
+			<version>2.0.4</version>
 		</dependency>
 
 		<dependency>
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -1,9 +1,11 @@
 package redis.clients.jedis;
 
 import java.io.Closeable;
+import java.io.File;
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.net.Socket;
+import java.net.SocketAddress;
 import java.net.SocketException;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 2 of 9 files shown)


pom.xml
```diff
@@ -64,6 +65,11 @@
 			<type>jar</type>
 			<scope>compile</scope>
 		</dependency>
+		<dependency>
+			<groupId>com.kohlschutter.junixsocket</groupId>
+			<artifactId>junixsocket-native-common</artifactId>
```
sazzad16: I don't like the idea of adding one more dependency* just to support one feature.

\* compile time

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -45,6 +50,10 @@ public Connection(final String host) {
     this.host = host;
   }
 
+  public Connection(final File unixDomainSocket) {
```
sazzad16: What if we take the `Socket` (and/or other objects) or a custom object which would create/connect a socket on demand? This would help to support many forms of sockets to be supported by Jedis at once.

pom.xml
mina-asham: Yeah, I wasn't too mad about it as well, but thought it would be fine since it's a small dependency. So I think we have a few options (ordered by most to least favourite by me, but open to suggestions):
* [My preferred suggestion] Make the dependency an optional one and do a reflection check to see if it is and if the feature can be supported
* Shade and include the dependency (also since it's tiny), this way it would be self included and no need for consumers to worry
* Accept the dependency as is now since it's tiny
* Not include it at all and follow your below comments passing a socket straight in the constructor. It's my least preferred because I think the consumer shouldn't need to worry about creating/configuring the socket, specially that Jedis has some good defaults there. In the below comment you mention *other* sockets, any specific ones in mind?

(Sorry for the delayed response, this escaped me somehow...)

src/main/java/redis/clients/jedis/Connection.java
mina-asham: See last point in above comment.

sazzad16: > you mention other sockets, any specific ones in mind?

Not anything specific, rather generic.

I just meant, if we had such constructor, sockets like unix domain socket, ssl or even plain socket with different configuration from Jedis could be used by users.

mina-asham: > I just meant, if we had such constructor, sockets like unix domain socket, ssl or even plain socket with different configuration from Jedis could be used by users.

We can do that, but I think the question is should we expose that to the user as part of the library?

Jedis tries to do such constructs under the hood and expose other options (ssl, passwords, etc...) and sets some defaults for each of these, we would risk exposing this and making the library a bit harder (or a bit more confusing) to use.

It doesn't make much difference to my case by the way, so I am not opposing this, I just want to make sure you would be okay with that (minimal) risk.

sazzad16: > we would risk exposing this and making the library a bit harder (or a bit more confusing) to use.

I'm not sure I follow you.

mina-asham: As a user, I would much rather set a minimal parameters (e.g. port number, UDS path, credentials), than creating/setting/maintaining a socket (more code and might not be setup with the best configuration unless I am an advanced user).

Not the strongest argument, but this is my intuition. What do you think?

sazzad16: - About dependency:
Most users are not enthusiastic about more dependencies. So I'm a bit reluctant on this.

- About _not easy_ option:
At least it is an option which would allow avid users to avoid long routes (like extending Jedis/Connection classes, building custom Jedis). Let's take `sendCommand` method (coming in 3.1.0) as an example. It would require some coding for the user but would allow them to use new commands in Redis (which are not in Jedis) more easily without waiting for Jedis to include those.

These are my viewpoints.

mina-asham: Alright, sounds good, I will refactor this to allow creating it with a Socket object for advanced use-cases, will update this soon :)

sazzad16: @mina-asham Please take a look at #2036. WDYT?

sazzad16:
Resolved by #2151

PR #1948, opened 2019-02-08
Javadocs minor

ullenius:
I ran the FindBugs tool and added missing:

```
@param
@return
@Override
```

in the javadoc comments.


marcosnils:
Let's leave this on master as it's just a doc fix.

PR #1954, opened 2019-02-25
Create Unflushed getObjectMultiBulkReply variant in Connection

werkt:
Ensure that the mechanism used by subscription clients does not flush
the connection. This allows client synchronization to prevent protocol
message interleaving when manipulating the subscription connection.

### DIFF
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -270,13 +270,18 @@ public class Connection implements Closeable {
     return (List<byte[]>) readProtocolWithCheckingBroken();
   }
 
+  /**
+   * This method is used by JedisPubSub::process and must not interact
+   * with the Connection outputStream in order to prevent protocol
+   * desynchronization.
+   */
   @SuppressWarnings("unchecked")
   public List<Object> getRawObjectMultiBulkReply() {
-    flush();
     return (List<Object>) readProtocolWithCheckingBroken();
   }
 
   public List<Object> getObjectMultiBulkReply() {
+    flush();
     return getRawObjectMultiBulkReply();
   }
 
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -270,6 +270,11 @@ public class Connection implements Closeable {
     return (List<byte[]>) readProtocolWithCheckingBroken();
   }
 
+  @Deprecated
+  public List<Object> getRawObjectMultiBulkReply() {
+    return getUnflushedObjectMultiBulkReply();
+  }
+
   @SuppressWarnings("unchecked")
   public List<Object> getUnflushedObjectMultiBulkReply() {
     return (List<Object>) readProtocolWithCheckingBroken();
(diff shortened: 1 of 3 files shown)


sazzad16:
@gkorland @werkt `flush()` is called before reading any other Redis response. IMHO, `getRawObjectMultiBulkReply()`, which may also be used by methods other than subscription, would be an exception. However, in case `flush()` is an issue for subscription, we should create a separate method hinting the purpose and/or use-case.

werkt:
There are no other internal users of `getRawObjectMultiBulkReply`, fwiw. The `Raw` part of its name seems to mean `Unflushed` by convention, but obviously, as you've said, the specific dependency on the non-interaction with the write side is not overtly as clear as, say, calling the method `getSubscribeReply`.

sazzad16:
@werkt Hmm. `Raw` part of the name could be there to mean `Unflushed`. But, the problem is, in Jedis `raw` generally means `byte[]`.

sazzad16:
@werkt Let's
1. create a new method `getUnflushedObjectMultiBulkReply`
2. use `getUnflushedObjectMultiBulkReply` everywhere instead of `getRawObjectMultiBulkReply`, even in `getObjectMultiBulkReply`
3. deprecate `getRawObjectMultiBulkReply`

WDYT?

@gkorland @marcosnils ping!

werkt:
Works for me, do you also mean to replace the existing uses of `readProtocolWithCheckingBroken()` throughout `Connection` with `getUnflushedObjectMultiBulkReply()`, or leave those untouched?

werkt:
Also, what is your preferred deprecation mechanism? Should the method be tagged? There are no other instances of `deprecate` (or variants) throughout jedis now.

sazzad16:
> Works for me, do you also mean to replace the existing uses of `readProtocolWithCheckingBroken()` throughout `Connection` with `getUnflushedObjectMultiBulkReply()`, or leave those untouched?

Leave those untouched.

> Also, what is your preferred deprecation mechanism? Should the method be tagged? There are no other instances of `deprecate` (or variants) throughout jedis now.

Any valid way will do.

werkt:
> > Also, what is your preferred deprecation mechanism? Should the method be tagged? There are no other instances of `deprecate` (or variants) throughout jedis now.
> 
> Any valid way will do.

Roger that, removed them.

PR is updated, commit text should be modified prior to import to reflect the Unflushed introduction and Raw removal, as well as including the original comment above.

marcosnils:
LGTM!

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -271,13 +271,13 @@ public Long getIntegerReply() {
   }
 
   @SuppressWarnings("unchecked")
-  public List<Object> getRawObjectMultiBulkReply() {
```
sazzad16: Keep this method, with `@Deprecated`, return `getUnflushedObjectMultiBulkReply()`.

werkt: Done.

qlaall:
After looking for the problem for a long time today, I finally found that the difference between 3.0.1 and 3.1.0 is that there is no flush method.

PR #1961, opened 2019-03-26
Test Fail

sazzad16:


gkorland:
@sazzad16 why is it fixing the tests?

marcosnils:
We always try to run Jedis against unstable and not pin it to any branch / tag. What's the reason behind this PR?

sazzad16:
@gkorland @marcosnils Sorry. I supposed to close this PR. Actually, I was trying to find out the cause of failure of some tests. As it seems, Redis has been changing behaviour of some commands in unstable branch, compared to the latest stable branch, 5.0.

PR #1968, opened 2019-04-10
implemented zpopmax, zpopmax with count

z0mg:
Added tests under redis.clients.jedis.tests.commands.SortedSetCommandsTest.
Tried mimicking other binary asserts, but it looks like you can use any value for `expected.add(SafeEncoder.encode("x"));`, they'll all pass.


src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -1673,6 +1673,20 @@ public Double zscore(final String key, final String member) {
     return BuilderFactory.DOUBLE.build(client.getOne());
   }
 
+  @Override
+  public List<String> zpopmax(final String key) {
```
sazzad16: Return type should be `Tuple`

src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -1673,6 +1673,20 @@ public Double zscore(final String key, final String member) {
...
+    checkIsInMultiOrPipeline();
+    client.zpopmax(key);
+    return client.getMultiBulkReply();
+  }
+
+  @Override
+  public List<String> zpopmax(final String key, final int count) {
```
sazzad16: Return type should be `Set<Tuple>`

src/main/java/redis/clients/jedis/Jedis.java
z0mg: Now that you've mentioned it, of course it should be :) I'll make the changes shortly.

z0mg: > Return type should be `Tuple`

Could you please help with this inside `BinaryJedis` - I can't figure out what it should look like. There's no single Tuple being returned anywhere that I could find, just Set<Tuple>. 

```
  public List<byte[]> zpopmax(final byte[] key) {
    checkIsInMultiOrPipeline();
    client.zpopmax(key);
    return client.getBinaryMultiBulkReply();
  }
```

sazzad16: There are 2 ways I can think of, right now.
1. Create `Set<Tuple>` from response (see other examples) and return the only element.
2. Create `List<byte[]>` of two elements and use `Tuple(byte[], Double)` constructor.

z0mg: How does this look under `BinaryJedis `?

```
  public Tuple zpopmax(final byte[] key) {
    checkIsInMultiOrPipeline();
    client.zpopmax(key);
    
    List<byte[]> result = client.getBinaryMultiBulkReply();
    
    return new Tuple(result.get(0), BuilderFactory.DOUBLE.build(result.get(1)));
  }
```

And what about the `PipelineBase` implementation?
  
```
  public Response<List<String>> zpopmax(final String key) {
    getClient(key).zpopmax(key);
    return getResponse(BuilderFactory.STRING_LIST);
  }
```

There already is a `BuilderFactory.Tuple_ZSET` type, but it returns a `Set<Tuple>`.

I see 2 options here: 
a) Make a new builder that returns a simple `Tuple`?
b) Use a Set<Tuple> for the zpopmax(String key) variant, that will always only have 1 element?

z0mg: Made the changes, please have a look.

src/main/java/redis/clients/jedis/Jedis.java
z0mg: Made the changes, please have a look.

src/test/java/redis/clients/jedis/tests/commands/SortedSetCommandsTest.java
```diff
@@ -542,7 +545,147 @@ public void zscore() {
...
+    
+    expected.add(new Tuple("a", 1d));
+    expected.add(new Tuple("c", 0.1d));
+    expected.add(new Tuple("e", 0.03d));
+    
+    assertEquals(expected, actual);
+    
+    // Empty
+    actual = jedis.zpopmax("foo", 1);
+    assertTrue(actual.isEmpty());
```
sazzad16: Please keep one coding pattern for binary and non-binary within same method. Personally, I liked your code for binary better. E.g.
```
expected.clear();
assertEquals(expected, actual);
```

src/test/java/redis/clients/jedis/tests/commands/SortedSetCommandsTest.java
```diff
@@ -542,7 +545,147 @@ public void zscore() {
...
+    wrapper = jedis.zpopmax("foo");
+    member = wrapper.getElement();
+    score = wrapper.getScore();
+    
+    assertEquals("c", member);
+    assertEquals((Double) 0.1d, score);
+    
+    // Empty
+    wrapper = jedis.zpopmax("foo");
+    assertTrue(wrapper == null);
```
sazzad16: Please keep one coding pattern for binary and non-binary within same method. Personally, I liked your code for binary better. E.g.
1. `actual`, `expected` variables
2. `assertNull(actual)` instead of `assertTrue(wrapper == null)`

src/test/java/redis/clients/jedis/tests/commands/SortedSetCommandsTest.java
```diff
@@ -4,13 +4,16 @@
...
+import static org.junit.Assert.assertTrue;
 import static redis.clients.jedis.ScanParams.SCAN_POINTER_START;
 import static redis.clients.jedis.ScanParams.SCAN_POINTER_START_BINARY;
 import static redis.clients.jedis.tests.utils.AssertUtil.assertByteArraySetEquals;
 
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
+import java.util.List;
```
sazzad16: We won't need this.

src/test/java/redis/clients/jedis/tests/commands/SortedSetCommandsTest.java
```diff
@@ -4,13 +4,16 @@
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
```
sazzad16: We may not need this after other changes.

src/main/java/redis/clients/jedis/BinaryJedis.java
```diff
@@ -1841,6 +1841,20 @@ public Double zscore(final byte[] key, final byte[] member) {
     return (score != null ? new Double(score) : null);
   }
 
+  public Tuple zpopmax(final byte[] key) {
+    checkIsInMultiOrPipeline();
+    client.zpopmax(key);
+    
+    List<byte[]> l = client.getBinaryMultiBulkReply();
+    return l.isEmpty() ? null : new Tuple(l.get(0), BuilderFactory.DOUBLE.build(l.get(1)));
```
sazzad16: You've created `BuilderFactory.TUPLE`. Can't we use it here to avoid code repetition?

src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -1673,6 +1673,21 @@ public Double zscore(final String key, final String member) {
     return BuilderFactory.DOUBLE.build(client.getOne());
   }
 
+  @Override
+  public Tuple zpopmax(final String key) {
+    checkIsInMultiOrPipeline();
+    client.zpopmax(key);
+    List<byte[]> l = client.getBinaryMultiBulkReply();
+    return l.isEmpty() ? null : new Tuple(l.get(0), BuilderFactory.DOUBLE.build(l.get(1)));
```
sazzad16: You've created `BuilderFactory.TUPLE`. Can't we use it here to avoid code repetition?

src/test/java/redis/clients/jedis/tests/commands/SortedSetCommandsTest.java
z0mg: In the spirit of using the same thing across multiple places, I hope this is OK? 
```
    expected = null;
    assertEquals(expected, actual);
```

Instead of `assertNull(actual)`.

sazzad16: At least I would prefer that over `assertTrue(wrapper == null)`.

src/main/java/redis/clients/jedis/BinaryJedis.java
z0mg: Done. With the mention that I changed the null check in `BuilderFactory.TUPLE` to look for empty casted List parameters instead of a null parameter. For now, it only receives non-null values.

z0mg:
Have a look, please.

sazzad16:
LGTM!

guptaaditya13:
Any ETA on when this will be merged?

src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -285,6 +285,24 @@ public String toString() {
 
   };
 
+  public static final Builder<Tuple> TUPLE = new Builder<Tuple>() {
```
gkorland: @z0mg can you please add a comment about the different between TUPLE_ZSET and TUPLE?
Or perhaps we better even rename this builder?

sazzad16: @gkorland ZPOPMAX and ZPOPMIN without count returns only one Tuple. Returning a Tuple should be better instead of returning a Set of single Tuple. This builder helps on that purpose.

PR #1791, opened 2018-03-17
Add redis command listener

SerCeMan:
As per https://groups.google.com/forum/?fromgroups#!topic/jedis_redis/UaJubt42f_0,

This PR adds a command listener that allows for 
* sane integration with tracing libraries like http://opentracing.io/
* capturing metrics and integration with metric libraries like http://metrics.dropwizard.io/4.0.0/
* integration with logging libraries

None of the public signatures was changed, so the change should be fully backwards compatible.

PR #1972, opened 2019-04-12
Redis command listener

gr8routdoors:
This builds on #1791 with the following:
* adds `JedisCommandListener.commandConnected()` to allow tracking of connection time metrics separately from command execution time
* adds coverage with unit tests
* rebases from the current master

More details from the PR #1791 description:
As per https://groups.google.com/forum/?fromgroups#!topic/jedis_redis/UaJubt42f_0,

This PR adds a command listener that allows for

sane integration with tracing libraries like http://opentracing.io/
capturing metrics and integration with metric libraries like http://metrics.dropwizard.io/4.0.0/
integration with logging libraries
None of the public signatures was changed, so the change should be fully backwards compatible.


src/main/java/redis/clients/jedis/JedisCommandListener.java
```diff
@@ -0,0 +1,51 @@
...
+ * implementation must be thread safe because the listener might be invoked from
+ * different threads.
+ * <p>
+ * The listener is guaranteed to call {@link #commandStarted} and one of the
+ * {@link #commandFinished} or {@link #commandFailed} once for each command.
+ * <p>
+ * The listener implementation must be exception free or else none of the aforementioned
+ * contracts will be enforced.
+ */
+public interface JedisCommandListener {
```
gkorland: Should extend EventListener?

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -87,6 +88,10 @@ public void setSoTimeout(int soTimeout) {
     this.soTimeout = soTimeout;
   }
 
+  public void setListener(JedisCommandListener listener) {
```
gkorland: should be addListener?

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -37,6 +37,7 @@
   private SSLSocketFactory sslSocketFactory;
   private SSLParameters sslParameters;
   private HostnameVerifier hostnameVerifier;
+  private JedisCommandListener listener = NoopJedisCommandListener.INSTANCE;
```
gkorland: Isn't it better to have EventListenerList?

src/main/java/redis/clients/jedis/JedisCommandListener.java
SerCeMan: Hey, @gkorland! Do you mean `java.util.EventListener`? Out of interest, what is the benefit of extending `EventListener`? As far as I'm aware, it doesn't have much application outside of the swing/AWT world.

src/main/java/redis/clients/jedis/Connection.java
SerCeMan: I believe this can result in an unnecessary dependency on `java.desctop` jigsaw module which can result in issues for library users on JDK 9+ and use compact java images.

gkorland: Perhaps it doesn't need to be EventListenerList but a the idea of managing a list of listeners I think is more robust

src/main/java/redis/clients/jedis/JedisCommandListener.java
gkorland: I guess if we're not going with EventListenerList then we don't need it to be an EventListener

src/main/java/redis/clients/jedis/Connection.java
gr8routdoors: That makes sense.  I'll update to supporting a `List<JedisCommandListener>`, with the default being an empty list.

gr8routdoors: OK, I've pushed a commit that supports multiple listeners @gkorland .  Please let me know what you think.

src/main/java/redis/clients/jedis/Connection.java
gr8routdoors: The new multi-listener interface contains:  `setListeners`, `addListener`, and `removeListener`

src/main/java/redis/clients/jedis/JedisCommandListener.java
gr8routdoors: Makes sense to me.  It's easy to implement this w/ a standard list and not include a swing dependency.  See my latest commit.

src/main/java/redis/clients/jedis/Connection.java
SerCeMan: Hey, @gr8routdoors! Wdyt of using `ArrayList` instead? In the vast majority of use cases, it allows for better performance and memory usage. There is a good discussion regarding this topic here: https://stackoverflow.com/questions/322715/when-to-use-linkedlist-over-arraylist-in-java

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -87,6 +89,48 @@ public void setSoTimeout(int soTimeout) {
     this.soTimeout = soTimeout;
   }
 
+  public void setListeners(List<JedisCommandListener> listeners) {
+    if (listeners == null) {
+      if (!this.listeners.isEmpty()) { // No reason to re-create an already empty list
+        this.listeners = new LinkedList<>();
+      }
+    } else {
+      this.listeners = listeners;
```
gkorland: that might be an issue and create a way for users to change the list without calling addListener/removeListener. I think listeners should be final and the listeners should be copied to the list.

gr8routdoors: OK, I've renamed this method `addListeners` and updated the behavior to just add new listeners, instead of resetting the list.

gr8routdoors:
@gkorland , any update here?  I believe I've addressed all your concerns.

gkorland:
Did you check my latest comments?

gr8routdoors:
@gkorland which comments?  AFAICT they've all been addressed.

src/main/java/redis/clients/jedis/Connection.java
gr8routdoors: Hi! My assumption was that the number of listeners would be very small, so I chose `LinkedList` for a smaller memory footprint.  Could easily change it to `ArrayList` if you think it better matches the expected usage.

gr8routdoors:
Any update here @gkorland ?

SerCeMan:
Hey, @gkorland! Would you be able to take a look at the PR? I believe it's almost there, just need to agree on some minor points :)

SerCeMan:
Hey, @gkorland, @sazzad16! Would you be able to take a look? There is clearly a need for this, https://github.com/xetorthio/jedis/pull/1791#issuecomment-515966697.

gkorland:
@SerCeMan can you please rebase the PR

gr8routdoors:
OK, I've updated to the latest from master @gkorland

src/test/java/redis/clients/jedis/tests/ConnectionTest.java
```diff
@@ -1,17 +1,23 @@
...
 import redis.clients.jedis.Protocol.Command;
 import redis.clients.jedis.commands.ProtocolCommand;
 import redis.clients.jedis.exceptions.JedisConnectionException;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
 public class ConnectionTest {
```
sazzad16: Connection operation or connection methods are not going to be tested. Move the test methods in a new test class.

src/main/java/redis/clients/jedis/JedisPool.java
```diff
@@ -40,18 +41,26 @@ public JedisPool(final String host) {
...
+  public JedisPool(final String host, final SSLSocketFactory sslSocketFactory,
+      final SSLParameters sslParameters, final HostnameVerifier hostnameVerifier,
+      final List<JedisCommandListener> listeners) {
     URI uri = URI.create(host);
     if (JedisURIHelper.isValid(uri)) {
       this.internalPool = new GenericObjectPool<Jedis>(new JedisFactory(uri,
-          Protocol.DEFAULT_TIMEOUT, Protocol.DEFAULT_TIMEOUT, null, sslSocketFactory, sslParameters,
-          hostnameVerifier), new GenericObjectPoolConfig());
+              Protocol.DEFAULT_TIMEOUT, Protocol.DEFAULT_TIMEOUT, null, sslSocketFactory, sslParameters,
+              hostnameVerifier, listeners), new GenericObjectPoolConfig());
```
sazzad16: Use two spaces per _tab_.

src/main/java/redis/clients/jedis/JedisPool.java
```diff
@@ -40,18 +41,26 @@ public JedisPool(final String host) {
...
     } else {
       this.internalPool = new GenericObjectPool<Jedis>(new JedisFactory(host,
-          Protocol.DEFAULT_PORT, Protocol.DEFAULT_TIMEOUT, Protocol.DEFAULT_TIMEOUT, null,
-          Protocol.DEFAULT_DATABASE, null, false, null, null, null), new GenericObjectPoolConfig());
+              Protocol.DEFAULT_PORT, Protocol.DEFAULT_TIMEOUT, Protocol.DEFAULT_TIMEOUT, null,
+              Protocol.DEFAULT_DATABASE, null, false, null, null, null,
+              listeners), new GenericObjectPoolConfig());
```
sazzad16: Use two spaces per _tab_.

src/main/java/redis/clients/jedis/JedisPool.java
```diff
@@ -40,18 +41,26 @@ public JedisPool(final String host) {
...
     }
   }
 
+
```
sazzad16: No need.

src/main/java/redis/clients/jedis/JedisFactory.java
```diff
@@ -50,16 +52,19 @@
...
   JedisFactory(final URI uri, final int connectionTimeout, final int soTimeout,
       final String clientName) {
-    this(uri, connectionTimeout, soTimeout, clientName, null, null, null);
+    this(uri, connectionTimeout, soTimeout, clientName, null, null, null,
+            null);
   }
 
   JedisFactory(final URI uri, final int connectionTimeout, final int soTimeout,
       final String clientName, final SSLSocketFactory sslSocketFactory,
-      final SSLParameters sslParameters, final HostnameVerifier hostnameVerifier) {
```
sazzad16: This breaks backward compatibility. Create a new constructor.

uglide:

PR #1979, opened 2019-05-09
the slots info refreshed by MOVED . not by cluster slots.

Leeviiii:
1. maybe make https://github.com/xetorthio/jedis/issues/1512 better
2. in our environment the "cluster slots" is a big problem

### DIFF
--- a/src/main/java/redis/clients/jedis/Client.java
+++ b/src/main/java/redis/clients/jedis/Client.java
@@ -1263,4 +1263,8 @@ public class Client extends BinaryClient implements Commands {
   }
 
  
+  public String toString() {
+    return getHost() + ":" + getPort();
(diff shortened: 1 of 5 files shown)


sazzad16:
@gkorland @marcosnils Are we ready to remove all those `Lock` implementations?

AnujTewari:
hoping this will help #1512 🤞 . @Leeviiii

sazzad16:
https://github.com/redis/jedis/blob/b5687de44fc67e344734dede5fab9bc6df47cf95/src/main/java/redis/clients/jedis/JedisClusterCommand.java#L29-L31

We can't accept above condition. We should the leave the option to use JedisCluster without any retry.

PR #1981, opened 2019-05-15
JedisPool exhausted #1920

raytz:
ShardedJedis.java has the same problem


marcosnils:
I've already asked for #1939 to be backported. This PR is not necessary.

sazzad16:
@marcosnils IMHO, we can use this PR as backport of #1939.

PR #1983, opened 2019-05-16
empty password mean no password

francisoliverlee:


gkorland:
It seems like might cause some backward issues

marcosnils:
@francisoliverlee this has been discussed before here https://github.com/xetorthio/jedis/pull/1809. Redis actually allows you to set an empty password and authenticate against that. Is there something here different that I'm missing? For what I see it seems the same thing.

PR #1988, opened 2019-06-06
Impl zpopmin command

kommradHomer:
implemented zpopmin command.
written a test method


pom.xml
```diff
@@ -191,6 +191,28 @@
 				    </execution>
 				</executions>		
 			</plugin>
+
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-assembly-plugin</artifactId>
```
sazzad16: Why this?

kommradHomer: Oh. I thought I removed that one. I'll remove that and commit again

kommradHomer:
removed the unnecessary plugin

pom.xml
```diff
@@ -191,6 +191,7 @@
 				    </execution>
 				</executions>		
 			</plugin>
+
```
sazzad16: Remove this.

gkorland:
what about JedisClusterCommands?

SuShuilian:
I am looking forward to seeing this new command available in release version.
When will it be available in release version, any plan?

PR #2003, opened 2019-06-22
Fix SetFromList serialization

### LINKED ISSUE #2000: BinaryJedis.SetFromList transient causes serialization error
### Description
When I serialized the result returned by jedis.zrevrange(), I lost the key. Because the field `list` of `redis.clients.jedis.BinaryJedis.SetFromList` has a transient attribute.

```
protected static class SetFromList<E> extends AbstractSet<E> implements Serializable {
    private static final long serialVersionUID = -2850347066962734052L;
    private final transient List<E> list;
    ...
}
```
I found that the reason for using `transient ` is to fix the code specification problem：#1304，but this will cause serialization errors.

### Steps to reproduce:
```
        Jedis jedis = new Jedis("127.0.0.1", 6379);

        jedis.zadd("zset", 1, "key1");
        jedis.zadd("zset", 2, "key2");
        jedis.zadd("zset", 3, "key3");

        try {
            String filename = "/tmp/set.ser";
            FileOutputStream fileOut = new FileOutputStream(filename);
            ObjectOutputStream out = new ObjectOutputStream(fileOut);
            // serialize zrevragene result to file
            out.writeObject(jedis.zrevrange("zset", 0, -1));

            FileInputStream fileIn = new FileInputStream(filename);
            ObjectInputStream in = new ObjectInputStream(fileIn);
            //deserialize result from file
            Set<String> ret = (Set<String>)in.readObject();
            if (ret != null) {
                // NPE
                System.out.println(ret.size());
            }
        } catch (Exception e) {
            e.printStackTrace();
        }
    }
```

Result：
```
java.lang.NullPointerException
	at redis.clients.jedis.BinaryJedis$SetFromList.size(BinaryJedis.java:3844)
```

sazzad16:
Resolves #2000 
Closes #2001

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryJedis.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedis.java
@@ -3825,7 +3825,7 @@ public class BinaryJedis implements BasicCommands, BinaryJedisCommands, MultiKey
    */
   protected static class SetFromList<E> extends AbstractSet<E> implements Serializable {
     private static final long serialVersionUID = -2850347066962734052L;
-    private final transient List<E> list;
+    private final List<E> list;
 
     private SetFromList(List<E> list) {
       if (list == null) {
--- a/src/test/java/redis/clients/jedis/tests/collections/SetFromListTest.java
+++ b/src/test/java/redis/clients/jedis/tests/collections/SetFromListTest.java
@@ -123,7 +123,7 @@ public class SetFromListTest {
   }
 
   @Test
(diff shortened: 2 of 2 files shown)


sazzad16:
@gkorland @marcosnils ping!

yangbodong22011:
Hi @sazzad16 does this PR not fix this  issue #1304?

sazzad16:
@yangbodong22011 Yes, but not entirely. This PR reverts half of #1304

PR #2008, opened 2019-06-25
Fix JedisByteHashMap transient

yangbodong22011:
Fix JedisByteHashMap transient bug and add test.
see also: #2003 #2000

@sazzad16 please reveiew this PR.

### DIFF
--- a/src/main/java/redis/clients/jedis/util/JedisByteHashMap.java
+++ b/src/main/java/redis/clients/jedis/util/JedisByteHashMap.java
@@ -11,7 +11,7 @@ import java.util.Set;
 
 public class JedisByteHashMap implements Map<byte[], byte[]>, Cloneable, Serializable {
   private static final long serialVersionUID = -6971431362627219416L;
-  private Map<ByteArrayWrapper, byte[]> internalMap = new HashMap<ByteArrayWrapper, byte[]>();
+  private final Map<ByteArrayWrapper, byte[]> internalMap = new HashMap<ByteArrayWrapper, byte[]>();
 
   @Override
   public void clear() {
--- a/src/main/java/redis/clients/jedis/util/JedisByteHashMap.java
+++ b/src/main/java/redis/clients/jedis/util/JedisByteHashMap.java
@@ -11,7 +11,7 @@ import java.util.Set;
 
 public class JedisByteHashMap implements Map<byte[], byte[]>, Cloneable, Serializable {
   private static final long serialVersionUID = -6971431362627219416L;
-  private transient Map<ByteArrayWrapper, byte[]> internalMap = new HashMap<ByteArrayWrapper, byte[]>();
+  private Map<ByteArrayWrapper, byte[]> internalMap = new HashMap<ByteArrayWrapper, byte[]>();
 
   @Override
   public void clear() {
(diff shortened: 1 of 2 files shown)


sazzad16:
@yangbodong22011 LGTM! Would you be able to add test with multiple key-values?

yangbodong22011:
@sazzad16 Of course, I will do it.

src/test/java/redis/clients/jedis/tests/collections/JedisByteHashMapTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+        ByteArrayOutputStream byteOut = new ByteArrayOutputStream();
+        ObjectOutputStream objOut = new ObjectOutputStream(byteOut);
+        objOut.writeObject(map);
+
+        ByteArrayInputStream byteIn = new ByteArrayInputStream(byteOut.toByteArray());
+        ObjectInputStream objIn = new ObjectInputStream(byteIn);
+        JedisByteHashMap mapRead = (JedisByteHashMap)objIn.readObject();
+
+        Assert.assertTrue(collectionSame(map.keySet(), mapRead.keySet()));
```
sazzad16: Instead of comparing keys and values separately, would you be able to compare entries?

yangbodong22011: Do you mean like this？

```
private boolean entryContainsKV(Set<Map.Entry<byte[], byte[]>> s, byte[] key, byte[] value) {
        for (Map.Entry<byte[], byte[]> en : s) {
            if (Arrays.equals(en.getKey() ,key) && Arrays.equals(en.getValue(), value)) {
                return true;
            }
        }
        return false;
    }

    private boolean entrySetSame(Set<Map.Entry<byte[], byte[]>> s1, Set<Map.Entry<byte[], byte[]>> s2) {
        for (Map.Entry<byte[], byte[]> en1 : s1) {
            if (!entryContainsKV(s2, en1.getKey(), en1.getValue())) {
                return false;
            }
        }
        for (Map.Entry<byte[], byte[]> en2 : s2) {
            if (!entryContainsKV(s1, en2.getKey(), en2.getValue())) {
                return false;
            }
        }

        return true;
    }
```

and we call `entrySetSame` to compare entries

```
entrySetSame(map.entrySet(), mapRead.entrySet())  
```

sazzad16: IMO, it's at least better than comparing keys and values separately.

yangbodong22011: Yes, it's more elegant.

sazzad16:
LGTM!

src/main/java/redis/clients/jedis/util/JedisByteHashMap.java
```diff
@@ -11,7 +11,7 @@
 
 public class JedisByteHashMap implements Map<byte[], byte[]>, Cloneable, Serializable {
   private static final long serialVersionUID = -6971431362627219416L;
-  private transient Map<ByteArrayWrapper, byte[]> internalMap = new HashMap<ByteArrayWrapper, byte[]>();
+  private Map<ByteArrayWrapper, byte[]> internalMap = new HashMap<ByteArrayWrapper, byte[]>();
```
gkorland: final?

yangbodong22011: Fixed it.

PR #2017, opened 2019-07-08
trim dnComponent white spaces

gkorland:
In case the CN is not the first attribute the test fails, e.g.
```
DN=sdsa, CN=localhost
```


src/test/java/redis/clients/jedis/tests/SSLJedisTest.java
```diff
@@ -318,11 +318,12 @@ private String getCommonName(X509Certificate peerCertificate) {
       String subjectDN = peerCertificate.getSubjectDN().getName();
       String[] dnComponents = subjectDN.split(",");
       for (String dnComponent : dnComponents) {
+        dnComponent = dnComponent.trim();
```
sazzad16: Why is it necessary?

gkorland: For some reason in some of the places where we run the test we get the CN= at the end so without trim we get " CN=localhost" which doesn't start with "CN="

PR #2019, opened 2019-07-14
byte[] conversion for numbers in Params classes

sazzad16:
This would also be required when upgrading to JDK 1.8.


gkorland:
@sazzad16 what is this PR about? why is it needed?

sazzad16:
@gkorland In Params classes values are encoded by `SafeEncoder.encode(String.valueOf(value))`. But the problem with `String.valueOf(value)` is that it doesn't allow numbers to be encoded properly. For example, boolean won't be encoded as `1` or `0`, double couldn't be encoded as `+inf` or `-inf`. This PR lets those values to be encoded to be properly compatible with Redis.

Furthermore, current code would get [`ClassCastException`](https://stackoverflow.com/questions/23155907/classcast-error-java-7-vs-java-8) if it's built on JDK1.8. This PR solves this issue.

PR #2020, opened 2019-07-14
Remove dependency on 'Locale' for Keyword enum

sazzad16:
There was already an issue in #1319. Having the opportunity, may it be better to skip it entirely?

Note: Introduction of `toLowerCase()` is dated back to #37.

### DIFF
--- a/src/main/java/redis/clients/jedis/Protocol.java
+++ b/src/main/java/redis/clients/jedis/Protocol.java
@@ -3,7 +3,6 @@ package redis.clients.jedis;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Locale;
 
 import redis.clients.jedis.commands.ProtocolCommand;
 import redis.clients.jedis.exceptions.*;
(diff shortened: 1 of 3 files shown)


src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -274,17 +273,18 @@ public static Object read(final RedisInputStream is) {
...
+    // Request keywords
+    AGGREGATE, ALPHA, ASC, BY, DESC, GET, LIMIT, NO, NOSORT, ONE, SET, STORE, WEIGHTS, WITHSCORES,
     RESETSTAT, REWRITE, RESET, FLUSH, EXISTS, LOAD, KILL, LEN, REFCOUNT, ENCODING, IDLETIME,
-    GETNAME, SETNAME, LIST, MATCH, COUNT, PING, PONG, UNLOAD, REPLACE, KEYS, PAUSE, DOCTOR, 
-    BLOCK, NOACK, STREAMS, KEY, CREATE, MKSTREAM, SETID, DESTROY, DELCONSUMER, MAXLEN, GROUP, 
-    IDLE, TIME, RETRYCOUNT, FORCE;
+    GETNAME, SETNAME, LIST, MATCH, COUNT, UNLOAD, REPLACE, KEYS, PAUSE, DOCTOR, BLOCK, NOACK,
+    STREAMS, CREATE, MKSTREAM, SETID, DESTROY, DELCONSUMER, MAXLEN, GROUP, IDLE, RETRYCOUNT, FORCE,
+    // Response keywords
+    OK, QUEUED, message, pmessage, psubscribe, punsubscribe, subscribe, unsubscribe, pong;
```
gkorland: why do you need to have the keywords in lowercase?

sazzad16: Because Redis responds those responses in lower cases.

gkorland:
```
$ redis-cli
127.0.0.1:6379> ping
PONG
127.0.0.1:6379>
```

sazzad16: I mean, in PUB/SUB.

sazzad16: https://github.com/xetorthio/jedis/blob/c4612a379ab8bad627b28193bb44604825f4b4bd/src/main/java/redis/clients/jedis/JedisPubSub.java#L163

gkorland: so won't we have issue with `ping`?

sazzad16: Because `ping` is a request and Redis accepts both upper and lower cased letters in case of requests.

gkorland: Sorry I meant `pong`

sazzad16: You mean, why we don't have issue with `pong` right now? Because it is lower cased in 
https://github.com/xetorthio/jedis/blob/c370114e84ea0fc27a20c6d2244d150826e31e71/src/main/java/redis/clients/jedis/Protocol.java#L286-L288 

This is dependant on `Locale` which I'm thinking about not using.

PR #2021, opened 2019-07-14
Unify non-clustered and clustered interfaces

sazzad16:
An attempt to address [Common interface for <strike>binary</strike> clustered and non-clustered redis needed](https://github.com/xetorthio/jedis/issues/1545)

Inspiration: #1545 

Note: JDK 1.8 was required for using `default`.


pom.xml
```diff
@@ -101,8 +101,8 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.1</version>
 				<configuration>
-					<source>1.7</source>
-					<target>1.7</target>
+					<source>1.8</source>
```
gkorland: are we killing the 1.7 support?

src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -3737,7 +3737,7 @@ public Long xlen(final String key) {
    * {@inheritDoc}
    */
   @Override
-  public long xack(final String key, final String group, final StreamEntryID... ids) {
+  public Long xack(final String key, final String group, final StreamEntryID... ids) {
```
gkorland: I think it can cause some backward issues

pom.xml
sazzad16: May be in future??

src/main/java/redis/clients/jedis/Jedis.java
sazzad16: That's why I've added this in major release. And it's necessary to maintain common/unified interface.

src/test/java/redis/clients/jedis/tests/commands/StreamsCommandsTest.java
```diff
@@ -82,7 +82,7 @@ public void xdel() {
     assertEquals(2L, jedis.xlen("xdel-stream").longValue());
 
 
-    assertEquals(1L, jedis.xdel("xdel-stream", id1));
+    assertEquals(1L, jedis.xdel("xdel-stream", id1).longValue());
```
gkorland: here is an example of backward...

sazzad16:
@gkorland @marcosnils Please review!

@HeartSaVioR @mp911de Your comments would invaluable!

@greenrd @ptahchiev PTAL!

mp911de:
It makes sense to expose a common API of base Redis commands for data structure and server interaction. Sentinel, Transactional, Cluster (CLUSTER SLOTS, CLUSTER MEET) and such commands are a separate issue. When working with Redis, the primary API concern should be working with the actual data structures regardless of the Redis operation mode. 

The operation mode is a separate aspect. A secondary concern is administrative API calls when using a Redis connector. A stand-alone API (e.g. new Jedis(...)) is good for regular data structure API calls as well as administrative calls. 

Therefore, I appreciate keeping segregated interfaces per Redis command group (how Redis documentation groups commands) and forming composite interfaces that are intended for use by applications (RedisCommands, RedisSentinelCommands, RedisStandaloneCommands, RedisClusterCommands). 

Does this help?

Issue #1987, opened 2019-06-03
Support TLS connections in Sentinel mode


ae6rt:
## Feature Request

Redis open source will [support TLS](https://github.com/antirez/redis/pull/4855) in the not distant future.  TLS support in the Sentinel will likely follow soon thereafter.

Please implement TLS support in Jedis to support Sentinel mode.  This would include not only connecting to the Sentinel over TLS, but also connecting to the sentinel-reported master over TLS.

#### Is your feature request related to a problem? Please describe

Financial services and enterprise users of Jedis, including those using Jedis by way of Spring Data Redis, are increasingly being asked to ensure all service traffic is encrypted via TLS.  

#### Describe alternatives you've considered

* removing calls to the Sentinel API in our code, migrating to the lower level Redis API for Redis Cluster (when it supports TLS - see link above) or Redis Enterprise.  This would be painful for the number of applications we have that program to the Sentinel API.
* stunnel and spiped, neither of which work well for how our services are deployed and maintained
* exotic IPSec solutions to encrypt traffic on the wire in lieu of first class TLS support in Sentinel mode

gkorland:
@ae6rt I think the way to support it, without adding new dependencies to Jedis is by adding AddressResolver API

gkorland:
@ae6rt is #2024 is that what you need?

ae6rt:
@gkorland Yes, this seems to be the same feature request.

gkorland:
@ae6rt so does that PR good enough? can you verify it's working on your environment?

ae6rt:
@gkorland I need a few days to study my code and whether the code in #2024 slots into the way I build the connection factory.

ae6rt:
I don't have a definitive answer for whether #2024 will work for me.  However, I can say that this is the nexus of how we define our sentinel-based connection factory:

```
JedisConnectionFactory factory = new JedisConnectionFactory(new RedisSentinelConfiguration(sentinel.master, sentinel.nodes), poolConfig);
```

I don't see obvious support in the #2024 PR for me sticking with this idiom.

```

PR #2024, opened 2019-07-17
Allowed connect to redis-master with ssl.

keddar1:
When I use redis-server with sentinel I can not use safe connect to redis-server with ssl because sentinel Pool in default create not safe connection.

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisFactory.java
+++ b/src/main/java/redis/clients/jedis/JedisFactory.java
@@ -35,6 +35,11 @@ class JedisFactory implements PooledObjectFactory<Jedis> {
     this(host, port, connectionTimeout, soTimeout, password, database, clientName,
         false, null, null, null);
   }
+  JedisFactory(final String host, final int port, final int connectionTimeout,
+               final int soTimeout, final String password, final int database, final String clientName, final boolean isRedisSslEnabled) {
+    this(host, port, connectionTimeout, soTimeout, password, database, clientName,
+            isRedisSslEnabled, null, null, null);
+  }
 
   JedisFactory(final String host, final int port, final int connectionTimeout,
       final int soTimeout, final String password, final int database, final String clientName,
--- a/src/main/java/redis/clients/jedis/JedisSentinelPool.java
+++ b/src/main/java/redis/clients/jedis/JedisSentinelPool.java
@@ -149,8 +149,6 @@ public class JedisSentinelPool extends JedisPoolAbstract {
         currentHostMaster = master;
         if (factory == null) {
           factory = new JedisFactory(master.getHost(), master.getPort(), connectionTimeout,
-              soTimeout, password, database, clientName);
-          factory = new JedisFactory(master.getHost(), master.getPort(), connectionTimeout,
                   soTimeout, password, database, clientName, isRedisSslEnabled);
           initPool(poolConfig, factory);
         } else {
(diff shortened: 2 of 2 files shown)


gkorland:
#1987

src/main/java/redis/clients/jedis/JedisSentinelPool.java
```diff
@@ -117,6 +150,8 @@ private void initPool(HostAndPort master) {
         if (factory == null) {
           factory = new JedisFactory(master.getHost(), master.getPort(), connectionTimeout,
```
gkorland: It seems like you forgot to delete this row

keddar1: fixed https://github.com/xetorthio/jedis/pull/2024/commits/f232be700ec54a6b4cdf0d258b06b5db1b821c30

src/main/java/redis/clients/jedis/JedisSentinelPool.java
```diff
@@ -68,30 +82,49 @@ public JedisSentinelPool(String masterName, Set<String> sentinels,
...
+       final String password, final int database) {
+    this(masterName, sentinels, poolConfig, timeout, soTimeout, password, database, null, false);
   }
 
   public JedisSentinelPool(String masterName, Set<String> sentinels,
       final GenericObjectPoolConfig poolConfig, final int timeout, final int soTimeout,
-      final String password, final int database) {
-    this(masterName, sentinels, poolConfig, timeout, soTimeout, password, database, null);
+      final String password, final int database, final Boolean isRedisSslEnabled) {
+    this(masterName, sentinels, poolConfig, timeout, soTimeout, password, database, null, isRedisSslEnabled);
```
gkorland: Don't you think there's a need to add a constructor with `final SSLSocketFactory sslSocketFactory, final SSLParameters sslParameters, final HostnameVerifier hostnameVerifier`

keddar1: Yes. It can be useful. I added this parameters to main constructor of JedisSentinelPool class.

keddar1: @gkorland  What do you think about changes from this https://github.com/xetorthio/jedis/pull/2024/commits/aa02bac03a64ff40ac45657016a5ac4246e063e7 commit?

src/main/java/redis/clients/jedis/JedisFactory.java
```diff
@@ -35,6 +35,11 @@
     this(host, port, connectionTimeout, soTimeout, password, database, clientName,
         false, null, null, null);
   }
+  JedisFactory(final String host, final int port, final int connectionTimeout,
+               final int soTimeout, final String password, final int database, final String clientName, final boolean isRedisSslEnabled) {
```
sazzad16: Do we need this constructor? Isn't following constructor enough?

I'm asking this because JedisFactory is kept to have fewest constructors.

src/main/java/redis/clients/jedis/JedisSentinelPool.java
```diff
@@ -13,13 +13,20 @@
...
+import javax.net.ssl.SSLParameters;
+import javax.net.ssl.SSLSocketFactory;
+
 public class JedisSentinelPool extends JedisPoolAbstract {
 
   protected GenericObjectPoolConfig poolConfig;
 
   protected int connectionTimeout = Protocol.DEFAULT_TIMEOUT;
   protected int soTimeout = Protocol.DEFAULT_TIMEOUT;
-
```
sazzad16: An empty line would make code more readable.

src/main/java/redis/clients/jedis/JedisSentinelPool.java
gkorland: :+1:

gkorland:
I think we should have a test for this use case

PR #2007, opened 2019-06-24
added general send command to Jedis

DvirDukhan:
Added general "send command" functionality to Jedis and Transaction objects. 
Input: Command, and command arguments.
Output: Object, as returned by `client.getOne()`. It is the caller responsibility to handle the returned object properly.
The main motivation for this addition: add modules functionality, including sending module command in MULTI/EXEC:
1. There is no need to call `jedis.getClient()` in order to directly access and expose the binary client, in order to send a module command or 3rd party command.
2. Allowing 3rd party commands to be sent during MULTI/EXEC session

PR #2025, opened 2019-07-18
General execute command

DvirDukhan:
minor code fixes + binary commands now include general command with no args.
an update to #2007

### DIFF
--- a/src/main/java/redis/clients/jedis/commands/BinaryJedisCommands.java
+++ b/src/main/java/redis/clients/jedis/commands/BinaryJedisCommands.java
@@ -359,5 +359,5 @@ public interface BinaryJedisCommands {
 
   Object sendCommand(ProtocolCommand cmd, byte[]... args);
 
-    Object sendCommand(ProtocolCommand cmd);
+  Object sendCommand(ProtocolCommand cmd);
 }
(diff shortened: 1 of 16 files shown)


src/main/java/redis/clients/jedis/commands/BinaryJedisCommands.java
```diff
@@ -358,4 +358,6 @@
   List<byte[]> xclaim(byte[] key, byte[] groupname, byte[] consumername, long minIdleTime, long newIdleTime, int retries, boolean force, byte[][] ids);
 
   Object sendCommand(ProtocolCommand cmd, byte[]... args);
```
sazzad16: I missed this earlier. I think, we shouldn't include these here. These interfaces imitates actual Redis commands. But `sendCommand` is not a Redis command. I.e. there isn't any `SEND COMMAND` command. We may have to come up with some other solutions for this.

DvirDukhan: @sazzad16 a proposal: I'll remove the sendCommand entries from all affected interfaces. The definition and implementation will be only in the classes themselves.

sazzad16: @DvirDukhan I'm okay with that, specially in beginning. We may need a new interface for this.

@gkorland thoughts?

DvirDukhan: @sazzad16 I will open an issue of the new interface(s) required, so we can discuss it there. Meanwhile, I will remove the additional commands from the interface

PR #2028, opened 2019-07-20
Infinite Socket Timeout can be specified

### LINKED ISSUE #426: Specify connection timeout for blocking calls
In some scenarios the TCP connection in blocking calls like brpop, blpop and brpoplpush gets stalled (netstat shows it as ESTABLISHED on the client side, but gone on the server side). In this case it would be beneficial to be able to specify a connection timeout, but blocking calls set it to infinite. I know that it's necessary in issue #108, but it should be possible to be able to set this timeout at least for calls that specify a timeout.

I suggest that for calls that specify a timeout this value (or a slightly larger one) is set for the connection timeout as well.

### LINKED ISSUE #1537: Allow subscribeRetryWaitTimeMillis to be passed as argument to the JedisSentinelPool
Hello, we are using your library with 1 master, 3 slaves and 4 sentinels in total. The code looks like:
`     

        JedisSentinelPool pool = null;  
        Jedis jedis = null;
        String[] s = { ip1:port, ip2:port, ip3:port, ip4:port };
        Set<String> sentinels = new HashSet<String>();
        for (String sentinel : s) { sentinels.add(sentinel); }

        JedisPoolConfig poolConfig = new JedisPoolConfig();
        poolConfig.setMaxTotal(1);   // test configuration
        pool = new JedisSentinelPool("redis-cs-cluster", sentinels, poolConfig);
        jedis = pool.getResource();

        // Test operations
        int i = 0;
        while(true) {
	        try {
	            jedis.set("test", "a" + i);
	            i++;
	            Thread.sleep(2000);
	        } catch (Exception e) {
	            System.out.println(e.getMessage());
	            if(jedis != null) {
	            	jedis.close();
	            }	
	            jedis = pool.getResource();
		}
        }
`

After getting the Jedis instance and executing some jedis.set operations, the redis master instance died (killed by us to execute this test), and sentinel executed the failover. However, even after the failover, the MasterListener inside Jedis keeps throwing 
Lost connection to Sentinel at 10.210.43.97:16380. Sleeping 5000ms and retrying.: redis.clients.jedis.exceptions.JedisConnectionException: java.net.ConnectException: Connection refused (Connection refused)
[Server:cluster-cs]     at redis.clients.jedis.Connection.connect(Connection.java:207)
[Server:cluster-cs]     at redis.clients.jedis.BinaryClient.connect(BinaryClient.java:93)
[Server:cluster-cs]     at redis.clients.jedis.Connection.setTimeoutInfinite(Connection.java:94)
[Server:cluster-cs]     at redis.clients.jedis.Jedis.subscribe(Jedis.java:2678)
[Server:cluster-cs]     at redis.clients.jedis.JedisSentinelPool$MasterListener.run(JedisSentinelPool.java:291)
after every 5s.
We know from our application, that we won't be able to reactivate that particular instance in less than 1 hour, causing too many messages to be added to the log file.

Is it possible to add subscribeRetryWaitTimeMillis as argument to some constructor of JedisSentinelPool?

We are using Java 8, Jedis 2.9.0 and Redis 3.2.8.

Thank you for your attention!

### LINKED ISSUE #2027: Why Jedis uses setTimeoutInfinite when it runs eval&evalSha command?
Could you please explain why Jedis uses setTimeoutInfinite (socket.setSoTimeout(0);) when it runs eval&evalSha command?
https://github.com/xetorthio/jedis/blob/c9c118aa86a1ff473b94dddd7fb28d640837e194/src/main/java/redis/clients/jedis/BinaryJedis.java#L3317

I faced with an issue when thread just stuck here..

### LINKED ISSUE #2217: [BUG REPORT] The Command Jedis#blpop may block the thread forever.
### Expected behavior

We should optimize the api Jedis#blpop and the other api with block feature. When we set the timeout parameter with the value 3 seconds for example, we expect it will return in 3 seconds, but in some cases(like network jitter, redis's bug with some version and so on), it will block your thread forever.

We hope the timeout parameter set in the command with block feature will effect in socket timeout. With now, we see inside this command, socket timeout is set with 0, this may be a dangerous operation.

### Actual behavior

A problem happened in our application. We use the api Jedis#blpop with timeout 0 to pull value in the redis list, but with the impact of network jitter, the connection closed, and our thread block forever. Without pulling from the redis list, the memory of the redis server running out in hours.

### Steps to reproduce:

1. Use`redis.clients.jedis.Jedis#blpop(int, java.lang.String)` to pull value in the redis list.
2. In the method `redis.clients.jedis.Connection#sendCommand(redis.clients.jedis.commands.ProtocolCommand, byte[]...)`, we close the connection before the method `redis.clients.jedis.Protocol#sendCommand()` is executed.
3. Push value to the redis list, but `redis.clients.jedis.Jedis#blpop(int, java.lang.String)` cannot pull the value, but block forever.

#### Jedis version:
3.3.0
#### Redis version:
6.0.5
#### Java version:
1.8.0_91

Thanks for reading.

sazzad16:
Resolves #426 
Resolves #1537 
Resolves #2027 
Closes #2207 
Resolves #2217


gkorland:
@sazzad16 shouldn't just overloading `public void setTimeoutInfinite() ` with timeout parameter will be better?  `public void setTimeoutInfinite(long timeout) `

sazzad16:
@gkorland How would we get the `timeout` to call `setTimeoutInfinite(timeout)`? I can see two other options and I prefer current implementation over both of those.

- Repeat all the methods that calls `setTimeoutInfinite()` with a new timeout parameter.

- Hold `timeout` in a variable in BinaryJedis. Still we'll need all those constructors. And (Binary)Jedis hardly holds any parameter.

rominzxx:
Hey, can someone review this? It's an important fix.

rominzxx (dismissed):

tokenmathguy:
+1 to the fix @sazzad16 @gkorland 

It looks like there are some merge conflicts. Anything the community can assist with?

mkozioro:
Hi @sazzad16 
Issue that is fixed by this PR is sometimes still occurring in our system. Do you think it will be merged some day?

sazzad16:
@gkorland PING

gkorland (approved):

mkozioro:
Oh my god. After so much time finally merged. It's a great day. :)

Issue #1614, opened 2017-09-29
Use  unwatch between multi and exec, and throw NullPointException when executing Exec


dismasson:
### Expected behavior
Unwatch can be exec before execution

### Actual behavior
Use  unwatch between multi and exec, and throw NullPointException when executing Exec

### Steps to reproduce:
        jedis.watch("key");
        Transaction transaction = jedis.multi();
        jedis.unwatch();
        transaction.set("key","value");
        transaction.exec();
#### Jedis version:
2.9.0
#### Java vcersion:
1.7

sazzad16:
Resolved by #2031 and #2032

PR #2031, opened 2019-07-22
More Multi or Pipeline check

sazzad16:
First half of solution of #1614

### DIFF
--- a/src/main/java/redis/clients/jedis/Jedis.java
+++ b/src/main/java/redis/clients/jedis/Jedis.java
@@ -252,6 +252,7 @@ public class Jedis extends BinaryJedis implements JedisCommands, MultiKeyCommand
 
   @Override
   public Long unlink(final String key) {
+    checkIsInMultiOrPipeline();
     client.unlink(key);
     return client.getIntegerReply();
   }
@@ -1677,6 +1678,7 @@ public class Jedis extends BinaryJedis implements JedisCommands, MultiKeyCommand
 
   @Override
   public String watch(final String... keys) {
+    checkIsInMultiOrPipeline();
     client.watch(keys);
     return client.getStatusCodeReply();
   }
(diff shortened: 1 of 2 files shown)


src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -2826,6 +2821,18 @@ public void psubscribe(final JedisPubSub jedisPubSub, final String... patterns)
     return params;
   }
 
+  @Override
+  public Object eval(final String script, final int keyCount, final String... params) {
+    checkIsInMultiOrPipeline();
+    client.eval(script, keyCount, params);
+    client.setTimeoutInfinite();
```
gkorland: don't you think it's better to move `client.setTimeoutInfinite();` into the try block?

sazzad16: No. `setTimeoutInfinite()` has it's own try-catch. And if there is any exception, it is decided to be broken. There is no gain to `rollbackTimeout()` a broken connection.

gkorland:
@sazzad16 why is it on 3.1.1? should we move it to 3.2.0?

sazzad16:
Because it's **backwards-compatible bug fix**.

sazzad16:
@gkorland 

> why is it on 3.1.1?

Because it's **backwards-compatible bug fix**.

> should we move it to 3.2.0?

Only if you're plan right now is not to have a release 3.1.1.

PS: In that case, we should drop third number while versioning.

gkorland:
I think the third number should be used for "service packs" releasing critical bug fixes for existing version.
The second should be used for "feature version" 
And the first for major versions with breaking backward compatibility.

If you think this issue is critical bug fix on top of 3.1.0 then it should be part of 3.1.1

sazzad16:
I saw may be two issues related this. So I won't call it *"critical"*.

Again, if you've decided that bug fix releases would be only for critical bugs, IMHO, it may be off dropping that third number.

PR #2032, opened 2019-07-22
Add missing UNWATCH in Transaction and Pipeline

sazzad16:
Second half of solution of #1614


gkorland:
@sazzad16 it seems like it conflicts with #2033

sazzad16:
@gkorland Yes, it does. If we merge anyone, I'll change other one accordingly.

Issue #2039, opened 2019-07-30
Upgrade to Commons Pool 2.7.0


mp911de:
This is a dependency upgrade ticket.

sazzad16:
Closing in favor of #2165

PR #2040, opened 2019-07-30
upgrade commons-pool2 to 2.7.0

gkorland:


gkorland:
fix #2039

sazzad16:
@gkorland Well, commons-pool2 2.7.0 doesn't support JDK 1.7! How about that?

gkorland:
oops I missed it, in that case I'll put it on 4.0.0 milestone

sazzad16:
@gkorland Does it mean, we are not going to support JDK 1.7 from `4.x`?

gkorland:
I think it should be opened for discussion, what do you think?

sazzad16:
@gkorland Check #1863

MichalCho:
I think JDK 7 can be skipped from `4.x`. Oracle has stopped supporting it and even OpenJDK 7 goes end of support in June 2020

sazzad16:
Closing in favor of #2165

PR #2053, opened 2019-09-03
Wait strategies (2.10)

nosqlgeek:
Hi, this is adding some wait strategies to Jedis 2.10.x . The main purpose was to have a smarter way to wait in case that a Sentinel is unavailable for a specific period of time. This especially useful if it's not possible to pass multiple Sentinel endpoints over for the initial connection. Here an example:

1. Let's assume that Redis is provided as DBaaS by exposing exactly one endpoint
2. Let's further assume that we can reach a Sentinel behind the FQN of this endpoint (on a different port than the DB endpoint itself)

We would like to be able to survive the failover of such a DB endpoint by then being able to fetch the DB master again via Sentinel. Such a failover can take longer than the default amount of time in the original source code which is the reason why we decided to replace the simple Thread.sleep by a wait strategy.


sazzad16:
Closed due to inactivity.

PR #2062, opened 2019-09-17
Remove non-ascii colon in JedisSentintelPool

NickFirmani:
This character will fail to compile unless compiled with the additional compiler option `-Dfile.encoding=UTF8`

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisSentinelPool.java
+++ b/src/main/java/redis/clients/jedis/JedisSentinelPool.java
@@ -287,8 +287,8 @@ public class JedisSentinelPool extends JedisPoolAbstract {
            */
           List<String> masterAddr = j.sentinelGetMasterAddrByName(masterName);  
           if (masterAddr == null || masterAddr.size() != 2) {
-            log.warn("Can not get master addr, master name: {}. Sentinel: {}：{}.",masterName,host,port);
-          }else{
+            log.warn("Can not get master addr, master name: {}. Sentinel: {}: {}.",masterName,host,port);
+          } else{
               initPool(toHostAndPort(masterAddr)); 
           }
 


src/main/java/redis/clients/jedis/JedisSentinelPool.java
```diff
@@ -287,8 +287,8 @@ public void run() {
            */
           List<String> masterAddr = j.sentinelGetMasterAddrByName(masterName);  
           if (masterAddr == null || masterAddr.size() != 2) {
-            log.warn("Can not get master addr, master name: {}. Sentinel: {}：{}.",masterName,host,port);
-          }else{
+            log.warn("Can not get master addr, master name: {}. Sentinel: {}: {}.",masterName,host,port);
```
gkorland: why did you add a white space? where is none ascii?

NickFirmani: The colon is the non ascii character here, GitHub just renders it as such. Before `：` after `: `

src/main/java/redis/clients/jedis/JedisSentinelPool.java
```diff
@@ -287,8 +287,8 @@ public void run() {
...
+          } else{
```
sazzad16: Please remove wrongful and/or unnecessary spaces.

sazzad16:
@gkorland We should not merge this right now, in https://github.com/xetorthio/jedis/pull/2062/files#diff-e6525ac1a247b59a49332a6584030684R290 it changes `host:port` to `host:<space>port`. This should not happen.

sazzad16:
@gkorland Please see #2107

gkorland:
@sazzad16 Ok I see your point, let's merge the other PR and close this one.

PR #2067, opened 2019-09-24
return list with null values fixes #2066

### LINKED ISSUE #2066: ERR wrong number of arguments for 'mget' command jedis
### Expected behavior
Calling `Jedis#mget()` with an empty array for keys i.e `Jedis#mget(new String[0]);` should rather return an empty list or list with single null value (which is more in line with it's documentation) rather than throwing an exception

### Actual behavior
It throws the exception below

```
redis.clients.jedis.exceptions.JedisDataException: ERR wrong number of arguments for 'mget' command
	at redis.clients.jedis.Protocol.processError(Protocol.java:127)
	at redis.clients.jedis.Protocol.process(Protocol.java:161)
	at redis.clients.jedis.Protocol.read(Protocol.java:215)
	at redis.clients.jedis.Connection.readProtocolWithCheckingBroken(Connection.java:340)
	at redis.clients.jedis.Connection.getBinaryMultiBulkReply(Connection.java:276)
	at redis.clients.jedis.Connection.getMultiBulkReply(Connection.java:269)
	at redis.clients.jedis.Jedis.mget(Jedis.java:407)
```

I know that this is what redis returns, but I believe that returning an empty list rather than an exception is one of the benefits of a library, also this would avoid hitting redis.

`Jedis.mget(new byte[0])` returns a list with a `null` element

### Steps to reproduce:

1. call `Jedis#mget(new String[0]);`

#### Java version:  1.8.0_202
Java(TM) SE Runtime Environment (build 1.8.0_202-b08)
Java HotSpot(TM) 64-Bit Server VM (build 25.202-b08, mixed mode)

Happy to submit a PR for this

IEnoobong:
Documentation of `Jedis.mget(String ...)` states that
> If one or more keys don't exist or is not of type String, a 'nil' value is returned

However calling the method with an empty array of keys throws a `JedisDataException` as seen in #2066 

This fixes it

### DIFF
--- a/src/main/java/redis/clients/jedis/Jedis.java
+++ b/src/main/java/redis/clients/jedis/Jedis.java
@@ -1,34 +1,16 @@
 package redis.clients.jedis;
 
-import java.net.URI;
-import java.util.AbstractMap;
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Map.Entry;
-import java.util.Set;
+import redis.clients.jedis.commands.*;
+import redis.clients.jedis.params.*;
+import redis.clients.jedis.util.SafeEncoder;
+import redis.clients.jedis.util.Slowlog;
 
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLParameters;
 import javax.net.ssl.SSLSocketFactory;
-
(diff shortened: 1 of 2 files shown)


sazzad16:
This is against [Redis Protocol specification](https://redis.io/topics/protocol). Closing.

PR #2069, opened 2019-10-01
Xinfo command

MichalCho:
Hi,

I've noticed that xinfo command is missing so I've started to implement it. 
This is just to start a discussion about how this should be fixed.
Currently I've implemented a single xinfo command that returns a map of <String,Objects>.
This would work for any variant of xinfo (stream,groups, consumers) but it's a bit generic. 

The other idea would be to create 3 xinfo commands, returning something like XInfoStream, XInfoConsumers etc. objects. That would give a better visibility of returned values.
I will be happy to work on it.


MichalCho:
@sazzad16 @marcosnils @gkorland could you give me some input?
Thanks

sazzad16:
My initial thought is that `Params` is a better fit for this.

MichalCho:
> My initial thought is that `Params` is a better fit for this.

But this will not solve the problem of returning very generic object. 
`jedis.xinfo(myKey, STREAM)` and `jedis.xinfo(myKey, CONSUMERS, myGroup)` would return the same type - just a map. Then user would to need to know the implementation to actually check the returned values example:

`jedis.xinfo(myKey, STREAM).get("first-entry);` - they would need to know that this is a `StreamElement` and do casting in their code.

What I had in mind is to have 3 functions returning different types which expose more clear getters
```
XStreamInfo streamInfo = jedis.xinfo(myKey, STREAM); //Type specific to information about a stream
streamInfo.getFirstEntry(); //returned type is StreamElement, no need for casting;
```

Other option would be to return something like `XInfoElement` that contains 3 fields:
```
XInfoStream
XinfoGroups
XInfoConsumers
```
exposing `getXInfoStream`, `getXInfoGroups` and `XInfoConsumers`.
They would be set when a specific info is requested. That would require if-not-null check.

gkorland:
Sorry my mistake... should be draft

MichalCho:
> Sorry my mistake... should be draft

No problem, I'm working on the new implementation, I hope I can have some commits ready soon

gkorland:
@MichalCho sorry to nudge, but are still working on this PR?

MichalCho:
@gkorland no problem. Sorry I got quite busy at the office. I will try to use holiday break now to craft some code. Will you be willing to do some code review?

gkorland:
@MichalCho no pressure, I just wanted to check if it's still relevant for you

startjava:
@MichalCho   sorry,,,i want know xinfo() method Progress？？ :) :) :D

MichalCho:
> @MichalCho sorry,,,i want know xinfo() method Progress？？ :) :) :D

Hi @startjava 
I will try to have some commits ready for review after New Year's Eve.
Are you waiting for this feature? :)

startjava:
@MichalCho 
not waiting this feature，But Always waiting this feature  :）:)   :D

MichalCho:
Hi,
@gkorland I've just pushed my initial idea how to implement this. It require a lot more testing (I've just created one happy case test case). Do you think you could have a look just at it  to give me feedback what do you think about usability? Then I can continue to work on it after New Year.
Thanks

src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -1,13 +1,15 @@
 package redis.clients.jedis;
 
 import java.util.ArrayList;
+import java.util.Collections;
```
sazzad16: Not used.

src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -527,6 +529,41 @@ public String toString() {
...
+      }
+      StreamEntry streamEntry = new StreamEntry(entryID, map);
+
+
+      return streamEntry;
+    }
+
+    @Override
+    public String toString() {
+      return "List<StreamEntry>";
```
sazzad16: Not List.

src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -1,13 +1,15 @@
...
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Stream;
```
sazzad16: Not used?

src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -555,6 +592,143 @@ public String toString() {
...
 
+  public static final Builder<StreamInfo> STREAM_INFO = new Builder<StreamInfo>() {
+    @Override
+    @SuppressWarnings("unchecked")
+    public StreamInfo build(Object data) {
+      if (null == data) {
+        return null;
+      }
+
+      Map<String,Builder> mappingFunctions = new HashMap<>();
```
sazzad16: You're creating this map during every object creation. Create it once and use rest of the time.

src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -555,6 +592,143 @@ public String toString() {
...
+
+
+
+
+      //throw new RuntimeException("Not implemented yet");
+    }
+
+    @Override
+    public String toString() {
+      return "Map<Key,StreamInfoElement>";
```
sazzad16: Doesn't match.

src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -555,6 +592,143 @@ public String toString() {
...
```
sazzad16: Doesn't match.

src/main/java/redis/clients/jedis/ShardedJedis.java
```diff
@@ -14,6 +14,7 @@
 import redis.clients.jedis.params.ZAddParams;
 import redis.clients.jedis.params.ZIncrByParams;
 import redis.clients.jedis.util.Hashing;
+import sun.reflect.generics.reflectiveObjects.NotImplementedException;
```
sazzad16: Not used.

src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -3830,6 +3830,48 @@ public long xtrim(final String key, final long maxLen, final boolean approximate
...
+    StreamInfo resultMap;
+    client.xinfo(key,type);
+    List<Object> streamsEntries = client.getObjectMultiBulkReply();
+    if(streamsEntries == null) {
+      return null;
+    }
+    resultMap = BuilderFactory.STREAM_INFO.build(streamsEntries);
+
+
+    return resultMap;
```
sazzad16: What if these lines are reduced into 
```
    client.xinfo(key,type);
    List<Object> streamEntries = client.getObjectMultiBulkReply();
    return BuilderFactory.STREAM_INFO.build(streamEntries);
```

Or, further reduced 
```
    client.xinfo(key,type);
    return BuilderFactory.STREAM_INFO.build(client.getObjectMultiBulkReply());
```

src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -3830,6 +3830,48 @@ public long xtrim(final String key, final long maxLen, final boolean approximate
...
+    List<StreamGroupInfo> resultMap;
+    client.xinfo(key,type);
+    List<Object> streamsEntries = client.getObjectMultiBulkReply();
+    if(streamsEntries == null) {
+      return null;
+    }
+    resultMap = BuilderFactory.STREAM_GROUP_INFO.build(streamsEntries);
+
+
+    return resultMap;
```
sazzad16: See comment on lines +3835 to +3844

src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -3830,6 +3830,48 @@ public long xtrim(final String key, final long maxLen, final boolean approximate
...
+    List<StreamConsumersInfo> resultMap;
+    client.xinfo(key,group,type);
+    List<Object> streamsEntries = client.getObjectMultiBulkReply();
+    if(streamsEntries == null) {
+      return null;
+    }
+    resultMap = BuilderFactory.STREAM_CONSUMERS_INFO.build(streamsEntries);
+
+
+    return resultMap;
```
sazzad16: See comment on lines +3835 to +3844

src/main/java/redis/clients/jedis/Client.java
```diff
@@ -1262,5 +1262,29 @@ public void xclaim(String key, String group, String consumername, long minIdleTi
     xclaim(SafeEncoder.encode(key), SafeEncoder.encode(group), SafeEncoder.encode(consumername), minIdleTime, newIdleTime, retries, force, bids);    
   }
 
+  @Override
+  public void xinfo(String key, StreamInfo.StreamInfoType type) {
+
+
+    xinfo(SafeEncoder.encode(key),SafeEncoder.encode(StreamInfo.STREAM_INFO));
+
```
sazzad16: Please try to avoid too many blank lines.

src/main/java/redis/clients/jedis/Client.java
```diff
@@ -1262,5 +1262,29 @@ public void xclaim(String key, String group, String consumername, long minIdleTi
...
+  }
+
+  @Override
+  public void xinfo(String key, StreamGroupInfo.StreamGroupInfoType type) {
+
+
+    xinfo(SafeEncoder.encode(key),SafeEncoder.encode(StreamGroupInfo.GROUP_INFO));
+
```
sazzad16: Please try to avoid too many blank lines.

src/main/java/redis/clients/jedis/Client.java
```diff
@@ -1262,5 +1262,29 @@ public void xclaim(String key, String group, String consumername, long minIdleTi
...
+  }
+
+  @Override
+  public void xinfo(String key, String group, StreamConsumersInfo.StreamConsumersInfoType type) {
+
+
+    xinfo(SafeEncoder.encode(key),SafeEncoder.encode(group),SafeEncoder.encode(StreamConsumersInfo.CONSUMERS_INFO));
+
```
sazzad16: Please try to avoid too many blank lines.

src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -3830,6 +3830,48 @@ public long xtrim(final String key, final long maxLen, final boolean approximate
     return BuilderFactory.STREAM_ENTRY_LIST.build(client.getObjectMultiBulkReply());
   }
 
+  @Override
+  public StreamInfo xinfo(String key, StreamInfo.StreamInfoType type) {
```
sazzad16: Put type first. Imitate [Redis order](https://redis.io/commands/xinfo) as much as possible.

src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -3830,6 +3830,48 @@ public long xtrim(final String key, final long maxLen, final boolean approximate
...
+      return null;
+    }
+    resultMap = BuilderFactory.STREAM_INFO.build(streamsEntries);
+
+
+    return resultMap;
+  }
+
+  @Override
+  public List<StreamGroupInfo> xinfo(String key, StreamGroupInfo.StreamGroupInfoType type) {
```
sazzad16: See comment on line +3834

src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -3830,6 +3830,48 @@ public long xtrim(final String key, final long maxLen, final boolean approximate
...
+      return null;
+    }
+    resultMap = BuilderFactory.STREAM_GROUP_INFO.build(streamsEntries);
+
+
+    return resultMap;
+  }
+
+  @Override
+  public List<StreamConsumersInfo> xinfo(String key, String group, StreamConsumersInfo.StreamConsumersInfoType type) {
```
sazzad16: See comment on line +3834

src/main/java/redis/clients/jedis/BinaryClient.java
```diff
@@ -1455,4 +1455,22 @@ public void xclaim(byte[] key, byte[] groupname, byte[] consumername, long minId
       sendCommand(XCLAIM, arguments.toArray(new byte[arguments.size()][]));
   }
 
+  public void xinfo(byte[] key, byte[] type) {
```
sazzad16: Change order: type, key

src/main/java/redis/clients/jedis/BinaryClient.java
```diff
@@ -1455,4 +1455,22 @@ public void xclaim(byte[] key, byte[] groupname, byte[] consumername, long minId
...
+
+    ArrayList<byte[]> arguments = new ArrayList<>(10);
+    arguments.add(type);
+    arguments.add(key);
+    sendCommand(XINFO,arguments.toArray(new byte[arguments.size()][]));
+
+  }
+
+  public void xinfo (byte[] key, byte[] group, byte[] type) {
```
sazzad16: Change order: type, key, group

src/main/java/redis/clients/jedis/BinaryClient.java
```diff
@@ -1455,4 +1455,22 @@ public void xclaim(byte[] key, byte[] groupname, byte[] consumername, long minId
...
+
+    ArrayList<byte[]> arguments = new ArrayList<>(10);
+    arguments.add(type);
+    arguments.add(key);
+    arguments.add(group);
+    sendCommand(XINFO,arguments.toArray(new byte[arguments.size()][]));
```
sazzad16: Simply 
```
    sendCommand(XINFO, type, key, group);
```

sazzad16:
@MichalCho Even though I've put some review comments, I'm liking the idea. Are we complicating it too much? As the [Redis documentation for XINFO](https://redis.io/commands/xinfo) says

> a well behaving client should fetch the whole list, and report it to the user, for example, **as a dictionary data structure**

Perhaps one method (per class) returning a `Map` is enough? @gkorland WDYT?

MichalCho:
@sazzad16 thanks a lot for the review. Regarding formatting comments - I was aware of most of them, I just wanted to show something before leaving for holidays :)


> @MichalCho Even though I've put some review comments, I'm liking the idea. Are we complicating it too much? As the [Redis documentation for XINFO](https://redis.io/commands/xinfo) says
> 
> > a well behaving client should fetch the whole list, and report it to the user, for example, **as a dictionary data structure**
> 
> Perhaps one method (per class) returning a `Map` is enough? @gkorland WDYT?

I was thinking about it. Then I had an idea to provide getters for all map members. Otherwise user will have to try to get on Object from the map and he will have to know what kind of object it is. 

something like (if xinfo returns just a map):
`Long streamLenght info = (Long)jedis.xinfo("xadd-stream1", StreamInfo.StreamInfoType.getStreamInfoType()).get("lenght");`

Another issue that I had in mind is - when we decode the answer we can not be fully flexible as we need to know how to decode the answer (to know that "length" is a integer, "last-generated-id" is a string and "first-entry" is a stream entry object).
One solution would be to try to decode first element that is a name of a field (let's say "length") and then try to decode next element using known type (let's say first try to decode as a string, if we fail, then try to decode as stream object, then as a integer etc) until we succeed. If we fail we throw some kind of decoding exception.
I think that would be the only way to be fully transparent and able to react on changes done on redis side. 
I will try to fix the stuff you have commented on and also craft my new approach for decoding an answer from redis. in few days. In the meantime if you have some thought about it, please let me know.

startjava:
@MichalCho hi~ what time xinfo method final over? thank you job !

MichalCho:
hi @sazzad16!

I've implemented your comments and changed the returned type just to be `Map<String,Object>`

During the weekend I will try to add more test cases and maybe make the code prettier.
Looking forward to any comments from you and @gkorland 
Thanks

PS @startjava it's ongoing, you can check the latest commit

MichalCho:
Hi @sazzad16 and @gkorland 

I've added more tests and worked on the review comments. I think it's ready for a final review from you guys.

Please share your thoughts.
Thanks!

src/main/java/redis/clients/jedis/BinaryClient.java
```diff
@@ -1455,4 +1455,22 @@ public void xclaim(byte[] key, byte[] groupname, byte[] consumername, long minId
       sendCommand(XCLAIM, arguments.toArray(new byte[arguments.size()][]));
   }
 
+  public void xinfo(byte[] type, byte[] key) {
+
+    ArrayList<byte[]> arguments = new ArrayList<>(10);
+    arguments.add(type);
+    arguments.add(key);
```
sazzad16: Note needed.

src/main/java/redis/clients/jedis/BinaryClient.java
```diff
@@ -1455,4 +1455,22 @@ public void xclaim(byte[] key, byte[] groupname, byte[] consumername, long minId
...
+    sendCommand(XINFO,type,key);
+
+  }
+
+  public void xinfo ( byte[] type, byte[] key, byte[] group) {
+
+    ArrayList<byte[]> arguments = new ArrayList<>(10);
+    arguments.add(type);
+    arguments.add(key);
+    arguments.add(group);
```
sazzad16: Note needed.

src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -527,6 +527,41 @@ public String toString() {
...
   };
+
+  public static final Builder<StreamEntry> STREAM_ENTRY = new Builder<StreamEntry>() {
+    @Override
+    @SuppressWarnings("unchecked")
+    public  StreamEntry build(Object data) {
+      if (null == data) {
+        return null;
+      }
+      ArrayList<Object> objectList = (ArrayList<Object>) data;
```
sazzad16: Shouldn't it be `List` instead of ArrayList? Under the hood, it is ArrayList but it isn't *guaranteed* to be so.

src/main/java/redis/clients/jedis/StreamGroupInfo.java
```diff
@@ -0,0 +1,61 @@
...
+  public final static String CONSUMERS = "consumers";
+  public final static String PENDING = "pending";
+  public final static String LAST_DELIVERED = "last-delivered-id";
+  public static final String GROUP_INFO = "groups";
+
+
+  /*private final String name;
+  private final long consumers;
+  private final long pending;
+  private final String lastDeliveredId;*/
```
sazzad16: May be, keep these variables? Along with the map, as well as in other classes.
The variables will ease implementation. The map would help during changes in Redis.
WDYT?

src/test/java/redis/clients/jedis/tests/commands/StreamsCommandsTest.java
```diff
@@ -306,6 +305,48 @@ public void xpendeing() {
...
+    Entry<String, StreamEntryID> streamQeury11 = new AbstractMap.SimpleImmutableEntry<>("xadd-stream1", new StreamEntryID("0-0"));
+
+    jedis.xreadGroup("G1", "myConsumer",1,0,false,streamQeury11);
+
+
+    List<StreamGroupInfo> info = jedis.xinfo(StreamGroupInfo.StreamGroupInfoType.getStreamGroupInfoType(),"xadd-stream1");
+    List<StreamConsumersInfo> consumersInfos = jedis.xinfo(StreamConsumersInfo.StreamConsumersInfoType.getStreamGroupInfoType(),
+        "xadd-stream1", "G1");
+
+    System.out.println(info.get(0).getGroupInfo().get("name"));
```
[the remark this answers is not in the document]
MichalCho: This has already been updated

src/main/java/redis/clients/jedis/StreamGroupInfo.java
MichalCho: Do you mean to keep them together with getters? So classes provide both `getXYZInfo()` method returning map and getters for already known fields? I like this idea.

src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -555,6 +590,202 @@ public String toString() {
...
+      return streamInfo;
+    }
+
+    @Override
+    public String toString() {
+      return "StreamInfo";
+    }
+  };
+
+  public static final Builder<List<StreamGroupInfo>> STREAM_GROUP_INFO = new Builder<List<StreamGroupInfo>>() {
```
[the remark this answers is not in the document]
MichalCho: done

src/main/java/redis/clients/jedis/BinaryClient.java
MichalCho: done

src/main/java/redis/clients/jedis/BinaryClient.java
MichalCho: done

src/main/java/redis/clients/jedis/BuilderFactory.java
MichalCho: done

MichalCho:
I've just committed changes according to review comments. 

I've restored getters for map elements, that are known to be present in a reply as per today.
In case some more elements are added in the future they can still be accessed through the map.

Still the question is - how shall we try to decode those unknown elements. As for now we try use a set of decoding functions that we know are needed to decode specific type of answer. Should all decoding function should be added? 
Anyway I feel we are getting closer to finalize that PR

startjava:
please speed final commit       I Wait long time------------------ 原始邮件 ------------------
发件人: "MichalCho"<notifications@github.com>
发送时间: 2020年1月14日(星期二) 晚上6:14
收件人: "xetorthio/jedis"<jedis@noreply.github.com>;
抄送: "startjava"<279377921@qq.com>;"Mention"<mention@noreply.github.com>;
主题: Re: [xetorthio/jedis] Xinfo command (#2069)



I've just committed changes according to review comments.
 
I've restored getters for map elements, that are known to be present in a reply as per today.
 In case some more elements are added in the future they can still be accessed through the map.
 
Still the question is - how shall we try to decode those unknown elements. As for now we try use a set of decoding functions that we know are needed to decode specific type of answer. Should all decoding function should be added?
 Anyway I feel we are getting closer to finalize that PR
 
&mdash;

src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -514,7 +514,7 @@ public String toString() {
         Iterator<byte[]> hashIterator = hash.iterator();
         Map<String, String> map = new HashMap<>(hash.size()/2);
         while(hashIterator.hasNext()) {
-          map.put(SafeEncoder.encode((byte[])hashIterator.next()), SafeEncoder.encode((byte[])hashIterator.next()));
+          map.put(SafeEncoder.encode(hashIterator.next()), SafeEncoder.encode((byte[])hashIterator.next()));
```
sazzad16: Either change none or change both.

src/main/java/redis/clients/jedis/StreamGroupInfo.java
sazzad16: @MichalCho 

> Do you mean to keep them together with getters? So classes provide both getXYZInfo() method returning map and getters for already known fields?

YES.

src/main/java/redis/clients/jedis/BuilderFactory.java
MichalCho: done

src/main/java/redis/clients/jedis/StreamGroupInfo.java
MichalCho: ok, the getters are there. And constants can be used to access map elements.
Thanks

MichalCho:
Hi @gkorland,

sorry for disturbing but do you think you will have some time anytime soon to review the changes? I'm thinking about adding some java docs to make it easier for users to start working with xinfo

startjava:
Not done yet?   It's almost two months.  :(

src/main/java/redis/clients/jedis/StreamInfo.java
```diff
@@ -0,0 +1,86 @@
...
+  public Map<String,Object> getStreamInfo() {
+    return streamInfo;
+  }
+
+  public static class StreamInfoType {
+
+    private static final StreamInfoType streamInfoType = new StreamInfoType();
+    private StreamInfoType() {
+      //Should not be used
+    };
```
gkorland: `;`?

gkorland:
I'm sorry it took me so long to really review the code, but I think we should rethink the design and really simplified it. i.e. the different `xinfo` methods/commands are practically different methods.
I would have define three methods 

```java
  @Override
  public StreamInfo xinfoStream(String key) {
    client.xinfoStream(key);

    return BuilderFactory.STREAM_INFO.build(client.getObjectMultiBulkReply());

  }

  @Override
  public List<StreamGroupInfo> xinfoGroup(String key) {
    client.xinfoGroup(key);

    return BuilderFactory.STREAM_GROUP_INFO_LIST.build(client.getObjectMultiBulkReply());

  }

  @Override
  public List<StreamConsumersInfo> xinfoConsumers(String key, String group) {
    client.xinfoConsumers(type,key,group);

    return BuilderFactory.STREAM_CONSUMERS_INFO_LIST.build(client.getObjectMultiBulkReply());

  }
```

MichalCho:
HI @gkorland,
thanks for your input. Regarding simplifying - I totally agree, it will be simpler and easier solution to use.
I've come up with my initial design just because I thought that we want to keep the same methods' names as one can use in redis cli (one xinfo command with different arguments).
Do you agree @sazzad16 ?
I will refactor asap.

MichalCho:
hi @gkorland,

I've implemented your suggestion - splitting xinfo into 3 different methods. I've also added some simple javadocs to make it easier for a user to understand how to use xinfo with jedis.

I had some doubts regarding how I create info objects, for example:

```
if (map!= null) {
     streamInfo = map;
     length = (Long) map.get(LENGTH);  
```
This can fail if length is not present in the map( for example future version of redis will remove this filed). But according to redis documentation:

> Note that you should not rely on the fields exact position, nor on the number of fields, new fields may be added in the future. 

so I guess it's save to assume that fields that exist now will always be there. Nothing is mentioned about removing fields.

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -80,6 +80,9 @@
   public static final byte[] POSITIVE_INFINITY_BYTES = "+inf".getBytes();
   public static final byte[] NEGATIVE_INFINITY_BYTES = "-inf".getBytes();
 
+  public static final String STREAM = "STREAM";
+  public static final String GROUPS = "GROUPS";
+  public static final String CONSUMERS = "CONSUMERS";
```
sazzad16: Prefix these, perhaps with `XINFO_`. Also put these along with `SENTINEL_...`, `CLUSTER_...`, `PUBSUB_...` variables.

src/main/java/redis/clients/jedis/commands/BinaryJedisCommands.java
```diff
@@ -356,4 +356,10 @@
   List<byte[]> xpending(byte[] key, byte[] groupname, byte[] start, byte[] end, int count, byte[] consumername);
 
   List<byte[]> xclaim(byte[] key, byte[] groupname, byte[] consumername, long minIdleTime, long newIdleTime, int retries, boolean force, byte[][] ids);
+
+  byte[] xinfoStream (byte[] key);
+
+  List<byte[]> xinfoGroup (byte[] key);
+
+  List<byte[]> xinfoConsumers (byte[] key, byte[] group);
```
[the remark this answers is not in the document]
MichalCho: I think we need some aligment here. I can see that some `BinaryJedisCommands` returns Objects (or Collections of Objects) - in some cases using builder factory directly like 
```@Override
  public List<GeoRadiusResponse> georadiusByMember(final byte[] key, final byte[] member, final double radius,
      final GeoUnit unit, final GeoRadiusParam param) {
    checkIsInMultiOrPipeline();
    client.georadiusByMember(key, member, radius, unit, param);
    return BuilderFactory.GEORADIUS_WITH_PARAMS_RESULT.build(client.getObjectMultiBulkReply());
  }
```
some other returns byte arrays or collection of byte arrays etc.
Maybe I missunderstood the purpose of binary client - I thought it was supposed to use just binary data, without trying to decode them. Could you please help me to understand that?

If we should return objects then I would propose following:
``` Object xinfoStream (byte[] key);

  List<Object> xinfoGroup (byte[] key);

  List<Object> xinfoConsumers (byte[] key, byte[] group);
```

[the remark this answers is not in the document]
sazzad16: Following would be good enough:

```
StreamInfo xinfoStream (byte[] key);
List<StreamGroupInfo> xinfoGroup (byte[] key);
List<StreamConsumersInfo> xinfoConsumers (byte[] key, byte[] group);
```

src/main/java/redis/clients/jedis/Protocol.java
MichalCho: done

src/main/java/redis/clients/jedis/commands/BinaryJedisCommands.java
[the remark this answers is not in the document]
MichalCho: I've refactored it that way and added a test for binary usage.

src/main/java/redis/clients/jedis/StreamInfo.java
MichalCho: The initial idea was to use that type just to overload xinfo function. I've made a constructor private to avoid creating this objects by users. Anyway it's not used anymore

MichalCho:
Pinging  @sazzad16 @gkorland :)

src/test/java/redis/clients/jedis/tests/commands/StreamsCommandsTest.java
```diff
@@ -306,6 +319,166 @@ public void xpendeing() {
...
+    assertNotNull(id2);
+
+    jedis.xgroupCreate(STREAM_NAME,G1, StreamEntryID.LAST_ENTRY,false);
+    Entry<String, StreamEntryID> streamQeury11 = new AbstractMap.SimpleImmutableEntry<>(STREAM_NAME, new StreamEntryID("0-0"));
+    jedis.xreadGroup(G1, MY_CONSUMER,1,0,false,streamQeury11);
+    List<StreamGroupInfo> groupInfo = jedis.xinfoGroup(STREAM_NAME);
+    List<StreamConsumersInfo> consumersInfo = jedis.xinfoConsumers(STREAM_NAME, G1);
+
+    //Stream info test
+    assertEquals(2L,streamInfo.getStreamInfo().get(LENGTH));
```
gkorland: Missing LAST_GENERATED_ID

src/test/java/redis/clients/jedis/tests/commands/StreamsCommandsTest.java
```diff
@@ -306,6 +319,166 @@ public void xpendeing() {
...
+    assertEquals(2L,streamInfo.getStreamInfo().get(RADIX_TREE_NODES));
+    assertEquals(0L,streamInfo.getStreamInfo().get(GROUPS));
+    assertEquals(V1,((StreamEntry)streamInfo.getStreamInfo().get(FIRST_ENTRY)).getFields().get(F1));
+    assertEquals(V2,((StreamEntry)streamInfo.getStreamInfo().get(LAST_ENTRY)).getFields().get(F1));
+
+    //Using getters
+    assertEquals(2,streamInfo.getLength());
+    assertEquals(1,streamInfo.getRadixTreeKeys());
+    assertEquals(2,streamInfo.getRadixTreeNodes());
+    assertEquals(0,streamInfo.getGroups());
```
gkorland: Missing getLastGeneratedId()

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -73,6 +73,10 @@
   public static final String PUBSUB_NUMSUB = "numsub";
   public static final String PUBSUB_NUM_PAT = "numpat";
 
+  public static final String XINFO_STREAM = "STREAM";
```
gkorland: Should be added to  `enum Keyword`

src/main/java/redis/clients/jedis/StreamInfo.java
```diff
@@ -0,0 +1,89 @@
...
+  public static final String RADIX_TREE_KEYS = "radix-tree-keys";
+  public static final String RADIX_TREE_NODES = "radix-tree-nodes";
+  public static final String GROUPS = "groups";
+  public static final String LAST_GENERATED_ID = "last-generated-id";
+  public static final String FIRST_ENTRY = "first-entry";
+  public static final String LAST_ENTRY = "last-entry";
+
+  private final long length;
+  private final long radixTreeKeys;
+
```
gkorland: remove white space

src/main/java/redis/clients/jedis/StreamInfo.java
```diff
@@ -0,0 +1,89 @@
...
+     streamInfo = map;
+     length = (Long) map.get(LENGTH);
+     radixTreeKeys = (Long) map.get(RADIX_TREE_KEYS);
+     radixTreeNodes = (Long) map.get(RADIX_TREE_NODES);
+     groups = (Long) map.get(GROUPS);
+     lastGeneratedId = (String) map.get(LAST_GENERATED_ID);
+     firstEntry = (StreamEntry) map.get(FIRST_ENTRY);
+     lastEntry = (StreamEntry) map.get(LAST_ENTRY);
+
+    } else throw  new IllegalArgumentException("InfoMap can not be null");
```
gkorland: If we are throwing runtime exception anyway when map is null isn't it better to just let it throw NullPointerException?

src/main/java/redis/clients/jedis/StreamGroupInfo.java
```diff
@@ -0,0 +1,65 @@
...
+   */
+  public StreamGroupInfo(Map<String, Object> map) {
+    if (map!= null && map.size()>0) {
+     groupInfo = map;
+     name = (String) map.get(NAME);
+     consumers = (long) map.get(CONSUMERS);
+     pending = (long) map.get(PENDING);
+     lastDeliveredId = (String) map.get(LAST_DELIVERED);
+
+    } else throw new IllegalArgumentException();
```
gkorland: If we are throwing runtime exception anyway when map is null isn't it better to just let it throw NullPointerException?

src/main/java/redis/clients/jedis/StreamConsumersInfo.java
```diff
@@ -0,0 +1,60 @@
...
+  public StreamConsumersInfo(Map<String,Object> map) {
+
+    if (map!= null && map.size()>0) {
+      consumerInfo = map;
+      name = (String) map.get(NAME);
+      idle = (long) map.get(IDLE);
+      pending = (long) map.get(PENDING);
+
+
+    } else throw new IllegalArgumentException();
```
gkorland: If we are throwing runtime exception anyway when map is null isn't it better to just let it throw NullPointerException?

src/test/java/redis/clients/jedis/tests/commands/StreamsCommandsTest.java
MichalCho: Done in a way that we check if it's not null - it is difficult to know what last-generated should be

src/test/java/redis/clients/jedis/tests/commands/StreamsCommandsTest.java
MichalCho: Done in a way that we check if it's not null - it is difficult to know what last-generated should be

src/main/java/redis/clients/jedis/Protocol.java
MichalCho: done

src/main/java/redis/clients/jedis/StreamInfo.java
MichalCho: done

src/main/java/redis/clients/jedis/StreamInfo.java
MichalCho: Done. My initial thought was that it is not nice to see NullPointers

src/main/java/redis/clients/jedis/StreamGroupInfo.java
MichalCho: Done. My initial thought was that it is not nice to see NullPointers

src/main/java/redis/clients/jedis/StreamConsumersInfo.java
MichalCho: Done. My initial thought was that it is not nice to see NullPointers

src/main/java/redis/clients/jedis/Protocol.java
gkorland: :+1:

src/test/java/redis/clients/jedis/tests/commands/StreamsCommandsTest.java
gkorland: shouldn't it be id2?

src/test/java/redis/clients/jedis/tests/commands/StreamsCommandsTest.java
gkorland: shouldn't it be id2?

src/test/java/redis/clients/jedis/tests/commands/StreamsCommandsTest.java
MichalCho: you are right. I've missed that we have that info already.
Maybe then I should change `getLastGeneratedId` method to return `StreamEntryID` instead of `String`?

src/test/java/redis/clients/jedis/tests/commands/StreamsCommandsTest.java
MichalCho: you are right. I've missed that we have that info already.
Maybe then I should change `getLastGeneratedId` method to return `StreamEntryID` instead of `String`?

gkorland:
Tests are failing https://travis-ci.org/xetorthio/jedis/jobs/648852944

MichalCho:
> Tests are failing https://travis-ci.org/xetorthio/jedis/jobs/648852944

I'm on it

gkorland:
@MichalCho sounds OK just sleep for 1ms

MichalCho:
I think we are done here

sazzad16:
Merged into master. Excellent job @MichalCho

MichalCho:
Thanks,
@gkorland @sazzad16 do we have a list of features/ issues to be implemented/fixed? I would like to contribute more?

gkorland:
@MichalCho thanks for the great help!
You can see here the tasks marked for the next versions https://github.com/xetorthio/jedis/milestones/

PR #2078, opened 2019-10-16
Format after #1988 and #1968

sazzad16:
-  Fix #1988 
-  Modify #1988 according to #1968 
- Format and re-order BuilderFactory

### DIFF
--- a/src/main/java/redis/clients/jedis/BuilderFactory.java
+++ b/src/main/java/redis/clients/jedis/BuilderFactory.java
@@ -13,6 +13,7 @@ import redis.clients.jedis.util.JedisByteHashMap;
 import redis.clients.jedis.util.SafeEncoder;
 
 public final class BuilderFactory {
+
   public static final Builder<Double> DOUBLE = new Builder<Double>() {
     @Override
     public Double build(Object data) {
@@ -32,6 +33,7 @@ public final class BuilderFactory {
       return "double";
     }
   };
+
   public static final Builder<Boolean> BOOLEAN = new Builder<Boolean>() {
     @Override
     public Boolean build(Object data) {
@@ -43,10 +45,11 @@ public final class BuilderFactory {
       return "boolean";
     }
   };
+
   public static final Builder<byte[]> BYTE_ARRAY = new Builder<byte[]>() {
     @Override
     public byte[] build(Object data) {
-      return ((byte[]) data); // deleted == 1
+      return (byte[]) data;
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


gkorland:
@sazzad16 it is really hard to see what is #2078 is adding on top of  #1968

sazzad16:
@gkorland 

1. Almost all of `Jedis` methods that takes a `count` param, receives it as an `int`. There are only two exceptions in Jedis class. As `zpopmin` is yet to release, I changed its `count` to int.

2. ZPOPMIN without count returns only one Tuple (or none). Instead of returning a `Set<Tuple>` with only one element, it'd be better to return `Tuple`. ZPOPMAX #1968 is changed to be like that. I included this change in this PR.

sazzad16:
@gkorland I forgot that I had changed the same changes in https://github.com/xetorthio/jedis/pull/1968/commits/915a8106e722679548f83a925b168567d891ba51 which are merged through #1968 just now.

gkorland:
So can we close this PR?

gkorland:
or at lease undo all the none relevant changes?

src/main/java/redis/clients/jedis/PipelineBase.java
```diff
@@ -1337,13 +1337,13 @@
   }
 
   @Override
-  public Response<Set<Tuple>> zpopmin(final byte[] key, final int count) {
+  public Response<Set<Tuple>> zpopmin(final String key, final int count) {
```
gkorland: undo?

src/main/java/redis/clients/jedis/PipelineBase.java
```diff
@@ -1337,13 +1337,13 @@
...
     getClient(key).zpopmin(key, count);
     return getResponse(BuilderFactory.TUPLE_ZSET);
   }
 
   @Override
-  public Response<Set<Tuple>> zpopmin(final String key, final int count) {
+  public Response<Set<Tuple>> zpopmin(final byte[] key, final int count) {
```
gkorland: undo?

src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -13,6 +13,7 @@
 import redis.clients.jedis.util.SafeEncoder;
 
 public final class BuilderFactory {
+
```
gkorland: undo?

src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -32,6 +33,7 @@ public String toString() {
       return "double";
     }
   };
+
```
gkorland: undo?

src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -43,10 +45,11 @@ public String toString() {
       return "boolean";
     }
   };
+
```
gkorland: undo?

sazzad16:
@gkorland 
> please undo uneeded changes

done!

gkorland:
@sazzad16 is still seems like a lot of code just moved around

Issue #2029, opened 2019-07-21
Create an interface for 3rd party commands


DvirDukhan:
create additional interfaces for 3ed party commands and general commands. this is a discussion following PR #2025

maksymkovalenko:
PR #2025 is now merged in, please close this issue.

sazzad16:
@maksymkovalenko This idea was generated after mentioned PR.

sazzad16:
Issue details:

We may need interface(s) for the 3rd party commands. Here are reference methods:

1. https://github.com/xetorthio/jedis/blob/c581682bda779f6cfe7e9f1270705e4e1844cb01/src/main/java/redis/clients/jedis/BinaryJedis.java#L4054

1. https://github.com/xetorthio/jedis/blob/c581682bda779f6cfe7e9f1270705e4e1844cb01/src/main/java/redis/clients/jedis/BinaryJedis.java#L4049

1. https://github.com/xetorthio/jedis/blob/c581682bda779f6cfe7e9f1270705e4e1844cb01/src/main/java/redis/clients/jedis/Jedis.java#L3833

gkorland:
@sazzad16 can we close this issue?

sazzad16:
@gkorland I don't think so. Don't you think we should have such separate interface(s)?

PR #2079, opened 2019-10-16
PipelineBase should not contain multi key operations

sazzad16:
PipelineBase should not *look like* it contains multi key operations.

This is modification of #2007 and #2025

### DIFF
--- a/src/main/java/redis/clients/jedis/PipelineBase.java
+++ b/src/main/java/redis/clients/jedis/PipelineBase.java
@@ -1975,15 +1975,13 @@ public abstract class PipelineBase extends Queable implements BinaryRedisPipelin
     return getResponse(BuilderFactory.BYTE_ARRAY_LIST);            
   }
 
-  public Response<Object> sendCommand(ProtocolCommand cmd, String... args){
-    String key = args.length > 0 ? args[0] : cmd.toString();
-    getClient(key).sendCommand(cmd, args);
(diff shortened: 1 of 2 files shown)


sazzad16:
@DvirDukhan PING! as you were the author of related PRs.

src/main/java/redis/clients/jedis/PipelineBase.java
```diff
@@ -1975,15 +1975,13 @@
     return getResponse(BuilderFactory.BYTE_ARRAY_LIST);            
   }
 
-  public Response<Object> sendCommand(ProtocolCommand cmd, String... args){
-    String key = args.length > 0 ? args[0] : cmd.toString();
-    getClient(key).sendCommand(cmd, args);
+  public Response<Object> sendCommand(final String sampleKey, final ProtocolCommand cmd, final String... args) {
```
gkorland: @sazzad16 I think we should just deprecate these methods, once you have them in the MultiKeyPipelineBase you don't need these.
I don't think we need those for ShardedJedisPipeline or just move them there with the sample keys.

gkorland:
@sazzad16 I think we should also add these methods to the interfaces MultiKeyBinaryRedisPipeline and MultiKeyCommandsPipeline

sazzad16:
@gkorland MultiKeyBinaryRedisPipeline and MultiKeyCommandsPipeline contains [**actual** Redis commands](https://redis.io/commands). Because of this conflict, we already have an open issue #2029 related to this.

PR #2083, opened 2019-10-27
Modify tests

sazzad16:
Changes mostly includes:
- Move utility methods (`arrayContains`, `setContains`) from `JedisCommandTestBase`
- `close()`/`disconnect()` Jedis objects

### DIFF
--- a/src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
+++ b/src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
@@ -149,10 +149,13 @@ public class JedisPoolTest {
     j.auth("foobared");
     j.select(2);
     j.set("foo", "bar");
+    j.close();
+
     JedisPool pool = new JedisPool("redis://:foobared@localhost:6380/2");
     Jedis jedis = pool.getResource();
     assertEquals("PONG", jedis.ping());
     assertEquals("bar", jedis.get("foo"));
+    jedis.close();
   }
 
   @Test
@@ -161,6 +164,8 @@ public class JedisPoolTest {
     j.auth("foobared");
     j.select(2);
     j.set("foo", "bar");
+    j.close();
+
     JedisPool pool = new JedisPool(new URI("redis://:foobared@localhost:6380/2"));
     Jedis jedis = pool.getResource();
     assertEquals("PONG", jedis.ping());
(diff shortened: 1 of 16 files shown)


src/test/java/redis/clients/jedis/tests/JedisPoolTest.java
```diff
@@ -149,10 +149,13 @@ public void startWithUrlString() {
     j.auth("foobared");
     j.select(2);
     j.set("foo", "bar");
+    j.close();
+
     JedisPool pool = new JedisPool("redis://:foobared@localhost:6380/2");
     Jedis jedis = pool.getResource();
     assertEquals("PONG", jedis.ping());
     assertEquals("bar", jedis.get("foo"));
+    jedis.close();
```
gkorland: why not using try resource block?

sazzad16: just finding out the *closeable* Jedis(es) first, before using try-resource.

gkorland: @sazzad16 I'm not sure I understand the answer

sazzad16: @gkorland In this PR, I tried only to find out unclosed objects and `close()` them. There are too many places where try-with-resources can be used which can be done in separate PR. IMHO, using those here and there, in a few random places, would not fulfil our willingness of using try-with-resources. Nevertheless, changes in this PR will also help us to find out the places to put try-with-resources.

src/test/java/redis/clients/jedis/tests/utils/AssertUtil.java
```diff
@@ -15,6 +14,15 @@
 
 public class AssertUtil {
 
+  public static boolean assertCollectionContains(Collection<byte[]> array, byte[] expected) {
+    for (byte[] bytes : array) {
+      if (Arrays.equals(bytes, expected)) {
```
gkorland: @sazzad16 shouldn't you use assertArrayEquals here?

sazzad16: @gkorland NO. We're checking whether `array` contains `expected`. In case where `expected` is present but not the first element in `array`, assertArrayEquals would throw an assertion error right away, causing the check to fail wrongfully.

sazzad16: @gkorland also, check the assertion error (`ComparisonFailure`)  thrown right after the loop.

gkorland: :+1:

PR #2089, opened 2019-11-05
fix #1959 add support for MEMORY USAGE

### LINKED ISSUE #1959: Jedis Memory commands, ex memory usage, memory stats
I want to get the memory usage per key that is being by redis cache. I tried using the Connection class method sendCommand but to no luck because i don't think it has existing functionality for the Memory commands. 

jedis:2.9.0
redis:4.0.11

gkorland:


src/main/java/redis/clients/jedis/commands/AdvancedBinaryJedisCommands.java
```diff
@@ -42,4 +42,9 @@
   String clientSetname(byte[] name);
 
   byte[] memoryDoctorBinary();
+  
+  byte[] memoryUsageBinary(byte[] key);
+
+  byte[] memoryUsageBinary(byte[] key, int samples);
```
sazzad16: `Binary` is not required because of presence of `byte[]`.

src/main/java/redis/clients/jedis/JedisCluster.java
```diff
@@ -2052,6 +2052,27 @@ public Long execute(Jedis connection) {
...
+  
+  @Override
+  public Long memoryUsage(final String key, final int samples) {
+    return new JedisClusterCommand<Long>(connectionHandler, maxAttempts) {
+      @Override
+      public Long execute(Jedis connection) {
+        return connection.memoryUsage(key, samples);
+      }
+    }.run(key);
+  }
```
sazzad16: Also add in BinaryJedisCluster.

src/main/java/redis/clients/jedis/commands/AdvancedBinaryJedisCommands.java
gkorland: @sazzad16 yes I know, but I noticed other methods are keeping this patter so I followed it

gkorland: done

src/main/java/redis/clients/jedis/JedisCluster.java
gkorland: done

PR #2095, opened 2019-11-11
add pipelined to JedisCluster

gkorland:


gkorland:
@sazzad16 please review

sazzad16:
@gkorland Are you sure to add this in 3.3.0?

src/main/java/redis/clients/jedis/commands/BasicCommands.java
```diff
@@ -3,6 +3,9 @@
 import redis.clients.jedis.DebugParams;
 
 public interface BasicCommands {
+  
+  RedisPipeline pipelined();
```
sazzad16: This interface contains only actual Redis commands. We should keep it this way.

gkorland:
> @gkorland Are you sure to add this in 3.3.0?

not sure, what do you say?

src/main/java/redis/clients/jedis/commands/BasicCommands.java
gkorland: :+1:

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
```diff
@@ -2270,4 +2270,14 @@ public Object execute(Jedis connection){
       }
     }.runBinary(sampleKey);
   }
+  
+
+  public Pipeline pipelined(final byte[] sampleKey) {
+    return new JedisClusterCommand<Pipeline>(connectionHandler, maxAttempts) {
+      @Override
+      public Pipeline execute(Jedis connection){
+        return connection.pipelined();
```
sazzad16: Returned `Pipeline` object holds the `Client` object from `Jedis` object (connection). But that `Jedis` object gets returned to pool by JedisClusterCommand.releaseConnection(). This can have severe consequences in (heavy) multi-threaded scenario.

sazzad16:
> Are you sure to add this in 3.3.0?
> Not sure, what do you suggest?

From our recent conversation, I was under impression that you may release 3.3.0 soon. So I asked the question whether you want to include this *critical* change hurriedly. That's all.

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
gkorland: @sazzad16 good cache! my bad! I'll rework this solution

yuluoqianmu:
hi，a very nice idea! why are you close the commit?

Issue #791, opened 2014-11-21
Exceptions in Jedis.close() cause empty JedisPool and then deadlock


svetob:
Exceptions (e.g. SocketTimeoutException) can occur when `Jedis.close()` is called. See this example using a try-with-resources:

``` java
try (Jedis jedis = jedisPool.getResources()) {    
    p.sadd(redisKey, serialized);
    p.expire(redisKey, ttlInSeconds);
}
```

Depending on at what point in the code this occurs, it is possible for the resource to not be returned to the resource pool. If this happens enough times you will eventually run out of resources and you will have a deadlock the next time you attempt to get a resource from the pool.

The offender is this function in JedisPool.java:

``` java
    public void returnResource(final Jedis resource) {
    if (resource != null) {
        resource.resetState();
        returnResourceObject(resource);
    }
    }
```

If an exception occurs during `resource.resetState()`, then `returnResourceObject(resource)` will not be called. This seems like it could be easily resolved with a try-catch here.

PR #2102, opened 2019-11-22
Handle exceptions on close in JedisSentinelPool

Talon876:
When returning a Jedis object to the pool, handle any exceptions and still return the connection to prevent deadlock.
This is similar to issue and fix for #791, but for the JedisSentinelPool.

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisSentinelPool.java
+++ b/src/main/java/redis/clients/jedis/JedisSentinelPool.java
@@ -237,8 +237,13 @@ public class JedisSentinelPool extends JedisPoolAbstract {
   @Override
   protected void returnResource(final Jedis resource) {
     if (resource != null) {
-      resource.resetState();
-      returnResourceObject(resource);
+      try {
+        resource.resetState();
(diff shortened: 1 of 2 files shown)


sazzad16:
LGTM!

devil5462:
@sazzad16 Hi, mate. Thanks for this Issue to solve. I faced that like you. However, I am using version Jedis 2.9.0. I understand version 2.10.2 is latest in version 2. Is it work well or a fixed issue?( #1819  ) I have to use the jedis 2.9.0. Let me know the way to use in version 2. I'm looking forward to your reply.

devil5462:
@sazzad16 Hi, mate. Thanks for this Issue to solve. I faced that like you. However, I am using version Jedis 2.9.0. I understand version 2.10.2 is latest in version 2. Is it work well or a fixed issue?( #1819 ) I have to use version 2. Let me know the way to use in version 2. I'm looking forward to your reply.

sazzad16:
@devil5462 yes, this merged PR solved connection leaking issue.

PR #2118, opened 2019-12-14
Implement multi keys sendCommand for cluster

yangbodong22011:
Implement multi keys sendCommand for Cluster, although single key command can pass multiple keys directly through `args`, like below:
```
public Object sendCommand(final String sampleKey, final ProtocolCommand cmd, final String... args) {
    return new JedisClusterCommand<Object>(connectionHandler, maxAttempts) {
      @Override
      public Object execute(Jedis connection){
        return connection.sendCommand(cmd, args);
      }
    }.run(sampleKey);
  }
```
user code：
```
jedisCluster.sendCommand(sampleKey1, cmd, sampleKey2, sampleKey3, sampleKey4...);
```
this will lack the detection of whether these keys are in the same slot, so I added this command.

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryJedisCluster.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedisCluster.java
@@ -2251,4 +2251,13 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
       }
     }.runBinary(sampleKey);
   }
+
+  public Object sendCommand(final ProtocolCommand cmd, final byte[]... keys) {
+    return new JedisClusterCommand<Object>(connectionHandler, maxAttempts) {
+      @Override
+      public Object execute(Jedis connection) {
+        return connection.sendCommand(cmd, keys);
+      }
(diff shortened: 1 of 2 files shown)


sazzad16:
@yangbodong22011 This is a very specific change. Redis commands which takes multiple arguments but all are keys are very rare. As you've mentioned, there is a way to achieve that rare feature. And I don't see any need to ease a rare feature.

yangbodong22011:
I understand what you mean, but if the key passed in is not in the same slot, the result will confuse the user, eg `MODULE.MGET key1 key2 key3`, return` value1 null value3`, because key1 and key3 are in the same slot, but key2 is not.

Should this be avoided by the user? Still jedis should provide a parameter check. Or do you have any better suggestions?

sazzad16:
@yangbodong22011 
> Should this be avoided by the user?

That's the plan.

yangbodong22011:
@sazzad16 Ok. Then I closed this PR

PR #2131, opened 2020-01-08
issue#2126 - Add support for 'client id' command

### LINKED ISSUE #2126: please implements clientId() method!
please implements clientId() method!

thank you very much !

amitg87:
Fixes #2126 
Added support for 'client id' command

**Reference** [client-id command doc](https://redis.io/commands/client-id): 
**Command**: > client id
**Return**: Integer - id of the client
**Test**: response from 'client id' should match the 'client list' self filter


src/main/java/redis/clients/jedis/BinaryJedis.java
```diff
@@ -3609,6 +3609,12 @@ public String clientSetname(final byte[] name) {
     return client.getBulkReply();
   }
 
+  public Long clientId() {
```
sazzad16: `@Override`

sazzad16:
@amitg87 Can we test

> If the ID of a connection is greater than the ID of another connection, it is guaranteed that the second connection was established with the server at a later time.

?

amitg87:
@sazzad16 - Added requested test case

src/test/java/redis/clients/jedis/tests/commands/ClientIdCommandsTest.java
```diff
@@ -0,0 +1,73 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import redis.clients.jedis.Jedis;
+
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Tests related to clientId
+ */
+public class ClientIdCommandsTest extends JedisCommandTestBase {
```
sazzad16: No need to create new class. `ClientCommandsTest` should be enough. So move the test methods in that class.

src/test/java/redis/clients/jedis/tests/commands/ClientIdCommandsTest.java
```diff
@@ -0,0 +1,73 @@
...
+    private final String clientName = "fancy_jedis_name";
+    private final String clientName2 = "fancy_jedis_another_name";
+
+
+    protected Jedis client;
+    private Jedis client2;
```
sazzad16: This is used only in `clientIdMultipleConnection()`. So create this in that method. Also declare `clientName2` there.

src/test/java/redis/clients/jedis/tests/commands/ClientIdCommandsTest.java
```diff
@@ -0,0 +1,73 @@
...
+     * but it is still the same connection.
+     */
+    @Test
+    public void clientReconnectTest() {
+        long clientId1 = client.clientId();
+        client.close();
+
+        client = new Jedis(hnp.getHost(), hnp.getPort(), 500);
+        client.auth("foobared");
+        client.clientSetname(clientName);
```
sazzad16: `client = new Jedis(hnp.getHost(), hnp.getPort(), 500);` means *recreate* not *reconnect*. What happens if you remove these three lines (aka stop recreating) entirely? Is it `clientId2 == clientId1`??

src/test/java/redis/clients/jedis/tests/commands/ClientIdCommandsTest.java
amitg87: Added test case to existing **ClientCommandsTest**

src/test/java/redis/clients/jedis/tests/commands/ClientIdCommandsTest.java
amitg87: Removed this.
PS - every new connection has distinct client id. And client-id is monotonically increasing, so every new connection will have client-id bigger than previous connections.

src/test/java/redis/clients/jedis/tests/commands/ClientCommandsTest.java
```diff
@@ -11,6 +11,7 @@
 import java.util.regex.Pattern;
 
 import org.junit.After;
+import org.junit.Assert;
```
sazzad16, line 14:
```suggestion
```

src/test/java/redis/clients/jedis/tests/commands/ClientCommandsTest.java
```diff
@@ -56,6 +57,43 @@ public void nameBinary() {
...
+  public void clientIdmultipleConnection() {
+    Jedis client2 = new Jedis(hnp.getHost(), hnp.getPort(), 500);
+    client2.auth("foobared");
+    client2.clientSetname("fancy_jedis_another_name");
+
+    long clientId1 = client.clientId();
+    long clientId2 = client2.clientId();
+
+    ///client-id is monotonically increasing
+    Assert.assertTrue(clientId1 < clientId2);
```
sazzad16, line 82:
```suggestion
    assertTrue(clientId1 < clientId2);
```

PR #2135, opened 2020-02-11
Implementation in Jedis cluster, use slot to get HostAndPort

yangbodong22011:
I am implementing the mset/mget command across slots in the redis cluster(JedisCluster for multiple keys, only execute if they all share the same connection slot).

I've tried to merge and forward by slot, but in the case of a small number of machines, there are too many slots on each machine (for example, 3 machines, 16384 / 3 ≈ 5460), so the efficiency of dividing keys by slot is too poor, so I need to divide them by IP:PORT.

But I can't get IP:PORT information according to key or slot, It is recorded in `redis.clients.jedis.JedisClusterInfoCache#slots`, so, I submit this PR.

```
import redis.clients.jedis.JedisFactory;

public class MyCluster extends JedisCluster {
    public MyCluster(HostAndPort node) {
        super(node);
    }

    public HostAndPort getHostAndPortFormSlot() {
        int slot = 1;
        return ((JedisFactory)connectionHandler.getSlots().get(slot).internalPool.getFactory())
            .getHostAndPort().get();
    }
}
```

@sazzad16  @gkorland Please review when you have time, I would appreciate any suggestions.

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisClusterConnectionHandler.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterConnectionHandler.java
@@ -45,6 +45,10 @@ public abstract class JedisClusterConnectionHandler implements Closeable {
     return cache.getNodes();
   }
 
+  public Map<Integer, JedisPool> getSlots() {
+    return cache.getSlots();
+  }
+
   private void initializeSlotsCache(Set<HostAndPort> startNodes,
       int connectionTimeout, int soTimeout, String password, String clientName,
       boolean ssl, SSLSocketFactory sslSocketFactory, SSLParameters sslParameters, HostnameVerifier hostnameVerifier) {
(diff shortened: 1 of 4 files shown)


sazzad16:
@yangbodong22011 

1. JedisCluster already has MGET https://github.com/xetorthio/jedis/blob/5089cbd3e495232a5ae57dfb9bc32727fbbcf35d/src/main/java/redis/clients/jedis/commands/MultiKeyJedisClusterCommands.java#L26 

2. Redis throws following ERROR for keys slotted to multiple hash slots even if those hash slots are allocated in same node.

        (error) CROSSSLOT Keys in request don't hash to the same slot

I don't see any purpose for the intention of this PR. Closing.

yangbodong22011:
> 2\. Redis throws following ERROR for keys slotted to multiple hash slots even if those hash slots are allocated in same node.

I understand, thank you for your reply.

PR #2139, opened 2020-02-13
Implement ability to connect to Sentinel with TLS

Talon876:
This change allows connecting to sentinel via TLS. Additionally, you can specify if you want to connect to the redis master with TLS or not.

When connecting to sentinel with TLS, it will still broadcast the non-TLS ports for redis. For this scenario to work, `toHostAndPort` was modified to be protected so consumers can sub-class JedisSentinelPool and implement your own port mapping behavior.

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisSentinelPool.java
+++ b/src/main/java/redis/clients/jedis/JedisSentinelPool.java
@@ -13,6 +13,10 @@ import org.slf4j.LoggerFactory;
 import redis.clients.jedis.exceptions.JedisConnectionException;
 import redis.clients.jedis.exceptions.JedisException;
 
+import javax.net.ssl.HostnameVerifier;
+import javax.net.ssl.SSLParameters;
+import javax.net.ssl.SSLSocketFactory;
+
 public class JedisSentinelPool extends JedisPoolAbstract {
 
   protected GenericObjectPoolConfig poolConfig;
@@ -27,6 +31,12 @@ public class JedisSentinelPool extends JedisPoolAbstract {
   protected int sentinelSoTimeout;
   protected String sentinelPassword;
   protected String sentinelClientName;
+
+  protected boolean isRedisSslEnabled;
+  protected boolean isSentinelSslEnabled;
+  protected SSLSocketFactory sslSocketFactory;
+  protected SSLParameters sslParameters;
+  protected HostnameVerifier hostnameVerifier;
 
   protected Set<MasterListener> masterListeners = new HashSet<MasterListener>();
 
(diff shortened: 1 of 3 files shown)


src/main/java/redis/clients/jedis/JedisSentinelPool.java
```diff
@@ -28,6 +32,12 @@
   protected String sentinelPassword;
   protected String sentinelClientName;
 
+  protected boolean isRedisSslEnabled;
```
sazzad16: `Master` would be better that `Redis`. E.g. `isMasterSslEnabled`. Also, apply this change entire PR.

src/main/java/redis/clients/jedis/JedisSentinelPool.java
```diff
@@ -89,14 +99,16 @@ public JedisSentinelPool(String masterName, Set<String> sentinels,
...
   }
 
   public JedisSentinelPool(String masterName, Set<String> sentinels,
       final GenericObjectPoolConfig poolConfig, final int connectionTimeout, final int soTimeout,
       final String password, final int database, final String clientName,
       final int sentinelConnectionTimeout, final int sentinelSoTimeout, final String sentinelPassword,
-      final String sentinelClientName) {
+      final String sentinelClientName, final boolean isRedisSslEnabled, final boolean isSentinelSslEnabled,
+      final SSLSocketFactory sslSocketFactory, final SSLParameters sslParameters,
+      final HostnameVerifier hostnameVerifier) {
```
sazzad16: DON'T change an existing constructor or public method. Create a new one.

src/main/java/redis/clients/jedis/JedisSentinelPool.java
```diff
@@ -274,23 +294,39 @@ protected void returnResource(final Jedis resource) {
...
+        long subscribeRetryWaitTimeMillis, boolean isSslEnabled, SSLSocketFactory sslSocketFactory,
+        SSLParameters sslParameters, HostnameVerifier hostnameVerifier) {
+      this(masterName, host, port, isSslEnabled, sslSocketFactory, sslParameters, hostnameVerifier);
+      this.subscribeRetryWaitTimeMillis = subscribeRetryWaitTimeMillis;
     }
 
     public MasterListener(String masterName, String host, int port,
-        long subscribeRetryWaitTimeMillis) {
-      this(masterName, host, port);
+                          long subscribeRetryWaitTimeMillis) {
```
sazzad16:
```suggestion
        long subscribeRetryWaitTimeMillis) {
```

PR #2151, opened 2020-02-26
Add support for custom socket factories

mina-asham:
This can enable UDS (Unix Domain Socket) or any other custom address resolution

- Related feature request: https://github.com/xetorthio/jedis/issues/1998
- Related pull requests: https://github.com/xetorthio/jedis/pull/2036 and https://github.com/xetorthio/jedis/pull/1942 and https://github.com/xetorthio/jedis/pull/1132

### DIFF
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -2,7 +2,6 @@ package redis.clients.jedis;
 
 import java.io.Closeable;
 import java.io.IOException;
-import java.net.InetSocketAddress;
 import java.net.Socket;
 import java.net.SocketException;
 import java.util.ArrayList;
@@ -10,7 +9,6 @@ import java.util.List;
 
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLParameters;
-import javax.net.ssl.SSLSocket;
 import javax.net.ssl.SSLSocketFactory;
 
 import redis.clients.jedis.commands.ProtocolCommand;
@@ -25,46 +23,37 @@ public class Connection implements Closeable {
 
   private static final byte[][] EMPTY_ARGS = new byte[0][];
 
-  private String host = Protocol.DEFAULT_HOST;
-  private int port = Protocol.DEFAULT_PORT;
+  private JedisSocketFactory jedisSocketFactory;
   private Socket socket;
   private RedisOutputStream outputStream;
   private RedisInputStream inputStream;
-  private int connectionTimeout = Protocol.DEFAULT_TIMEOUT;
-  private int soTimeout = Protocol.DEFAULT_TIMEOUT;
   private boolean broken = false;
-  private boolean ssl;
-  private SSLSocketFactory sslSocketFactory;
-  private SSLParameters sslParameters;
-  private HostnameVerifier hostnameVerifier;
 
   public Connection() {
+    this(Protocol.DEFAULT_HOST);
   }
 
   public Connection(final String host) {
-    this.host = host;
+    this(host, Protocol.DEFAULT_PORT);
   }
 
   public Connection(final String host, final int port) {
-    this.host = host;
-    this.port = port;
+    this(host, port, false);
   }
 
   public Connection(final String host, final int port, final boolean ssl) {
-    this.host = host;
-    this.port = port;
-    this.ssl = ssl;
+    this(host, port, ssl, null, null, null);
   }
 
   public Connection(final String host, final int port, final boolean ssl,
       SSLSocketFactory sslSocketFactory, SSLParameters sslParameters,
       HostnameVerifier hostnameVerifier) {
-    this.host = host;
-    this.port = port;
-    this.ssl = ssl;
-    this.sslSocketFactory = sslSocketFactory;
-    this.sslParameters = sslParameters;
-    this.hostnameVerifier = hostnameVerifier;
+    this(new DefaultJedisSocketFactory(host, port, Protocol.DEFAULT_TIMEOUT,
+        Protocol.DEFAULT_TIMEOUT, ssl, sslSocketFactory, sslParameters, hostnameVerifier));
+  }
+
+  public Connection(final JedisSocketFactory jedisSocketFactory) {
+    this.jedisSocketFactory = jedisSocketFactory;
   }
 
   public Socket getSocket() {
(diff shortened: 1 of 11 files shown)


mina-asham:
@sazzad16 @gkorland can you have a look when you have time? This should unblock any custom connection

sazzad16:
@mina-asham Please add a test or two. A test for UDS would be a great addition.

mina-asham:
@sazzad16 I didn't want to add another dependency per our discussion in https://github.com/xetorthio/jedis/pull/1942, but thinking about it now, I think it makes sense to add the dependency as a **test** dependency, so will do that and port my test from there!

mina-asham:
Test is failing because of spop test due to a fix in redis/unstable, see: https://github.com/xetorthio/jedis/pull/2152

sazzad16:
> Test is failing because of spop test due to a fix in redis/unstable

@mina-asham Don't worry about it. There are a couple of Redis commands that are in back-and-forth changes.

sazzad16:
LGTM!

@mina-asham Not necessary, but can give a short explanation of change in `SSLJedisClusterTest`? (I'm just being lazy :P)

mina-asham:
> LGTM!
> 
> @mina-asham Not necessary, but can give a short explanation of change in `SSLJedisClusterTest`? (I'm just being lazy :P)

Ha, the main difference is that the previous code used to:
* Create and assign the socket field
* Then fail trying to wrap it in SSL (but at that point the socket field was already assigned, so the isConnected check would pass)

The new logic is all encapsulated in the factory, so you either get a socket or an exception, so the isConnected method would return false (which I believe is the right behavior), so the cluster failure is that there are no active nodes instead of trying to use a node it thinks is active but it's not

mina-asham:
@gkorland do you have any comments on this one?

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -334,4 +296,117 @@ protected Object readProtocolWithCheckingBroken() {
     }
     return responses;
   }
+
+  private static class DefaultJedisSocketFactory implements JedisSocketFactory {
```
gkorland: Shouldn't we set this class as public? So users can inherent and override if needed?

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -334,4 +296,117 @@ protected Object readProtocolWithCheckingBroken() {
...
+        socket.setKeepAlive(true); // Will monitor the TCP connection is
+        // valid
+        socket.setTcpNoDelay(true); // Socket buffer Whetherclosed, to
+        // ensure timely delivery of data
+        socket.setSoLinger(true, 0); // Control calls close () method,
+        // the underlying socket is closed
+        // immediately
+        // <-@wjw_add
+
+        socket.connect(new InetSocketAddress(host, port), connectionTimeout);
```
gkorland: I this you should use the getHost(), getPort(), getConnectionTimeout()... so a user can override these methods

src/main/java/redis/clients/jedis/JedisSocketFactory.java
```diff
@@ -0,0 +1,27 @@
+package redis.clients.jedis;
+
+import java.io.IOException;
+import java.net.Socket;
+
+public interface JedisSocketFactory {
```
gkorland: Can you please add some JDoc here?

mina-asham: Will do!

src/main/java/redis/clients/jedis/Connection.java
mina-asham: I think of it as an implementation detail, users should create their own versions if they want something different, I am okay exposing it if there is a strong reasoning or you think it's mostly re-usable?

src/main/java/redis/clients/jedis/Connection.java
mina-asham: Good point, will update.

gkorland: This is why I wanted DefaultJedisSocketFactory to be public, otherwise no-one can override it.

mina-asham: Oh, I missed that, tell you what, I will just pull the class out and folks can do whatever logic they want.

src/main/java/redis/clients/jedis/Connection.java
mina-asham: Will extract.

gkorland:
@mina-asham it LGTM, but it's pretty hard to review the changes this way with the force-pushed. Next time please keep the incremental commits.

src/main/java/redis/clients/jedis/JedisSocketFactory.java
gkorland: thanks

src/test/java/redis/clients/jedis/tests/SSLJedisClusterTest.java
```diff
@@ -191,11 +191,11 @@ public void connectWithCustomHostNameVerifier() {
       jc2 = new JedisCluster(new HostAndPort("127.0.0.1", 8379), DEFAULT_TIMEOUT, DEFAULT_TIMEOUT,
           DEFAULT_REDIRECTIONS, "cluster", null, DEFAULT_CONFIG, true, 
           null, null, hostnameVerifier, portMap);
-      Assert.fail("The code did not throw the expected NullPointerException.");
-    } catch (NullPointerException e) {
-      // Null pointer exception occurs from closing Jedis object that did not connect due to custom
-      // hostname validation. When closing this Jedis object, the RedisOutputStream in the underlying 
-      // Connection object is null and causes this exception
+      jc2.get("foo");
```
gkorland: @mina-asham  `jc2.get("foo");` what is for?

gkorland: @mina-asham `jc2.get("foo");` what is for?

mina-asham: It's to trigger the failure, the previous logic used to assign the `socket` variable before failing on hostname verification making the `isConnected` method return `true`, but since now we only assign it if it succeeds the whole thing, it wouldn't behave exactly the same when connecting, so had to trigger the failure some other way.

mina-asham:
> @mina-asham it LGTM, but it's pretty hard to review the changes this way with the force-pushed. Next time please keep the incremental commits.

Will do, thanks for reviewing!

src/test/java/redis/clients/jedis/tests/SSLJedisClusterTest.java
gkorland: Isn't that a backward compatibility issue? 
@sazzad16

mina-asham: For a broken state yes, if you initiate a SSL cluster connection that fails due to host verification you wouldn't get the NPE on connection, but your connection will report broken instead.

Since this is an error state in anyway, I think the backward compatibility issue is not bad, what do you think?

sazzad16: @gkorland Kinda! Only if a user is catching `NullPointerException` without catching `JedisConnectionException`/`JedisException`. IMHO, this could be very rare and shouldn't be a significant issue.

mina-asham:
Can we merge this one?

sazzad16:
@mina-asham We will merge it, but there is a confusion about which version we'll be adding it in. That's all.

gkorland:
I think we are missing here a solution for the JedisPool & JedisCluster

mina-asham:
You're right, missed the respective constructors there, will add them and uodate this PR!

sazzad16:
@gkorland I'm going ahead and merging this.
@mina-asham Please do further improvements in separate PR. Thanks for you contribution!

PR #2152, opened 2020-02-26
Fix broken spop with count test

mina-asham:
- According to redis specifications, a list/set response will always be an empty list/set instead of nil
- Tested on CLI:

```
$ redis-cli
127.0.0.1:6379> flushall
OK
127.0.0.1:6379> spop a
(nil)
127.0.0.1:6379> spop a 2
(empty list or set)
```

### DIFF
--- a/src/test/java/redis/clients/jedis/tests/commands/SetCommandsTest.java
+++ b/src/test/java/redis/clients/jedis/tests/commands/SetCommandsTest.java
@@ -162,7 +162,7 @@ public class SetCommandsTest extends JedisCommandTestBase {
     assertEquals(1, members.size());
     assertEquals(superSet, members);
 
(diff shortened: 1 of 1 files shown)


mina-asham:
Ignore, I had an older version redis on my local machine.

mina-asham:
Actually, this was the right thing, it was a bug in redis/unstable and was just fixed yesterday: https://github.com/antirez/redis/commit/7e64aa80d8c222064e916b0cbda2d67dfabaddb7

gkorland:
@mina-asham can we close this PR?

mina-asham:
@gkorland this fixes a test that was previously succeeding due to a bug in redis/unstable, after that bug was fixed in redis this test will keep on failing unless we merge this one (or any other fix for the test)

PR #2158, opened 2020-03-04
Update README.md remove ambiguous cluster support

raphaelauv:
The current README is really confusing about redis cluster support


sazzad16:
@raphaelauv Thanks for reporting.

PR #2167, opened 2020-03-15
some quick code cleanups

gkorland:


src/test/java/redis/clients/jedis/tests/commands/TransactionCommandsTest.java
```diff
@@ -413,7 +413,7 @@ public void unwatchWithinMulti() {
     jedis.set(key, val);
     jedis.watch(key);
 
-    List exp = new ArrayList();
+    List<String> exp = new ArrayList<>();
```
sazzad16: It should be `List<Object>`.

gkorland: :+1:

Issue #2170, opened 2020-03-16
please implements ACL log method !


startjava:
please implements ACL log method !

tgrall:
Working on it

PR #2172, opened 2020-03-17
add support for ACL LOG command - issue #2170

tgrall:
I have added the ACL LOG method.

For the `jedis.aclLog()` that returns the list of log entries I have selected to return a basic and generic structure that is  `List<Map<String,String>>`.

The other option, would be to create a specific data structure that return a list of ACLLogEntries with proper types and sub structure.

I have selected simplicity since this method wont be used that much in Java app.

Let me know if you want me to change it


src/main/java/redis/clients/jedis/commands/AdvancedBinaryJedisCommands.java
```diff
@@ -64,5 +64,11 @@
 
   List<byte[]> aclCat(byte[] category);
 
+  List<byte[]> aclLogBinary();
```
sazzad16: `List<byte[]>` doesn't seem right. Take a look at `hgetAll(byte[])`. If nothing works, simple `Object` may be better.

src/main/java/redis/clients/jedis/commands/AdvancedBinaryJedisCommands.java
```diff
@@ -64,5 +64,11 @@
...
+
+  List<byte[]> aclLogBinary(int limit);
```
sazzad16: Same as line 67.

src/main/java/redis/clients/jedis/commands/AdvancedJedisCommands.java
```diff
@@ -64,5 +65,11 @@
 
   List<String> aclCat(String category);
 
+  List<Map<String,String>> aclLog();
```
sazzad16: It should be `List<Map<String,Object>>`.

src/main/java/redis/clients/jedis/commands/AdvancedJedisCommands.java
```diff
@@ -64,5 +65,11 @@
...
+
+  List<Map<String,String>> aclLog(int limit);
```
sazzad16: Same as line 70.

src/test/java/redis/clients/jedis/tests/commands/AccessControlListCommandsTest.java
```diff
@@ -315,6 +316,84 @@ public void aclCatTest() {
...
+    jedis.aclLog("RESET");
+    assertNotNull(jedis.aclLog()); // after reset the log should be null
+
+    // create new user and cconnect
+    jedis.aclSetUser("antirez", ">foo", "on", "+set", "~object:1234");
+    jedis.aclSetUser("antirez", "+eval", "+multi", "+exec");
+    jedis.auth("antirez", "foo");
+
+    // generate an error (antirez user does not have the permission to access foo)
+    try { jedis.get("foo");  } catch(JedisAccessControlException e) {}
```
sazzad16: Also use `fail()`.

src/test/java/redis/clients/jedis/tests/commands/AccessControlListCommandsTest.java
```diff
@@ -315,6 +316,84 @@ public void aclCatTest() {
...
+
+    // test the ACL Log
+    jedis.auth("default", "foobared");
+    assertEquals("Number of log messages ", 1, jedis.aclLog().size());
+    assertEquals("10", jedis.aclLog().get(0).get("count"));
+    assertEquals("get", jedis.aclLog().get(0).get("object"));
+
+    // Generate another type of error
+    jedis.auth("antirez", "foo");
+    try { jedis.set("somekeynotallowed", "1234");} catch (JedisAccessControlException e) {}
```
sazzad16: Also use `fail()`.

src/test/java/redis/clients/jedis/tests/commands/AccessControlListCommandsTest.java
```diff
@@ -315,6 +316,84 @@ public void aclCatTest() {
     }
   }
 
+  @Test
+  public void aclLogTest() {
+    jedis.aclLog("RESET");
+    assertNotNull(jedis.aclLog()); // after reset the log should be null
```
sazzad16: In comment, the world *null* may be confusing. Perhaps, *cleared* or something else?

src/test/java/redis/clients/jedis/tests/commands/AccessControlListCommandsTest.java
```diff
@@ -315,6 +316,84 @@ public void aclCatTest() {
...
+    assertEquals("Number of log messages ", 1, jedis.aclLog().size());
+    assertEquals("1", jedis.aclLog().get(0).get("count"));
+    assertEquals("antirez", jedis.aclLog().get(0).get("username"));
+    assertEquals("toplevel", jedis.aclLog().get(0).get("context"));
+    assertEquals("command", jedis.aclLog().get(0).get("reason"));
+    assertEquals("get", jedis.aclLog().get(0).get("object"));
+
+    // Capture similar event
+    jedis.aclLog("RESET");
+    assertNotNull(jedis.aclLog()); // after reset the log should be null
```
sazzad16: Same as line 322.

src/main/java/redis/clients/jedis/commands/AdvancedJedisCommands.java
sazzad16: @gkorland Should we go for a new custom object here?

-
```diff
@@ -0,0 +1,17 @@
...
+user default on #1b58ee375b42e41f0e48ef2ff27d10a5b1f6924a9acdcdba7cae868e7adce6bf ~* +@all
+dir "/Users/tgrall/projects/redis/clients/jedis"
+requirepass 1b58ee375b42e41f0e48ef2ff27d10a5b1f6924a9acdcdba7cae868e7adce6bf
+
+timeout 0
+
+client-output-buffer-limit normal 0 0 0
+masterauth "foobared"
+
+replicaof 127.0.0.1 6381
```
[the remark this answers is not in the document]
tgrall: Thank you for the review

Oh I did not realize that clean does not delete it, and it was not ignored in git

src/main/java/redis/clients/jedis/commands/AdvancedBinaryJedisCommands.java
tgrall: The reason why I have create a list is because of the way the log is organized: (and I can if you think it is better create an object like we do for StreamEntry - but his list of "map" is quite generic and simple)
```
127.0.0.1:6379> acl log
1)  1) "count"
    2) (integer) 1
    3) "reason"
    4) "key"
    5) "context"
    6) "toplevel"
    7) "object"
    8) "foo-2"
    9) "username"
   10) "antirez"
   11) "age-seconds"
   12) "25.146999999999998"
   13) "client-info"
   14) "id=34 addr=127.0.0.1:53509 fd=11 name= age=0 idle=0 flags=N db=0 sub=0 psub=0 multi=-1 qbuf=33 qbuf-free=32735 obl=0 oll=0 omem=0 events=r cmd=set user=antirez"
2)  1) "count"
    2) (integer) 5
    3) "reason"
    4) "command"
    5) "context"
    6) "toplevel"
    7) "object"
    8) "incr"
    9) "username"
   10) "antirez"
   11) "age-seconds"
   12) "25.146999999999998"
   13) "client-info"
   14) "id=34 addr=127.0.0.1:53509 fd=11 name= age=0 idle=0 flags=N db=0 sub=0 psub=0 multi=-1 qbuf=23 qbuf-free=32745 obl=0 oll=0 omem=0 events=r cmd=incr user=antirez"
```

sazzad16:
@tgrall BTW, there is an error

    AccessControlListCommandsTest.aclLogTest » StackOverflow

tgrall:
Yes I saw that and investigating in the same time and looking at your
review/feedback.

src/main/java/redis/clients/jedis/commands/AdvancedBinaryJedisCommands.java
tgrall: Oh yes... I oversimplified this part and I will add the proper processing and more tests about binary

tgrall: I was looking at `BinaryJedis.xrange()`, that returns also a list of `byte[]`.

However,  I will create a new class `AccessControlLogEntry` that will be cleaner overall, and I will use it also in this binary class. (for example `BinaryJedis.xinfoGroup` returns List<StreamGroupInfo>

I think the ACL LOG is more useful was a specific class even in binary class

-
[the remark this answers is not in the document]
tgrall: File removed

src/main/java/redis/clients/jedis/Client.java
```diff
@@ -1167,6 +1167,18 @@ public void aclCat(final String category) {
     aclCat(SafeEncoder.encode(category));
   }
 
+  public void aclLog() {
+    aclLog();
+  }
```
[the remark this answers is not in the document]
tgrall: Method removed as it is part the parent BinaryClient.

src/main/java/redis/clients/jedis/Client.java
```diff
@@ -1167,6 +1167,18 @@ public void aclCat(final String category) {
...
+
+  public void aclLog(final int limit) {
+    aclLog(toByteArray(limit));
+  }
```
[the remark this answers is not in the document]
tgrall: Method removed as it is part the parent BinaryClient.

src/main/java/redis/clients/jedis/commands/AdvancedBinaryJedisCommands.java
tgrall: Remove the binary methods that are not necessary now.

src/main/java/redis/clients/jedis/commands/AdvancedJedisCommands.java
tgrall: I have created a custom object for the log entry see `AccessControlLogEntry` class

src/main/java/redis/clients/jedis/commands/AdvancedJedisCommands.java
tgrall: I have created a custom object for the log entry see AccessControlLogEntry class

src/test/java/redis/clients/jedis/tests/commands/AccessControlListCommandsTest.java
tgrall: Added

src/test/java/redis/clients/jedis/tests/commands/AccessControlListCommandsTest.java
tgrall: Added

src/test/java/redis/clients/jedis/tests/commands/AccessControlListCommandsTest.java
tgrall: Removed and changed the test to `isEmpty()`

src/main/java/redis/clients/jedis/AccessControlLogEntry.java
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     * Convert the client-info string into a Map of String.
+     * When the value is empty, the value in the map is set to an empty string
+     * The key order is maintained to reflect the string return by Redis
+     * @param clientInfo
+     * @return A Map with all client info
+     */
+    private Map<String,String> getMapFromRawClientInfo( String clientInfo) {
+        Map<String,String> clientInfoMap =  new LinkedHashMap<>();
```
gkorland: Why is LinkedHashMap required? 

I think the following is more efficient:
```java
        String[] entries = clientInfo.split(" ");
        Map<String,String> clientInfoMap =  new LinkedHashMap<>(entries.length);
        for (String entry : entries) {
            String[] kvArray = entry.split("=");
            clientInfoMap.put(kvArray[0], (kvArray.length ==2)?kvArray[1]:"" );
        }
```

src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -524,6 +524,55 @@ public String toString() {
 
   };
 
+  /**
+   * Create an Access Control Log Entry
+   * Result of ACL LOG command
+   */
+  public static final Builder<List<AccessControlLogEntry>> ACCESS_CONTROL_LOG_ENTRY_LIST = new Builder<List<AccessControlLogEntry>>() {
+
+    Map<String,Builder> mappingFunctions = createDecoderMap();
```
gkorland:
```java
private static final Map<String,Builder> mappingFunctions = createDecoderMap(); 

private static Map<String, Builder> createDecoderMap() { 
```

src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -524,6 +524,55 @@ public String toString() {
...
+    @Override
+    public List<AccessControlLogEntry> build(Object data) {
+
+      if (null == data) {
+        return null;
+      }
+
+      List<AccessControlLogEntry> list = new ArrayList<>();
+      List<Object> logEntries = (List<Object>)data;
+      Iterator<Object> logsArray = logEntries.iterator();
```
gkorland:
```java
List<List<Object>> logEntries = (List<List<Object>>)data; 
for  (List<Object> logEntryData : logEntries) {
```

src/main/java/redis/clients/jedis/commands/AdvancedBinaryJedisCommands.java
```diff
@@ -64,5 +65,7 @@
 
   List<byte[]> aclCat(byte[] category);
 
+  String aclLog(byte[] options);
```
sazzad16: Why not `byte[] aclLog(byte[] options)`?

sazzad16: `aclLog()` and `aclLog(int limit)` can also be added here.

src/main/java/redis/clients/jedis/AccessControlLogEntry.java
tgrall: +1 on the changes
About the LinkedHashMap
I used a LinkedHashMap to keep the fields in the same order as the string that is returned by the ACL LOG comment. (when people print or iterate over the keys)

*, To be honest, I do not know yet how people will use the log from a Java application so try to keep it as close as possible from the command returned value*

src/main/java/redis/clients/jedis/BuilderFactory.java
tgrall: I used the same notation as the `StreamConsumersInfo`, should I open a small issue for improving other part of the code to?

src/main/java/redis/clients/jedis/commands/AdvancedBinaryJedisCommands.java
tgrall: I looked at various methods in the BinaryJedisCommand, and still not sure what is the *rule*

Many of the methods are return the status as a string for example `set`,`type` `setex`, `psetex`...

*Since I do not know how the binary methods are used I chose the simplity*

tgrall: Since I have chosen to return a `List<AccessControlLogEntry>` I have not added these methods here.

If I return `byte[]` I can add them, but like on previous comment it is not clear to me what is the good practice for binary API

sazzad16: In BinaryJedis, we should be able to do most if not all operations using `byte[]` instead of String. The common exceptions, in case of parameters, are IP, password, etc. And common exceptions, in case of return types, are definitive Strings i.e. `"OK"` in set, setex, psetex, etc. A case can be made for `type`. But that's historical implementation and the users don't seem to bothered about it because isn't any issue submitted about it.

sazzad16: `List<AccessControlLogEntry>` should be fine in BinaryJedis as well. If there is user demand, we can update `AccessControlLogEntry` to serve `byte[]` too. For example, `Tuple` is used in both binary and string codes.

src/main/java/redis/clients/jedis/BuilderFactory.java
tgrall: * done for `private` & `final`
* cannot use `static` (static inner class)

src/main/java/redis/clients/jedis/AccessControlLogEntry.java
tgrall: done for the loop

I am keeping the LinkedMap easier when printing the information

src/main/java/redis/clients/jedis/BuilderFactory.java
tgrall: Done

src/main/java/redis/clients/jedis/commands/AdvancedBinaryJedisCommands.java
tgrall: Done.

tgrall: I have added 

```
  List<byte[]> aclLogBinary();

  List<byte[]> aclLogBinary(int limit);
```
Returning a simple list of `byte[]`, in case developer wants to use the raw data.

src/main/java/redis/clients/jedis/commands/AdvancedBinaryJedisCommands.java
sazzad16: *Considering you want to stick to this,* `List<byte[]>` isn't proper return type. It has to be at least `List<Object>`. So just return `Object`*.
\* You may have to consider Connection.getOne().

sazzad16 (approved):

gkorland (approved):

gkorland:
@tague did you delete your original repo? can you please resolve the conflict?

sazzad16:
@gkorland PING #2302

PR #2174, opened 2020-03-31
Implements BITFIELD_RO command

yangbodong22011:
In the redis unstable branch, the `bitfield_ro` command is added, refer: [this](https://github.com/antirez/redis/commit/493a7f9823dae5303208be2de778942815df24ba) .


yangbodong22011:
> What if there's a new read sub-command for BITFIELD command?

Can you give a specific code example?

sazzad16:
Sorry, there was a typo in my previous comment.

About example, in `JedisCommands` interface, instead of 

    List<Long> bitfieldReadonly(String key, String type, String offset);

my suggestion is to use

    List<Long> bitfieldReadonly(String key, String... arguments);

yangbodong22011:
> List<Long> bitfieldReadonly(String key, String... arguments);

Is this for scalability? Currently bitfield_ro only supports the `GET` subcommand, So i hardcoded it.

```
public void bitfieldReadonly(final byte[] key, final byte[] type, final byte[] offset) {
    sendCommand(BITFIELD_RO, key, SafeEncoder.encode("GET"), type, offset);
}
```

For scalability reasons, I think your suggestion is good.

sazzad16:
Yes.

yangbodong22011:
@sazzad16 Thanks，updated.

src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -3746,6 +3746,13 @@ public String aclGenPass() {
     return client.getIntegerMultiBulkReply();
   }
 
+  @Override
+  public List<Long> bitfieldReadonly(final String key, final String...arguments) {
```
sazzad16, line 3750:
```suggestion
  public List<Long> bitfieldReadonly(final String key, final String... arguments) {
```

src/main/java/redis/clients/jedis/BinaryClient.java
```diff
@@ -1320,6 +1320,10 @@ public void bitfield(final byte[] key, final byte[]... value) {
     sendCommand(BITFIELD, joinParameters(key, value));
   }
 
+  public void bitfieldReadonly(final byte[] key, final byte[]... value) {
+    sendCommand(BITFIELD_RO, joinParameters(key, value));
+  }
```
sazzad16, line 1325:
```suggestion
  public void bitfieldReadonly(final byte[] key, final byte[]... arguments) {
    sendCommand(BITFIELD_RO, joinParameters(key, arguments));
  }
```

PR #2195, opened 2020-05-20
[Ehancement]Try run command again after renew the redis cluster slot cache

diracccc:
Consider such a scenario in redis cluster, one master node down and the replica node promote to master, which means the cluster topological changed. So the request to the previous master node will fail and retry. But in this scenario, retry will never success.
Currently, Jedis will renew the slots cache in the last retry, release this invalid jedis instance from pool, then just throws a `JedisClusterMaxAttemptsException`, so this request get failed.
Maybe it's better to give a chance to retry run command after renew the cluster's slots cache, so I add this logic: in the second last attempt, renew slots cache, then run last retry.

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisClusterCommand.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterCommand.java
@@ -9,10 +9,13 @@ import redis.clients.jedis.exceptions.JedisNoReachableClusterNodeException;
 import redis.clients.jedis.exceptions.JedisRedirectionException;
 import redis.clients.jedis.util.JedisClusterCRC16;
 
+import java.util.concurrent.atomic.AtomicBoolean;
+
 public abstract class JedisClusterCommand<T> {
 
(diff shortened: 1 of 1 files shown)


wangminpku:
Dear diraccc(@diracccc ) and Jedis team (@xetorthio ), (@HeartSaVioR ), (@sazzad16 )

We are working on automatically identifying  and clustering the related pull requests for code review. We have found that [PR#1763](https://github.com/xetorthio/jedis/pull/1763) and [PR#1728](https://github.com/xetorthio/jedis/pull/1728) might be related to this one, because both of them modified  src/main/java/redis/clients/jedis/JedisClusterCommand.java for the similar purpose. 

We would really appreciate it if you could help us to validate if the information from [PR#1763](https://github.com/xetorthio/jedis/pull/1763) and [PR#1728](https://github.com/xetorthio/jedis/pull/1728) is helpful when you review the current PR. 

Thank you very much for your time!

sazzad16:
Same goal is now achieved due to #2358

PR #2203, opened 2020-06-06
Make `getConnection` abstract methods protected

jchambers:
Because the `getConnection` methods in `JedisClusterConnectionHandler` are package-private, they can only be implemented by classes inside the `redis.clients.jedis` package, and the prevents third parties from subclassing `JedisClusterConnectionHandler`.

Third parties may want to subclass `JedisClusterConnectionHandler` and related classes to add things like metrics or fault-tolerance measures. Making these methods public allows them to do so.


sazzad16:
@jchambers according to your statement, `protected` should be enough.

sazzad16 (changes requested):
according to the statement, `protected` should be enough

gkorland (approved):

sazzad16 (approved):

PR #2204, opened 2020-06-06
Add hincrByFloat in JedisCluster

### LINKED ISSUE #2202: JedisCluster lock the method "hincrByFloat(String key, String field, double value)"
### Expected behavior

JedisCluster has the method: hincrByFloat(String key, String field, double value)

### Actual behavior

jedisCluster don't have method: hincrByFloat(String key, String field, double value)

just has the method: hincrByFloat(final byte[] key, final byte[] field, final double value)

but Jedis has both methods

### Steps to reproduce:

check the JedisCluster

sazzad16:
Resolves #2202


gkorland:
@sazzad16 I think you should add a test here

sazzad16:
@gkorland done

gkorland (approved):

Issue #2183, opened 2020-04-14
 Add support for custom socket factories in JedisPool 


gkorland:
ref #2151

sazzad16:
Resolved by #2293

PR #2221, opened 2020-07-23
ref #2183 add socket factory to JedisPool

gkorland:


src/main/java/redis/clients/jedis/JedisFactory.java
```diff
@@ -19,17 +19,13 @@
  * PoolableObjectFactory custom impl.
  */
 class JedisFactory implements PooledObjectFactory<Jedis> {
-  private final AtomicReference<HostAndPort> hostAndPort = new AtomicReference<HostAndPort>();
-  private final int connectionTimeout;
-  private final int soTimeout;
+  
+  private final JedisSocketFactory jedisSocketFactory;
```
sazzad16, line 23: Shouldn't we use `AtomicReference`, may be in JedisSocketFactory, to match previous implementation?

gkorland: @sazzad16 I'm not sure why it was there to begin with

sazzad16: @gkorland #681

sazzad16:
@gkorland Close due to #2151 ?

Issue #2227, opened 2020-08-20
Concrete jedis exceptions are discarded when client falls back and fails to get a random node from slot-based connection handler


bigmarvin:
### Expected behavior

When experiencing no-reachable-node-in-cluster exception from getConnection() in JedisSlotBasedConnectionHandler, I'm expecting to get concrete exceptions thrown when individual jedis instance is borrowed, if any.

### Actual behavior

These concrete exceptions are discarded, which complicates the investigation of the no-reachable-node-in-cluster exception mentioned previously. Client has to guess whether something goes wrong on the way or the expected PONG is not delivered, or so.

### Steps to reproduce:

Please create a reproducible case of your problem. Make sure
that case repeats consistently and it's not random
1. Configure contact points of cluster with some invalid/unreachable hostname
2. Try to borrow some jedis instance based on slot

### Redis / Jedis Configuration

N/A

#### Jedis version:

3.3.0

#### Redis version:

Any

#### Java version:

Any

PR #2228, opened 2020-08-20
(#2227) carry suppressed exceptions in the one thrown eventually

bigmarvin:
The proposed change would be simply putting concrete exceptions in the one thrown eventually.

As the change is simple and has little impact in functionality, and I couldn't find any unit case using this JedisSlotBasedConnectionHandler directly, I'm not adding any case against this change. Please feel free to let me know if any concern or so around this.


bigmarvin:
@sazzad16, may I ask for your review and comments around this PR?

It's some simple change that brings additional info in logged stacks, which helps investigation, so I assume the feedback would be simple and straightforward. Thanks!

sazzad16:
@bigmarvin Please check my take on this issue in #2233

bigmarvin:
> @bigmarvin Please check my take on this issue in #2233

Thanks, so I assume the primary concern would be too many suppressed exceptions collected and logged in clusters of large scale, correct? I would have to say this is some good point, and one exception is some fair start point of investigation. Please proceed that way and I'm closing this PR.

sazzad16:
Also, `JedisNoReachableClusterNodeException` is being created only when it's necessary.

PR #2229, opened 2020-08-20
adds support for list lpos command

kenus34:
This commit adds support for the `lpos` command added in Redis 6.0.6


kenus34:
@sazzad16 Thanks a lot. I have made the changes and pushed the same.

src/main/java/redis/clients/jedis/commands/Commands.java
```diff
@@ -17,6 +17,8 @@
 import redis.clients.jedis.params.SetParams;
 import redis.clients.jedis.params.ZAddParams;
 import redis.clients.jedis.params.ZIncrByParams;
+import redis.clients.jedis.params.LPosParams;
+import redis.clients.jedis.util.SafeEncoder;
```
sazzad16, line 21: Not needed.
```suggestion
```

src/test/java/redis/clients/jedis/tests/commands/ListCommandsTest.java
```diff
@@ -328,6 +329,86 @@ public void lpop() {
 
   }
 
+  @Test
+  public void lpos() {
```
sazzad16, line 333: Test empty LPosParams.

Also, move this test method after all List related tests.

sazzad16 (changes requested):
Almost done!

src/main/java/redis/clients/jedis/commands/Commands.java
kenus34: Removed stray import

src/test/java/redis/clients/jedis/tests/commands/ListCommandsTest.java
kenus34: Refactored test towards the done and added a assertion with an empty paras list.

kenus34:
Thanks for the suggestion. I have resolved the comments.

sazzad16 (approved):

gkorland (approved):

PR #2231, opened 2020-08-24
Add toString method to the Params class

tgrall:
* As I am looking at CSC implementation I think it is interesting to add this method.
* I have not added any test as it is very basic  update, let me know if I need to

### DIFF
--- a/src/main/java/redis/clients/jedis/params/Params.java
+++ b/src/main/java/redis/clients/jedis/params/Params.java
@@ -67,4 +67,10 @@ public abstract class Params {
     params.put(name, null);
   }
 
+  @Override public String toString() {
+    if (params == null) {
+      params = new HashMap<>();
+    }
+    return params.toString();
+  }
 }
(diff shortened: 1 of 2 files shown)


sazzad16 (commented):
Params may contain `byte[]`. This solution does not handle this.

tgrall:
@sazzad16 thanks for your review.

I have updated the PR with the following:

* replace the simple Map.toString() using a List (that is closer to how params are used by applications)
* manage `byte[]` 
* add a simple test

src/main/java/redis/clients/jedis/params/Params.java
```diff
@@ -67,4 +67,21 @@ protected void addParam(String name) {
...
+    ArrayList<String> paramsAsString = new ArrayList<>();
+    if (params != null) {
+      for (Entry<String, Object> param : params.entrySet()) {
+        paramsAsString.add(param.getKey());
+        Object value = param.getValue();
+        if (value != null) {
+          if (value instanceof byte[]) {
+            paramsAsString.add( SafeEncoder.encode((byte[]) value));
+          } else {
+            paramsAsString.add(value.toString());
```
sazzad16, line 80: I think `paramsAsString.add(value);` is enough. Later `paramsAsString.toString();` will handle `value.toString()` anyway.

sazzad16 (changes requested):

sazzad16 (approved):

PR #2233, opened 2020-08-25
Carry first suppressed exception

### LINKED ISSUE #2227: Concrete jedis exceptions are discarded when client falls back and fails to get a random node from slot-based connection handler

sazzad16:
within JedisNoReachableClusterNodeException

Resolves #2227 
Closes #2228 
Closes #2307


gkorland:
@sazzad16 why only the first exception and not all the exceptions like in https://github.com/xetorthio/jedis/pull/2228?
If we go with only the first exception why not set the exception as the cause? 

```java
edisNoReachableClusterNodeException noReachableNode = new JedisNoReachableClusterNodeException("No reachable node in cluster", superssed);
```

sazzad16:
@gkorland This is to start a discussion about the approach. My disliking about #2228 is creating an exception (`JedisNoReachableClusterNodeException`) even when it is not necessary.

So, WDYT should be our approach? Suppressing all exception / first exception as cause / anything else?

sazzad16:
@gkorland Please check #2228 regarding same issue.

gkorland:
@sazzad16 what about collecting all the suppressed exceptions in a list and only in case we need to throw an exception use this list?

sazzad16:
@gkorland WDYT about #2228 ?

gkorland:
Yes I thought of some compromise, avoiding creating the exception in advance and still collect all the suppressed

gkorland (approved):

PR #2238, opened 2020-08-28
#2220 client-side-caching-poc2

tgrall:
A different approach to the PR #2236 

* here the client should use a different class `ClientSideCaching` and pass 2 jedis connection (the data and the invalidation / RESP2)
* this class implement "some" of the JedisCommands and manage the cache

Is that better than the other idea? 
It is a quick PoC, should I continue this path? 

If not interesting, I can simply start with a small blog post showing how to use CSC in a Jedis application (does not have to be "in the client lib" to start.


sazzad16:
@tgrall Looks good to me. WDYT about adding very limited number of constructors to create Jedis internally?

wplvlwc:
Can we implement client-side caching based on **resp3** and **"push"** message? 🧐

sazzad16:
Closing in favor of #3757

PR #2240, opened 2020-08-31
Update README.md

dengliming:
Currently we already release the verison `3.3.0`. See [https://github.com/xetorthio/jedis/releases/tag/jedis-3.3.0](https://github.com/xetorthio/jedis/releases/tag/jedis-3.3.0) , So we should update Official Releases version in README.


sazzad16:
Fixed.

Thanks @dengliming

PR #2245, opened 2020-09-05
This class should not be package private but should be possible to ex…

momomo:
…tend JedisFactory adnd JedisPool. Further, jedispool should create factory through a method to allow for overriding, alternatively be added as a constructor parameter. Current impl is bogged by constructors which  should really be taking a ConfigParam instead.


momomo:
Why did the tests fail?

gkorland (dismissed):

gkorland:
@sazzad16 any objections?

sazzad16:
@gkorland without any test, I'm a bit concerned. but you can go ahead if want.

PR #2247, opened 2020-09-12
Branch 20200912

momomo:


sazzad16:
@momomo It's hard to review a PR without any description and use case.

momomo:
This took some time! 

There are comments per commit. Do you not see them? 

You can however ignore: 

Extract new Jedis(...) to a method of its own to allow someone to override it should they desire to.

Since we overrid the entire method instead.

sazzad16:
@momomo 

> This took some time!

It wouldn't take that much time if it was easy recognize what it is.

About `newJedis` method:
Only use case I can see is in `makeObject()` but that method itself can be easily overridden. So I'd say no about this.

Your changes in JedisPool seems simple but necessary. Can you do this separate PR?

Also, can you create a separate PR with your changes in JedisPubSub with some tests?

momomo:
JedisPubSub should already have test cases for that. So should the constructors. The others are just turning things from package scope to protected / public. 

Yes, newJedis is not required. However, if that method makeObject starts to grow, you might not want to reimplement all of it, just might want to supply another class instance that implements the same base class. 

I will see when i will get around to do separate merges. I am not sure on how to do that.

sazzad16:
Closing due to lack of description and multiple _split-able_ PRs in one.

PR #2251, opened 2020-09-17
avoid send RST message when Socket.close() invoked

chimney-lee:
This setting socket.setSoLinger(true, 0)  will cause jedis client sent RST message when it receives FIN message from redis-server; which is an abortive connection release; 
Should we use an Orderly connection release here ?
https://docs.oracle.com/javase/1.5.0/docs/guide/net/articles/connection_release.html


chimney-lee:
@ewhauser  @wjw465150  thanks if have look at this issue

sazzad16:
@chimney-lee Is there any reason to close this PR?

sazzad16:
@dengliming Do you want to reopen?

sazzad16:
@chimney-lee Ah, thank you for the info.

PR #2254, opened 2020-09-23
JedisPubSub and BinaryJedisPubSub PING don't support optional argument

sabbey37:
According to the Redis docs at https://redis.io/commands/ping PING takes an optional message, and indeed the Jedis client does have both overloads with and without the message parameter. The overload with message is missing from JedisPubSub and BinaryJedisPubSub.

Relevant issues:
- https://github.com/xetorthio/jedis/issues/2049
- https://github.com/xetorthio/jedis/issues/1446
- https://github.com/xetorthio/jedis/pull/1448


sabbeyPivotal:
Test failure in Travis CI is unrelated to this PR and addressed in https://github.com/xetorthio/jedis/issues/2223

src/test/java/redis/clients/jedis/tests/commands/PublishSubscribeCommandsTest.java
```diff
@@ -82,7 +84,7 @@ public void onSubscribe(String channel, int subscribedChannels) {
   }
 
   @Test
-  public void pubSubChannelWithPingPong() throws InterruptedException {
+  public void pubSubChannelWithPingPong() {
```
sazzad16, line 87: Undo this. Keep the specific to issue/PR.

src/test/java/redis/clients/jedis/tests/commands/PublishSubscribeCommandsTest.java
```diff
@@ -301,6 +338,74 @@ public void onPMessage(byte[] pattern, byte[] channel, byte[] message) {
...
+    jedis.subscribe(new BinaryJedisPubSub() {
+
+      @Override
+      public void onSubscribe(byte[] channel, int subscribedChannels) {
+        publishOne("testchan1", "hello");
+      }
+
+      @Override
+      public void onMessage(byte[] channel, byte[] message) {
+        this.ping("hi!".getBytes());
```
sazzad16, line 388: Instread of String.getBytes(), use SafeEncoder.encode(String).

Alternate idea, keep SafeEncoder.encode(message) in variable and use it wherever you want.

src/test/java/redis/clients/jedis/tests/commands/PublishSubscribeCommandsTest.java
```diff
@@ -301,6 +338,74 @@ public void onPMessage(byte[] pattern, byte[] channel, byte[] message) {
...
+
+      @Override
+      public void onUnsubscribe(byte[] channel, int subscribedChannels) {
+        latchUnsubscribed.countDown();
+      }
+    }, SafeEncoder.encode("testchan1"));
+
+    assertEquals(0L, latchReceivedPong.getCount());
+    assertEquals(0L, latchUnsubscribed.getCount());
+    assertEquals("hi!", new String(pongPatterns.get(0)));
```
sazzad16, line 406: compare binary array with assertArrayEquals.

sazzad16 (changes requested):

src/test/java/redis/clients/jedis/tests/commands/PublishSubscribeCommandsTest.java
sabbey37: Good call. We also added `throws InterruptedException` to the other PingPong tests.

src/test/java/redis/clients/jedis/tests/commands/PublishSubscribeCommandsTest.java
sabbey37: That's a good idea!  Just updated it.

src/test/java/redis/clients/jedis/tests/commands/PublishSubscribeCommandsTest.java
sabbey37: Good suggestion!

src/main/java/redis/clients/jedis/BinaryJedisPubSub.java
```diff
@@ -10,6 +11,7 @@
 import java.util.Arrays;
 import java.util.List;
 
+import redis.clients.jedis.exceptions.JedisConnectionException;
```
sazzad16, line 14: unnecessary

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/BinaryJedisPubSub.java
sabbey37: Just removed it.

sazzad16 (approved):

sabbey37:
@gkorland I think @sazzad16 also wanted you to review this PR if possible.

src/main/java/redis/clients/jedis/BinaryJedisPubSub.java
```diff
@@ -115,6 +129,9 @@ private void process(Client client) {
         subscribedChannels = ((Long) reply.get(2)).intValue();
         final byte[] bpattern = (byte[]) reply.get(1);
         onPUnsubscribe(bpattern, subscribedChannels);
+      } else if (Arrays.equals(PONG.raw, resp)) {
+        final byte[] bpattern = (byte[]) reply.get(1);
```
gkorland, line 133: What does this `reply.get(1)` return when `ping()` is called without an argument?

sabbey37: Good question!  It actually returns an empty string as bytes.  However, we currently return `null` in `JedisPubSub` if no argument is provided:
```
else if (Arrays.equals(PONG.raw, resp)) {
        final byte[] bpattern = (byte[]) reply.get(1);
        final String strpattern = (bpattern == null) ? null : SafeEncoder.encode(bpattern);
        onPong(strpattern);
}
```

Perhaps we should return null in `BinaryJedisPubSub` as well instead of an empty string?

gkorland: @sazzad16 do we need to protect this use case?

sazzad16: @gkorland Yes. PING with message replies PONG with message here. Check `onPong` in `binaryPubSubChannelWithPingPongWithArgument` method.

gkorland (approved):

PR #2257, opened 2020-09-24
fix eval & evalsha in pipeline

minisancy:
add 'targetKey' as the first parameter to find the correct JedisClient from the cluster client list,
the targerKey should be match with the keys in lua script


sazzad16 (changes requested):
It would take some time to put a proper review. Nevertheless, for starter, this PR breaks backward compatibility. This should not happen <strike>unless it is absolute necessity</strike>. <strike>Right now, it does not seem like it.</strike>

sazzad16:

PR #2268, opened 2020-10-20
Jedis.hget document is outdated when key or field is not exists

hnourhani:


sazzad16:
@hnourhani Thank you for your interest in Jedis.

Jedis documentation is almost entirely based on Redis commands documentation. As Redis uses `nil`, that word has become common in Jedis even though it is not a Java term. Translating `nil` to Java `null` is no brainer but even if we change that everywhere in Jedis, it will again pop up due to new commands or some other changes in Redis. This would become a silly headache for maintainers. Because of all these, I'm rejecting this PR. Other existences of `null` in documentation have been removed already.

Please don't let this to discourage you from reporting issues and crafting PRs in Jedis. Thanks again :)

PR #2269, opened 2020-10-25
Add support for SET … GET option when version >=6.2.0 -resolves #2264

### LINKED ISSUE #2264: Add support for SET … GET option
Redis 6.2 supports a GET option to return the old value stored at key, or nil when key did not exist. See [https://redis.io/commands/set](https://redis.io/commands/set)

Jayice-zjw:
added support for SET ...GET option
See https://redis.io/commands/set
resolves #2264

### DIFF
--- a/src/main/java/redis/clients/jedis/params/SetParams.java
+++ b/src/main/java/redis/clients/jedis/params/SetParams.java
@@ -11,6 +11,7 @@ public class SetParams extends Params {
   private static final String NX = "nx";
   private static final String PX = "px";
   private static final String EX = "ex";
+  private static final String GET = "get";
 
   public SetParams() {
   }
@@ -57,6 +58,15 @@ public class SetParams extends Params {
     return this;
   }
 
+  /**
+   * Return the old value stored at key, or nil when key did not exist.
+   * @return SetParams
+   */
+  public SetParams get() {
+    addParam(GET);
+    return this;
+  }
+
   public byte[][] getByteParams(byte[]... args) {
     ArrayList<byte[]> byteParams = new ArrayList<>();
     for (byte[] arg : args) {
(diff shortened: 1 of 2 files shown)


sazzad16:
@Jayice-zjw Please add test codes.

Jayice-zjw:
> @Jayice-zjw Please add test codes.

done

src/main/java/redis/clients/jedis/params/SetParams.java
```diff
@@ -69,6 +79,9 @@ public SetParams xx() {
     if (contains(XX)) {
       byteParams.add(SafeEncoder.encode(XX));
     }
+    if (contains(GET)) {
+      byteParams.add(SafeEncoder.encode(GET));
+    }
```
sazzad16, line 84: Move this `if` block after all `if` blocks.

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/params/SetParams.java
Jayice-zjw: done.
btw, does the order of the if block here need to follow the order supported by the version?

sazzad16 (approved):

src/main/java/redis/clients/jedis/params/SetParams.java
sazzad16: not necessarily. just being safe here

sazzad16:
@Jayice-zjw I'd appreciate if you could also test `nil` when the key is absent.

Jayice-zjw:
> @Jayice-zjw I'd appreciate if you could also test `nil` when the key is absent.

all right.added

gkorland (approved):

PR #2275, opened 2020-11-11
throw exception if JedisCluster cannot connect to any host-port

chinafzy:
```
new JedisCluster(startNodes)
```
If none of the `startNodes` works:
+ JedisCluster does not work.
+ No message/exception comes up.
+ JedisCluster does not `retry` later.

---
`Throwing exception` seems to be a good solution.


sazzad16:
@chinafzy Sorry for late response. Are you available? Could you please update your branch and resolve the conflict?

PR #2289, opened 2020-11-27
Fix SSL cluster

gkorland:


sazzad16 (approved):
@gkorland Hats off :1st_place_medal:

sazzad16 (approved):

PR #2293, opened 2020-11-28
Add support to the use of JedisSocketFactory using a pool

mina-asham:
- Support for JedisSocketFactory has already been added to the lowest level Jedis to support adding any custom socket factory (e.g. UDS), this propagates the support in the JedisPool too

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryJedis.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedis.java
@@ -279,9 +279,9 @@ public class BinaryJedis implements BasicCommands, BinaryJedisCommands, MultiKey
   }
 
   /**
-   * @deprecated This will constructor will be removed in future releases
+   * @deprecated This constructor will be removed in future major release.
    *
-   * Use {@link BinaryJedis#BinaryJedis(redis.clients.jedis.JedisSocketFactory, redis.clients.jedis.JedisClientConfig)}
+   * Use {@link BinaryJedis#BinaryJedis(redis.clients.jedis.JedisSocketFactory, redis.clients.jedis.JedisClientConfig)}.
    */
   @Deprecated
   public BinaryJedis(final JedisSocketFactory jedisSocketFactory) {
--- a/src/main/java/redis/clients/jedis/BinaryJedis.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedis.java
@@ -278,8 +278,19 @@ public class BinaryJedis implements BasicCommands, BinaryJedisCommands, MultiKey
     }
   }
 
+  /**
+   * @deprecated This will constructor will be removed in future releases
+   *
+   * Use {@link BinaryJedis#BinaryJedis(redis.clients.jedis.JedisSocketFactory, redis.clients.jedis.JedisClientConfig)}
+   */
+  @Deprecated
   public BinaryJedis(final JedisSocketFactory jedisSocketFactory) {
     client = new Client(jedisSocketFactory);
+  }
+
+  public BinaryJedis(final JedisSocketFactory jedisSocketFactory, final JedisClientConfig clientConfig) {
+    client = new Client(jedisSocketFactory);
+    initializeFromClientConfig(clientConfig);
   }
 
   public boolean isConnected() {
--- a/src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
+++ b/src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
@@ -105,6 +105,21 @@ public class DefaultJedisSocketFactory implements JedisSocketFactory {
     }
   }
 
+  @Override
+  public JedisSocketFactory copyWith(HostAndPort hostAndPort) {
+    JedisClientConfig config =
+            DefaultJedisClientConfig.builder()
+                                    .connectionTimeoutMillis(this.connectionTimeout)
+                                    .socketTimeoutMillis(this.socketTimeout)
+                                    .ssl(this.ssl)
+                                    .sslSocketFactory(this.sslSocketFactory)
+                                    .sslParameters(this.sslParameters)
+                                    .hostnameVerifier(this.hostnameVerifier)
+                                    .hostAndPortMapper(this.hostAndPortMapper)
+                                    .build();
+    return new DefaultJedisSocketFactory(hostAndPort, config);
+  }
+
   public HostAndPort getSocketHostAndPort() {
     HostAndPortMapper mapper = getHostAndPortMapper();
     HostAndPort hostAndPort = getHostAndPort();
--- a/src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
+++ b/src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
@@ -3,6 +3,7 @@ package redis.clients.jedis;
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.net.Socket;
+import java.util.concurrent.atomic.AtomicReference;
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLParameters;
 import javax.net.ssl.SSLSocket;
@@ -16,7 +17,7 @@ public class DefaultJedisSocketFactory implements JedisSocketFactory {
   protected static final HostAndPort DEFAULT_HOST_AND_PORT = new HostAndPort(Protocol.DEFAULT_HOST,
       Protocol.DEFAULT_PORT);
 
-  private HostAndPort hostAndPort = DEFAULT_HOST_AND_PORT;
+  private final AtomicReference<HostAndPort> hostAndPort = new AtomicReference<>(DEFAULT_HOST_AND_PORT);
   private int connectionTimeout = Protocol.DEFAULT_TIMEOUT;
   private int socketTimeout = Protocol.DEFAULT_TIMEOUT;
   private boolean ssl = false;
@@ -36,7 +37,7 @@ public class DefaultJedisSocketFactory implements JedisSocketFactory {
   public DefaultJedisSocketFactory(String host, int port, int connectionTimeout, int socketTimeout,
       boolean ssl, SSLSocketFactory sslSocketFactory, SSLParameters sslParameters,
       HostnameVerifier hostnameVerifier) {
-    this.hostAndPort = new HostAndPort(host, port);
+    this.hostAndPort.set(new HostAndPort(host, port));
     this.connectionTimeout = connectionTimeout;
     this.socketTimeout = socketTimeout;
     this.ssl = ssl;
@@ -46,7 +47,7 @@ public class DefaultJedisSocketFactory implements JedisSocketFactory {
   }
 
   public DefaultJedisSocketFactory(HostAndPort hostAndPort, JedisClientConfig config) {
-    this.hostAndPort = hostAndPort;
+    this.hostAndPort.set(hostAndPort);
     if (config != null) {
       this.connectionTimeout = config.getConnectionTimeoutMillis();
       this.socketTimeout = config.getSocketTimeoutMillis();
@@ -106,18 +107,8 @@ public class DefaultJedisSocketFactory implements JedisSocketFactory {
   }
 
   @Override
-  public JedisSocketFactory copyWith(HostAndPort hostAndPort) {
-    JedisClientConfig config =
-            DefaultJedisClientConfig.builder()
-                                    .connectionTimeoutMillis(this.connectionTimeout)
-                                    .socketTimeoutMillis(this.socketTimeout)
-                                    .ssl(this.ssl)
-                                    .sslSocketFactory(this.sslSocketFactory)
-                                    .sslParameters(this.sslParameters)
-                                    .hostnameVerifier(this.hostnameVerifier)
-                                    .hostAndPortMapper(this.hostAndPortMapper)
-                                    .build();
-    return new DefaultJedisSocketFactory(hostAndPort, config);
+  public void updateHostAndPort(HostAndPort hostAndPort) {
+    this.hostAndPort.set(hostAndPort);
   }
 
   public HostAndPort getSocketHostAndPort() {
@@ -133,11 +124,11 @@ public class DefaultJedisSocketFactory implements JedisSocketFactory {
   }
 
   public HostAndPort getHostAndPort() {
-    return this.hostAndPort;
+    return this.hostAndPort.get();
   }
 
   public void setHostAndPort(HostAndPort hostAndPort) {
-    this.hostAndPort = hostAndPort;
+    this.hostAndPort.set(hostAndPort);
   }
 
   @Override
@@ -147,22 +138,22 @@ public class DefaultJedisSocketFactory implements JedisSocketFactory {
 
   @Override
   public String getHost() {
-    return this.hostAndPort.getHost();
+    return this.hostAndPort.get().getHost();
   }
 
   @Override
   public void setHost(String host) {
-    this.hostAndPort = new HostAndPort(host, this.hostAndPort.getPort());
+    this.hostAndPort.set(new HostAndPort(host, this.hostAndPort.get().getPort()));
   }
 
   @Override
   public int getPort() {
-    return this.hostAndPort.getPort();
+    return this.hostAndPort.get().getPort();
   }
 
   @Override
   public void setPort(int port) {
-    this.hostAndPort = new HostAndPort(this.hostAndPort.getHost(), port);
+    this.hostAndPort.set(new HostAndPort(this.hostAndPort.get().getHost(), port));
   }
 
   @Override
--- a/src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
+++ b/src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
@@ -3,7 +3,6 @@ package redis.clients.jedis;
 import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.net.Socket;
-import java.util.concurrent.atomic.AtomicReference;
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLParameters;
 import javax.net.ssl.SSLSocket;
@@ -17,7 +16,7 @@ public class DefaultJedisSocketFactory implements JedisSocketFactory {
   protected static final HostAndPort DEFAULT_HOST_AND_PORT = new HostAndPort(Protocol.DEFAULT_HOST,
       Protocol.DEFAULT_PORT);
 
-  private final AtomicReference<HostAndPort> hostAndPort = new AtomicReference<>(DEFAULT_HOST_AND_PORT);
+  private volatile HostAndPort hostAndPort = DEFAULT_HOST_AND_PORT;
   private int connectionTimeout = Protocol.DEFAULT_TIMEOUT;
   private int socketTimeout = Protocol.DEFAULT_TIMEOUT;
   private boolean ssl = false;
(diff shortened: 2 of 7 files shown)


src/main/java/redis/clients/jedis/JedisFactory.java
```diff
@@ -81,21 +73,20 @@
...
     this.password = JedisURIHelper.getPassword(uri);
     this.database = JedisURIHelper.getDBIndex(uri);
     this.clientName = clientName;
-    this.ssl = JedisURIHelper.isRedisSSLScheme(uri);
-    this.sslSocketFactory = sslSocketFactory;
-    this.sslParameters = sslParameters;
-    this.hostnameVerifier = hostnameVerifier;
   }
 
   public void setHostAndPort(final HostAndPort hostAndPort) {
```
sazzad16, line 87: This should be `synchronized` for reliablity as there is no AtomicReference.

src/main/java/redis/clients/jedis/JedisPool.java
```diff
@@ -283,6 +283,19 @@ public JedisPool(final GenericObjectPoolConfig poolConfig, final URI uri,
...
+    super(poolConfig, new JedisFactory(jedisSocketFactory, user, password, database, clientName));
+  }
+
+  public JedisPool(final GenericObjectPoolConfig poolConfig, final JedisSocketFactory jedisSocketFactory) {
+    super(poolConfig, new JedisFactory(jedisSocketFactory, null, null, Protocol.DEFAULT_DATABASE, null));
+  }
+
+  public JedisPool(final JedisSocketFactory jedisSocketFactory) {
+    super(new GenericObjectPoolConfig(), new JedisFactory(jedisSocketFactory, null, null, Protocol.DEFAULT_DATABASE, null));
+  }
```
sazzad16, line 297: Use `this` instead of `super`.

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/JedisPool.java
mina-asham: Good point, will update.

src/main/java/redis/clients/jedis/JedisFactory.java
mina-asham: That's a good point, but even if we add `synchronized` it would still be prone to some corner cases in the other methods since make/validate can still happen mid-updating the host/port.

I can think of a few solutions here (no particular ordering beside preference towards first one):
* We put the jedis socket factory in an atomic reference, we make the setHostAndPort copy/update/reset the jedis socket factory
* We put `synchronized` on all methods (overkill and I really think that's a bad option, but just listing)
* We keep an instance of HostAndPort in an atomic reference and cross check/validate with that (but then we duplicate the state and make it a bit too complex)
* The `setHostAndPort` method is only used in [sentinel pool](https://github.com/redis/jedis/blob/f24a817987fe1990a9ff65b31dba7fa4b44969e6/src/main/java/redis/clients/jedis/JedisSentinelPool.java#L180), we can mark the method as deprecated with a comment for removal in a future release, and change the sentinel bit to re-create the whole pool (although I don't know the sentinel code that well so maybe it's a bad idea)

What do you think?

sazzad16: Hmm, first one does seem like a prominent one if you have actually meant copy+update+reset.

mina-asham: Yup, I meant all not either (shouldn't have used `/` here :D), so something like:

```java
JedisSocketFactory newFactory = jedisSocketFactory.get().clone();
newFactory.setHost(...);
newFactory.setPort(...);
jedisSocketFactory.set(newFactory);
```

sazzad16 (approved):

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
```diff
@@ -121,4 +121,12 @@ public int getSoTimeout() {
   public void setSoTimeout(int soTimeout) {
     this.soTimeout = soTimeout;
   }
+
+  @Override
+  public JedisSocketFactory copy() {
```
gkorland, line 126: Why copy and not clone and clonable?

src/main/java/redis/clients/jedis/JedisSocketFactory.java
```diff
@@ -34,4 +34,6 @@
   int getSoTimeout();
 
   void setSoTimeout(int soTimeout);
+
+  JedisSocketFactory copy();
```
gkorland, line 38: Why copy and not clone and clonable?

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
mina-asham: I have always found clone/Cloneable not that helpful, specially in its contract:
* Where it strongly insist that you shouldn't call any constructors
* It throws `CloneNotSupportedException` and I don't want that propagating everywhere (although that can be suppress in subclass with try/catch/throw runtime but that's usually code smell)
* Objects should implement equality (not really wrong but not all objects need that)

That being said, I am not specifically against clone here if that's something more consistent with the rest of the codebase or if we think it's better for the use case here, what do you think?

src/main/java/redis/clients/jedis/JedisSocketFactory.java
mina-asham: Commented above, will resolve this conversation in favour of the above one but whatever we agree will be made here too.

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
gkorland: I think it's better to implement Clonable and clone(), notice you don't have to have to declare CloneNotSupportedException.

And just declare:
```
@override
public JedisSocketFactory clone() {
 return new DefaultJedisSocketFactory(
      this.host, this.port, this.connectionTimeout, this.soTimeout,
      this.ssl, this.sslSocketFactory, this.sslParameters, this.hostnameVerifier
    );
}
```

mina-asham: Well, that's the problem I meant, this implementation breaks the `clone` contract:
1. It calls a constructor
2. It doesn't call `super.clone()`

The implementation of clone that doesn't break the `Cloneable` contract would be:

```java
@Override
public DefaultJedisSocketFactory clone() throws CloneNotSupportedException {
  // forced to throw above because we need to call super.clone()
  DefaultJedisSocketFactory clone = (DefaultJedisSocketFactory) super.clone();
  clone.soTimeout = 123; // Apply any additional settings through fields directly (this line should go away, it's just here as an example)
  return clone;
}
```
```

Comment (mina-asham) [reply src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java:126]: @gkorland pinging here as a reminder, what do you think about the above?

**Comment by dengliming** [comment]
Please resolve the conflicts.

File: src/main/java/redis/clients/jedis/JedisSocketFactory.java [hunk 5]
```diff
@@ -24,6 +24,8 @@
    */
   Socket createSocket() throws IOException, JedisConnectionException;
 
+  JedisSocketFactory copyWith(HostAndPort hostAndPort);
```
sazzad16, line 27: I'm not that happy with this one. What could be other options?

Frankly, I was working with this very PR while you pushed the resolution of conflicts.
My very last thought process was to use a completely new JedisSocketFactory containing `Socket createSocket(HostAndPort)`.

mina-asham: > I'm not that happy with this one. What could be other options?

I agree, it's an annoying one and I couldn't (yet?) find a perfect solution here

> Frankly, I was working with this very PR while you pushed the resolution of conflicts.

Sorry for that, it just had too many changes after rebasing on top of your earlier config changes, so I had to abandon the branch and force a new one.

> My very last thought process was to use a completely new JedisSocketFactory containing `Socket createSocket(HostAndPort)`.

That would kind of break the concept as for UDS there is no real "host and port", though maybe that's okay since UDS is a very rare usecase but in that case we would need to also store a reference to hostandport in addition to the socket factory which is not the best thing to do IMO.

mina-asham: One alternative is to deprecate the setHostAndPort (currently only used in the sentinel pool I think and cleanup that completely)

sazzad16: > Sorry for that, ...

There's nothing to be sorry about :)

> ... deprecate the setHostAndPort ...

We have to keep sentinel under consideration as well, where master (HostAndPort) can be changed. If not with this pattern, may be with a different pattern.

mina-asham: > We have to keep sentinel under consideration as well, where master (HostAndPort) can be changed. If not with this pattern, may be with a different pattern.

I didn't mean kill the sentinel at all (though it looks like Redis 7 might retire it? https://github.com/redis/redis/projects/4), I meant replace that with something else that's more manageable (though maybe that's just moving the problem around?).

I will have a think about it, but I am also open to suggestions here!

sazzad16: > it looks like Redis 7 might retire it?

We are yet to retire sharding, sentinel would be a long shot :(

sazzad16: Just another idea.
```suggestion
  default JedisSocketFactory updateHostAndPort(HostAndPort hostAndPort) {
    return this;
  }
```

mina-asham: I don't object to the name, if we are okay with the JedisSocketFactory knowing explicitly about the hostandport then I think it's fine, I don't think we should provide a default here because that can lead to nasty bugs though, I would prefer it breaks compling if someone is already using the socket factory feature (which is a very niche feature and I don't think anyone is actually using it). I will add another commit with these changes and we can have a look.

src/main/java/redis/clients/jedis/JedisFactory.java
```diff
@@ -24,7 +24,7 @@
 
   private static final Logger logger = LoggerFactory.getLogger(JedisFactory.class);
 
-  private final AtomicReference<HostAndPort> hostAndPort = new AtomicReference<>();
+  private final AtomicReference<JedisSocketFactory> jedisSocketFactory = new AtomicReference<>();
```
sazzad16, line 27: It seems we don't have any more constructor where JedisSocketFactory is not being set.
```suggestion
  private final JedisSocketFactory jedisSocketFactory;
```

mina-asham: We do 😢 , specifically these three:

```java
/**
 * {@link #setHostAndPort(redis.clients.jedis.HostAndPort) setHostAndPort} must be called later.
 */
protected JedisFactory(final int connectionTimeout, final int soTimeout, final int infiniteSoTimeout,
    final String user, final String password, final int database, final String clientName) {
  this(connectionTimeout, soTimeout, infiniteSoTimeout, user, password, database, clientName, false, null, null, null);
}
```

```java
/**
 * {@link #setHostAndPort(redis.clients.jedis.HostAndPort) setHostAndPort} must be called later.
 */
protected JedisFactory(final int connectionTimeout, final int soTimeout, final int infiniteSoTimeout,
    final String user, final String password, final int database, final String clientName, final boolean ssl,
    final SSLSocketFactory sslSocketFactory, final SSLParameters sslParameters, final HostnameVerifier hostnameVerifier) {
  this(DefaultJedisClientConfig.builder().connectionTimeoutMillis(connectionTimeout)
      .socketTimeoutMillis(soTimeout).blockingSocketTimeoutMillis(infiniteSoTimeout).user(user)
      .password(password).database(database).clientName(clientName)
      .ssl(ssl).sslSocketFactory(sslSocketFactory)
      .sslParameters(sslParameters).hostnameVerifier(hostnameVerifier).build());
}
```

```java
/**
 * {@link #setHostAndPort(redis.clients.jedis.HostAndPort) setHostAndPort} must be called later.
 */
protected JedisFactory(final JedisClientConfig clientConfig) {
  this.clientConfig = clientConfig;
}
```

sazzad16: Ahh, I missed the last one.

sazzad16: How about a new constructor?

    DefaultJedisSocketFactory(JedisClientConfig config)

mina-asham: Got you, so we basically move the NPE to the socket factory instead the jedis factory, I think that makes sense, will update.

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
```diff
@@ -46,7 +51,7 @@ public DefaultJedisSocketFactory(String host, int port, int connectionTimeout, i
   }
 
   public DefaultJedisSocketFactory(HostAndPort hostAndPort, JedisClientConfig config) {
-    this.hostAndPort = hostAndPort;
+    this.hostAndPort.set(hostAndPort);
```
sazzad16, line 54: Shouldn't we check null here?
It might be better to keep the default HostAndPort.

mina-asham: For consistency with the rest of the Jedis code I would opt for not null checking here, all the constructors I have seen in the codebase do accept null (and while it would be a bad idea) it is only possible if the user does that on purpose.

I don't mind adding it though if you feel strongly about it

sazzad16: I'm asking because of this https://github.com/redis/jedis/pull/2293/files#diff-5c1bd2b61d33731ac162212486900a5408ba527baddcdb5decdf41d1e4f26c53R37 

IMO, we should either check null here or avoid sending null there (`this(null, config);`).

mina-asham: Ah, I think I misunderstood you, I though you meant `if (hostAndPort == null) throw new ArgumentException("HostAndPort cannot be null"` but now I think you meant `if (hostAndPort != null) this.hostAndPort.set(hostAndPort)`?

If it's the latter then I totally agree and I think it makes perfect sense, is that what you meant?

sazzad16: YES

sazzad16 (approved):
I'm not 100% about the change in 

    src/main/java/redis/clients/jedis/JedisSocketFactory.java

I'm leaving this discussion/debate to other reviewers.

mp911de:
Introducing `JedisSocketFactory` is a neat idea. However, it raises the question of how addressing should work. Right now, `HostAndPort` was the only mechanism to obtain a connection. With domain sockets, you need another way to connect to a node. I'd suggest introducing an interface and replace all usages of `HostAndPort` (breaking change, potentially a 4.0 topic) with that interface. Behind that interface, one can put addressing details so one could e.g. connect to Sentinel using domain sockets (or do other crazy things). You could even have an SSL implementation of `JedisSocketFactory` that uses SSL details from a `SecureHostAndPort`.

mina-asham:
> Introducing `JedisSocketFactory` is a neat idea. However, it raises the question of how addressing should work. Right now, `HostAndPort` was the only mechanism to obtain a connection. With domain sockets, you need another way to connect to a node. I'd suggest introducing an interface and replace all usages of `HostAndPort` (breaking change, potentially a 4.0 topic) with that interface. Behind that interface, one can put addressing details so one could e.g. connect to Sentinel using domain sockets (or do other crazy things). You could even have an SSL implementation of `JedisSocketFactory` that uses SSL details from a `SecureHostAndPort`.

I think that makes sense, it will be a major overhaul though so I don't think we should block this here, having something that works for 3.x (specially that we already support that in the Jedis class) and then we work on something bigger for 4.x

src/main/java/redis/clients/jedis/BinaryJedis.java
```diff
@@ -278,10 +278,21 @@ private void initializeFromURI(URI uri) {
     }
   }
 
+  /**
+   * @deprecated This will constructor will be removed in future releases
```
gkorland, line 282:
```suggestion
   * @deprecated This constructor will be removed in future releases
```

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
```diff
@@ -16,7 +17,7 @@
   protected static final HostAndPort DEFAULT_HOST_AND_PORT = new HostAndPort(Protocol.DEFAULT_HOST,
       Protocol.DEFAULT_PORT);
 
-  private HostAndPort hostAndPort = DEFAULT_HOST_AND_PORT;
+  private final AtomicReference<HostAndPort> hostAndPort = new AtomicReference<>(DEFAULT_HOST_AND_PORT);
```
gkorland, line 20: Do you really need AtomicReference or a `volatile` will do?

src/main/java/redis/clients/jedis/BinaryJedis.java
```diff
@@ -278,10 +278,21 @@ private void initializeFromURI(URI uri) {
...
+   *
+   * Use {@link BinaryJedis#BinaryJedis(redis.clients.jedis.JedisSocketFactory, redis.clients.jedis.JedisClientConfig)}
```
sazzad16, line 284:
```suggestion
   * @deprecated This constructor will be removed in future major release.
   *
   * Use {@link BinaryJedis#BinaryJedis(redis.clients.jedis.JedisSocketFactory, redis.clients.jedis.JedisClientConfig)}.
```

src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -153,10 +153,20 @@ public Jedis(final URI uri, JedisClientConfig config) {
     super(uri, config);
   }
 
+  /**
+   * @deprecated This will constructor will be removed in future releases
+   *
+   * Use {@link Jedis#Jedis(redis.clients.jedis.JedisSocketFactory, redis.clients.jedis.JedisClientConfig)}
```
sazzad16, line 159:
```suggestion
   * @deprecated This constructor will be removed in future major release.
   *
   * Use {@link Jedis#Jedis(redis.clients.jedis.JedisSocketFactory, redis.clients.jedis.JedisClientConfig)}.
```

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
mina-asham: volatile will work fine, we are not doing any CAS operations so it should be the same

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
```diff
@@ -105,6 +111,11 @@ public Socket createSocket() throws JedisConnectionException {
     }
   }
 
+  @Override
+  public void updateHostAndPort(HostAndPort hostAndPort) {
```
sazzad16, line 115: Should we use `synchronized` here?

gkorland: why? if you're using `volatile` you don't need.

gkorland (approved):

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
sazzad16: @gkorland If we had check-and-set, we should have. But we don't have it here. So it's okay.
(sorry, for a second I forgot that CAS is absent)

sazzad16 (approved):

Issue #2303, opened 2020-12-06
Use Keywords in Param classes


sazzad16:
> I would have set these (STORE &STOREDIST) as Keywords so you won't have to re-serialize them on each call

_Originally posted by @gkorland in https://github.com/redis/jedis/pull/2157#discussion_r392667981_

> None of the Params classes use Keyword yet. May be we can create a separate issue to improve this?

_Originally posted by @sazzad16 in https://github.com/redis/jedis/pull/2157#discussion_r396091103_

MrFrydae:
As I am unfamiliar with this codebase, could you please give me an example of how you would want this to be implemented?

Instead of ``addParam(STORE, key);`` with ``STORE`` being a String, would you want it to be converted to a Keyword?

sazzad16:
@MrFrydae Yes.

dengliming:
Currently, All fields of `Params` were put into a `HashMap`. e.g.

```
public class ZIncrByParams extends Params {

  private static final String XX = "xx";
  private static final String NX = "nx";
  private static final String INCR = "incr";
  ....
  public byte[][] getByteParams(byte[] key, byte[]... args) {
    ArrayList<byte[]> byteParams = new ArrayList<>();
    byteParams.add(key);

    if (contains(NX)) {
      byteParams.add(SafeEncoder.encode(NX));
    }
    if (contains(XX)) {
      byteParams.add(SafeEncoder.encode(XX));
    }

    byteParams.add(SafeEncoder.encode(INCR));

    Collections.addAll(byteParams, args);
    return byteParams.toArray(new byte[byteParams.size()][]);
  }
}

```
Wouldn't it be better if we changed it to the following? no need to call `contains(XX)` and `get(xxx)` from a `Map` and something like `INCR` will be place in `Keyword`.

```
public class ZIncrByParams extends Params {

  private boolean xx = false;
  private boolean nx = false;
  private boolean incr = false;

  public ZIncrByParams nx() {
    this.nx = true;
    return this;
  }
  .....

  public byte[][] getByteParams(byte[] key, byte[]... args) {
    ArrayList<byte[]> byteParams = new ArrayList<>();
    byteParams.add(key);

    if (nx) {
      byteParams.add(SafeEncoder.encode(NX));
    }
    if (xx) {
      byteParams.add(SafeEncoder.encode(XX));
    }
    if (incr) {
      byteParams.add(SafeEncoder.encode(INCR));
    }

    Collections.addAll(byteParams, args);
[...]

PR #2305, opened 2020-12-07
XRead(Group) Params with allowing block=0

### LINKED ISSUE #2277: xread with block=0 doesn't block
### Expected behavior

According to [redis documentation](https://redis.io/topics/streams-intro) when passing block=0 to xread command, the operation should block forever (until has data to read).

From the documentation:
> "... I specified the new BLOCK option with a timeout of 0 milliseconds (that means to never timeout)"

In Jedis, when I pass block=0 it seems like the BLOCK param is omitted at all from the sent command thus the xread command won't block at all.

**BinaryClient.java:**
```java
public void xread(final int count, final long block, final Map<byte[], byte[]> streams) {
    ...
    if(block > 0) {
        params[streamsIndex++] = Keyword.BLOCK.raw;
        params[streamsIndex++] = toByteArray(block);
    }
    ...
```

I also validate this behavior with the python client and redis-cli and indeed block=0 means block forever.

I guess the block argument to Jedis.xread should be optional, when provided it should be passed to redis for any value >= 0 (negative value are not allowed)

#### Java version: 11

Thanks.

sazzad16:
Resolves #2277


src/main/java/redis/clients/jedis/params/XReadGroupParams.java
```diff
@@ -0,0 +1,31 @@
+package redis.clients.jedis.params;
+
+public class XReadGroupParams extends Params {
+
+  private static final String COUNT = "COUNT";
+  private static final String BLOCK = "BLOCK";
+  private static final String NOACK = "NOACK";
+
```
dengliming, line 8: Should we keep these in `keyword`? (not sure : )

dengliming (commented):
Personally, I think we can do this logic by simply determining whether the `block ` is greater than 0 or not, and in general I usually think that if a block is equal to 0 or less than 0, you don't need a block, and we can explain that in comment. Using `XReadParams` is good too. can avoid having too many arguments for this method. and we can also comment or deprecated `void xread(int count, long block, Entry<String, StreamEntryID>... streams); ` since the block param is not working.  Thanks. : )

src/main/java/redis/clients/jedis/params/XReadGroupParams.java
sazzad16: Yes, absolutely! And we plan to do that #2303. But yet to get a PR (from community) or manage some (of our) time :(

sazzad16:
@dengliming 
> we can also comment or deprecated `void xread(int count, long block, Entry<String, StreamEntryID>... streams);` since the block param is not working

Nice suggestion, thanks

src/main/java/redis/clients/jedis/params/XReadGroupParams.java
dengliming: yeah. Community development is really time consuming. Thank you for your efforts 👍

sazzad16: @dengliming Could you please also make comments on #2360 and #2375? We are certain on having the changes but in real dilemma on what approach to follow regarding deprecation. Thanks again :)

dengliming:
@sazzad16 Can you resolve the conflicts so that we can take this pr here? : )

src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -4023,6 +4009,23 @@ public Long xlen(final String key) {
...
+
+    if (!xReadParams.hasBlock()) {
+      return BuilderFactory.STREAM_READ_RESPONSE.build(client.getObjectMultiBulkReply());
+    }
+
+    client.setTimeoutInfinite();
+    try {
+      return BuilderFactory.STREAM_READ_RESPONSE.build(client.getObjectMultiBulkReply());
+    } finally {
+      client.rollbackTimeout();
```
dengliming, line 4124: Does the `infiniteSoTimeout` will affect the block time? (Not sure)

sazzad16: @dengliming No. It effects the socket time(out).

sazzad16: https://github.com/redis/jedis/blob/f35484c9abae2350b6eed6e557830edc30d0ff50/src/test/java/redis/clients/jedis/tests/JedisTest.java#L116-L129

dengliming (approved):
LGTM

dengliming (dismissed):

dengliming (approved):

src/main/java/redis/clients/jedis/params/XReadGroupParams.java
```diff
@@ -0,0 +1,31 @@
...
+  public static XReadGroupParams xReadGroupParams() {
```
gkorland, line 9: @sazzad16 I would have consider a different name for `xReadGroupParams()` perhaps `create()`

dengliming: Seems that static method is not really necessary. Because user can also new Object by `none args Constructor`. Maybe we can also keep `none args Constructor` private?

sazzad16: @gkorland It is there for method chaining. Any _good enough_ name is okay to me.

@dengliming Empty constructor is still an option to the users if they want to use. I don't to hide it until it is hurting the development process.

sazzad16: @gkorland We are carrying this pattern probably since Jedis 2.10.0. Each of the Params classes have a static method with name at this pattern. Some of already merged PR's for 3.6.0 include this pattern. If we want replace this pattern with something like `create()`, we'll always have time to do that. IMHO, we shouldn't keep this PR on hold just for this matter if we agree on the implementation of this PR.

gkorland: > Each of the Params classes have a static method with name at this pattern.

If that is the common pattern then lest leave it like that

gkorland (approved):

PR #2314, opened 2020-12-13
Add ability to reset password in JedisFactory for JedisPool and JedisSentinelPool

### LINKED ISSUE #533: Make JedisFactory as public class
Beacause we need to write the JedisPool/SentinelPool ourself in the project the match the project requirements, we hope to reuse the JedisFactory, but it is not a public class,

### LINKED ISSUE #1602: Request New Feature: allowing the ability to reset password during runtime for Jedis's auth
### Expected behavior

When Redis's password changed, without redeploy, the JedisFactory can get the new password during making a new incoming Object.

### Actual behavior

JedisFactory is package protected, we couldn't extend from it to make our own customized factory.

### Our proposal

1, Inside JedisPool.java, add a new constractor
  ```
 public JedisPool(final GenericObjectPoolConfig poolConfig, final PooledObjectFactory<Jedis> factory) {
		  super(poolConfig, factory);
  }
```
2,  Change JedisFactory.java to public if possible
3, Inside JedisFactory.java (Optional)
  * add  `private final AtomicReference<String> password = new AtomicReference<String>();`
  * add `setPassword(final String password)` method

If above is approved/accepted, our application can use the new constructor to pass in JedisFactory or our own customized factory to allow reset password.

#### Jedis version:
3.0.0-SNAPSHOT
#### Redis version:
2.8/3.0
#### Java version:
1.8

Thanks,
Yan

### PR DESCRIPTION
**Pull request by sazzad16**
Original author: @yapei123 (#1606) 

Closes #1602 
Closes #1606 

Closes #533 
Closes #2245

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisFactory.java
+++ b/src/main/java/redis/clients/jedis/JedisFactory.java
@@ -37,31 +37,31 @@ public class JedisFactory implements PooledObjectFactory<Jedis> {
     this(host, port, connectionTimeout, soTimeout, password, database, clientName, false, null, null, null);
   }
 
-  JedisFactory(final String host, final int port, final int connectionTimeout,
+  public JedisFactory(final String host, final int port, final int connectionTimeout,
                final int soTimeout, final String user, final String password, final int database, final String clientName) {
     this(host, port, connectionTimeout, soTimeout, 0, user, password, database, clientName);
   }
 
-  JedisFactory(final String host, final int port, final int connectionTimeout, final int soTimeout,
+  public JedisFactory(final String host, final int port, final int connectionTimeout, final int soTimeout,
       final int infiniteSoTimeout, final String user, final String password, final int database, final String clientName) {
     this(host, port, connectionTimeout, soTimeout, infiniteSoTimeout, user, password, database, clientName, false, null, null, null);
   }
 
-  JedisFactory(final String host, final int port, final int connectionTimeout,
+  public JedisFactory(final String host, final int port, final int connectionTimeout,
       final int soTimeout, final String password, final int database, final String clientName,
       final boolean ssl, final SSLSocketFactory sslSocketFactory, final SSLParameters sslParameters,
       final HostnameVerifier hostnameVerifier) {
     this(host, port, connectionTimeout, soTimeout, null, password, database, clientName, ssl, sslSocketFactory, sslParameters, hostnameVerifier);
   }
 
-  JedisFactory(final String host, final int port, final int connectionTimeout,
+  public JedisFactory(final String host, final int port, final int connectionTimeout,
                final int soTimeout, final String user, final String password, final int database, final String clientName,
                final boolean ssl, final SSLSocketFactory sslSocketFactory, final SSLParameters sslParameters,
                final HostnameVerifier hostnameVerifier) {
     this(host, port, connectionTimeout, soTimeout, 0, user, password, database, clientName, ssl, sslSocketFactory, sslParameters, hostnameVerifier);
   }
 
-  JedisFactory(final String host, final int port, final int connectionTimeout, final int soTimeout,
+  public JedisFactory(final String host, final int port, final int connectionTimeout, final int soTimeout,
       final int infiniteSoTimeout, final String user, final String password, final int database,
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/main/java/redis/clients/jedis/JedisFactory.java [hunk 1]
```diff
@@ -18,50 +18,50 @@
...
  */
-class JedisFactory implements PooledObjectFactory<Jedis> {
+public class JedisFactory implements PooledObjectFactory<Jedis> {
   private final AtomicReference<HostAndPort> hostAndPort = new AtomicReference<>();
   private final int connectionTimeout;
   private final int soTimeout;
   private final int infiniteSoTimeout;
   private final String user;
-  private final String password;
+  private final AtomicReference<String> password = new AtomicReference<>();
```
gkorland, line 27: why do you need AtomicReference? volatile should do here, no?

sazzad16: > why do you need AtomicReference?

because I have borrowed that piece of code from #1606 

> volatile should do here, no?

yes. both of those are same in java, isn't it?

guykorlandredis: for get/set you're right but you just don't need it.

sazzad16: addressed

src/main/java/redis/clients/jedis/JedisFactory.java
```diff
@@ -116,6 +116,20 @@ public void setHostAndPort(final HostAndPort hostAndPort) {
...
 
+  public void setPassword(final String password) throws IllegalArgumentException {
+    if (this.user != null) {
+      throw new IllegalArgumentException();
+    }
+    this.password = password;
+  }
+
+  public void setPassword(final String user, final String password) throws IllegalArgumentException {
+    if (!java.util.Objects.equals(this.user, user)) {
```
gkorland, line 127: @sazzad16 why do we force same user?

src/main/java/redis/clients/jedis/JedisSentinelPool.java
```diff
@@ -175,6 +175,27 @@ public JedisSentinelPool(String masterName, Set<String> sentinels,
...
+    this.clientName = null;
+    this.sentinelConnectionTimeout = Protocol.DEFAULT_TIMEOUT;
+    this.sentinelSoTimeout = Protocol.DEFAULT_TIMEOUT;
+    this.sentinelUser = null;
+    this.sentinelPassword = null;
+    this.sentinelClientName = null;
+    this.factory = factory;
+
+    HostAndPort master = initSentinels(sentinels, masterName);
+    initPool(poolConfig, factory);
```
gkorland, line 195: @sazzad16 perhaps I get it wrong, but do you need to call initPool twice?

src/main/java/redis/clients/jedis/JedisFactory.java
sazzad16: @gkorland In redis-cli on Redis ACL, after doing `AUTH user pass` once, if we want to `AUTH` again, we have to do `AUTH user newpass` not just `AUTH newpass`. I tried to imitate that here.
I also thought, this would help to us to avoid any mistaken password change as a wrong password (but a valid one for other user) would break the whole pool.

src/main/java/redis/clients/jedis/JedisSentinelPool.java
sazzad16: @gkorland `initPool(poolConfig, factory);` was usually get called within `initPool(master);` (Line 219). But as we are doing `this.factory = factory;` (Line 192) now, the `if (factory == null)` check (Line 216) would not let `initPool(poolConfig, factory);` to happen at usual place. So we're doing it here.

gkorland: Please at least leave a comment in the code explaining this double call.

src/main/java/redis/clients/jedis/JedisFactory.java
gkorland: Please add a JavaDoc explaining it

Issue #2091, opened 2019-11-07
jedisCluster不支持批量读写有解决方案么


chenbinglong:
### Expected behavior

jediscluster支持批量读写

### Actual behavior

jediscluster目前不支持批量读写 mget满足不了

### Steps to reproduce:

Please create a reproducible case of your problem. Make sure
that case repeats consistently and it's not random
1.
2.
3.

startjava:
i know it ..not support!

mekingshow:
+1

sazzad16:
Resolved by https://github.com/redis/jedis/commit/858c8051301854e8343eef66aa68c6b9dab1c328 

Introduced ClusterPipeline

PR #2324, opened 2020-12-23
Support Pipelining in a Cluster HashSlot

sazzad16:
Ref #2091


yangbodong22011:
After thinking twice, I don't think we need to provide this feature, maybe we only need to provide an example as follows:
```
try {
  // NOTICE: Use cluster pipeline, your keys must be in same slot

  // step1: Define a hashtag, refer: https://redis.io/topics/cluster-spec#keys-hash-tags
  String hashtag = "{hashtag}";
  // step2: Get a jedis connection
  jedis = jedisCluster.getConnectionFromSlot(JedisClusterCRC16.getSlot(hashtag));
  // step3: Use pipeline with hashtag+key
   Pipeline p = jedis.pipelined();
   p.set(hashtag + ".string", "foo");
   p.get(hashtag + ".string");
   p.sync();
} finally {
  if (jedis != null) {
    jedis.close();  // return jedis to cluster JedisPool
  }
}
```


Another important point is that after calling `JedisCluster.startPipeline()`, we must call `pipeline.close()` to return the connection to the JedisPool, but for Jedis users, the habit is to call `Jedis.close(`) instead of `pipeline.close()` , I think this is important.

sazzad16:
@yangbodong22011 

Firstly,

> Another important point is that after calling JedisCluster.startPipeline(), we must call pipeline.close() to return the connection to the JedisPool, but for Jedis users, the habit is to call Jedis.close() instead of pipeline.close() , I think this is important.

I would argue that the habit is actually to call `Closeable.close()`.

---

Coming to the point, at some point in history, users had to call `pool.returnResource()` / `pool.returnBrokenResource()` to return the connection. Taking the advantage of Java try-with-resources feature and familiarity of `Closeable.close()` method with the Java users and to reduce users' hassle, `close()` method was manipulated. Now, I'm trying the same thing for Pipeline.

Let's simple comparison with your code:

```java
  // NOTICE: Use cluster pipeline, your keys must be in same slot

  // step1: Define a hashtag, refer: https://redis.io/topics/cluster-spec#keys-hash-tags
  String hashtag = "{hashtag}";
  // step2: Use pipeline with hashtag+key
  try (Pipeline p = jedisCluster.startPipeline(hashtag)) {
    p.set(hashtag + ".string", "foo");
    p.get(hashtag + ".string");
    p.sync();
  }
```

If you can see here, the difference is not only in the number of lines.

- Users would have to know about fewer things. E.g. without knowledge of
  - `jedisCluster.getConnectionFromSlot()`
  - `JedisClusterCRC16.getSlot(hashtag)`
- Users would have to handle fewer objects. E.g. no need to handle
  - a Jedis object [1]

[1] It is worth mentioning that codes/programs that uses JedisCluster do not handle Jedis objects (directly).

From my experience, I can say that when it is the matter of adaptability, the above things really come into play. So, even after your argument, I'm still backing this PR :)

yangbodong22011:
@sazzad16 Thanks for your response, I agree with you, let's keep going.

About code, why we need `synchronized` on `Jedis#setDataSource` and `Jedis#unsetDataSource`. I know that `JedisPool#getResouce` may be called in multi threads, but `GenericObjectPool#borrowObject()` should be thread safe, and the code has never been `synchronized`. Is it because you found some bugs or other reasons, maybe adding comments would be better.

sazzad16:
@yangbodong22011 Thank you for noticing. I was trying some other things but both of these changes shouldn't be in same PR. I'll undo.

yangbodong22011 (dismissed):
LGTM

src/main/java/redis/clients/jedis/BinaryJedis.java
```diff
@@ -222,6 +222,16 @@ private void initializeClientFromURI(URI uri, final SSLSocketFactory sslSocketFa
     }
   }
 
+  /**
+   * @deprecated Use {@link Jedis#startPipeline() Jedis.startPipeline()}
+   * @return 
+   */
+  public Pipeline pipelined() {
```
gkorland, line 229:
```suggestion
   */
  @Deprecated
  public Pipeline pipelined() {
```

src/main/java/redis/clients/jedis/Pipeline.java
```diff
@@ -62,15 +74,31 @@ public void addResponse(Response<?> response) {
     }
   }
 
+  /**
+   * @deprecated This will be removed in future.
+   * @param client
+   */
```
gkorland, line 80:
```suggestion
   */
   @Deprecated 
```

src/main/java/redis/clients/jedis/Pipeline.java
```diff
@@ -62,15 +74,31 @@ public void addResponse(Response<?> response) {
...
   @Override
   protected Client getClient(byte[] key) {
     return client;
   }
 
+  /**
+   * @deprecated This will be final in future.
+   * @param key
+   * @return 
+   */
```
gkorland, line 101:
```suggestion
   */
   @Deprecated 
```

src/main/java/redis/clients/jedis/Pipeline.java
```diff
@@ -62,15 +74,31 @@ public void addResponse(Response<?> response) {
...
+    if (this.resource == null) {
+      this.client = client;
+    }
   }
 
+  /**
+   * @deprecated This will be final in future.
+   * @param key
+   * @return 
+   */
```
gkorland, line 91:
```suggestion
   */
  @Deprecated 
```

gkorland (changes requested):

yangbodong22011 (approved):

agnarok:
Hey guys!
Does anyone know when they intend to merge this PR? :eyes:

sazzad16:
@agnarok Hi!
My idea was to introduce both pipeline and transaction in a hash slot. As I couldn't do the transaction, we (actually myself) have backed down from merging this.

Issue #2105, opened 2019-11-24
Exception=redis.clients.jedis.exceptions.JedisConnectionException: java.net.SocketTimeoutException: Read timed out


parag90-zz:
Sometimes I am getting these exceptions on my application connecting to Redis.  

I am running a service with 40 containers (spring-boot) 
Env: AWS
Redis: AWS managed
Jedis version: 3.1.0
connect.timeout: 700
  write.timeout: 1000
  maxIdle: 20
  minIdle: 20
  maxActive: 20
Node type:cache.r4.large
Engine Version: 4.0.10
Java version 1.8
Traffic on the service: 5-7000 Transactions per second. 
Issue frequency: Sometimes in 10 days sometimes once a week, No consistency observed. No changes in the incoming traffic pattern at that time. 

Observations: New connections at Redis side spikes to 25-30k and the whole system becomes unstable. Huge spike in network packets in during the issue. 

Need help in debugging the issue, This causes a live site the service availability going down. 

@Configuration
@ConfigurationProperties(prefix = "redis")
public class CacheConfiguration {

    Map<String, String> hostname = new HashMap<>();

    @Value("${redis.port:6379}")
    int redisPort;

    @Value("${redis.connect.timeout:1000}")
    int connectTimeOut;

    @Value("${redis.write.timeout:1000}")
    int writeTimeOut;

    @Value("${redis.enabled:true}")
    boolean isRedisEnabled;

    @Value("${redis.maxIdle:5}")
    int maxIdle;

    @Value("${redis.minIdle:1}")
    int minIdle;

    @Value("${redis.maxActive:10}")
    int maxActive;

    public Map<String, String> getHostname() {
        return hostname;
    }

    /**
     * Bean for elasticache.
     *
     * @return configured jedisPool
     */
    @Bean
    public JedisPool getRedisConfiguration() {
        JedisPool jedisPool = null;
        if (isRedisEnabled) {
            final JedisPoolConfig jedisPoolConfig = new JedisPoolConfig();
            jedisPoolConfig.setMaxWaitMillis(writeTimeOut);
            jedisPoolConfig.setMaxTotal(maxActive);
            jedisPoolConfig.setMaxIdle(maxIdle);
            jedisPoolConfig.setMinIdle(minIdle);
            final Region region = (Regions.getCurrentRegion() != null) ? Regions.getCurrentRegion() : Region.getRegion(Regions.US_WEST_2);
[...]

PR #2336, opened 2021-01-04
Avoid QUIT for broken connections

### LINKED ISSUE #2108: Pooled Connections shouldn't be closed if they have socket timeouts
The following line can throw a SocketTimeoutException and in that case, the connection is marked as broken.
https://github.com/xetorthio/jedis/blob/6126d30efccb8921dcf9a03776299a504e022a3e/src/main/java/redis/clients/jedis/Connection.java#L126

Then when the close() method is called (the intention is to add the connection back to the pool), then the broken client is eventually closed:
https://github.com/xetorthio/jedis/blob/6126d30efccb8921dcf9a03776299a504e022a3e/src/main/java/redis/clients/jedis/Jedis.java#L3458

The above call will end up in the invalidateObject() method in the commons-pool2:
https://github.com/apache/commons-pool/blob/f4455dcb8afaf9ae7054589110f1082a7a8a282c/src/main/java/org/apache/commons/pool2/impl/GenericObjectPool.java#L607

which eventually will call:
https://github.com/xetorthio/jedis/blob/6126d30efccb8921dcf9a03776299a504e022a3e/src/main/java/redis/clients/jedis/JedisFactory.java#L94

This is the cause of issue https://github.com/xetorthio/jedis/issues/2105, where the guys see a large number of new connections. I was able to reproduce it by lowering the socket timeout.

sazzad16:
Note: Found it hard to write proper test codes. We may need to mock for this purpose.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -351,7 +351,7 @@ stop:
 	kill `cat /tmp/redis_cluster_node5.pid` || true
 	kill `cat /tmp/redis_uds.pid` || true
 	kill `cat /tmp/stunnel.pid` || true
-	# kill `cat /tmp/redis_unavailable.pid` || true
+	[ -f /tmp/redis_unavailable.pid ] && kill `cat /tmp/redis_unavailable.pid` || true
 	rm -f /tmp/sentinel1.conf
 	rm -f /tmp/sentinel2.conf
 	rm -f /tmp/sentinel3.conf
--- a/Makefile
+++ b/Makefile
@@ -236,6 +236,17 @@ save ""
 appendonly no
 endef
 
+# UNAVAILABLE REDIS NODES
+define REDIS_UNAVAILABLE_CONF
+daemonize yes
+protected-mode no
+port 6400
+pidfile /tmp/redis_unavailable.pid
+logfile /tmp/redis_unavailable.log
+save ""
+appendonly no
+endef
+
 #STUNNEL
 define STUNNEL_CONF
 cert = src/test/resources/private.pem
@@ -278,6 +289,7 @@ export REDIS_CLUSTER_NODE3_CONF
 export REDIS_CLUSTER_NODE4_CONF
 export REDIS_CLUSTER_NODE5_CONF
 export REDIS_UDS
+export REDIS_UNAVAILABLE_CONF
 export STUNNEL_CONF
 export STUNNEL_BIN
 
@@ -309,6 +321,7 @@ start: stunnel cleanup
 	echo "$$REDIS_CLUSTER_NODE4_CONF" | redis-server -
 	echo "$$REDIS_CLUSTER_NODE5_CONF" | redis-server -
 	echo "$$REDIS_UDS" | redis-server -
+	echo "$$REDIS_UNAVAILABLE_CONF" | redis-server -
 
 cleanup:
 	- rm -vf /tmp/redis_cluster_node*.conf 2>/dev/null
@@ -338,6 +351,7 @@ stop:
 	kill `cat /tmp/redis_cluster_node5.pid` || true
 	kill `cat /tmp/redis_uds.pid` || true
 	kill `cat /tmp/stunnel.pid` || true
+	# kill `cat /tmp/redis_unavailable.pid` || true
 	rm -f /tmp/sentinel1.conf
 	rm -f /tmp/sentinel2.conf
 	rm -f /tmp/sentinel3.conf
--- a/src/main/java/redis/clients/jedis/BinaryShardedJedis.java
+++ b/src/main/java/redis/clients/jedis/BinaryShardedJedis.java
@@ -4,6 +4,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.regex.Pattern;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import redis.clients.jedis.commands.BinaryJedisCommands;
 import redis.clients.jedis.commands.ProtocolCommand;
@@ -18,6 +20,8 @@ import redis.clients.jedis.util.Sharded;
 
 public class BinaryShardedJedis extends Sharded<Jedis, JedisShardInfo> implements
     BinaryJedisCommands {
+
+  private final Logger logger = LoggerFactory.getLogger(getClass());
 
   private final byte[][] dummyArray = new byte[0][];
 
@@ -41,14 +45,19 @@ public class BinaryShardedJedis extends Sharded<Jedis, JedisShardInfo> implement
     for (Jedis jedis : getAllShards()) {
       if (jedis.isConnected()) {
         try {
-          jedis.quit();
+          // need a proper test, probably with mock
+          if (!jedis.isBroken()) {
+            jedis.quit();
+          }
         } catch (JedisConnectionException e) {
           // ignore the exception node, so that all other normal nodes can release all connections.
+          logger.error("Error while QUIT", e);
         }
         try {
           jedis.disconnect();
         } catch (JedisConnectionException e) {
           // ignore the exception node, so that all other normal nodes can release all connections.
+          logger.error("Error while disconnect", e);
         }
       }
     }
--- a/src/main/java/redis/clients/jedis/BinaryShardedJedis.java
+++ b/src/main/java/redis/clients/jedis/BinaryShardedJedis.java
@@ -21,7 +21,7 @@ import redis.clients.jedis.util.Sharded;
 public class BinaryShardedJedis extends Sharded<Jedis, JedisShardInfo> implements
     BinaryJedisCommands {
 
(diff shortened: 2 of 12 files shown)


yangbodong22011:
Hi, @sazzad16, the code look good to me, but i don't think it will solve #2108 #2105

Under current jedis netWork protocol, if a operation timeouted, will set `borken = true`, then the socket will be closed. The reason is the socket may be return data after sometime, this may cause next operation get error response.

About this issue, i think the only way to avoid close connection is increase `soTimeout`.

sazzad16:
@yangbodong22011 The object will be returned to the pool as a broken resource. The pool will remove it and never use it again. If necessarily, the pool will create a new resource which will have a new socket. Do you think it will still be a problem?

yangbodong22011:
> @yangbodong22011 The object will be returned to the pool as a broken resource. The pool will remove it and never use it again. If necessarily, the pool will create a new resource which will have a new socket. Do you think it will still be a problem?

This is the current logic, and it is correct. but `create a new resource` will cause https://github.com/redis/jedis/issues/2105#issuecomment-557901749 . 
One way to avoid connection broken is to avoid timeout, that is, increase `soTimeout`.

sazzad16:
> but `create a new resource` will cause https://github.com/redis/jedis/issues/2105#issuecomment-557901749 . 
> One way to avoid connection broken is to avoid timeout, that is, increase `soTimeout`.

The goal of this PR is not to avoid creating new connections and solve #2105. The goal here is to avoid executing a Redis command (QUIT) with a connection which is already broken while trying to execute some other command.

If you are not convinced that this will resolve #2108, we can still keep the issue open and move forward with this PR.

Makefile
```diff
@@ -338,6 +351,7 @@ stop:
 	kill `cat /tmp/redis_cluster_node5.pid` || true
 	kill `cat /tmp/redis_uds.pid` || true
 	kill `cat /tmp/stunnel.pid` || true
+	# kill `cat /tmp/redis_unavailable.pid` || true
```
mina-asham, line 354: Need to uncomment

src/main/java/redis/clients/jedis/BinaryShardedJedis.java
```diff
@@ -19,6 +21,8 @@
 public class BinaryShardedJedis extends Sharded<Jedis, JedisShardInfo> implements
     BinaryJedisCommands {
 
+  private final Logger logger = LoggerFactory.getLogger(getClass());
```
mina-asham, line 24: Loggers should be `static` otherwise they could introduce performance issues. (Applies in a few other places too)

src/main/java/redis/clients/jedis/BinaryShardedJedis.java
```diff
@@ -41,14 +45,19 @@ public void disconnect() {
...
       if (jedis.isConnected()) {
         try {
-          jedis.quit();
+          // need a proper test, probably with mock
+          if (!jedis.isBroken()) {
+            jedis.quit();
+          }
         } catch (JedisConnectionException e) {
           // ignore the exception node, so that all other normal nodes can release all connections.
+          logger.error("Error while QUIT", e);
```
mina-asham, line 54: Can we reduce this to WARN or INFO, lots of people will have alarms on ERRORs, so if we start adding these it might be tempting for people to turn off the logger completely and they might lose some good information here (Applies in a few other places too)

src/main/java/redis/clients/jedis/BinaryShardedJedis.java
```diff
@@ -41,14 +45,19 @@ public void disconnect() {
...
         }
         try {
           jedis.disconnect();
```
mina-asham, line 57: Can we actually disconnect in a broken state?

mina-asham (changes requested):

Makefile
sazzad16: NO.
This redis instance is supposed to be shutdown at `UnavailableConnectionTest`.
I've added this line just to trace the instances. I'm not sure whether we should though.

src/main/java/redis/clients/jedis/BinaryShardedJedis.java
sazzad16: Yes, we can. It actually flushes socket stream and closes socket.

Makefile
mina-asham: But what if the test fails that will leave a rogue running instance, I think we should try and kill it here since this happens *after* running all the tests and would pass still if the process is already dead because of `|| true` bit

src/main/java/redis/clients/jedis/BinaryShardedJedis.java
mina-asham: But wouldn't flush fail in that case? (I guess the `finally` in `Connection.disconnet` still makes sure the socket is closed if possible so it might be fine), if your test and manual run works fine then that's okay then.

sazzad16: Jedis is disconnecting broken connections since ages. It is yet to be pin-pointed as an issue :laughing:

src/main/java/redis/clients/jedis/JedisFactory.java
```diff
@@ -129,12 +134,17 @@ public void destroyObject(PooledObject<Jedis> pooledJedis) throws Exception {
...
     if (jedis.isConnected()) {
       try {
-        try {
+        // need a proper test, probably with mock
+        if (!jedis.isBroken()) {
           jedis.quit();
-        } catch (Exception e) {
         }
+      } catch (Exception e) {
+        logger.error("Error while QUIT", e);
```
sazzad16, line 142:
```suggestion
        logger.warn("Error while QUIT", e);
```

src/main/java/redis/clients/jedis/JedisFactory.java
```diff
@@ -129,12 +134,17 @@ public void destroyObject(PooledObject<Jedis> pooledJedis) throws Exception {
...
+      }
+      try {
         jedis.disconnect();
       } catch (Exception e) {
+        logger.error("Error while disconnect", e);
```
sazzad16, line 147:
```suggestion
        logger.warn("Error while disconnect", e);
```

src/main/java/redis/clients/jedis/ShardedJedisPool.java
```diff
@@ -79,14 +85,17 @@ public void destroyObject(PooledObject<ShardedJedis> pooledShardedJedis) throws
...
           try {
-            try {
+            // need a proper test, probably with mock
+            if (!jedis.isBroken()) {
               jedis.quit();
-            } catch (Exception e) {
-
             }
+          } catch (Exception e) {
+            logger.error("Error while QUIT", e);
```
sazzad16, line 93:
```suggestion
            logger.warn("Error while QUIT", e);
```

src/main/java/redis/clients/jedis/ShardedJedisPool.java
```diff
@@ -79,14 +85,17 @@ public void destroyObject(PooledObject<ShardedJedis> pooledShardedJedis) throws
...
+          }
+          try {
             jedis.disconnect();
           } catch (Exception e) {
-
+            logger.error("Error while disconnect", e);
```
sazzad16, line 98:
```suggestion
            logger.warn("Error while disconnect", e);
```

src/main/java/redis/clients/jedis/BinaryShardedJedis.java
mina-asham: I yield to empirical evidence then 😅

src/main/java/redis/clients/jedis/JedisFactory.java
```diff
@@ -19,6 +21,9 @@
  * PoolableObjectFactory custom impl.
  */
 class JedisFactory implements PooledObjectFactory<Jedis> {
+
+  private final Logger logger = LoggerFactory.getLogger(getClass());
```
mina-asham, line 25: Missed one

Makefile
sazzad16: AFAICR, I commented it because it failed.

src/main/java/redis/clients/jedis/JedisFactory.java
sazzad16, line 25:
```suggestion
  private static final Logger logger = LoggerFactory.getLogger(JedisFactory.class);
```

Makefile
mina-asham: You're right, Redis deletes the pid file on exit, how about this to check if the file exists first:
```suggestion
	[ -f /tmp/redis_unavailable.pid ] && kill `cat /tmp/redis_unavailable.pid` || true
```

mina-asham (approved):

PR #2339, opened 2021-01-14
Support Sentinel with TLS

sazzad16:
Original author: @Talon876 (#2139)

Resolves #1987 

Closes #2024 
Closes #2139 
Closes #2232

PR #2342, opened 2021-01-18
Introduce Config pattern

### LINKED ISSUE #1824: if uri contains dbIndex, timeout configuration doesn't work
### Expected behavior
set timeout to 10000 ms

### Actual behavior
timeout is 2000 ms as default

### Steps to reproduce:

1 .URI uri = new URI("redis://13.229.188.59:6379/1");//use fake address to test timeout
    Jedis jedis = new Jedis(uri, 10000);
2. when time passed 2 seconds, throws SocketTimeoutException, That's not my expected.
3. URI uri = new URI("redis://13.229.188.59:6379");//use fake address to test timeout
    Jedis jedis = new Jedis(uri, 10000);
    jedis.flushDB();
4. when time passed 10 seconds, throws SocketTimeoutException, That's expected.

sazzad16:
If this is approved, we can extend it to JedisPool, JedisCluster, everywhere else.

Resolves #1824

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -129,8 +129,8 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.8.1</version>
 				<configuration>
-					<source>1.7</source>
-					<target>1.7</target>
+					<source>1.8</source>
+					<target>1.8</target>
 				</configuration>
 			</plugin>
 			<plugin>
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -179,6 +179,9 @@ public class Connection implements Closeable {
         broken = true;
         throw new JedisConnectionException("Failed connecting to "
             + socketFactory.getDescription(), ex);
+      } catch(Exception ex) {
+        broken = true;
+        throw ex;
       }
     }
   }
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -39,7 +39,7 @@ public class Connection implements Closeable {
   }
 
   public Connection(final String host, final int port) {
-    this(host, port, null);
+    this(host, port, DefaultJedisSocketConfig.DEFAULT_SOCKET_CONFIG);
   }
 
   @Deprecated
@@ -57,7 +57,11 @@ public class Connection implements Closeable {
   }
 
   public Connection(final String host, final int port, final JedisSocketConfig jedisSocketConfig) {
-    this(new DefaultJedisSocketFactory(host, port, jedisSocketConfig));
+    this(new HostAndPort(host, port), jedisSocketConfig);
+  }
+
+  public Connection(final HostAndPort hostAndPort, final JedisSocketConfig jedisSocketConfig) {
+    this(new DefaultJedisSocketFactory(hostAndPort, jedisSocketConfig));
   }
 
   public Connection(final JedisSocketFactory jedisSocketFactory) {
@@ -177,10 +181,14 @@ public class Connection implements Closeable {
         inputStream = new RedisInputStream(socket.getInputStream());
       } catch (IOException ioe) {
         broken = true;
-        throw new JedisConnectionException("Failed connecting to " + socketFactory.getDescription(), ioe);
+        throw new JedisConnectionException("Failed to create input/output stream", ioe);
       } catch (JedisConnectionException jce) {
         broken = true;
         throw jce;
+      } finally {
+        if (broken) {
+          IOUtils.closeQuietly(socket);
+        }
       }
     }
   }
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -175,10 +175,12 @@ public class Connection implements Closeable {
 
         outputStream = new RedisOutputStream(socket.getOutputStream());
         inputStream = new RedisInputStream(socket.getInputStream());
-      } catch (Exception ex) {
+      } catch (IOException ioe) {
         broken = true;
-        throw new JedisConnectionException("Failed connecting to "
-            + socketFactory.getDescription(), ex);
+        throw new JedisConnectionException("Failed connecting to " + socketFactory.getDescription(), ioe);
+      } catch (JedisConnectionException jce) {
+        broken = true;
+        throw jce;
       }
     }
   }
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -23,7 +23,7 @@ public class Connection implements Closeable {
 
   private static final byte[][] EMPTY_ARGS = new byte[0][];
 
-  private JedisSocketFactory jedisSocketFactory;
+  private final JedisSocketFactory socketFactory;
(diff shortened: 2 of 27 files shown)


src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -179,6 +179,9 @@ public void connect() {
         broken = true;
         throw new JedisConnectionException("Failed connecting to "
             + socketFactory.getDescription(), ex);
+      } catch(Exception ex) {
```
mina-asham, line 182: What other exceptions can this throw? If we are just generalizing, should we just roll this in the above?

pom.xml
```diff
@@ -129,8 +129,8 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.8.1</version>
 				<configuration>
-					<source>1.7</source>
-					<target>1.7</target>
+					<source>1.8</source>
+					<target>1.8</target>
```
mina-asham, line 133: This is a breaking change, should we also bump the major version to 4.0.0?

src/main/java/redis/clients/jedis/DefaultJedisSocketConfig.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    public HostnameVerifier getHostnameVerifier() {
+      return hostnameVerifier;
+    }
+
+    public HostAndPortMapper getHostAndPortMapper() {
+      return hostAndPortMapper;
+    }
+  }
```
mina-asham, line 156: Should we introduce lombok here and avoid all this boiler plate code? https://projectlombok.org/

src/main/java/redis/clients/jedis/JedisSocketConfig.java
```diff
@@ -0,0 +1,22 @@
+package redis.clients.jedis;
+
+import javax.net.ssl.HostnameVerifier;
+import javax.net.ssl.SSLParameters;
+import javax.net.ssl.SSLSocketFactory;
+
+public interface JedisSocketConfig {
```
mina-asham, line 7: Is this basically the socket factory without the host/port?

mina-asham (commented):
I like the Config pattern in general, but I feel like this is something that we need to put more thought into, I have always wanted to see a more immutable version of Jedis where you create the client/pool/cluster and that's it, if you want to modify it you create another one that with a builder-config pattern is really awesome I think.

I think we should split fixing #1824 and this config change as well as it's easier to fix that bug.

src/main/java/redis/clients/jedis/JedisSocketConfig.java
sazzad16: Mostly, yes. I want to keep socket factory dedicated for creating socket and this one for holding config parameters.

src/main/java/redis/clients/jedis/DefaultJedisSocketConfig.java
sazzad16: Would it add an extra dependency?
- If yes, then, no we won't use that.
- If no, then, may be you can help us to add it after this PR gets approved.

pom.xml
sazzad16: IMO, no. Many projects if not most of them (i.e. commons-pool) did not bump the major version when they moved to java 8.

src/main/java/redis/clients/jedis/Connection.java
sazzad16: > What other exceptions can this throw?

`JedisConnectionException`.

Actually, using `DefaultJedisSocketFactory`, it is never `IOException`. Because `DefaultJedisSocketFactory.createSocket()` wraps the IOException in a JedisConnectionException.

> should we just roll this in the above?

I tried. But the compiler is forcing to handle `IOException` separately. So if we want to roll this in above, `throws IOException` had to be added with the method which I wanted to avoid.

pom.xml
mina-asham: I mean it's a breaking change by definition, as in JDK 7 will not be supported anymore, which is totally fine by me btw but I don't know if there are consumers depending on that. Bumping to 4.x isn't necessary, it's just the usual versioning scheme but there is not strict rule that we have to follow it.

I am okay with the change anyway, but if we do this we should drop this bit from the CI: https://github.com/redis/jedis/blob/master/.circleci/config.yml#L20-L22

src/main/java/redis/clients/jedis/DefaultJedisSocketConfig.java
mina-asham: Yes, and no. It's an extra dependency, but it's a compile-only dependency so it doesn't propagate to consumers in any mean, it just runs during compilation and does its magic.

Will see if I can add it in a separate PR :)

src/main/java/redis/clients/jedis/Connection.java
mina-asham: > Actually, using DefaultJedisSocketFactory, it is never IOException. Because DefaultJedisSocketFactory.createSocket() wraps the IOException in a JedisConnectionException.

It can throw IOException, the `Socket.connect` can cause that and then we have a catch that just rethrows after attempting to close the socket if it was opened.

> I tried. But the compiler is forcing to handle IOException separately. So if we want to roll this in above, throws IOException had to be added with the method which I wanted to avoid.

I think it should work fine, this works and compile for me:

```java
public void connect() {
  if (!isConnected()) {
    try {
      socket = jedisSocketFactory.createSocket();

      outputStream = new RedisOutputStream(socket.getOutputStream());
      inputStream = new RedisInputStream(socket.getInputStream());
    } catch (Exception ex) {
      broken = true;
      throw new JedisConnectionException("Failed connecting to "
          + jedisSocketFactory.getDescription(), ex);
    }
  }
}
```

You can also replace `Exception` with `IOException | JedisConnectionException` if you want to be explicit about the exact types of exceptions.

src/main/java/redis/clients/jedis/JedisSocketConfig.java
mina-asham: Does HostAndPortMapper make sense though for other types of sockets? (e.g. AFUNIX)

src/main/java/redis/clients/jedis/Connection.java
sazzad16: 1. Sorry, I misunderstood.
2. You're right. I probably had some other code that was affecting this code.

pom.xml
sazzad16: Agreed.

PS: that piece is meaningless because of https://github.com/redis/jedis/blob/master/.circleci/config.yml#L73-L74

sazzad16: That piece is meaningless because of https://github.com/redis/jedis/blob/master/.circleci/config.yml#L73-L74 

However, I agree to remove. Will do it separately.

src/main/java/redis/clients/jedis/JedisSocketConfig.java
sazzad16: YES. I am generalizing it for all sorts of connections.

This is my comment from #2339 

https://github.com/redis/jedis/blob/af6ec7d2c705968f946804e9dec19eb2c4098073/src/main/java/redis/clients/jedis/JedisSentinelPool.java#L342-L344

PR #2346, opened 2021-01-21
Testing with Java 11

sazzad16:


dengliming (approved):

dengliming:
I guess there are still some configuration adjustments that need to be made. (not sure). When CI is executed, one of the Java versions fails.

Issue #2347, opened 2021-01-21
JedisCluster keeps reconnecting to the same cluster node when it has gone down


jensgreen:
### Expected behavior

When a cluster node that redis is writing to goes down, JedisCluster recovers by attempting to connect to another node in the cluster, in a user-agnostic manner. A few reconnection attempts occur, with a backoff strategy.

### Actual behavior

When a cluster node that redis is writing to goes down, JedisCluster immediately retries the connection to the same node several times without backoff. After `maxAttempts`, it crashes with `JedisClusterMaxAttemptsException: No more cluster attempts left.`.

Stack trace:
```
Exception in thread "main" redis.clients.jedis.exceptions.JedisClusterMaxAttemptsException: No more cluster attempts left.
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:86)
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:124)
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:124)
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:124)
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:124)
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:124)
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:124)
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:124)
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:124)
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:124)
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:124)
	at redis.clients.jedis.JedisClusterCommand.run(JedisClusterCommand.java:25)
	at redis.clients.jedis.JedisCluster.incr(JedisCluster.java:443)
	at JedisClusterFailover.main(JedisClusterFailover.java:23)
```

One observation: `runWithRetries` has no backoff, so all retries will occur immediately. Even if connection to another node was attempted, the cluster has no time to recover by electing a new master.

### Steps to reproduce:

[...]

PR #2350, opened 2021-01-22
Add logging for cluster retries logic

walles:
Done as part of researching #2347.


walles:
CI fails, don't understand how it would be related to this PR, help!

walles:
Superseded by #2358, never mind.

PR #2351, opened 2021-01-22
Fix retries logic

walles:
Done as part of researching #2347.

This change adds tests for `JedisClusterCommand.runWithRetries()`, and fixes two logic errors in said method.

It contains no backoff, we thought it best to do this piecemeal and do that in another PR.

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisClusterCommand.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterCommand.java
@@ -119,7 +119,7 @@ public abstract class JedisClusterCommand<T> {
         this.connectionHandler.renewSlotCache();
       }
 
-      return runWithRetries(slot, attempts - 1, tryRandomNode, redirect);
+      return runWithRetries(slot, attempts - 1, true, null);
     } catch (JedisRedirectionException jre) {
       // if MOVED redirection occurred,
       if (jre instanceof JedisMovedDataException) {
--- a/pom.xml
+++ b/pom.xml
@@ -91,6 +91,12 @@
 			<version>2.3.2</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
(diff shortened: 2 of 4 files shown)


src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -119,7 +119,7 @@ private T runWithRetries(final int slot, int attempts, boolean tryRandomNode, Je
         this.connectionHandler.renewSlotCache();
       }
 
-      return runWithRetries(slot, attempts - 1, tryRandomNode, redirect);
+      return runWithRetries(slot, attempts - 1, true, null);
```
sazzad16, line 122: This would heavily affect current applications. Connection exceptions can happen for silliest of the reasons. So we can't backoff right after a connection exception. There should a balance between retry and backoff. For example, a slot based node or a redirected node should be given at least 2/3 consecutive tries before backing off.

In all of the cases, `maxAttempts`, remaining `attempts`, `renewSlotCache()` should be kept in mind.

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/JedisClusterCommand.java
walles: I'll just try to rephrase what you said so I know I understand:
* With this change, we'd start trying random nodes on the first connection exception
* In your opinion, what we should do is:
  1. First, try the exact same operation a few times in a row
  1. If that doesn't help, start trying random nodes

I think this sounds reasonable, just want to ensure I change the right thing!

sazzad16: @walles Yes, you've got it right.

walles:
Let's just revisit this after #2355 has gone in.

PR #1005, opened 2015-05-24
Atty303 fix cluster connection handling

marcosnils:
Adds test cases for #1000. @HeartSaVioR once you approve you can go ahead and merge

PR #2355, opened 2021-01-25
Split JedisClusterCommand into multiple methods

walles:
No behavior changes, just a refactoring.

Changes:
* Replaces recursion with a `for` loop
* Extract redirection handling into its own method
* Extract connection-failed handling into its own method

I did try to keep the order of the new methods the same order as the previous `catch` blocks.

Note that `tryWithRandomNode` is gone, it was never `true` so it and its code didn't survive the refactoring.

I think redirection handling looks a bit strange, but it didn't change from before. If it *is* wrong then that should be fixed in another PR, this one is just a refactoring.

# Reviewer notes
The `git diff` isn't great for this change.

If you want to see what changed it's probably easier to bring the old and the new code up side by side and compare that way instead.

Here's the old version for reference: <https://github.com/redis/jedis/blob/master/src/main/java/redis/clients/jedis/JedisClusterCommand.java>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -129,8 +129,8 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.8.1</version>
 				<configuration>
-					<source>1.7</source>
-					<target>1.7</target>
+					<source>1.8</source>
+					<target>1.8</target>
 				</configuration>
 			</plugin>
 			<plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -129,8 +129,8 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.8.1</version>
 				<configuration>
-					<source>1.8</source>
-					<target>1.8</target>
+					<source>1.7</source>
+					<target>1.7</target>
 				</configuration>
 			</plugin>
 			<plugin>
--- a/src/main/java/redis/clients/jedis/JedisClusterCommand.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterCommand.java
@@ -1,5 +1,6 @@
 package redis.clients.jedis;
 
+import java.util.function.Supplier;
 import redis.clients.jedis.exceptions.JedisAskDataException;
 import redis.clients.jedis.exceptions.JedisClusterMaxAttemptsException;
 import redis.clients.jedis.exceptions.JedisClusterOperationException;
@@ -79,36 +80,27 @@ public abstract class JedisClusterCommand<T> {
     }
   }
 
-  private interface ConnectionGetter {
-    Jedis getConnection();
-  }
-
   private T runWithRetries(final int slot) {
-    ConnectionGetter connectionGetter = new ConnectionGetter() {
-      @Override
-      public Jedis getConnection() {
-        return connectionHandler.getConnectionFromSlot(slot);
-      }
-    };
+    Supplier<Jedis> connectionSupplier = () -> connectionHandler.getConnectionFromSlot(slot);
 
     // If we got one redirection, stick with that and don't try anything else
-    ConnectionGetter redirected = null;
+    Supplier<Jedis> redirectionSupplier = null;
 
     for (int currentAttempt = 0; currentAttempt < this.maxAttempts; currentAttempt++) {
       Jedis connection = null;
       try {
-        if (redirected != null) {
-          connection = redirected.getConnection();
+        if (redirectionSupplier != null) {
+          connection = redirectionSupplier.get();
         } else {
-          connection = connectionGetter.getConnection();
+          connection = connectionSupplier.get();
         }
         return execute(connection);
       } catch (JedisNoReachableClusterNodeException e) {
         throw e;
       } catch (JedisConnectionException e) {
-        connectionGetter = handleConnectionProblem(slot, currentAttempt);
+        connectionSupplier = handleConnectionProblem(slot, currentAttempt);
       } catch (JedisRedirectionException e) {
-        redirected = handleRedirection(connection, e);
+        redirectionSupplier = handleRedirection(connection, e);
       } finally {
         releaseConnection(connection);
       }
@@ -117,7 +109,7 @@ public abstract class JedisClusterCommand<T> {
     throw new JedisClusterMaxAttemptsException("No more cluster attempts left.");
   }
 
-  private ConnectionGetter handleConnectionProblem(final int slot, int currentAttempt) {
+  private Supplier<Jedis> handleConnectionProblem(final int slot, int currentAttempt) {
     int attemptsLeft = (maxAttempts - currentAttempt) - 1;
     if (attemptsLeft <= 1) {
       //We need this because if node is not reachable anymore - we need to finally initiate slots
@@ -128,15 +120,10 @@ public abstract class JedisClusterCommand<T> {
       this.connectionHandler.renewSlotCache();
     }
 
-    return new ConnectionGetter() {
-      @Override
-      public Jedis getConnection() {
-        return connectionHandler.getConnectionFromSlot(slot);
-      }
-    };
+    return () -> connectionHandler.getConnectionFromSlot(slot);
   }
 
-  private ConnectionGetter handleRedirection(Jedis connection, final JedisRedirectionException jre) {
+  private Supplier<Jedis> handleRedirection(Jedis connection, final JedisRedirectionException jre) {
     // if MOVED redirection occurred,
     if (jre instanceof JedisMovedDataException) {
       // it rebuilds cluster's slot cache recommended by Redis cluster specification
@@ -146,17 +133,14 @@ public abstract class JedisClusterCommand<T> {
     // release current connection before iteration
     releaseConnection(connection);
 
-    return new ConnectionGetter() {
-      @Override
-      public Jedis getConnection() {
-        Jedis connection = connectionHandler.getConnectionFromNode(jre.getTargetNode());
-        if (jre instanceof JedisAskDataException) {
-          // TODO: Pipeline asking with the original command to make it faster....
-          connection.asking();
-        }
-
-        return connection;
+    return () -> {
+      Jedis redirectedConnection = connectionHandler.getConnectionFromNode(jre.getTargetNode());
+      if (jre instanceof JedisAskDataException) {
+        // TODO: Pipeline asking with the original command to make it faster....
+        redirectedConnection.asking();
       }
+
+      return redirectedConnection;
     };
   }
 
--- a/src/main/java/redis/clients/jedis/JedisClusterCommand.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterCommand.java
@@ -110,9 +110,7 @@ public abstract class JedisClusterCommand<T> {
       } catch (JedisRedirectionException e) {
         redirected = handleRedirection(connection, e);
       } finally {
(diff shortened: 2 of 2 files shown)


src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -79,62 +79,87 @@ public T runWithAnyNode() {
...
 
-    Jedis connection = null;
-    try {
+  private T runWithRetries(final int slot) {
+    ConnectionGetter connectionGetter = new ConnectionGetter() {
+      @Override
+      public Jedis getConnection() {
+        return connectionHandler.getConnectionFromSlot(slot);
+      }
+    };
```
sazzad16, line 92: Repeated code (i).

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -79,62 +79,87 @@ public T runWithAnyNode() {
...
-      // if MOVED redirection occurred,
-      if (jre instanceof JedisMovedDataException) {
-        // it rebuilds cluster's slot cache recommended by Redis cluster specification
-        this.connectionHandler.renewSlotCache(connection);
+    return new ConnectionGetter() {
+      @Override
+      public Jedis getConnection() {
+        return connectionHandler.getConnectionFromSlot(slot);
       }
+    };
```
sazzad16, line 138: Repeated code (ii).

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -79,62 +79,87 @@ public T runWithAnyNode() {
...
+      } catch (JedisNoReachableClusterNodeException e) {
+        throw e;
+      } catch (JedisConnectionException e) {
+        connectionGetter = handleConnectionProblem(slot, currentAttempt);
+      } catch (JedisRedirectionException e) {
+        redirected = handleRedirection(connection, e);
+      } finally {
+        if (connection != null) {
+          releaseConnection(connection);
         }
```
sazzad16, line 115: Skip null checking because it is done in `releaseConnection` anyway.

sazzad16 (changes requested):
This is more of a code suggestion. We'll almost always need a normal connection getter. Why not have one `final` defaultConnectionGetter created within constructor and use it everywhere else?

src/main/java/redis/clients/jedis/JedisClusterCommand.java
walles: The `ConnectionGetter` uses the `slot`, so I don't think there can be a default one.

With Java 8 lambdas this code would be shorter at least, but currently the source code is Java 7 so that's a no go.

src/main/java/redis/clients/jedis/JedisClusterCommand.java
walles: Done.

src/main/java/redis/clients/jedis/JedisClusterCommand.java
walles: Also, a `Thread.sleep()` is likely to go into the other one soon, so the code won't stay duplicated for long.

src/main/java/redis/clients/jedis/JedisClusterCommand.java
walles: See above, `slot` can be different between these two so they just *look* identical.

src/main/java/redis/clients/jedis/JedisClusterCommand.java
sazzad16: > The ConnectionGetter uses the slot, so I don't think there can be a default one.

Can't you take a slot parameter?

> a Thread.sleep() is likely to go into the other one

In that case, you should make it configurable, meaning user can choose to use sleep or not. It should be `not` by default as this is the current behavior.

jensgreen: We applied the changes from #2357 use lambdas instead of ConnectionGetters now.

I don't think `() -> connectionHandler.getConnectionFromSlot(slot)` needs to be de-duplicated, what do you think @sazzad16?

sazzad16: We'll look into it later. Don't worry about it for now.

walles:
AFAWCT we have no outstanding tasks here and are awaiting review.

So for clarity, do you think this is ready for review / merge @sazzad16, or are you expecting any more changes from our side?

sazzad16 (dismissed):

sazzad16:
@walles According to community rule, we need approval of at least two collaborators. Let me see what I can do.
In the mean time, you can wait or continue to work on another PR based on this PR.

walles:
@sazzad16 is there anything we can do to get this reviewed?

Given that me and @jensgreen pair programmed this and you reviewed it, we are three people who approve of these changes so far.

sazzad16:
@walles I don't know anything else but asking the reviewers (which I've done). Please understand that they are doing it voluntarily, using their free time.

pom.xml
```diff
@@ -129,8 +129,8 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.8.1</version>
 				<configuration>
-					<source>1.7</source>
-					<target>1.7</target>
+					<source>1.8</source>
```
yangbodong22011, line 132: I'm not sure if updating it to 1.8 will cause compatibility problems for some users of 1.7. I think we decided to update it to 1.8, which should be done when a major version is released.

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -79,62 +80,68 @@ public T runWithAnyNode() {
...
-      releaseConnection(connection);
+  private Supplier<Jedis> handleRedirection(Jedis connection, final JedisRedirectionException jre) {
+    // if MOVED redirection occurred,
+    if (jre instanceof JedisMovedDataException) {
+      // it rebuilds cluster's slot cache recommended by Redis cluster specification
+      this.connectionHandler.renewSlotCache(connection);
     }
+
+    // release current connection before iteration
+    releaseConnection(connection);
```
yangbodong22011, line 134:
```suggestion
```

yangbodong22011: In the previous code, release connection can solve a connection leak, see:https://github.com/redis/jedis/pull/1005 
but i don't think we need it, because we do not have recursion, and `releaseConnection ` in `finally`.

walles:
@yangbodong22011 / @sazzad16, with @yangbodong22011's suggested change applied (good catch Yang!), do you think this is ready for merge?

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -79,62 +80,66 @@ public T runWithAnyNode() {
...
-
-      if (attempts <= 1) {
-        //We need this because if node is not reachable anymore - we need to finally initiate slots
-        //renewing, or we can stuck with cluster state without one node in opposite case.
-        //But now if maxAttempts = [1 or 2] we will do it too often.
-        //TODO make tracking of successful/unsuccessful operations for node - do renewing only
-        //if there were no successful responses from this node last few seconds
-        this.connectionHandler.renewSlotCache();
-      }
+    throw new JedisClusterMaxAttemptsException("No more cluster attempts left.");
```
mina-asham, line 109: Minor enhancement: can we track the last exception and add it as the cause exception here?

```suggestion
    throw new JedisClusterMaxAttemptsException("No more cluster attempts left.", lastException);
```

mina-asham (dismissed):
Looks good, just suggesting a minor enhancement there but not a blocker for me.

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -79,62 +80,66 @@ public T runWithAnyNode() {
...
-
-      if (attempts <= 1) {
-        //We need this because if node is not reachable anymore - we need to finally initiate slots
-        //renewing, or we can stuck with cluster state without one node in opposite case.
-        //But now if maxAttempts = [1 or 2] we will do it too often.
-        //TODO make tracking of successful/unsuccessful operations for node - do renewing only
-        //if there were no successful responses from this node last few seconds
-        this.connectionHandler.renewSlotCache();
-      }
+    throw new JedisClusterMaxAttemptsException("No more cluster attempts left.", lastException);
```
mina-asham, line 109: Sorry I should have provided a better suggestion here, I meant keeping track of `lastException` as well, this variable doesn't exist at the moment, so we can initialize it at the top and just override in `catch` clauses.

mina-asham (changes requested):

src/main/java/redis/clients/jedis/JedisClusterCommand.java
walles: Fixed!

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -98,15 +99,17 @@ private T runWithRetries(final int slot) {
...
       } catch (JedisRedirectionException e) {
+        lastException = e;
         redirectionSupplier = handleRedirection(connection, e);
       } finally {
         releaseConnection(connection);
       }
     }
 
-    throw new JedisClusterMaxAttemptsException("No more cluster attempts left.");
+    throw new JedisClusterMaxAttemptsException("No more cluster attempts left.", lastException);
```
sazzad16, line 112: Just adding another idea. What if we add the `lastException` as suppressed, not as cause?
@mina-asham WDYT?

mina-asham: Ah, that's an even better idea as it allows us to have the full list of exceptions, let's do that!

sazzad16: NO. Let's start with one and see if uses want more.

Check resolves and closes of #2233. You'll get an idea.

mina-asham: It doesn't cost much to add all of them and it's more informative (but I might be biases by my own opinion as a user here), in anyway I don't mind either approach for now, it's a two way door in either case :)

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -79,62 +80,69 @@ public T runWithAnyNode() {
...
-
-      if (attempts <= 1) {
-        //We need this because if node is not reachable anymore - we need to finally initiate slots
-        //renewing, or we can stuck with cluster state without one node in opposite case.
-        //But now if maxAttempts = [1 or 2] we will do it too often.
-        //TODO make tracking of successful/unsuccessful operations for node - do renewing only
-        //if there were no successful responses from this node last few seconds
-        this.connectionHandler.renewSlotCache();
-      }
+    throw new JedisClusterMaxAttemptsException("No more cluster attempts left.", lastException);
```
sazzad16, line 112:
```suggestion
    JedisClusterMaxAttemptsException maxAttemptsException =  new JedisClusterMaxAttemptsException("No more cluster attempts left.");
    maxAttemptsException.addSuppressed(lastException);
    throw maxAttemptsException;
```

PR #2356, opened 2021-01-25
Document JedisClusterCommand.releaseConnection() as Nullable

walles:
This helps both people (me!) and machines (IntelliJ) know how to call this method.

Also, for IntelliJ users, if you find the `Constant conditions & exceptions` inspection and enable `Suggest @Nullable annotation ... where nullable values are used` you will be informed about potential runtime `NullPointerException`s.

A lot of things could probably be annotated, but I wanted to start with *something* to encourage future PRs to add nullability annotations where appropriate.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,6 +65,11 @@
 			<type>jar</type>
 			<scope>compile</scope>
 		</dependency>
+		<dependency>
+			<groupId>com.google.code.findbugs</groupId>
+			<artifactId>jsr305</artifactId>
+			<version>3.0.2</version>
+		</dependency>
 
 		<dependency>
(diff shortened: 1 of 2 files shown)


walles:
Would have helped me in #2355.

sazzad16:
@walles Include the change in your actual PR. But it is worth mentioning that it is unlikely to pass. We are trying to keep Jedis with least number of dependencies.

walles:
I would like this and #2355 to be reviewed independently and merged (or not) independently.

While having had this PR merged first would have saved me some time in #2355, they are functionally independent.

sazzad16:
In that case, consider it rejected because of the new dependency.

PR #2358, opened 2021-01-29
Retry with backoff on cluster connection failures

walles:
Before this change, if there were connection failures to the cluster, we did all our retries without any backoff.

With this change in place:
* We first do the previous no-backoff tactic for one third of our `maxAttempts` (see the `shouldBackOff()` method)
* Then we start backing off as determined by the `getBackoffSleepMillis()` method

Additionally, this change adds unit tests for the retries / backoff logic.

This change is based on the changes in #2355 (approved, not yet merged, currently waiting for more reviewers).

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryJedisCluster.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedisCluster.java
@@ -15,6 +15,7 @@ import redis.clients.jedis.util.KeyMergeUtil;
 import redis.clients.jedis.util.SafeEncoder;
 
 import java.io.Closeable;
+import java.time.Duration;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -32,6 +33,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
   protected static final int DEFAULT_MAX_ATTEMPTS = 5;
 
   protected int maxAttempts;
+  protected final Duration timeout;
 
   protected JedisClusterConnectionHandler connectionHandler;
 
@@ -65,6 +67,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
     this.connectionHandler = new JedisSlotBasedConnectionHandler(jedisClusterNode, poolConfig,
         connectionTimeout, soTimeout, user, password, clientName);
     this.maxAttempts = maxAttempts;
+    this.timeout = Duration.ofMillis(Math.max(connectionTimeout, soTimeout));
   }
 
   public BinaryJedisCluster(Set<HostAndPort> jedisClusterNode, int connectionTimeout, int soTimeout,
@@ -72,6 +75,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
     this.connectionHandler = new JedisSlotBasedConnectionHandler(jedisClusterNode, poolConfig,
         connectionTimeout, soTimeout, infiniteSoTimeout, user, password, clientName);
     this.maxAttempts = maxAttempts;
+    this.timeout = Duration.ofMillis(Math.max(connectionTimeout, soTimeout));
   }
 
   public BinaryJedisCluster(Set<HostAndPort> jedisClusterNode, int connectionTimeout, int soTimeout, int maxAttempts, String password, String clientName, GenericObjectPoolConfig poolConfig,
@@ -97,6 +101,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
     this.connectionHandler = new JedisSlotBasedConnectionHandler(jedisClusterNode, poolConfig,
         connectionTimeout, soTimeout, user, password, clientName, ssl, sslSocketFactory, sslParameters, hostnameVerifier, hostAndPortMap);
     this.maxAttempts = maxAttempts;
+    this.timeout = Duration.ofMillis(Math.max(connectionTimeout, soTimeout));
   }
 
   public BinaryJedisCluster(Set<HostAndPort> jedisClusterNode, int connectionTimeout, int soTimeout,
@@ -105,6 +110,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
     this.connectionHandler = new JedisSlotBasedConnectionHandler(jedisClusterNode, poolConfig,
         connectionTimeout, soTimeout, infiniteSoTimeout, user, password, clientName, ssl, sslSocketFactory, sslParameters, hostnameVerifier, hostAndPortMap);
     this.maxAttempts = maxAttempts;
+    this.timeout = Duration.ofMillis(Math.max(connectionTimeout, soTimeout));
   }
 
   @Override
@@ -124,7 +130,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
 
   @Override
   public String set(final byte[] key, final byte[] value) {
-    return new JedisClusterCommand<String>(connectionHandler, maxAttempts) {
+    return new JedisClusterCommand<String>(connectionHandler, maxAttempts, this.timeout) {
       @Override
       public String execute(Jedis connection) {
         return connection.set(key, value);
@@ -134,7 +140,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
 
   @Override
   public String set(final byte[] key, final byte[] value, final SetParams params) {
-    return new JedisClusterCommand<String>(connectionHandler, maxAttempts) {
+    return new JedisClusterCommand<String>(connectionHandler, maxAttempts, this.timeout) {
       @Override
       public String execute(Jedis connection) {
         return connection.set(key, value, params);
@@ -144,7 +150,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
 
   @Override
   public byte[] get(final byte[] key) {
-    return new JedisClusterCommand<byte[]>(connectionHandler, maxAttempts) {
+    return new JedisClusterCommand<byte[]>(connectionHandler, maxAttempts, this.timeout) {
       @Override
       public byte[] execute(Jedis connection) {
         return connection.get(key);
@@ -154,7 +160,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
 
   @Override
   public Long exists(final byte[]... keys) {
-    return new JedisClusterCommand<Long>(connectionHandler, maxAttempts) {
+    return new JedisClusterCommand<Long>(connectionHandler, maxAttempts, this.timeout) {
       @Override
       public Long execute(Jedis connection) {
         return connection.exists(keys);
@@ -164,7 +170,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
 
   @Override
   public Boolean exists(final byte[] key) {
-    return new JedisClusterCommand<Boolean>(connectionHandler, maxAttempts) {
+    return new JedisClusterCommand<Boolean>(connectionHandler, maxAttempts, this.timeout) {
       @Override
       public Boolean execute(Jedis connection) {
         return connection.exists(key);
@@ -174,7 +180,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
 
   @Override
   public Long persist(final byte[] key) {
-    return new JedisClusterCommand<Long>(connectionHandler, maxAttempts) {
+    return new JedisClusterCommand<Long>(connectionHandler, maxAttempts, this.timeout) {
       @Override
       public Long execute(Jedis connection) {
         return connection.persist(key);
@@ -184,7 +190,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
 
   @Override
   public String type(final byte[] key) {
-    return new JedisClusterCommand<String>(connectionHandler, maxAttempts) {
+    return new JedisClusterCommand<String>(connectionHandler, maxAttempts, this.timeout) {
       @Override
       public String execute(Jedis connection) {
         return connection.type(key);
@@ -194,7 +200,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
 
   @Override
   public byte[] dump(final byte[] key) {
-    return new JedisClusterCommand<byte[]>(connectionHandler, maxAttempts) {
+    return new JedisClusterCommand<byte[]>(connectionHandler, maxAttempts, this.timeout) {
       @Override
       public byte[] execute(Jedis connection) {
         return connection.dump(key);
@@ -204,7 +210,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
 
   @Override
   public String restore(final byte[] key, final int ttl, final byte[] serializedValue) {
-    return new JedisClusterCommand<String>(connectionHandler, maxAttempts) {
+    return new JedisClusterCommand<String>(connectionHandler, maxAttempts, this.timeout) {
       @Override
       public String execute(Jedis connection) {
         return connection.restore(key, ttl, serializedValue);
@@ -214,7 +220,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
 
   @Override
   public Long expire(final byte[] key, final int seconds) {
-    return new JedisClusterCommand<Long>(connectionHandler, maxAttempts) {
+    return new JedisClusterCommand<Long>(connectionHandler, maxAttempts, this.timeout) {
       @Override
       public Long execute(Jedis connection) {
         return connection.expire(key, seconds);
@@ -224,7 +230,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
 
   @Override
   public Long pexpire(final byte[] key, final long milliseconds) {
-    return new JedisClusterCommand<Long>(connectionHandler, maxAttempts) {
+    return new JedisClusterCommand<Long>(connectionHandler, maxAttempts, this.timeout) {
       @Override
       public Long execute(Jedis connection) {
         return connection.pexpire(key, milliseconds);
@@ -234,7 +240,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
 
   @Override
   public Long expireAt(final byte[] key, final long unixTime) {
-    return new JedisClusterCommand<Long>(connectionHandler, maxAttempts) {
+    return new JedisClusterCommand<Long>(connectionHandler, maxAttempts, this.timeout) {
       @Override
       public Long execute(Jedis connection) {
         return connection.expireAt(key, unixTime);
@@ -244,7 +250,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
 
   @Override
   public Long pexpireAt(final byte[] key, final long millisecondsTimestamp) {
-    return new JedisClusterCommand<Long>(connectionHandler, maxAttempts) {
+    return new JedisClusterCommand<Long>(connectionHandler, maxAttempts, this.timeout) {
       @Override
       public Long execute(Jedis connection) {
         return connection.pexpireAt(key, millisecondsTimestamp);
@@ -254,7 +260,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
 
   @Override
   public Long ttl(final byte[] key) {
-    return new JedisClusterCommand<Long>(connectionHandler, maxAttempts) {
+    return new JedisClusterCommand<Long>(connectionHandler, maxAttempts, this.timeout) {
       @Override
       public Long execute(Jedis connection) {
         return connection.ttl(key);
@@ -264,7 +270,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
 
   @Override
   public Long pttl(final byte[] key) {
-    return new JedisClusterCommand<Long>(connectionHandler, maxAttempts) {
+    return new JedisClusterCommand<Long>(connectionHandler, maxAttempts, this.timeout) {
       @Override
       public Long execute(Jedis connection) {
         return connection.pttl(key);
@@ -274,7 +280,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
 
   @Override
   public Long touch(final byte[] key) {
-    return new JedisClusterCommand<Long>(connectionHandler, maxAttempts) {
+    return new JedisClusterCommand<Long>(connectionHandler, maxAttempts, this.timeout) {
       @Override
       public Long execute(Jedis connection) {
         return connection.touch(key);
@@ -284,7 +290,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
 
   @Override
   public Long touch(final byte[]... keys) {
-    return new JedisClusterCommand<Long>(connectionHandler, maxAttempts) {
+    return new JedisClusterCommand<Long>(connectionHandler, maxAttempts, this.timeout) {
       @Override
       public Long execute(Jedis connection) {
         return connection.touch(keys);
@@ -294,7 +300,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
 
   @Override
   public Boolean setbit(final byte[] key, final long offset, final boolean value) {
-    return new JedisClusterCommand<Boolean>(connectionHandler, maxAttempts) {
+    return new JedisClusterCommand<Boolean>(connectionHandler, maxAttempts, this.timeout) {
       @Override
       public Boolean execute(Jedis connection) {
         return connection.setbit(key, offset, value);
@@ -304,7 +310,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
 
   @Override
   public Boolean setbit(final byte[] key, final long offset, final byte[] value) {
-    return new JedisClusterCommand<Boolean>(connectionHandler, maxAttempts) {
+    return new JedisClusterCommand<Boolean>(connectionHandler, maxAttempts, this.timeout) {
       @Override
       public Boolean execute(Jedis connection) {
         return connection.setbit(key, offset, value);
@@ -314,7 +320,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
 
   @Override
   public Boolean getbit(final byte[] key, final long offset) {
-    return new JedisClusterCommand<Boolean>(connectionHandler, maxAttempts) {
+    return new JedisClusterCommand<Boolean>(connectionHandler, maxAttempts, this.timeout) {
       @Override
       public Boolean execute(Jedis connection) {
         return connection.getbit(key, offset);
@@ -324,7 +330,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
 
   @Override
   public Long setrange(final byte[] key, final long offset, final byte[] value) {
-    return new JedisClusterCommand<Long>(connectionHandler, maxAttempts) {
+    return new JedisClusterCommand<Long>(connectionHandler, maxAttempts, this.timeout) {
       @Override
       public Long execute(Jedis connection) {
         return connection.setrange(key, offset, value);
@@ -334,7 +340,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
 
   @Override
   public byte[] getrange(final byte[] key, final long startOffset, final long endOffset) {
-    return new JedisClusterCommand<byte[]>(connectionHandler, maxAttempts) {
+    return new JedisClusterCommand<byte[]>(connectionHandler, maxAttempts, this.timeout) {
       @Override
       public byte[] execute(Jedis connection) {
         return connection.getrange(key, startOffset, endOffset);
@@ -344,7 +350,7 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
 
   @Override
   public byte[] getSet(final byte[] key, final byte[] value) {
-    return new JedisClusterCommand<byte[]>(connectionHandler, maxAttempts) {
+    return new JedisClusterCommand<byte[]>(connectionHandler, maxAttempts, this.timeout) {
       @Override
       public byte[] execute(Jedis connection) {
         return connection.getSet(key, value);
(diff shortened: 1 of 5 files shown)


walles:
✅ 👀  Ready for review!

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -11,18 +17,23 @@
...
 
+  private static final Logger LOG = LoggerFactory.getLogger(JedisClusterCommand.class);
+
   private final JedisClusterConnectionHandler connectionHandler;
   private final int maxAttempts;
+  private final Duration timeout;
 
-  public JedisClusterCommand(JedisClusterConnectionHandler connectionHandler, int maxAttempts) {
+  public JedisClusterCommand(JedisClusterConnectionHandler connectionHandler, int maxAttempts,
+      Duration timeout) {
```
sazzad16, line 27: This breaks backward compatibility.

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
```diff
@@ -105,6 +110,7 @@ public BinaryJedisCluster(Set<HostAndPort> jedisClusterNode, int connectionTimeo
     this.connectionHandler = new JedisSlotBasedConnectionHandler(jedisClusterNode, poolConfig,
         connectionTimeout, soTimeout, infiniteSoTimeout, user, password, clientName, ssl, sslSocketFactory, sslParameters, hostnameVerifier, hostAndPortMap);
     this.maxAttempts = maxAttempts;
+    this.timeout = Duration.ofMillis(Math.max(connectionTimeout, soTimeout));
```
sazzad16, line 113: 1. We already have three different timeouts. Use a better name than `timeout` so that it is easily understandable.
2. There should be an option to set this timeout. It's fine to add another parameter in constructor (in a new constructor of course).
3. `connectionTimeout` should not be a part of this. `soTimeout` can be used as default.

sazzad16 (changes requested):
This PR breaks backward compatibility. Breaking backward compatibility means it won't be released until next major release. As of this moment, next major release for Jedis is 4.0.0 which, you can imagine, is a long away.

Try to find a backward compatible solution. Don't make the code too ugly for that purpose though :)

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
walles: Done!

src/main/java/redis/clients/jedis/JedisClusterCommand.java
walles: Done!

walles:
Thank you for your short turnaround time in reviewing, I really appreciate that @sazzad16!

walles:
> Try to find a backward compatible solution. Don't make the code too ugly for that purpose though :)

Another constructor is needed either way (I think).

But if #2364 would get merged before this PR, that constructor could be made `private` and wouldn't have to clutter the public API.

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
```diff
@@ -28,11 +29,22 @@
     MultiKeyBinaryJedisClusterCommands, JedisClusterBinaryScriptingCommands, Closeable {
 
   public static final int HASHSLOTS = 16384;
-  protected static final int DEFAULT_TIMEOUT = 2000;
-  protected static final int DEFAULT_MAX_ATTEMPTS = 5;
+
+  /**
+   * Default timeout in milliseconds.
+   */
+  public static final int DEFAULT_TIMEOUT = 2000;
```
walles, line 36: `public` makes these reachable from `JedisClusterCommand.java` for its default timeout.

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -79,62 +100,131 @@ public T runWithAnyNode() {
     }
   }
 
-  private T runWithRetries(final int slot, int attempts, boolean tryRandomNode, JedisRedirectionException redirect) {
-    if (attempts <= 0) {
-      throw new JedisClusterMaxAttemptsException("No more cluster attempts left.");
+  private boolean shouldBackOff(int attemptsLeft) {
+    int attemptsDone = maxAttempts - attemptsLeft;
+    return attemptsDone >= maxAttempts / 3;
+  }
```
sazzad16, line 106: I don't agree with the math here. It'd be a long comment to explain. Rather, let me find some time and put my spin over this PR.

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/JedisClusterCommand.java
walles: I invited you as a collaborator to https://github.com/walles/jedis so that you can push to the PR branch.

sazzad16 (dismissed):

sazzad16:
@jensgreen Are you satisfied at the current state of this PR?

jensgreen:
@sazzad16 I will take a look

jensgreen (commented):
Some nitpicks, and I don't understand the `this.maxAttempts < 3` check.

I think it's sound overall!

src/test/java/redis/clients/jedis/tests/SSLJedisClusterWithCompleteCredentialsTest.java
```diff
@@ -4,6 +4,7 @@
 import redis.clients.jedis.*;
 import redis.clients.jedis.exceptions.JedisClusterMaxAttemptsException;
 import redis.clients.jedis.exceptions.JedisConnectionException;
+import redis.clients.jedis.exceptions.JedisNoReachableClusterNodeException;
```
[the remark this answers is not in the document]
walles:
```suggestion
```

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -79,61 +101,97 @@ public T runWithAnyNode() {
...
-        //TODO make tracking of successful/unsuccessful operations for node - do renewing only
-        //if there were no successful responses from this node last few seconds
+  private boolean handleConnectionProblem(int attemptsLeft, int consecutiveConnectionFailures, Instant doneDeadline) {
+    if (this.maxAttempts < 3) {
+      // keep the old implementations as is?
+      if (attemptsLeft == 0) {
         this.connectionHandler.renewSlotCache();
       }
+      return false;
+    }
```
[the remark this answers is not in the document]
sazzad16: At current implementation, there is always a `renewSlotCache()` if all possible attempts are failed. I've tried to keep that for the smallest `maxAttempts` values as these are likely to miss `renewSlotCache()`.

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -79,61 +101,97 @@ public T runWithAnyNode() {
...
+        // if MOVED redirection occurred,
+        if (jre instanceof JedisMovedDataException) {
+          // it rebuilds cluster's slot cache recommended by Redis cluster specification
+          this.connectionHandler.renewSlotCache(connection);
+        }
+      } finally {
+        releaseConnection(connection);
+      }
+      if (Instant.now().isAfter(deadline)) {
+        throw new JedisClusterMaxAttemptsException("Deadline exceeded"); // TODO: change to JedisClusterException
```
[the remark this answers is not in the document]
sazzad16: throwing _Max Attempts_ exception for exceeding deadline does not sound right. But current implementations are not prepared for the existing one, not the new one.

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -145,18 +145,29 @@ private T runWithRetries(final int slot) {
...
+   * @param doneDeadline
+   * @return true - if some actions are taken
+   * <br /> false - if no actions are taken
+   */
   private boolean handleConnectionProblem(int attemptsLeft, int consecutiveConnectionFailures, Instant doneDeadline) {
     if (this.maxAttempts < 3) {
       // keep the old implementations as is?
       if (attemptsLeft == 0) {
         this.connectionHandler.renewSlotCache();
+        return true;
```
walles, line 170: At this point there are no more attempts left (`attemptsLeft == 0`), so it doesn't matter what we return here.

sazzad16: I know. But I wanted to maintain `true - if some actions are taken`.

src/main/java/redis/clients/jedis/JedisClusterCommand.java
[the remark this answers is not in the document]
walles: Another way of ensuring a `renewSlotCache()` call here would be to just drop the `consecutiveConnectionFailures` logic.

Then this `if` statement could be removed as well.

But to me this is really a maintainer call, do the `consecutiveConnectionFailures` upsides outweigh the extra complexity (large or small) required by it?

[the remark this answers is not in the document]
sazzad16: `consecutiveConnectionFailures` is just a way that I could think of to reduce back-off. I'd welcome any other approach to do that.

But I just can't agree on backing off at every connection exception.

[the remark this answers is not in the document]
walles: Then I'm good with this 👍

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -79,61 +101,109 @@ public T runWithAnyNode() {
...
+   *
+   * @param attemptsLeft
+   * @param consecutiveConnectionFailures
+   * @param doneDeadline
+   * @return true - if some actions are taken
+   * <br /> false - if no actions are taken
+   */
+  private boolean handleConnectionProblem(int attemptsLeft, int consecutiveConnectionFailures, Instant doneDeadline) {
+    if (this.maxAttempts < 3) {
+      // keep the old implementations as is?
```
walles, line 167: As @jensgreen said, this comment might want rephrasing, right now it needs this PR as context.

sazzad16: @walles I won't mind if you rephrase it :)

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -79,61 +101,109 @@ public T runWithAnyNode() {
...
+        // if MOVED redirection occurred,
+        if (jre instanceof JedisMovedDataException) {
+          // it rebuilds cluster's slot cache recommended by Redis cluster specification
+          this.connectionHandler.renewSlotCache(connection);
+        }
+      } finally {
+        releaseConnection(connection);
+      }
+      if (Instant.now().isAfter(deadline)) {
+        // TODO: change to JedisClusterOperationException or a new sub-class of it
```
sazzad16, line 148: @jensgreen throwing _Max Attempts_ exception for exceeding deadline does not sound right. We either have to create new exception or use `JedisClusterOperationException`. But current user implementations may break in either case. So I'm a bit confused about what to do here.

src/main/java/redis/clients/jedis/JedisClusterCommand.java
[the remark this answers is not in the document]
sazzad16: Let's continue here https://github.com/redis/jedis/pull/2358/files#r575791185

src/main/java/redis/clients/jedis/JedisClusterCommand.java
walles:
```suggestion
      // Since we only renew the slots cache after two consecutive connection
      // failures (see consecutiveConnectionFailures above), we need to special
      // case the situation where we max out after two or fewer attempts.
      //
      // Otherwise, on two or fewer max attempts, the slots cache would never be
      // renewed.
```

walles:
@sazzad16 do you think you could e-mail me at `walles at spotify.com`?

If we could have a video call where we decide on what exactly this PR should look like I think that would save us lots of calendar time!

src/main/java/redis/clients/jedis/JedisClusterCommand.java
jensgreen: I see the predicament.

I can see only two ways forward that does not break current usage:
1) throw `JedisClusterMaxAttemptsException`; or
2) use a generous interpretation of `JedisClusterMaxAttemptsException` that is "I tried the maximum number of attempts allowed by the specified constraints", and throw a new exception `JedisClusterDeadlineExceededException extends JedisClusterMaxAttemptsException` that means "I tried the maximum number of attempts possible before the deadline constraint was violated".

Now that said, I think a `JedisClusterDeadlineExceededException extends JedisClusterOperationException` would be cleaner.

Which one is the way to go I think depends mostly on how you as maintainers value backwards compatibility.

sazzad16:
@walles okay

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
```diff
@@ -121,4 +121,9 @@ public int getSoTimeout() {
   public void setSoTimeout(int soTimeout) {
     this.soTimeout = soTimeout;
   }
+
+  @Override
+  public String toString() {
+    return "DefaultJedisSocketFactory{" + host + ":" + +port + "}";
```
mina-asham, line 127: Typo: repeated `+` before port

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
```diff
@@ -28,11 +29,22 @@
...
+  public static final int DEFAULT_MAX_ATTEMPTS = 5;
 
   protected int maxAttempts;
 
+  /**
+   * After this amount of time we will do no more retries and report the operation as failed.
+   *
+   * Defaults to {@link #DEFAULT_TIMEOUT} if unset, or {@code soTimeout} if available.
```
mina-asham, line 44: Where does it default to DEFAULT_TIMEOUT?

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -79,61 +101,114 @@ public T runWithAnyNode() {
...
+    long millisLeft = Duration.between(Instant.now(), deadline).toMillis();
+    if (millisLeft < 0) {
+      // TODO: change to JedisClusterOperationException or a new sub-class of it
+      throw new JedisClusterMaxAttemptsException("Deadline exceeded");
+    }
+
+    return millisLeft / (attemptsLeft * attemptsLeft);
+  }
+
+  protected void sleep(long sleepMillis) {
```
mina-asham, line 207: Should this be `private`?

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -79,61 +101,114 @@ public T runWithAnyNode() {
...
+      return 0;
+    }
+
+    long millisLeft = Duration.between(Instant.now(), deadline).toMillis();
+    if (millisLeft < 0) {
+      // TODO: change to JedisClusterOperationException or a new sub-class of it
+      throw new JedisClusterMaxAttemptsException("Deadline exceeded");
+    }
+
+    return millisLeft / (attemptsLeft * attemptsLeft);
```
mina-asham, line 204: Why divide by the square of attempts? (Just wondering if it is specific or an arbitrary decision?)

mina-asham (commented):
I like the idea of backoff before retries, but I think it needs a bit more though:
* It's not really conventional to have a max allowed "time" and throw an exception if we exceed it
* I don't get the backoff strategy here, the common strategies uses are usually uniform or exponential backoffs

Can we change this to avoid the max time and define some backoff strategies the user can select from?

I imagine it would be something like:

```
new JedisCluster(..., new ExponentionalBackoffStrategy(...));
```

We can even define a simple backoff strategy interface so that users can define whatever they want, and I suggest we keep the default behavior as is (i.e. uniform with 0 wait)

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
walles:
```suggestion
    return "DefaultJedisSocketFactory{" + host + ":" + port + "}";
```

src/main/java/redis/clients/jedis/JedisClusterCommand.java
walles: No, `protected` enables overriding in unit tests.

src/main/java/redis/clients/jedis/JedisClusterCommand.java
walles: Kind of arbitrary. Not squared would work as well. The important thing is that the sleeps get longer and longer the fewer iterations that you have left, and that the last sleep uses up all of the time remaining.

**Example**: Let's say you have max 10s and max 5 attempts.

At time 0.0, 5 attempts left, sleep `(10s/(5*5))` = **0.4s**
At time 0.4, 4 attempts left, sleep `(9.6s/(4*4))` = **0.6s**
At time 1.0, 3 attempts left, sleep `(9s/(3*3))` = **1.0s**
At time 2.0, 2 attempts left, sleep `(8/(2*2))` = **2.0s**
At time 6.0, 1 attempts left, sleep `(4/(1*1))` = **4.0s**

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
walles: It's pretty opaque, but If you follow the `DEFAULT_TIMEOUT`'s way from the `public BinaryJedisCluster(Set<HostAndPort> nodes)` constructor, you can see that it eventually makes its way into the `maxTotalRetriesDuration` instance variable.

sazzad16:
@walles Could please rebase this from #2355? (If not, merge)

sazzad16:
@walles Could you please rebase this from #2355? If not merge?

walles:
> We can even define a simple backoff strategy interface so that users can define whatever they want, and I suggest we keep the default behavior as is (i.e. uniform with 0 wait)

Good idea @mina-asham, we just did exactly that in <https://github.com/redis/jedis/pull/2377>!

sazzad16:
@walles Did you get my email (sent about a week ago)?

walles:
Unable to rebase / merge, but I added the missing code from #2355 in a23d6029e3b0dc93b456a71b3ca8e1741d68896f.

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -124,22 +145,70 @@ private T runWithRetries(final int slot) {
       } finally {
         releaseConnection(connection);
       }
+      if (Instant.now().isAfter(deadline)) {
+        // TODO: change to JedisClusterOperationException or a new sub-class of it
+        throw new JedisClusterOperationException("Retry deadline exceeded");
```
sazzad16, line 150:
```suggestion
        throw new JedisClusterOperationException("Retry deadline exceeded");
```

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -124,22 +145,70 @@ private T runWithRetries(final int slot) {
       } finally {
         releaseConnection(connection);
       }
+      if (Instant.now().isAfter(deadline)) {
+        // TODO: change to JedisClusterOperationException or a new sub-class of it
```
sazzad16, line 149:
```suggestion
```

sazzad16 (dismissed):

sazzad16:
@mina-asham If _backoff strategy interface_ is your only major concern, IMHO, we should go ahead with this PR as it is, after resolving the minor ones of course. We can do the interface in a separate PR. Otherwise this PR would become more complex. Also, merging it would just ease crafting the interface.

mina-asham:
> @mina-asham If _backoff strategy interface_ is your only major concern, IMHO, we should go ahead with this PR as it is, after resolving the minor ones of course. We can do the interface in a separate PR. Otherwise this PR would become more complex. Also, merging it would just ease crafting the interface.

Sorry for the long pause, I agree with that I think it's a separate issue and I was only suggesting it as an improvement not as a blocker :)

mina-asham (approved):

mp911de:
I'm always cautious with retries because the particular implementation is pretty often subject to discussion as it rarely meets users where they are. In doubt, it makes sense to have a `RetryHandler` abstraction that can be populated with the actual logic to run a retry (e.g. resilience4j, Spring Retry, a custom retry).

Other than that, it makes sense to consider a `Delay` or `Backoff` abstraction. In Lettuce we've learned that recoveries from network partitions cause a spike in Redis load because all application servers suddenly can reach the target node. For other arrangements, a simple delay is just enough. Check out https://github.com/lettuce-io/lettuce-core/blob/main/src/main/java/io/lettuce/core/resource/Delay.java for some inspiration.

sazzad16 (dismissed):

ynevet:
@gkorland @yangbodong22011 Can you please approve & merge this PR? Otherwise, we're stuck with a PROD issue that impacts our overall latency due to this bug.
Thank you!

yangbodong22011:
Sorry for my late comment.

- I think `maxTotalRetriesDuration` should be an independent configuration in `JedisClientConfig`, although the default value can be `soTimout`, but in most cases, users may set `maxTotalRetriesDuration> timeout`, but in the current PR, only [this method](https://github.com/redis/jedis/pull/2358/files#diff-229425c6d7b7269c74710dac6f5ca74577b3d1346bfa615db8ab35141e53a5e7R165) can set `maxTotalRetriesDuration`, Even `JedisCluster.java` has no way to set an independent maxTotalRetriesDuration.

- Some JedisCluster commands, such as `copy`, `getDel`, `getEx`, do not have the configuration of `maxTotalRetriesDuration`. Is it because we forgot to add it or I missed something?

```
@Override
  public String getDel(final String key) {
    return new JedisClusterCommand<String>(connectionHandler, maxAttempts) {
    // should be: return new JedisClusterCommand<String>(connectionHandler, maxAttempts, this.maxTotalRetriesDuration) {
      @Override
      public String execute(Jedis connection) {
        return connection.getDel(key);
      }
    }.run(key);
  }
```

sazzad16:
@yangbodong22011 

> maxTotalRetriesDuration should be an independent configuration in JedisClientConfig

I disagree. Firstly, it doesn't suit there. Secondly, when we'd try to improve this (targeting Jedis 4.0.0), this would mess the config interface and/or could be bottlenecked by it.

> users may set maxTotalRetriesDuration> timeout

Exactly. This is one of our ultimately goal. But #2377 and mp911de's comment, makes me think that a good enough solution is likely to be a breaking change and thus be targeting 4.0.0. This PR is at least bringing (somewhat not-customizable) sleep time in 3.x. Considering we don't have any sort of sleep, something should be better than nothing.

> Some JedisCluster commands, such as copy, getDel, getEx, do not have the configuration of maxTotalRetriesDuration

It's just that those commands were implemented & merged after this PR is crafted and simple git merge doesn't add those. We'll always have time to add those.

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -85,7 +100,10 @@ public T runWithAnyNode() {
   }
 
   private T runWithRetries(final int slot) {
+    Instant deadline = Instant.now().plus(maxTotalRetriesDuration);
```
gkorland, line 103: Taking the the time on each successful call seems like a waste and might impact the performance.

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
```diff
@@ -69,6 +81,7 @@ public BinaryJedisCluster(Set<HostAndPort> jedisClusterNode, int connectionTimeo
     this.connectionHandler = new JedisSlotBasedConnectionHandler(jedisClusterNode, poolConfig,
         connectionTimeout, soTimeout, user, password, clientName);
     this.maxAttempts = maxAttempts;
+    this.maxTotalRetriesDuration = Duration.ofMillis(soTimeout);
```
gkorland, line 84: Why are we connecting soTimeout with maxTotalRetriesDuration?
I think we should have a separate argument for the maxTotalRetriesDuration and perhaps for enabling backoff at all to avoid backward issues (at least in 3.6)

sazzad16: @gkorland 

> Why are we connecting soTimeout with maxTotalRetriesDuration?

Because max duration for one single try IS EQUALS TO soTimeout.

yangbodong22011:
> I disagree. Firstly, it doesn't suit there. Secondly, when we'd try to improve this (targeting Jedis 4.0.0), this would mess the config interface and/or could be bottlenecked by it.

@sazzad16 Okay, If we have an improvement plan, then I agree to continue, but I still think the default value of `maxTotalRetriesDuration` should be: `maxAttempts * soTimeout`, not equal to soTimeout.

> It's just that those commands were implemented & merged after this PR is crafted and simple git merge doesn't add those. We'll always have time to add those.

This is the responsibility of this PR, and maxTotalRetriesDuration should be added to the new command before merged.

sazzad16:
@yangbodong22011 

> the default value of maxTotalRetriesDuration should be: maxAttempts * soTimeout

agreed

> maxTotalRetriesDuration should be added to the new command before merged

We can do this after the PR is approved.

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -16,12 +18,25 @@
...
 
+  private static final Duration DEFAULT_MAX_TOTAL_RETRIES_DURATION = Duration.ofMillis(
+      BinaryJedisCluster.DEFAULT_TIMEOUT * BinaryJedisCluster.DEFAULT_MAX_ATTEMPTS);
+
   private final JedisClusterConnectionHandler connectionHandler;
   private final int maxAttempts;
+  private final Duration maxTotalRetriesDuration;
 
   public JedisClusterCommand(JedisClusterConnectionHandler connectionHandler, int maxAttempts) {
+    this(connectionHandler, maxAttempts, DEFAULT_MAX_TOTAL_RETRIES_DURATION);
```
sazzad16, line 29:
```suggestion
    this(connectionHandler, maxAttempts, Duration.ofMillis((long) BinaryJedisCluster.DEFAULT_TIMEOUT * maxAttempts));
```

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -16,12 +18,25 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(JedisClusterCommand.class);
 
+  private static final Duration DEFAULT_MAX_TOTAL_RETRIES_DURATION = Duration.ofMillis(
+      BinaryJedisCluster.DEFAULT_TIMEOUT * BinaryJedisCluster.DEFAULT_MAX_ATTEMPTS);
+
```
sazzad16, line 23:
```suggestion
```

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -1,8 +1,10 @@
 package redis.clients.jedis;
 
+import java.time.Duration;
+import java.time.Instant;
+import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
 import redis.clients.jedis.exceptions.JedisAskDataException;
```
sazzad16, line 8:
```suggestion

import redis.clients.jedis.exceptions.JedisAskDataException;
```

src/main/java/redis/clients/jedis/JedisClusterCommand.java
sazzad16: @gkorland According to https://www.alibabacloud.com/blog/performance-issues-related-to-localdatetime-and-instant-during-serialization-operations_595605 

Throughput of Instant.now+atZone+**format**+DateTimeFormatter.ofPattern is 6816922.578 ops/sec.
Without any formatting, throughput of Instant.now+plus should be much higher. Shouldn't it be enough?

src/main/java/redis/clients/jedis/JedisClusterCommand.java
yangbodong22011: The change here is invalid under the current situation, `maxTotalRetriesDuration` will be modified to `soTimeout` again in `BinaryJedisCluster.java`. Because user code always needs `new JedisCluster(xxx)`.

I think there are three solutions:

1. Modify the `BinaryJedisCluster` constructor
```
this.maxTotalRetriesDuration = Duration.ofMillis(soTimeout * maxAttempts);
```

2. Add a constructor in `JedisCluster.java`, so we can directly specify `maxTotalRetriesDuration`

```
public JedisCluster(Set<HostAndPort> jedisClusterNode, int connectionTimeout, int soTimeout,
      int infiniteSoTimeout, int maxAttempts, String user, String password, String clientName,
      final GenericObjectPoolConfig<Jedis> poolConfig, boolean ssl,
      SSLSocketFactory sslSocketFactory, SSLParameters sslParameters,
      HostnameVerifier hostnameVerifier, JedisClusterHostAndPortMap hostAndPortMap, Duration maxTotalRetriesDuration) {
    super(jedisClusterNode, connectionTimeout, soTimeout, infiniteSoTimeout, maxAttempts, user,
        password, clientName, poolConfig, ssl, sslSocketFactory, sslParameters, hostnameVerifier,
        hostAndPortMap, maxTotalRetriesDuration);
  }
```

3. Set `maxTotalRetriesDuration` as a separate configuration (discussed, not working for this PR)

sazzad16:
@gkorland @yangbodong22011 Please check #2490. Hopefully that PR addresses your concerns.

ggivo:
Hi @nitinware 

Took a brief look at the history shows that this [PR](https://github.com/redis/jedis/pull/2358) is already part of  5.1.0.
**"No more cluster attempts left."** is a pretty generic error. Looking at Jedis code the actual cause is stored as a **suppressed** exception inside **JedisClusterOperationException** [here](https://github.com/ggivo/jedis/blob/464a81362a2567bb0629b9bd9b4a7621e5cea3b8/src/main/java/redis/clients/jedis/executors/ClusterCommandExecutor.java#L139-L139). 

I see you are using spring framework and it wraps the original Jedis exception, probably somewhere down the stack there should be **JedisClusterOperationException** and with an actual error causing the failure inside **suppressed**.

Hope it helps

PR #2360, opened 2021-01-30
Introduce JedisBatchOperationException (backward compatible)

sazzad16:
This is a backward compatible version of #1733.


yangbodong22011:
comments at here : https://github.com/redis/jedis/pull/1733#discussion_r570824204

src/main/java/redis/clients/jedis/exceptions/JedisBatchOperationException.java
```diff
@@ -0,0 +1,16 @@
+package redis.clients.jedis.exceptions;
+
+/**
+ * @deprecated This class is extending {@link JedisDataException} to maintain backward
+ * compatibility. But this will not do so in future. Instead, This would extend
+ * {@link JedisException} directly.
+ */
+@Deprecated
```
mina-asham, line 8: The comment doesn't mean it's deprecated though? What's the recommendation here if it's deprecated, is it that the user should catch JedisException intead?

src/main/java/redis/clients/jedis/util/SafeEncoder.java
```diff
@@ -27,7 +26,7 @@ private SafeEncoder(){
   public static byte[] encode(final String str) {
     try {
       if (str == null) {
-        throw new JedisDataException("value sent to redis cannot be null");
+        throw new JedisException("null value cannot be sent to redis");
```
mina-asham, line 29: I agree with the change, but this bit is **not** backward compatible

sazzad16: Nice catch.

sazzad16: This method throws `JedisException` from catch. Check `throw new JedisException(e);`. Would it still be backward incompatible?

mina-asham: But that's only catching `UnsupportedEncodingException` which could be thrown from `str.getBytes(Protocol.CHARSET)`, so doesn't cover **this** throw

sazzad16: Hmm... Well, from a different perspective, **this** throws exception only when `str == null`, which, IMO, is highly unlikely. Would it be worth adding this in a non-major release rather than waiting for a major release?

mina-asham: I agree, it's fairly low risk case, I think it's fine in a minor release, we should mention it in the changelog explicitly as technically it is a breaking change.

mina-asham (approved):

gkorland:
It feel a bit strange that we introduce a new class JedisBatchOperationException and it's marked as deprecated.

sazzad16:
@gkorland Reference https://github.com/redis/jedis/pull/1733#discussion_r239427291 

I haven't found any other way to let users know and prepare for upcoming _supposed_ breaking change.

PS: I'm not supporting deprecation strongly. It just seems a safer/saner approach.

sazzad16:
@mp911de It'd be a great help if you could participate in this conversation. Your comments are always invaluable.

The history of this PR:
- https://github.com/redis/jedis/issues/1183#issuecomment-166799828 
- https://github.com/redis/jedis/issues/1183#issuecomment-169230449 

First attempt #1733

mp911de:
I'm wondering how useful `JedisBatchOperationException` really is. Depending on whether a failure happened within or outside batching/pipelining doesn't seem to be really significant as long as the underlying mechanism throws appropriately an exception. In Lettuce, we do not perform client-side validation on certain commands (e.g. `DISCARD` without `MULTI`) but rather rely on Redis command responses.

Looking at `Please close pipeline or multi block before calling this method` and friends, these are failed validations caused by method invocations in a state that doesn't permit such a state. IMO these are ideal candidates for reusing what Java provides (e.g. `IllegalStateException`). Furthermore, I think Jedis should (in the case of #1183) fix the issue by reporting a different message so callers receive the correct error message.

sazzad16:
@mp911de It was a helpful comment. Thank you.
Could you please check #2393?

PR #2361, opened 2021-01-30
Deprecate JedisPoolAbstract

sazzad16:


src/main/java/redis/clients/jedis/JedisPool.java
```diff
@@ -333,21 +330,20 @@ public Jedis getResource() {
...
   @Override
-  protected void returnResource(final Jedis resource) {
+  public void returnResource(final Jedis resource) {
     if (resource != null) {
       try {
         resource.resetState();
         returnResourceObject(resource);
       } catch (Exception e) {
         returnBrokenResource(resource);
-        throw new JedisException("Resource is returned to the pool as broken", e);
```
yangbodong22011, line 350: I agree not to throw a `JedisException`.
but we should logger.error for it (normal resource became to broken)

sazzad16: acknowledged

src/main/java/redis/clients/jedis/JedisPool.java
```diff
@@ -333,21 +332,14 @@ public Jedis getResource() {
...
+        log.debug("Resource is returned to the pool as broken", e);
```
yangbodong22011, line 342: I think we should use `log.warn` or `log.error`, `log.debug` is generally not used in a production environment, but we need to record errors.

sazzad16, line 342:
```suggestion
        log.warn("Resource is returned to the pool as broken", e);
```

yangbodong22011 (approved):

gkorland (approved):

PR #2033, opened 2019-07-22
Remove WATCH from Transaction

sazzad16:
Redis doesn't support WATCH within MULTI

Resolves #2030

PR #2362, opened 2021-01-30
Disable WATCH within Transaction

### LINKED ISSUE #2030: ClassCastException when transaction.watch(...) is used
Jedis code expects a String return value from `transaction.watch(..)` call, while redis server doesn't return one.
This leads to "shift" in [Queuable.java](https://github.com/xetorthio/jedis/blob/jedis-3.0.1/src/main/java/redis/clients/jedis/Queable.java#L14)'s "pipelanedResponse", and then to ClassCastException (see stacktrace below), because Response<String> (crated for "watch") is used to parse Long result of "incr" call.

By "expects" I mean [MultiKeyCommandsPipeline.java](https://github.com/xetorthio/jedis/blob/jedis-3.0.1/src/main/java/redis/clients/jedis/commands/MultiKeyCommandsPipeline.java#L58) and it's implementation MultiKeyPipelineBase.java

Code:
```   
        try (var cli = new Jedis("127.0.0.1");
            var transaction = cli.multi()) {
            transaction.watch("bar");
            transaction.incr("foo");
            transaction.exec();
        }
    }
```

### Expected behavior

Executed successfully

### Actual behavior

```
Exception in thread "main" java.lang.ClassCastException: class java.lang.Long cannot be cast to class [B (java.lang.Long and [B are in module java.base of loader 'bootstrap')
	at redis.clients.jedis.BuilderFactory$5.build(BuilderFactory.java:73)
	at redis.clients.jedis.BuilderFactory$5.build(BuilderFactory.java:70)
	at redis.clients.jedis.Response.build(Response.java:61)
	at redis.clients.jedis.Response.get(Response.java:37)
	at redis.clients.jedis.Transaction.exec(Transaction.java:53)
	at test.RedisCCE.main(RedisCCE.java:13)
```

### Steps to reproduce:

1. Start local redis instance 
2. Execute `main` code above

#### Java version: openjdk-12.0.2

sazzad16:
This is a backward compatible **alternate** of #2033.

Though #2033 is already approved by two reviewers and tagged `READY TO MERGE`, I am having a second thought about it.

This seems to me a better approach than #2033.

Resolves #2030 
Closes #2033


yangbodong22011:
@sazzad16 i vote for #2033, because this is the style of Jedis for unsupported commands, let them not exist instead of `UnsupportedOperationException`, for example: `JedisCluster.java`, there are no `select` or `flushAll` commands in.

sazzad16:
@yangbodong22011 Thank you for your vote. Let's say, the score is 1-1 for now.

sazzad16:
@yangbodong22011 FYI

> ... this is the style of Jedis for unsupported commands, let them not exist instead of `UnsupportedOperationException`, ...

The style can change in (near?!) future... #1545 https://github.com/redis/jedis/issues/1545#issuecomment-347687167

gkorland (approved):

PR #2364, opened 2021-02-02
Create a BinaryJedisCluster.Builder

walles:
To combat the constructor inflation.

With Java 8 (#2357) we could have used `Duration` rather than `int`s for the timeouts which would probably be nicer in the long run.


sazzad16:
@walles Currently we're trying to implement config pattern for the entire project (e.g. #2368). So, right now I'd consider only if it is done for all the necessary classes.

walles:
Sounds reasonable.

Issue #2367, opened 2021-02-05
Flaky tests


walles:
### Expected behavior

Running `JedisTest.timeoutConnectionWithURI` over and over should yield the same result every time.

### Actual behavior

I ran it five times from inside of IntelliJ. Attempts 1 and 2 passed, 3 failed, 4 and five passed. 

### Steps to reproduce:

Run `JedisTest.timeoutConnectionWithURI` over and over in your favorite IDE.

#### Jedis version:

0ad0b4f08258af9b6c0bc7ff6d4ca1c332f12f6b

#### Redis version:

`Redis server v=6.0.10 sha=00000000:0 malloc=libc bits=64 build=e1b2cac03875a8ff`

#### Java version:

```
openjdk version "11.0.9" 2020-10-20
OpenJDK Runtime Environment (build 11.0.9+11)
OpenJDK 64-Bit Server VM (build 11.0.9+11, mixed mode)
```

walles:
Same thing with `JedisTest.timeoutConnection()`, but that test I had to run about 15 times before it failed the first time.

walles:
To add some structure I wrote [this script](https://github.com/redis/jedis/files/5944197/mvn-testalot.py.gz) and ran it on aa0158c17f72ca95eecef41173a5d2cb2d8c202f:
```
mvn-testalot.py 50
```

While that is running, you can do...
```
mvn-testalot.py report target/testalot
```
... to get an intermediate result, starts being relevant after the two first complete `mvn test` invocations.

Note that the script just does `mvn test` over and over, so the Redis backend isn't restarted between the runs as in the `Makefile` / CI. This may or may not be skewing the results.

Also, I disabled `JedisSentinelPoolTest.returnResourceDestroysResourceOnException()` after the first few runs because I felt it was too slow.

Anyway, this found the tests listed above and some others. Looking into some of these might make CI more reliable.

# Flaky tests

`.` = pass, `x` = fail, `E` = error

| Result | Name |
|--------|------|
| `x................xx..x........x..............x..` | `redis.clients.jedis.tests.JedisPoolTest.testCloseConnectionOnMakeObject()` |
| `....E.EE.EEE..EEEE.E.EE.E.E.EEEEE.E.E.EEE.EE.EEE` | `redis.clients.jedis.tests.JedisSentinelPoolTest.checkCloseableConnections()` |
[...]

PR #2368, opened 2021-02-06
Introduce and extend Config pattern (III/2)

### LINKED ISSUE #1780: Implement builder pattern of JedisPool and JedisCluster
### Expected behavior

Currently, the `JedisCluster` constructor has at most 7 parameter.
Now
```java
  public JedisCluster(HostAndPort node, int connectionTimeout, int soTimeout,
          int maxAttempts, String password, String clientName, final GenericObjectPoolConfig poolConfig) {
  }
```

After adding support SSL in `JedisCluster` #1550, the number become 12

```java
  public JedisCluster(Set<HostAndPort> jedisClusterNode, int connectionTimeout, int soTimeout,
                      int maxAttempts, String password, String clientName, final GenericObjectPoolConfig poolConfig,
                      boolean ssl, SSLSocketFactory sslSocketFactory, SSLParameters sslParameters, 
                      HostnameVerifier hostnameVerifier, JedisClusterHostAndPortMap hostAndPortMap) {
  }
```

We should implement builder pattern to make it easy to build.

Same as `JedisPool`

### LINKED ISSUE #1824: if uri contains dbIndex, timeout configuration doesn't work

### LINKED ISSUE #2234: Support host and port mapping for cluster connections (not only SSL)
### Expected behavior

When connecting to a cluster which is run inside Docker (e.g. using http://testcontainers.org library, which maps ports dynamically depending on available ones) `JedisCluster` should take a mapping function to properly map discovered nodes host/port to their real ports.

### Actual behavior

Jedis connects to the node(s) using host/port specified in `JedisCluster` constructor, then discovers all nodes and fails when trying to connect to them directly.

### Proposed solution

Use `JedisClusterHostAndPortMap` not only for SSL connections. Is it a way to go?

sazzad16:
Resolves #1780 
Resolves #1824 
Resolves #2234 

Closes #1787 
Closes #2237 

Closes #2342 
Closes #2349 
Closes #2352 
Closes #2353 
Closes #2363


src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
```diff
@@ -1,80 +1,122 @@
...
+  public DefaultJedisSocketFactory(HostAndPort hostAndPort, JedisClientConfig config) {
+    setHostAndPort(hostAndPort);
+    if (config != null) {
+      setConnectionTimeout(config.getConnectionTimeout());
+      setSoTimeout(config.getSoTimeout());
+      setSsl(config.isSsl());
+      setSslSocketFactory(config.getSslSocketFactory());
+      setSslParameters(config.getSslParameters());
+      setHostnameVerifier(config.getHostnameVerifier());
+      setHostAndPortMapper(config.getHostAndPortMapper());
```
mina-asham, line 52: It's a bit unconventional to use setters in constructors, any advantage of doing that?

src/test/java/redis/clients/jedis/tests/JedisClusterTest.java
```diff
@@ -510,24 +566,35 @@ public void testCloseable() throws IOException {
...
+//    Iterator<JedisPool> poolIterator = jc.getClusterNodes().values().iterator();
+//    while (poolIterator.hasNext()) {
+//      JedisPool pool = poolIterator.next();
+//      try {
+//        pool.getResource();
+//        fail("JedisCluster's internal pools should be already destroyed");
+//      } catch (JedisConnectionException e) {
+//        // ok to go...
+//      }
+//    }
```
mina-asham, line 586: I think you forgot to uncomment this?

src/test/java/redis/clients/jedis/tests/JedisTest.java
```diff
@@ -190,14 +179,40 @@ public void allowUrlWithNoDBAndNoPassword() {
...
+    URI uri = new URI("redis://fakehost:6378/1");
+    long startTime = System.nanoTime();
+    try (Jedis j = new Jedis(uri, 5000)) {
+      j.ping();
+    } catch (Exception ex) {
+      assertEquals(JedisConnectionException.class, ex.getClass());
+      assertEquals(java.net.UnknownHostException.class, ex.getCause().getClass());
+    }
+    long stopTime = System.nanoTime();
+    assertTrue(stopTime - startTime > 4000);
```
mina-asham, line 193: I don't get this one, the timeout is 5 seconds, but we are validating the runtime is greater than 4 **milli**seconds?

src/test/java/redis/clients/jedis/tests/ShardedJedisPoolTest.java
```diff
@@ -136,6 +137,7 @@ public void checkFailedJedisServer() {
     pool.destroy();
   }
 
+  @Ignore
```
mina-asham, line 140: Why ignore it? If the test doesn't make sense anymore can we remove it instead?

src/test/java/redis/clients/jedis/tests/ShardedJedisPoolWithCompleteCredentialsTest.java
```diff
@@ -145,6 +146,7 @@ public void checkFailedJedisServer() {
     pool.destroy();
   }
 
+  @Ignore
```
mina-asham, line 149: Ditto

mina-asham (changes requested):
Looks good, thanks for going through the comments from the prevision revision! Just a few more minor comments here and then I am happy to sign off

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
sazzad16: Not exactly. I saw this style is being used in commons-pool2. So I put it here at least to have some discussions :)

src/test/java/redis/clients/jedis/tests/JedisTest.java
sazzad16: Socket connections are dependent on OS. We can't strictly check the socket timeouts. I'm using the threshold to reduce this test's probability to become another [flaky](https://github.com/redis/jedis/issues/2367) test.

src/test/java/redis/clients/jedis/tests/ShardedJedisPoolTest.java
sazzad16: I couldn't make this work with my current changes. What if someone else can? So I kept this.

src/test/java/redis/clients/jedis/tests/ShardedJedisPoolWithCompleteCredentialsTest.java
sazzad16: Ditto

src/test/java/redis/clients/jedis/tests/JedisClusterTest.java
sazzad16: Nope.

I debugged and found that `jc.getClusterNodes().values()` is actually empty, even without my changes. So above codes are meaningless.

Also check that following `assertEquals(0, jc.getClusterNodes().size());` is passing.

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
mina-asham: IMO, still unconventional commons-pool2 doesn't make it okay :P

src/test/java/redis/clients/jedis/tests/JedisClusterTest.java
mina-asham: If that part is useless let's delete it then, commenting it out leave unanswered questions

src/test/java/redis/clients/jedis/tests/JedisTest.java
mina-asham: Got you, so that would make it "less" flaky, but potentially still flaky, do we want to avoid that check completely? We are kind of checking the JDK's Socket timeout works as expected here which is not really adding much, I think just checking the exception is enough in that case

src/test/java/redis/clients/jedis/tests/ShardedJedisPoolTest.java
mina-asham: If it worked before your change then there are two cases:
1. We are testing the wrong thing, in that case let's drop the test
2. We broke something so w need to look into it

I think in either cases we need to look into this one and the other one below

src/test/java/redis/clients/jedis/tests/JedisTest.java
sazzad16: I really want to have test for #1824 as this is a reported issue. And I can't find any other way to test it.

mina-asham: This should do the trick? (a bit more involved, but should be a lot more stable)

```suggestion
    int port = 6677;
    int timeoutMillis = 5000;
    int deltaMillis = 500;

    URI uri = new URI("redis://localhost:" + port + "/1");
    ServerSocket server = new ServerSocket(port);
    Instant start = Instant.now();
    try (Jedis jedis = new Jedis(uri, timeoutMillis)) {
        jedis.ping();
    } catch (JedisConnectionException ex) {
        assertEquals(SocketTimeoutException.class, ex.getCause().getClass());
        assertEquals(timeoutMillis, Duration.between(start, Instant.now()).toMillis(), deltaMillis);
        return;
    } finally {
        server.close();
    }
    fail();
```

sazzad16: I'll give that a try.

src/main/java/redis/clients/jedis/DefaultJedisClientConfig.java
```diff
@@ -0,0 +1,201 @@
...
+
+import javax.net.ssl.HostnameVerifier;
+import javax.net.ssl.SSLParameters;
+import javax.net.ssl.SSLSocketFactory;
+
+public final class DefaultJedisClientConfig implements JedisClientConfig {
+
+  private final int connectionTimeout;
+  private final int soTimeout;
+  private final int infiniteSoTimeout;
```
walles, line 11: With Java 8 (#2357) `Duration` could be used rather than `int`s for the timeouts, which would probably be a better API in the long run.

walles: I assume this goes for all places where `Protocol.DEFAULT_TIMEOUT` is used really, as it is now the unit for the timeouts is not obvious.

sazzad16: @walles Let's merge this PR first and then we'll think about it :)

mina-asham (approved):

src/main/java/redis/clients/jedis/BinaryClient.java
```diff
@@ -59,16 +65,28 @@ public BinaryClient(final String host, final int port) {
...
+  /**
+   * @deprecated This constructor will be removed in future.
+   */
+  @Deprecated
   public BinaryClient(final String host, final int port, final boolean ssl) {
     super(host, port, ssl);
   }
 
+  /**
+   * @deprecated This constructor will be removed in future.
```
gkorland, line 77: @sazzad16 please point to the alternative Constructor

src/main/java/redis/clients/jedis/BinaryClient.java
```diff
@@ -59,16 +65,28 @@ public BinaryClient(final String host, final int port) {
     super(host, port);
   }
 
+  /**
+   * @deprecated This constructor will be removed in future.
```
gkorland, line 69: @sazzad16 please point to the alternative Constructor

src/main/java/redis/clients/jedis/BinaryClient.java
```diff
@@ -81,23 +99,20 @@ public boolean isInWatch() {
...
-
-  private byte[][] joinParameters(byte[] first, byte[] second, byte[][] rest) {
-    byte[][] result = new byte[rest.length + 2][];
-    result[0] = first;
-    result[1] = second;
-    System.arraycopy(rest, 0, result, 2, rest.length);
-    return result;
+  /**
+   * @param user
+   * @deprecated This method will be removed in future.
```
gkorland, line 104: @sazzad16 please point to the alternative Constructor

src/main/java/redis/clients/jedis/BinaryClient.java
```diff
@@ -81,23 +99,20 @@ public boolean isInWatch() {
...
+  @Deprecated
+  public void setUser(final String user) {
+    this.user = user;
   }
 
-  public void setUser(final String user) { this.user = user; }
-
+  /**
+   * @param password
+   * @deprecated This method will be removed in future.
```
gkorland, line 113: @sazzad16 please point to the alternative Constructor

src/test/java/redis/clients/jedis/tests/ShardedJedisPoolTest.java
```diff
@@ -136,49 +137,6 @@ public void checkFailedJedisServer() {
     pool.destroy();
   }
 
-  @Test
-  public void shouldReturnActiveShardsWhenOneGoesOffline() {
```
gkorland, line 140: @sazzad16 why did your remove this test?

src/test/java/redis/clients/jedis/tests/ShardedJedisPoolWithCompleteCredentialsTest.java
```diff
@@ -145,49 +146,6 @@ public void checkFailedJedisServer() {
     pool.destroy();
   }
 
-  @Test
-  public void shouldReturnActiveShardsWhenOneGoesOffline() {
```
gkorland, line 149: @sazzad16 why did your remove this test?

src/test/java/redis/clients/jedis/tests/ConnectionTest.java
```diff
@@ -4,51 +4,60 @@
...
 
-  @Before
-  public void setUp() throws Exception {
-    client = new Connection();
-  }
-
   @After
   public void tearDown() throws Exception {
-    client.close();
+    if (client != null) {
```
gkorland, line 20: @sazzad16 why do you still need this `@After` if the connection is not created in the `@Before` any more?

src/main/java/redis/clients/jedis/JedisSocketFactory.java
```diff
@@ -15,23 +16,29 @@
  */
 public interface JedisSocketFactory {
```
gkorland, line 17: @sazzad16 it seems like all the methods are deprecated so how come the interface isn't?

src/test/java/redis/clients/jedis/tests/ShardedJedisTest.java
```diff
@@ -270,9 +270,9 @@ public void testMasterSlaveShardingConsistency() {
         Hashing.MURMUR_HASH);
 
     List<JedisShardInfo> otherShards = new ArrayList<JedisShardInfo>(3);
-    otherShards.add(new JedisShardInfo("otherhost", Protocol.DEFAULT_PORT));
-    otherShards.add(new JedisShardInfo("otherhost", Protocol.DEFAULT_PORT + 1));
-    otherShards.add(new JedisShardInfo("otherhost", Protocol.DEFAULT_PORT + 2));
+    otherShards.add(new JedisShardInfo("127.0.0.1", Protocol.DEFAULT_PORT));
```
gkorland, line 273: why did you change all the host names to IPs?

src/main/java/redis/clients/jedis/BinaryClient.java
sazzad16: @gkorland There won't be any.

src/main/java/redis/clients/jedis/BinaryClient.java
sazzad16: @gkorland There won't be any direct alternate.

src/main/java/redis/clients/jedis/JedisSocketFactory.java
sazzad16: @gkorland `createSocket()` method is here to stay. But instead of throwing IOException, it will throw only JedisConnectionException.

src/test/java/redis/clients/jedis/tests/ConnectionTest.java
sazzad16: @gkorland This test class mostly consists of constructor tests. So connections are supposed to be created in each test methods. I'm closing those objects here in `@After`.

src/main/java/redis/clients/jedis/BinaryClient.java
sazzad16: @gkorland There won't be any direct alternate. Other than config class and, may be, factory class, no other class will hold user data.

src/main/java/redis/clients/jedis/BinaryClient.java
sazzad16: @gkorland There won't be any direct alternate. Other than config class and, may be, factory class, no other class will hold user data.

sazzad16:
@gkorland 

1. ShardedJedisPoolTest.shouldReturnActiveShardsWhenOneGoesOffline()
2. ShardedJedisPoolWithCompleteCredentialsTest.shouldReturnActiveShardsWhenOneGoesOffline()
3. ShardedJedisTest.testMasterSlaveShardingConsistency()
4. ShardedJedisTest.testMasterSlaveShardingConsistencyWithShardNaming()

These tests used to work because underlying connections were not always created given the combination of parameters. All the Jedis classes were suffering from this situation. The underlying connections of objects representing **a connection** either should always be created from constructors or should NEVER be created. In this PR I implemented the former approach.

I was able to keep the tests with a valid host address in (3) and (4).

But I could not make (1) and (2) to work. Even though I `@Ignore`d first, I removed those after review. A portion of (1) and (2):
```
    shards.set(1, new JedisShardInfo("localhost", 1234));
    pool = new ShardedJedisPool(redisConfig, shards);
    jedis = pool.getResource();
```
This is more logical that `pool.getResource()` would fail because `1234` is an invalid port. But the tests are expecting `pool.getResource()` without any error. That's why these two tests won't work with this PR.

IMHO, we shouldn't care about impractical tests. Still, I can `@Ignore` those if you prefer.

src/test/java/redis/clients/jedis/tests/ShardedJedisPoolTest.java
sazzad16: @gkorland https://github.com/redis/jedis/pull/2368#issuecomment-790262400

src/test/java/redis/clients/jedis/tests/ShardedJedisPoolWithCompleteCredentialsTest.java
sazzad16: @gkorland https://github.com/redis/jedis/pull/2368#issuecomment-790262400

src/test/java/redis/clients/jedis/tests/ShardedJedisTest.java
sazzad16: @gkorland https://github.com/redis/jedis/pull/2368#issuecomment-790262400

src/main/java/redis/clients/jedis/BinaryClient.java
gkorland: So you should state in the JDoc

src/main/java/redis/clients/jedis/BinaryClient.java
gkorland: So you should state it in the JDoc

src/main/java/redis/clients/jedis/JedisSocketFactory.java
gkorland: Yes, since you are not going to really deprecate it

gkorland (approved):

PR #2370, opened 2021-02-07
Added Rawable interface

sazzad16:
Jedis should have a common interface to represent all commands and keywords (and, perhaps, arguments). Closest thing we currently have is `ProtocolCommand` interface. But the problem is that it contains the word COMMAND which would be misleading to use for something other than commands.

One option is to rename the existing interface. But this would be a breaking change and so it can't be available before at least Jedis 4.0. Even that can be received negatively by the users who could be using the existing interface.

Having a parent interface, on the other hand, does not have such issues and can be included in upcoming non-major release.

I went with name `Raw`, instead of other names like `ProtocolKeyword` or `ProtocolArgument`, because it is short and fits nicely with our go to method `getRaw()`.

I have also put it in a separate package because with growing number of classes, we'd be able to put some classes in that package.


mina-asham (commented):
Do you have a future use of this in mind?

sazzad16:
Yeah. Will do it later.

sazzad16:
A recent example is #2424, where `UnblockType` could implement the common interface.

dengliming (commented):
Looks good. Do we have a better naming option than `Raw` for the interface? 😃

sazzad16:
> Do we have a better naming option than `Raw` for the interface?

@dengliming There are **many** other options.

- ProtocolArgument
- ProtocolKeyword
- RawArgument
- and many more

Is any of these **better**? I'm not sure.

sazzad16:
> Do we have a better naming option than `Raw` for the interface?

@dengliming There are **many** other options.

- ProtocolArgument
- ProtocolKeyword
- RawArgument
- and many more

Is any of these **better**? I'm not sure. 😆

dengliming:
Not sure. but I prefer `ProtocolKeyword`. : )

dengliming (dismissed):
`Raw ` and `ProtocolKeyword ` are both good to me. : )

gkorland:
I like `Raw` I think it captures the functionality of it, or if we want to be more "Java" `Rawable`

sazzad16:
@dengliming What are your preferences now? (including Rawable)

dengliming:
> Rawable

+1 : )

dengliming (approved):

PR #2375, opened 2021-02-12
JedisNoReachableClusterNodeException should extend JedisClusterOperationException

sazzad16:


src/main/java/redis/clients/jedis/exceptions/JedisNoReachableClusterNodeException.java
```diff
@@ -1,17 +1,22 @@
 package redis.clients.jedis.exceptions;
 
+/**
+ * @deprecated This will not extend {@link JedisConnectionException} in future. It can extend
```
mina-asham, line 4: `can` -> `will`?

mina-asham (dismissed):

gkorland:
@sazzad16 I don't think deprecation is the right way to go here, it will generate useless warnings to users

sazzad16:
@gkorland It's possible that users are catching JedisConnectionException to actually catch JedisNoReachableClusterNodeException. But it doesn't make sense for JNRCNE to be a sub-class to JCE because there may be zero connections involved in this exception, e.g. when empty `Set` is provided for seed nodes. JedisClusterOperationException is custom-made for this kind of scenario that a cluster exception have occurred which is not a response from Redis (1). If you think that it is okay for JNRCNE to be a sub-class of JCE, then the existence of JCOE doesn't make sense. What can we do now?

(1) We have a custom-made exception as well, JedisClusterException.

An idea: What if JNRCNE is a sub-class of JCOE and JCOE is a sub-class of JCE?

dengliming:
agree with @gkorland 

> @sazzad16 I don't think deprecation is the right way to go here, it will generate useless warnings to users

and 

> An idea: What if JNRCNE is a sub-class of JCOE and JCOE is a sub-class of JCE?

@sazzad16 That seems make sense to me.

sazzad16:
> What if JNRCNE is a sub-class of JCOE and JCOE is a sub-class of JCE?

@dengliming BTW, I suggested that as a temporary solution. What would we do when we want to undo "JCOE is a sub-class of JCE" part for next major release?

PS: Sorry about incomplete idea above.

sazzad16:
Resolved at #2409

PR #2377, opened 2021-02-19
Pluggable Retry Mechanism

walles:
Inspired by @mina-asham in <https://github.com/redis/jedis/pull/2358#pullrequestreview-592136327>:
> We can even define a simple backoff strategy interface so that users can define whatever they want, and I suggest we keep the default behavior as is (i.e. uniform with 0 wait)

This PR does exactly that.

It adds [a new `Retryer` interface](https://github.com/walles/jedis/blob/j/pluggable-retries/src/main/java/redis/clients/jedis/Retryer.java) that people can implement. `Retryer` instances can be passed to two new `BinaryJedisCluster` and `JedisCluster` constructors.

[The existing retry code has been copied into `DefaultRetryer`](https://github.com/walles/jedis/blob/j/pluggable-retries/src/main/java/redis/clients/jedis/DefaultRetryer.java). This retryer is what's used by `BinaryJedisCluster` and `JedisCluster` for any old code.

The previous `JedisClusterCommand` class has been split into:
* `runWithRetries()` went (unchanged) into `DefaultRetryer`
* Some scaffolding `run...()` methods went into [`BinaryJedisCluster` (search for `runBinary`)](https://github.com/walles/jedis/blob/j/pluggable-retries/src/main/java/redis/clients/jedis/BinaryJedisCluster.java), which is the only place from which they were called
* Some other scaffolding `run...()` methods went into [`JedisCluster` (search for `run`)](https://github.com/walles/jedis/blob/j/pluggable-retries/src/main/java/redis/clients/jedis/JedisCluster.java), which is the only place from which they were called

One point with this PR is that it does *not* change any behavior. This will simplify review, since backoffs are hard.

With this change merged, Jedis users can roll their own backoff strategies.

And if they work well, those strategies can be easily turned into PRs.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -129,8 +129,8 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.8.1</version>
 				<configuration>
-					<source>1.7</source>
-					<target>1.7</target>
+					<source>1.8</source>
+					<target>1.8</target>
 				</configuration>
 			</plugin>
 			<plugin>
--- a/src/main/java/redis/clients/jedis/BinaryJedisCluster.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedisCluster.java
@@ -24,6 +24,14 @@ import javax.net.ssl.SSLSocketFactory;
 
 import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
 
+/**
+ * Interface to a Jedis cluster.
+ * <p/>
+ * Uses {@link DefaultRetryer}, or you can inject your own using
+ * {@link BinaryJedisCluster#BinaryJedisCluster(Retryer)}.
+ *
+ * @see JedisCluster
+ */
 public class BinaryJedisCluster implements BinaryJedisClusterCommands,
     MultiKeyBinaryJedisClusterCommands, JedisClusterBinaryScriptingCommands, Closeable {
 
--- a/src/main/java/redis/clients/jedis/BinaryJedisCluster.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedisCluster.java
@@ -105,6 +105,10 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
     retryer = new DefaultRetryer(connectionHandler, maxAttempts);
   }
 
+  public BinaryJedisCluster(Retryer retryer) {
+    this.retryer = retryer;
+  }
+
   @Override
   public void close() {
     retryer.close();
--- a/src/main/java/redis/clients/jedis/BinaryJedisCluster.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedisCluster.java
@@ -111,11 +111,11 @@ public class BinaryJedisCluster implements BinaryJedisClusterCommands,
   }
 
   public Map<String, JedisPool> getClusterNodes() {
-    return connectionHandler.getNodes();
+    return retryer.getClusterNodes();
   }
 
   public Jedis getConnectionFromSlot(int slot) {
-	  return  this.connectionHandler.getConnectionFromSlot(slot);
+    return retryer.getConnectionFromSlot(slot);
   }
 
   @Override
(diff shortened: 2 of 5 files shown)


mina-asham (commented):
I think this is closer in spirit but not what I meant, the retryer is really tightly coupled with the Jedis cluster internals, I meant is as a more generic thing that we can later plugin as we need to, I imagined something like this like this:
```java
class Retryer {
    <T> T runWithRetries(Callable<T> callable, RetryOptions options) {
        // Do retries
    }
    void runWithRetries(Runnable runnable, RetryOptions options) {
        // Do retries, probably wrap the runnable and call the above
    }
}

interface RetryOptions {
    Duration nextSleepAfterFailure(int retryNumber);
    <T extends Exception> boolean shouldRetry(T ex);
}
```

and then people would just plugin their retry options or something to the class, we don't have to make the `Retryer` class and we can just keep the retry logic in the cluster class if we don't have another use case, but I think the `RetryOptions` (or whatever we call it, and not necessarily exactly like above) is what I had in mind.

walles:
Something like this? <https://github.com/resilience4j/resilience4j#61-circuitbreaker-retry-and-fallback>

mina-asham:
> Something like this? https://github.com/resilience4j/resilience4j#61-circuitbreaker-retry-and-fallback

A simplified version yes, one of Jedis' nice properties is that it doesn't have a huge dependecy tree (only commons pool pretty much), so I wouldn't want to introduce new libraries there, but a very simple and light weight retry mechanism would be nice here.

walles:
> I think this is closer in spirit but not what I meant, the retryer is really tightly coupled with the Jedis cluster internals, I meant is as a more generic thing that we can later plugin as we need to

If you look at #2358 there are three retry reasons, not just one:
* `JedisConnectionException`
* `JedisMovedDataException`
* `JedisAskDataException`

They all get different treatment. I think this would be hard to express using a generic retry mechanism.

Also, the logic for when to `sleep()` and not, as well as when to `renewSlotCache()` or not is pretty involved, and could also be tricky to express using a generic retry mechanism.

So I think this PR is as generic as you will get.

But I'm happy to be proven wrong here :)

sazzad16:
@gkorland @dengliming @mina-asham @yangbodong22011 

IMO, this would be a great inclusion to have in JedisCluster. Of course, a generic interface would be even better thing to have. But having nothing due to in search of generic would be a loss. So I request you to review this with a more liberal view.

---

@walles I have some requests.

Keeping in mind of future generic interface:

1. Create a new package for cluster _execution_ related classes, i.e. `cluster`, and put the new classes and interfaces there.
2. Rename `Retryer` to something cluster specific.
3. Add a _no sleep retryer_ to imitate current behavior and use it by default.
4. Have the tests to use default retryer so that users can know they have that option and how to use it.
5. As you're going to break codes around `JedisClusterCommand`, consider its sub-classes and internal classes as well and compress them least possible number of classes. For example, we don't need separate `abstract JedisClusterConnectionHandler` and `JedisSlotBasedConnectionHandler` classes anymore.

dengliming:
> Rename Retryer to something cluster specific.

Agree, If we only use it for Jedis cluster internals. Maybe we can name `ClusterRetryer`? not sure : )

I also suggest rename `runWithRetries ` to `run ` or `call`. We already know this explicitly to retry the call in `Retryer`

walles:
From my perspective this was another shot at getting backoff in.

But since backoff is now in (#2358) I won't be putting any effort into this PR.

Closing this for now. If somebody else is interested, feel free to reopen and hack away!

walles:
@dengliming, to enable you to make changes I invited you as a collaborator to the source branch in <https://github.com/walles/jedis>.

@sazzad16 you were already in there since before.

If somebody else wants access, do ask in here!

sazzad16:
@walles You don't have make everyone of us collaborator. Keeping your fork and this specific branch is more than enough.

walles:
Closing as I have moved on to other things, and I'm unlikely to look at any conflicts.

Issue #2261, opened 2020-10-03
xrange make count optional


SevereOverfl0w:
### Expected behavior

Able to skip the count parameter to xrange, or otherwise pass it null.  I think this was raised initially on the introducing PR.

### Actual behavior

No such argument

### Steps to reproduce:

Call xrange()

### Redis / Jedis Configuration

JedisPool

#### Java version: 11.0.9

PR #2380, opened 2021-02-27
xrange make count optional

dengliming:
closes [https://github.com/redis/jedis/issues/2261](https://github.com/redis/jedis/issues/2261)


sazzad16:
@SevereOverfl0w Does this approach fulfills yours requirement?

sazzad16:
@dengliming It'd be helpful if you could comment on #2305. Thanks

sazzad16:
@dengliming Do we still need this after #2305 ?

SevereOverfl0w:
This would work, although a doc update would be helpful so it's clear this feature exists.  Additionally, having an arity which drops the count altogether would make it clear that this method is available.

sazzad16:
@dengliming I gave a long thought and finally I have decided to prefer #2460.

PR #2393, opened 2021-03-01
Use IllegalStateException to replace invalid JedisDataException

sazzad16:
Reference https://github.com/redis/jedis/pull/2360#issuecomment-787722937 

Closes #2360 
Closes #1733


src/main/java/redis/clients/jedis/Pipeline.java
```diff
@@ -20,8 +20,8 @@
       List<Object> values = new ArrayList<>();
 
       if (list.size() != responses.size()) {
-        throw new JedisDataException("Expected data size " + responses.size() + " but was "
-            + list.size());
+        throw new IllegalStateException(
```
mp911de, line 23: One could argue whether `JedisDataException` is more appropriate than `IllegalStateException`. In any case, there's no way to recover from here as that case could only be a result of a bug within Jedis or Redis.

src/main/java/redis/clients/jedis/util/SafeEncoder.java
```diff
@@ -27,7 +26,7 @@ private SafeEncoder(){
   public static byte[] encode(final String str) {
     try {
       if (str == null) {
-        throw new JedisDataException("value sent to redis cannot be null");
+        throw new JedisException("null value cannot be sent to redis");
```
mp911de, line 29: How about `IllegalArgumentException`?

mp911de (dismissed):
Looks good.

src/main/java/redis/clients/jedis/Pipeline.java
sazzad16: We[*] are trying to dedicate `JedisDataException` only for a direct response from Redis, i.e. `-ERR` response. From that perspective, it's not valid here.

[*] Frankly, it's just me right now. But the idea was generated and supported way back by both @HeartSaVioR and @marcosnils. But it is always getting stuck because of the backward incompatibility.

src/main/java/redis/clients/jedis/util/SafeEncoder.java
sazzad16, line 29:
```suggestion
        throw new IllegalArgumentException("null value cannot be sent to redis");
```

src/test/java/redis/clients/jedis/tests/JedisTest.java
```diff
@@ -114,7 +113,7 @@ public void infiniteTimeout() throws Exception {
     jedis.close();
   }
 
-  @Test(expected = JedisDataException.class)
+  @Test(expected = JedisException.class)
```
sazzad16, line 116:
```suggestion
  @Test(expected = IllegalArgumentException.class)
```

mp911de:
A thought on backward-compatibility: These exceptions aren't really ones where an application can recover from, especially when trying to send `null`. The only case where I can see it would matter if there's an interactive application such as a Java CLI for Redis built using Jedis.

sazzad16:
@mp911de Would it be a good idea to mark this PR for 3.6.0?

gkorland (approved):

dengliming (approved):

src/test/java/redis/clients/jedis/tests/PipeliningTest.java
```diff
@@ -351,27 +351,27 @@ public void multiUnwatch() {
     assertEquals(expect, pipe.syncAndReturnAll());
   }
 
-  @Test(expected = JedisDataException.class)
-  public void pipelineExecShoudThrowJedisDataExceptionWhenNotInMulti() {
+  @Test(expected = IllegalStateException.class)
+  public void pipelineExecShoudThrowBatchExceptionWhenNotInMulti() {
```
sazzad16, line 355:
```suggestion
  public void pipelineExecWhenNotInMulti() {
```

src/test/java/redis/clients/jedis/tests/PipeliningTest.java
```diff
@@ -351,27 +351,27 @@ public void multiUnwatch() {
...
     Pipeline pipeline = jedis.pipelined();
     pipeline.exec();
   }
 
-  @Test(expected = JedisDataException.class)
-  public void pipelineDiscardShoudThrowJedisDataExceptionWhenNotInMulti() {
+  @Test(expected = IllegalStateException.class)
+  public void pipelineDiscardShoudThrowBatchExceptionWhenNotInMulti() {
```
sazzad16, line 361:
```suggestion
  public void pipelineDiscardWhenNotInMulti() {
```

src/test/java/redis/clients/jedis/tests/PipeliningTest.java
```diff
@@ -351,27 +351,27 @@ public void multiUnwatch() {
...
     Pipeline pipeline = jedis.pipelined();
     pipeline.discard();
   }
 
-  @Test(expected = JedisDataException.class)
-  public void pipelineMultiShoudThrowJedisDataExceptionWhenAlreadyInMulti() {
+  @Test(expected = IllegalStateException.class)
+  public void pipelineMultiShoudThrowBatchExceptionWhenAlreadyInMulti() {
```
sazzad16, line 367:
```suggestion
  public void pipelineMultiWhenAlreadyInMulti() {
```

src/test/java/redis/clients/jedis/tests/PipeliningTest.java
```diff
@@ -351,27 +351,27 @@ public void multiUnwatch() {
...
     Pipeline pipeline = jedis.pipelined();
     pipeline.multi();
     pipeline.set("foo", "3");
     pipeline.multi();
   }
 
-  @Test(expected = JedisDataException.class)
+  @Test(expected = IllegalStateException.class)
   public void testJedisThowExceptionWhenInPipeline() {
```
sazzad16, line 375:
```suggestion
  public void testJedisThrowExceptionWhenInPipeline() {
```

PR #2395, opened 2021-03-02
Fix parameter types and return types

### LINKED ISSUE #2374: BUG: restore and restoreReplace not using correct native type for the TTL parameter
### Expected behavior

The restore* group of functions use int as the TTL parameter, this should actually be long because the redis restore command sets the TTL as milliseconds NOT seconds.

As per the redis docs: https://redis.io/commands/restore
"If ttl is 0 the key is created without any expire, otherwise the specified expire time (in milliseconds) is set".

### Actual behavior

Integer overflow when passing a large enough TTL as milliseconds

### Steps to reproduce:
```java
String key = "testkey";
int ttl = (int)TimeUnit.DAYS.toSeconds(30); // 30 days as seconds
jedis.set(key, "test-value");
jedis.expire(key, ttl);

// dump record and TTL as milliseconds
byte[] dump = jedis.dump(key);
long pttl = jedis.pttl(key);
jedis.restoreReplace(key, Math.toIntExact(pttl), dump); // results in: ArithmeticException: integer overflow
//jedis.restoreReplace(key, (int)pttl, dump); // unsafe cast results in: JedisDataException: ERR Invalid TTL value, must be >= 0
```

#### Jedis version: 
latest

#### Redis version: 
6.0.9

#### Java version: 
openjdk version "11.0.9.1"

### LINKED ISSUE #2394: Should we keep the parameter types of BinaryJedis PipelineBase stay the same?
e.g. 

`BinaryJedis`
public  List<byte[]> xclaim(byte[] key, byte[] groupname, byte[] consumername, long minIdleTime, long newIdleTime, int retries, boolean force, `byte[][] ids`)

`PipelineBase:`
public Response<List<byte[]>> xclaim(byte[] key, byte[] group, byte[] consumername, long minIdleTime, long newIdleTime, int retries, boolean force, `byte[]... ids`)

`BinaryJedis`
public List<byte[]> xrange(byte[] key, byte[] start, byte[] end, `long count`)

`Jedis`
public List<StreamEntry> xrange(final String key, final StreamEntryID start, final StreamEntryID end, `final int count`)

and BTW
should we return `Response<List<byte[]>>` ？
`redis.clients.jedis.PipelineBase#Response<List<StreamPendingEntry>> xpending(byte[] key, byte[] groupname, byte[] start, byte[] end, int count, byte[] consumername)`

sazzad16:
Resolves #2374 
Resolves #2394

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -129,8 +129,8 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>3.8.1</version>
 				<configuration>
-					<source>1.7</source>
-					<target>1.7</target>
+					<source>1.8</source>
+					<target>1.8</target>
 				</configuration>
 			</plugin>
 			<plugin>
--- a/src/main/java/redis/clients/jedis/BinaryClient.java
+++ b/src/main/java/redis/clients/jedis/BinaryClient.java
@@ -193,7 +193,12 @@ public class BinaryClient extends Connection {
     sendCommand(DBSIZE);
   }
 
+  @Deprecated
   public void expire(final byte[] key, final int seconds) {
+    sendCommand(EXPIRE, key, toByteArray(seconds));
+  }
+
+  public void expire(final byte[] key, final long seconds) {
     sendCommand(EXPIRE, key, toByteArray(seconds));
   }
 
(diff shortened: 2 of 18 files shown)


sazzad16:
Closing in favor of #2396

PR #2396, opened 2021-03-02
Fix parameter types and return types

### LINKED ISSUE #2374: BUG: restore and restoreReplace not using correct native type for the TTL parameter

### LINKED ISSUE #2394: Should we keep the parameter types of BinaryJedis PipelineBase stay the same?

sazzad16:
Resolves #2374 
Resolves #2394

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryClient.java
+++ b/src/main/java/redis/clients/jedis/BinaryClient.java
@@ -193,7 +193,12 @@ public class BinaryClient extends Connection {
     sendCommand(DBSIZE);
   }
 
+  @Deprecated
   public void expire(final byte[] key, final int seconds) {
+    sendCommand(EXPIRE, key, toByteArray(seconds));
+  }
+
+  public void expire(final byte[] key, final long seconds) {
     sendCommand(EXPIRE, key, toByteArray(seconds));
   }
 
@@ -237,7 +242,12 @@ public class BinaryClient extends Connection {
     sendCommand(SETNX, key, value);
   }
 
+  @Deprecated
   public void setex(final byte[] key, final int seconds, final byte[] value) {
+    sendCommand(SETEX, key, toByteArray(seconds), value);
+  }
+
+  public void setex(final byte[] key, final long seconds, final byte[] value) {
     sendCommand(SETEX, key, toByteArray(seconds), value);
   }
 
@@ -1068,11 +1078,21 @@ public class BinaryClient extends Connection {
     sendCommand(DUMP, key);
   }
 
+  @Deprecated
   public void restore(final byte[] key, final int ttl, final byte[] serializedValue) {
     sendCommand(RESTORE, key, toByteArray(ttl), serializedValue);
   }
 
+  public void restore(final byte[] key, final long ttl, final byte[] serializedValue) {
+    sendCommand(RESTORE, key, toByteArray(ttl), serializedValue);
+  }
+
+  @Deprecated
   public void restoreReplace(final byte[] key, final int ttl, final byte[] serializedValue) {
+    sendCommand(RESTORE, key, toByteArray(ttl), serializedValue, Keyword.REPLACE.getRaw());
+  }
+
+  public void restoreReplace(final byte[] key, final long ttl, final byte[] serializedValue) {
     sendCommand(RESTORE, key, toByteArray(ttl), serializedValue, Keyword.REPLACE.getRaw());
   }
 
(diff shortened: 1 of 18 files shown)


dengliming (commented):
Looks great.

src/main/java/redis/clients/jedis/commands/BinaryJedisCommands.java
```diff
@@ -41,11 +41,26 @@
...
   byte[] dump(byte[] key);
 
-  String restore(byte[] key, int ttl, byte[] serializedValue);
+  @Deprecated
+  default String restore(byte[] key, int ttl, byte[] serializedValue) {
+    return restore(key, (long) ttl, serializedValue);
+  }
 
-  String restoreReplace(byte[] key, int ttl, byte[] serializedValue);
+  String restore(byte[] key, long ttl, byte[] serializedValue);
```
dengliming: nice 👍

src/main/java/redis/clients/jedis/BinaryShardedJedis.java
```diff
@@ -1124,14 +1124,15 @@ public Long xtrim(byte[] key, long maxLen, boolean approximateLength) {
   }
 
   @Override
-  public List<byte[]> xpending(byte[] key, byte[] groupname, byte[] start, byte[] end, int count, byte[] consumername) {
+  public List<Object> xpending(byte[] key, byte[] groupname, byte[] start, byte[] end, int count,
```
dengliming, line 1127: Any reason to not use `List<byte[]>` here?

sazzad16: From [Redis doc](https://redis.io/commands/xpending):
```
> XPENDING mystream group55
1) (integer) 1
2) 1526984818136-0
3) 1526984818136-0
4) 1) 1) "consumer-123"
      2) "1"
```
The response can't be represented by `List<byte[]>`.

src/main/java/redis/clients/jedis/BinaryClient.java
```diff
@@ -193,10 +193,15 @@ public void dbSize() {
     sendCommand(DBSIZE);
   }
 
+  @Deprecated
```
gkorland, line 196: Add deprecated JDoc pointing to the other methods

src/main/java/redis/clients/jedis/PipelineBase.java
```diff
@@ -2088,11 +2088,10 @@
   }
   
   @Override
-  public Response<List<StreamPendingEntry>> xpending(byte[] key, byte[] groupname, byte[] start, byte[] end, int count, byte[] consumername){
+  public Response<List<Object>> xpending(byte[] key, byte[] groupname, byte[] start, byte[] end, int count, byte[] consumername){
```
gkorland, line 2091: @sazzad16 revert?

gkorland (changes requested):

src/main/java/redis/clients/jedis/PipelineBase.java
sazzad16: @gkorland Not exactly but nice catch. Keeping the old version while adding a new method.

src/main/java/redis/clients/jedis/BinaryClient.java
gkorland: ?

sazzad16: @gkorland Ooops!! PTAL again!

gkorland (approved):

src/main/java/redis/clients/jedis/BinaryJedis.java
```diff
@@ -539,7 +539,7 @@ public Long dbSize() {
    *         2.1.3, Redis &gt;= 2.1.3 will happily update the timeout), or the key does not exist.
    */
   @Override
-  public Long expire(final byte[] key, final int seconds) {
+  public Long expire(final byte[] key, final long seconds) {
```
walles, line 542: This is already merged now, but [Duration](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html) might be a better unit for the timeout.

sazzad16: @walles I'm not sure if you noticed, this not related to socket timeout. This is actually [EXPIRE](https://redis.io/commands/expire) command.

Still, this can be considered. But let's rebuild Jedis first :)

PR #2400, opened 2021-03-04
Unify non-clustered and clustered interfaces

### LINKED ISSUE #1545: Common interface for binary clustered and non-clustered redis needed

sazzad16:
Previous attempt #2021 

Resolves #1545 
Closes #2021


gkorland:
@sazzad16 no breaking changes?

sazzad16:
@gkorland Nope. Just some `UnsupportedOperationException` thrown from some newly added methods only in cluster mode.

PS: XINFO methods in cluster mode can be implemented. We can do those after resolving everything else.

PR #2402, opened 2021-03-07
Modify config pattern

sazzad16:
1. Added the word `millis` to the timeout parameters. I have seen people are confused about the unit of timeout parameters in almost everywhere (stackoverflow, github issues, mailing list). Adding the unit in config param name would be self explanatory.
2. After upgrading to JDK 1.8, it is now possible to use `default` which would allow users to code less.

I have put above changes in one PR because merging any of these would cause the other to be git-conflicted.

### DIFF
--- a/src/main/java/redis/clients/jedis/DefaultJedisClientConfig.java
+++ b/src/main/java/redis/clients/jedis/DefaultJedisClientConfig.java
@@ -22,13 +22,13 @@ public final class DefaultJedisClientConfig implements JedisClientConfig {
 
   private final HostAndPortMapper hostAndPortMapper;
 
-  private DefaultJedisClientConfig(int connectionTimeout, int soTimeout, int infiniteSoTimeout,
-      String user, String password, int database, String clientName,
+  private DefaultJedisClientConfig(int connectionTimeoutMillis, int soTimeoutMillis,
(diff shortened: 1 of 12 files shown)


src/main/java/redis/clients/jedis/DefaultJedisClientConfig.java
```diff
@@ -26,9 +26,9 @@ private DefaultJedisClientConfig(int connectionTimeout, int soTimeout, int infin
       String user, String password, int database, String clientName,
       boolean ssl, SSLSocketFactory sslSocketFactory, SSLParameters sslParameters,
       HostnameVerifier hostnameVerifier, HostAndPortMapper hostAndPortMapper) {
-    this.connectionTimeout = connectionTimeout;
-    this.soTimeout = soTimeout;
-    this.infiniteSoTimeout = infiniteSoTimeout;
+    this.connectionTimeoutMillis = connectionTimeout;
```
gkorland, line 29: @sazzad16 should you rename the input arguments too?

gkorland (approved):

src/main/java/redis/clients/jedis/DefaultJedisClientConfig.java
sazzad16: @gkorland yes, I should.

gkorland (approved):

PR #2403, opened 2021-03-07
Support Sentinel with TLS (II)

sazzad16:


src/main/java/redis/clients/jedis/JedisSentinelPool.java
```diff
@@ -15,30 +17,15 @@
...
+  protected final JedisClientConfig sentinelClientConfig;
 
-  protected int sentinelConnectionTimeout;
-  protected int sentinelSoTimeout;
-  protected String sentinelUser;
-  protected String sentinelPassword;
-  protected String sentinelClientName;
-
-  protected final Set<MasterListener> masterListeners = new HashSet<>();
+  protected final Collection<MasterListener> masterListeners = new LinkedList<>();
```
gkorland, line 28: @sazzad16 why did you move from Set to List?

sazzad16: @gkorland I didn't find any logic to put `MasterListener`s in a Set. For example, to put in a HashSet, the respective class should override hashCode and equals methods.

As this is a breaking change, I included it here. Even if we consider that there are codes that use masterListeners, hopefully this won't break many codes as Collection is super class to Set.

sazzad16:
@gkorland Please check #2445

src/main/java/redis/clients/jedis/JedisSentinelPool.java
mina-asham: Doesn't even seem like MasterListener implements equals/hashcode, so this wouldn't have been working as expected anyway.

Why LinkedList specifically though? ArrayList are usually more efficient

src/main/java/redis/clients/jedis/JedisSentinelPool.java
```diff
@@ -160,25 +147,32 @@ public JedisSentinelPool(String masterName, Set<String> sentinels,
...
-    this.sentinelClientName = sentinelClientName;
+    this.masterClientConfig = masteClientConfig;
+    this.sentinelClientConfig = sentinelClientConfig;
 
     HostAndPort master = initSentinels(sentinels, masterName);
     initPool(master);
   }
 
+  private static Set<HostAndPort> parseHostAndPorts(Set<String> strings) {
+    return strings.parallelStream().map(str -> HostAndPort.parseString(str)).collect(Collectors.toSet());
```
mina-asham, line 173: I would avoid `parallelStream` here:
1. I don't really think it's that intensive of an operation that you are going to get a gain
2. It's going to use the shared forkjoin pool and might affect other operations

Also we can use the method reference for the map operation.

```suggestion
    return strings.stream().map(HostAndPort::parseString).collect(Collectors.toSet());
```

mina-asham (changes requested):

src/main/java/redis/clients/jedis/JedisSentinelPool.java
sazzad16: @mina-asham IMO, the number of sentinels are usually very low, may be 3-5 in most use cases. Java would create 16-size array with the empty constructor of ArrayList. So I though to avoid that.

Only usage of this collection/list is in shutdown and only operation is a for-each loop. There's no special gain for ArrayList in this case.

mina-asham: Default size for array list is 10, but I think it's still more efficient from a size point of view in anyway, each LinkedList item is stored in a node that stores 3 pointer (data+next+prev), so if you have 3-5 items that means 9-15 pointers, the array will have 1 pointer per item regardless, so it can be less/more by a few bytes here, so I don't think the size comparison matter much.

The iteration perfomance is usually way better though because the array is contiguous in memory, but again for 3-5 items it wouldn't matter much here.

So I think both are more or less the same for the size of the data, and because ArrayList is the one that's used for most cases and using other ones IMO usually mean there is a specific reason for it, that's why I would opt for ArrayList.

mina-asham (dismissed):
I still think the ArrayList is the right one to use here, but it doesn't really change anything so will leave that up to you :D

src/main/java/redis/clients/jedis/JedisSentinelPool.java
```diff
@@ -16,33 +17,13 @@
...
-  @Deprecated protected int sentinelConnectionTimeout;
-  @Deprecated protected int sentinelSoTimeout;
-  @Deprecated protected String sentinelUser;
-  @Deprecated protected String sentinelPassword;
-  @Deprecated protected String sentinelClientName;
-
   private final JedisClientConfig sentinelClientConfig;
 
-  protected final Set<MasterListener> masterListeners = new HashSet<>();
+  protected final Collection<MasterListener> masterListeners = new ArrayList<>();
```
dengliming, line 26: Is there any reason to use `Collection` and not `List`?

sazzad16: There is no specific reason for it to be `List`. That's why `Collection`.

dengliming: I hope you don't mind. As a Java developer, I see this as a programming habit. Just like I wouldn't use an ArrayList directly because I need to subtract an import.

sazzad16: XD

sazzad16: Me too.
I only shift if it is in only one method and its life is only within that method.

sazzad16: I (hopefully) understand what you're indicating. I should've avoided that.

dengliming: That's just a personal opinion. I wouldn't object to your approach. : )

mina-asham (approved):

dengliming (approved):

PR #2404, opened 2021-03-08
Proper use of JedisClusterOperationException

sazzad16:
`JedisClusterMaxAttemptsException` and `JedisNoReachableClusterNodeException` seem unnecessary. These two exceptions created for two very specific use cases which can be easily represented by `JedisClusterOperationException` with proper error messages.


src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -102,8 +100,9 @@ private T runWithRetries(final int slot) {
 
         return execute(connection);
 
-      } catch (JedisNoReachableClusterNodeException jnrcne) {
-        throw jnrcne;
+      } catch (JedisClusterOperationException jcoe) {
+        // cluster state is unstable, so throw it instantly
+        throw jcoe;
```
sazzad16, line 105:
```suggestion
```

gkorland:
@sazzad16 I don't see the reason to remove these exceptions, it seems like no harm can happen by giving the user more specific exceptions, matching String is never a good idea.

sazzad16:

PR #2409, opened 2021-03-09
JedisNoReachableClusterNodeException should extend JedisClusterOperationException (II)

sazzad16:
`JedisClusterMaxAttemptsException` and `JedisNoReachableClusterNodeException` seem like luxury. These two exceptions created for two very specific use cases which can be easily represented by `JedisClusterOperationException` with proper error message.

This is an alternate to #2375. This can considered as a good balance between #2375 and #2404 if we target this for Jedis 4.0.

Note for reviewers:
---
Choose either #2404 or this PR.


src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -102,8 +101,9 @@ private T runWithRetries(final int slot) {
 
         return execute(connection);
 
-      } catch (JedisNoReachableClusterNodeException jnrcne) {
-        throw jnrcne;
+      } catch (JedisClusterOperationException jcoe) {
+        // cluster state is unstable, so throw the exception immediately
+        throw jcoe;
```
sazzad16, line 106:
```suggestion
```

mina-asham (approved):

yangbodong22011:
@sazzad16 LGTM，but I noticed that `JedisNoReachableClusterNodeException` has `@deprecated`, so if we will remove it in future,  #2404 also need be merged?

sazzad16:
@yangbodong22011 Yes, in Jedis 5.

In other words, if we decide that #2404 is too early for Jedis 4, we can push it to Jedis 5. So we can merge this for Jedis 4.

Frankly, even though I created this PR, I'm not liking it. I'd avoid hanging things for a long time. So I prefer #2404 over this one for Jedis 4.

yangbodong22011 (approved):
@sazzad16 OK, let's merge it first.

PR #2410, opened 2021-03-09
Add support PXAT/EXAT arguments to SET command

dengliming:
Resolves [https://github.com/redis/jedis/issues/2386](https://github.com/redis/jedis/issues/2386)

### DIFF
--- a/src/main/java/redis/clients/jedis/params/SetParams.java
+++ b/src/main/java/redis/clients/jedis/params/SetParams.java
@@ -12,6 +12,8 @@ public class SetParams extends Params {
   private static final String NX = "nx";
   private static final String PX = "px";
   private static final String EX = "ex";
+  private static final String EXAT = "exat";
+  private static final String PXAT = "pxat";
   private static final String KEEPTTL = "keepttl";
   private static final String GET = "get";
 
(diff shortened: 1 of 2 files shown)


src/main/java/redis/clients/jedis/params/SetParams.java
```diff
@@ -71,6 +73,26 @@ public SetParams xx() {
     return this;
   }
 
+  /**
+   * Set the specified Unix time at which the key will expire, in seconds.
+   * @param seconds
+   * @return SetParams
+   */
+  public SetParams exat(long seconds) {
```
sazzad16, line 81:
```suggestion
  public SetParams exAt(long seconds) {
```

src/main/java/redis/clients/jedis/params/SetParams.java
```diff
@@ -71,6 +73,26 @@ public SetParams xx() {
...
+    addParam(EXAT, seconds);
+    return this;
+  }
+
+  /**
+   * Set the specified Unix time at which the key will expire, in milliseconds.
+   * @param milliseconds
+   * @return SetParams
+   */
+  public SetParams pxat(long milliseconds) {
```
sazzad16, line 91:
```suggestion
  public SetParams pxAt(long milliseconds) {
```

sazzad16 (changes requested):
Mostly LGTM. Suggested two minor changes.

src/main/java/redis/clients/jedis/params/SetParams.java
dengliming: Thanks for quick reply. That was done.

sazzad16 (approved):

Issue #2114, opened 2019-12-11
please implements ZAddParams.incr() method!


startjava:
please implements ZAddParams.incr() method!

dengliming:
You should use ZIncrByParams.

dengliming:
@sazzad16 Seems we can close this one. : )

sazzad16:
@dengliming I'm not sure. Redis has INCR option in ZADD command. So we should keep it under consideration.

And it is not possible to pass ZIncrByParams in zadd(...) method. So I may not close this issue without resolution :|

dengliming:
Thanks. you were totally right. Redis doesn't have multi parameters for ZINCRBY.

dengliming:
@sazzad16 But notice this mention about `INCR`.

_INCR: When this option is specified ZADD acts like ZINCRBY. `Only one score-element pair can be specified in this mode.`_

So  I suggest using `zincrby` can resolve this issue. WDYT?

sazzad16:
@dengliming 

> I suggest using `zincrby` can resolve this issue.

This logic applies to Redis as much as it applies to Jedis. There must be reasons that Redis have that option. And as long as Redis has it we shouldn't completely disregard it.

PR #2415, opened 2021-03-10
Add support INCR argument to ZADD command

dengliming:
Resolves [https://github.com/redis/jedis/issues/2114](https://github.com/redis/jedis/issues/2114)


src/main/java/redis/clients/jedis/params/ZAddParams.java
```diff
@@ -66,6 +67,15 @@ public ZAddParams lt() {
     return this;
   }
 
+  /**
+   * Acts like ZINCRBY and only one score-element pair can be specified in this mode.
+   * @return ZAddParams
+   */
+  public ZAddParams incr() {
```
sazzad16, line 74: `incr()` is not required here. Because in `zaddIncr(...)`, we already know that `INCR` argument is required.
Also, the added benefit of not adding `incr()` here is that user won't be able to use it in normal `zadd(...)`.

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/params/ZAddParams.java
dengliming: 👍 you were right. thanks for so quick reply. : )

sazzad16 (changes requested):
Is there any reason that you followed two styles in test codes?

Between,
```
r = z(...);
assert(r);
```
and
```
assert(z(...));
```
Is it possible to use only one?

dengliming:
@sazzad16 Thanks. One line is fine for me. : )

sazzad16 (approved):

Issue #2294, opened 2020-11-29
Add support for clientUnblock command


awaizman1:
Hi,

I couldn't find the [clientUnblock](https://redis.io/commands/client-unblock) command in jedis.
This is very useful when working with streams blocking xread commands.

Is there a plan to add this command?

thanks.

gkorland:
@awaizman1 thanks for the feature request, we'll try to add it. Meanwhile, feel free to submit a PR.

PR #2424, opened 2021-03-12
Add support for clientUnblock command

dengliming:
Resolves [https://github.com/redis/jedis/issues/2294](https://github.com/redis/jedis/issues/2294)


src/main/java/redis/clients/jedis/UnblockType.java
```diff
@@ -0,0 +1,16 @@
+package redis.clients.jedis;
+
+import redis.clients.jedis.util.SafeEncoder;
+
+/**
+ * Unblock type for {@code CLIENT UNBLOCK} command.
+ */
+public enum UnblockType {
```
sazzad16, line 8: Put this in args package and implement Rawable interface.

src/main/java/redis/clients/jedis/UnblockType.java
```diff
@@ -0,0 +1,16 @@
...
+  TIMEOUT, ERROR;
+
+  public final byte[] raw;
```
sazzad16, line 11:
```suggestion
  private final byte[] raw;
```

sazzad16 (changes requested):

dengliming:
@sazzad16 The package name `args` is very similar to `params`

src/main/java/redis/clients/jedis/args/UnblockType.java
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Unblock type for {@code CLIENT UNBLOCK} command.
+ */
+public enum UnblockType implements Rawable {
+  TIMEOUT, ERROR;
+
+  private final byte[] raw;
+
+  UnblockType() {
+    raw = SafeEncoder.encode(this.name().toLowerCase());
```
sazzad16, line 14: Any reason to make it lower?
```suggestion
    raw = SafeEncoder.encode(this.name());
```

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/args/UnblockType.java
dengliming: No, Let me remove it.

sazzad16:
> The package name `args` is very similar to `params` 

@dengliming What can you suggest us do now? `params` should stay dedicated for _options_. And it is not something that I to break things.

sazzad16 (commented):
One last thing. WDYT about using `@Nullable` before `UnblockType`?

dengliming:
> One last thing. WDYT about using @Nullable before UnblockType?

Introduce `@Nullable` maybe better. Let user clearly know the params can be null. but we also need to import the annotation api(e.g. `javax.annotation-api`) right?

dengliming:
> @dengliming What can you suggest us do now? params should stay dedicated for options. And it is not something that I to break things.

Not sure. I can't found a better way to do now. : )

sazzad16:
> Introduce `@Nullable` maybe better. Let user clearly know the params can be null. but we also need to import the annotation api(e.g. `javax.annotation-api`) right? 

Hmm, you're right. Could you please add javadocs (in all methods) stating that `UnblockType` can be `null`? It's either this or adding new methods without `UnblockType`.

sazzad16 (approved):

PR #2431, opened 2021-03-14
Fix bug when COUNT > Integer.MAX_VALUE

sazzad16:


yangbodong22011:
@sazzad16 Does the bug mean that Math.min should be used instead of Math.max?

But is there any problem with count passing long directly?

Redis also treats count as long, see [here](https://github.com/redis/redis/blob/unstable/src/t_stream.c#L1878).

sazzad16:
@yangbodong22011 That bug means if you send long instead of int, the params would mess up.

For example, calling

    jedis.xrange(binaryKey, binaryStart, binaryEnd, 5);

was fine. But calling

    jedis.xrange(binaryKey, binaryStart, binaryEnd, 5L);

would actually convert the call to 

    jedis.xrange(binaryKey, binaryStart, binaryEnd, (int) Math.max(5L, (long) Integer.MAX_VALUE));

which is

    jedis.xrange(binaryKey, binaryStart, binaryEnd, Integer.MAX_VALUE);

Again,

    jedis.xrange(binaryKey, binaryStart, binaryEnd, 10L + Integer.MAX_VALUE);

would be converted to 

    jedis.xrange(binaryKey, binaryStart, binaryEnd, -2147483639);

The later one I tried to cover by the test.

---

> Redis also treats count as long,

Redis (almost) never uses int. As both int and long are present in our codebase, I opened #2539 to decide on only one whether it's int or long.

I, personally, think that `count` is better represented by int because we/users are unlikely to get/need more than 2147483647 elements in a single call.

yangbodong22011:
I noticed that Lettuce uses [`Limit`](https://github.com/lettuce-io/lettuce-core/blob/72b533a405/src/main/java/io/lettuce/core/Limit.java) in [xrange](https://github.com/lettuce-io/lettuce-core/blob/main/src/main/java/io/lettuce/core/api/sync/RedisStreamCommands.java#L282) to mask out the details, and the bottom layer uses long, but for Jedis, it seems that introducing `Limit` is a burden.

I agree with you that the `int` range is sufficient, but it does have a logical bug, so I voted for long.
@gkorland @mina-asham @dengliming WDYT?

Issue #2435, opened 2021-03-15
Jedis.xrevrange behaves different from PipelineBase.xrevrange


mp911de:
### Expected behavior

Both methods behave the same in terms of `start`/`end` arguments

### Actual behavior

`BinaryJedis` calls `xrevrange(key, end, start, count)` while `PipelineBase` calls `xrevrange(key, start, end, count)` (note that `end` and `start` have different parameter ordering.

The pipelined method should be fixed.

#### Jedis version

3.5.1

PR #2438, opened 2021-03-15
Cherrypick xrevrange fix to 3.5

gkorland:
ref #2435


dengliming:
@gkorland Seems the latest version `3.5.2` can't be found in https://mvnrepository.com/artifact/redis.clients/jedis

sazzad16:
@dengliming It takes some time to sync. But maven would fetch 3.5.2 properly.

PR #2442, opened 2021-03-17
Add support for HRANDFIELD command

dengliming:


src/test/java/redis/clients/jedis/tests/commands/HashesCommandsTest.java
```diff
@@ -464,4 +466,37 @@ public void testBinaryHstrLen() {
...
+
+    jedis.hset("foo", hash);
+
+    assertTrue(hash.containsKey(jedis.hrandfield("foo")));
+    assertEquals(2, jedis.hrandfield("foo", 2).size());
+
+    Map<String, String> actual = jedis.hrandfieldWithValues("foo", 2);
+    assertNotNull(actual);
+    assertEquals(2, actual.size());
+    assertTrue(hash.containsKey(actual.entrySet().iterator().next().getKey()));
```
sazzad16, line 484: Check a value as well.
```
Map.Entry entry = actual.entrySet().iterator().next();
assertEquals(hash.get(entry.getKey()), entry.getValue());
```

src/main/java/redis/clients/jedis/BinaryClient.java
```diff
@@ -1,5 +1,6 @@
 package redis.clients.jedis;
 
+import static redis.clients.jedis.Protocol.Keyword.WITHVALUES;
```
sazzad16, line 3: This is a chance to sort the imports :)

sazzad16 (changes requested):
Mostly LGTM

src/main/java/redis/clients/jedis/BinaryClient.java
dengliming: Should we avoid `import .*`?

sazzad16: No.

`*` is avoided for Keywords because of conflicts with Commands. You are most welcome to use `*` resolving those conflicts.

sazzad16 (approved):

PR #2232, opened 2020-08-24
Sentinel tls support

MattiasLass:
I've implemented the requested changes from #2139, as the progress on the original PR seems to have stopped.

PR #2445, opened 2021-03-17
Support Sentinel with TLS (II/bc)

### LINKED ISSUE #1987: Support TLS connections in Sentinel mode

sazzad16:
Thanks @Talon876 for the test instances in the Makefile (#2139) 

Resolves #1987 

Closes #2024 
Closes #2139 
Closes #2232 
Closes #2339


src/main/java/redis/clients/jedis/JedisSentinelPool.java
```diff
@@ -177,14 +180,33 @@ public JedisSentinelPool(String masterName, Set<String> sentinels,
...
+
     this.poolConfig = poolConfig;
     this.factory = factory;
+    this.sentinelClientConfig = sentinelClientConfig;
 
     HostAndPort master = initSentinels(sentinels, masterName);
     initMaster(master);
   }
 
+  private static Set<HostAndPort> parseHostAndPorts(Set<String> strings) {
```
dengliming, line 206: Can we remove `static ` here?

sazzad16: Why?

dengliming: We usually do this to provide another static method to use. and of course you can do that in this case. Never mind. : )

dengliming (approved):

Issue #2447, opened 2021-03-17
Add support for ZRANDMEMBER command


dengliming:
see https://redis.io/commands/zrandmember

PR #2448, opened 2021-03-17
Add support for ZRANDMEMBER command

dengliming:
Resolves [https://github.com/redis/jedis/issues/2447](https://github.com/redis/jedis/issues/2447)


src/test/java/redis/clients/jedis/tests/commands/SortedSetCommandsTest.java
```diff
@@ -1445,4 +1444,47 @@ public void bzpopmin() {
     actual = jedis.bzpopmin(0, bbar, bfoo);
     assertEquals(new KeyedTuple(bbar, bc, 0.1d), actual);
   }
+
+  @Test
+  public void zrandmember() {
+    Map<String, Double> hash = new HashMap<>();
```
sazzad16, line 1450:
```suggestion
  public void zrandmember() {

    assertNull(jedis.zrandmember("foo"));
    assertNotNull(jedis.zrandmember("foo", 1));
    assertNotNull(jedis.zrandmemberWithScores("foo", 1));

    Map<String, Double> hash = new HashMap<>();
```

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/BinaryJedis.java
```diff
@@ -2865,7 +2886,7 @@ public Long zcount(final byte[] key, final byte[] min, final byte[] max) {
 
   protected Set<Tuple> getTupledSet() {
     List<byte[]> membersWithScores = client.getBinaryMultiBulkReply();
-    if (membersWithScores.isEmpty()) {
+    if (membersWithScores == null || membersWithScores.isEmpty()) {
```
sazzad16, line 2889: For which case are you getting `membersWithScores == null`?

dengliming: When the key was not exist, it will throw NPE.

sazzad16, line 2889: If response from Redis `nil`, we should return `null`.

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/BinaryJedis.java
dengliming: OK

sazzad16: return `null` in that particular case.

sazzad16 (approved):

PR #2450, opened 2021-03-17
Remove unused comment

dengliming:
@sazzad16 @gkorland Can you help to confirm it? Seems we already handled `consumername == NULL`  case


sazzad16:
@dengliming `consumername == NULL` is _**handled**_, but there's no test to support that.

dengliming:
@sazzad16 Thanks. I will add test for it later.

Issue #2452, opened 2021-03-18
Avoid NullPointException from SetFromList class


dengliming:
The following code throws an exception when list is null.
```
protected static class SetFromList<E> extends AbstractSet<E> implements Serializable {
    private static final long serialVersionUID = -2850347066962734052L;
    private final List<E> list;

    private SetFromList(List<E> list) {
      if (list == null) {
        throw new NullPointerException("list");
      }
      this.list = list;
    }

    protected static <E> SetFromList<E> of(List<E> list) {
      return new SetFromList<>(list);
    }
  }
```
I suggest the following code instead
```
protected static <E> SetFromList<E> of(List<E> list) {
      if (list == null) {
         return null;
      }
      return new SetFromList<>(list);
    }
```
and remove `throw new NullPointerException("list");`

PR #2454, opened 2021-03-18
Avoid NullPointException from SetFromList class

dengliming:
closes [https://github.com/redis/jedis/issues/2452](https://github.com/redis/jedis/issues/2452)


sazzad16:
@dengliming I just want to be safe and set this PR set for 4.0.0. WDYT?

dengliming:
> @dengliming I just want to be safe and set this PR set for 4.0.0. WDYT?

@sazzad16 agree with this way.

sazzad16 (approved):

PR #2462, opened 2021-03-18
Add support for ZDIFFSTORE command

dengliming:


sazzad16 (commented):
IMO, we should name it `zdiffStore`. I understand we already have zunionstore, zinterstore, etc. But we can't do much about those right now. But we can do something about this.

@gkorland

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -249,7 +249,7 @@ public static Object read(final RedisInputStream is) {
     SETEX, MSET, MSETNX, DECRBY, DECR, INCRBY, INCR, APPEND, SUBSTR, HSET, HGET, HSETNX, HMSET,
     HMGET, HINCRBY, HEXISTS, HDEL, HLEN, HKEYS, HVALS, HGETALL, HRANDFIELD, RPUSH, LPUSH, LLEN, LRANGE, LTRIM,
     LINDEX, LSET, LREM, LPOP, RPOP, RPOPLPUSH, SADD, SMEMBERS, SREM, SPOP, SMOVE, SCARD, SISMEMBER,
-    SINTER, SINTERSTORE, SUNION, SUNIONSTORE, SDIFF, SDIFFSTORE, SRANDMEMBER, ZADD, ZDIFF, ZRANGE, ZREM,
+    SINTER, SINTERSTORE, SUNION, SUNIONSTORE, SDIFF, SDIFFSTORE, SRANDMEMBER, ZADD, ZDIFF, zdiffStore, ZRANGE, ZREM,
```
sazzad16, line 252: I didn't mean this one. Enums are better as capitalized.

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/Protocol.java
dengliming: Sorry. My mistake.

sazzad16 (approved):

PR #2470, opened 2021-03-21
Add support IDLE arg to XPENDING command

dengliming:
Resolves [https://github.com/redis/jedis/issues/2389](https://github.com/redis/jedis/issues/2389)

### DIFF
--- a/src/main/java/redis/clients/jedis/params/XPendingParams.java
+++ b/src/main/java/redis/clients/jedis/params/XPendingParams.java
@@ -1,0 +1,83 @@
+package redis.clients.jedis.params;
+
+import static redis.clients.jedis.Protocol.Keyword.IDLE;
+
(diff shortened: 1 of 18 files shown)


src/main/java/redis/clients/jedis/params/XPendingParams.java
```diff
@@ -0,0 +1,83 @@
...
+    this.start = start;
+    return this;
+  }
+
+  public XPendingParams end(StreamEntryID end) {
+    this.end = end;
+    return this;
+  }
+
+  public XPendingParams count(long count) {
```
sazzad16, line 43:
```suggestion
  public XPendingParams count(int count) {
```

sazzad16 (changes requested):

sazzad16 (approved):

Issue #2471, opened 2021-03-21
COPY command that copies keys


dengliming:
https://redis.io/commands/copy

PR #2472, opened 2021-03-21
Add support for COPY command

dengliming:
Resolves [https://github.com/redis/jedis/issues/2471](https://github.com/redis/jedis/issues/2471)


src/main/java/redis/clients/jedis/params/CopyParams.java
```diff
@@ -0,0 +1,47 @@
...
+
+  private Long destinationDb;
+
+  private boolean replace;
+
+  public static CopyParams copyParams() {
+    return new CopyParams();
+  }
+
+  public CopyParams destinationDb(long destinationDb) {
```
sazzad16, line 22: `db` can't be used in cluster mode. But `CopyParams` is directly used in cluster classes. This is contradictory.

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/params/CopyParams.java
dengliming: em. Do you have any suggestions?

sazzad16 (changes requested):
IMO, we can skip CopyParams for now.

dengliming:
But I think `REPLACE` still needed.

src/main/java/redis/clients/jedis/commands/MultiKeyCommands.java
```diff
@@ -21,6 +22,11 @@
 import java.util.Set;
 
 public interface MultiKeyCommands {
+
+  Long copy(String srcKey, String dstKey, CopyParams params);
```
sazzad16, line 26:
```suggestion
  boolean copy(String srcKey, String dstKey, Integer db, boolean replace);
```
WDYT?

dengliming: I agree with returning `boolean`.  but `db` is not required. people need to pass null If they don't want it(That's why i introduce the param class). anyway They all fine to me. : )

sazzad16: If they don't want it, they can use the other method.

sazzad16:
```suggestion
  boolean copy(String srcKey, String dstKey, int db, boolean replace);
```

sazzad16:
```suggestion
  boolean copy(String srcKey, String dstKey, int db, boolean replace);
```
Sorry, my suggestion could be even better.

dengliming:
```
boolean copy(String srcKey, String dstKey, int db, boolean replace);// not for cluster
boolean copy(String srcKey, String dstKey, boolean replace);
```
So, all we need to do is add the above two methods. Is that right?

sazzad16: yes

sazzad16 (approved):

PR #2481, opened 2021-03-23
Double timeout, avoid Tuple and more

sazzad16:
(1) Implement `double` timeout:
---
Since Redis 6.0, timeout parameter of some blocking commands now supports `double` instead of integer.
- Param type is changed for the methods that are yet to be released.
- Similar methods are added for the existing (released) methods.
- BRPOPLPUSH command is skipped as it is deprecated since Redis 6.2.

(2) Move away from Tuple:
---
The word Tuple means any finite number of elements. The way it is used not _grammatically_ correct. We should move away from the class, introduce a proper name and avoid introducing wrong usage of the term.
- KeyedTuple is just added targetting 3.6.0 for BZPOPMAX and BZPOPMIN commands. It has been renamed.

(3) Binary return type:
---
For binary methods, should we return custom objects or binary array(s) `byte[]`/`List<byte[]>`?


src/main/java/redis/clients/jedis/resps/KeyedZSetElement.java
```diff
@@ -0,0 +1,43 @@
+package redis.clients.jedis.resps;
```
dengliming, line 1: I am not sure whether putting in `redis.clients.jedis.resps` this package is a good solution or not, currently lots of model class in `redis.clients.jedis` such as `StreamPendingSummary`、`Tuple`

src/main/java/redis/clients/jedis/resps/KeyedListElement.java
```diff
@@ -0,0 +1,45 @@
+package redis.clients.jedis.resps;
+
+import redis.clients.jedis.util.SafeEncoder;
+
+public class KeyedListElement {
+
+  private final String key;
+  private final String element;
+
```
dengliming, line 12: I am thinking about it is possible to do with generic?

```
public class KeyedListElement<K, E> {

  private final K key;
  private final E element;
```

src/main/java/redis/clients/jedis/resps/KeyedZSetElement.java
sazzad16: We can move the new ones (targeted 3.6.0) here and plan to gradually move the existing ones.

src/main/java/redis/clients/jedis/resps/KeyedListElement.java
sazzad16: It is a matter of discussion. I avoided using generics for just two commands.

src/main/java/redis/clients/jedis/resps/KeyedZSetElement.java
dengliming: 👍 Is it also possible to add `package-info.java` to let people know what it's for?

sazzad16: sure!

dengliming:
> For binary methods, should we return custom objects or binary array(s) byte[]/List<byte[]>?

👍 That really should be discussed. and It is weird to see both `byte` and `non-byte` attributes in the same class.

mp911de:
I'd suggest splitting the ticket into two topics and avoid breaking changes (return types) until Jedis 4.0. It would also make sense to have an overloaded method accepting `double` in addition to the `long` to not break existing code.

sazzad16:
@mp911de There are no _exact_ breaking changes. For methods that exist in 3.5.2, there are overloading methods. Methods that are seemingly breaking are just merged for 3.6.0.

mp911de:
I see, thanks.

src/main/java/redis/clients/jedis/resps/KeyedListElement.java
```diff
@@ -0,0 +1,45 @@
+package redis.clients.jedis.resps;
+
+import redis.clients.jedis.util.SafeEncoder;
+
+public class KeyedListElement {
```
gkorland, line 5: @sazzad16 I think some JavaDoc that will describe this class is needed

src/main/java/redis/clients/jedis/resps/KeyedZSetElement.java
```diff
@@ -0,0 +1,43 @@
...
+
+import redis.clients.jedis.Tuple;
+import redis.clients.jedis.util.SafeEncoder;
+
+public class KeyedZSetElement extends Tuple {
```
gkorland, line 6: @sazzad16 I think some JavaDoc that will describe this class is needed

src/main/java/redis/clients/jedis/commands/MultiKeyBinaryJedisClusterCommands.java
```diff
@@ -31,15 +30,19 @@
...
+  byte[] blmove(byte[] srcKey, byte[] dstKey, ListDirection from, ListDirection to, double timeout);
 
   List<byte[]> blpop(int timeout, byte[]... keys);
 
+  List<byte[]> blpop(double timeout, byte[]... keys);
+
   List<byte[]> brpop(int timeout, byte[]... keys);
 
-  KeyedTuple bzpopmax(int timeout, byte[]... keys);
+  List<byte[]> brpop(double timeout, byte[]... keys);
```
gkorland, line 41: @sazzad16 I think you should deprecate the interface methods and not only the implementation methods

src/main/java/redis/clients/jedis/commands/MultiKeyBinaryRedisPipeline.java
```diff
@@ -30,15 +29,19 @@
...
+  Response<byte[]> blmove(byte[] srcKey, byte[] dstKey, ListDirection from, ListDirection to, double timeout);
 
   Response<List<byte[]>> blpop(byte[]... args);
 
+  Response<List<byte[]>> blpop(double timeout, byte[]... args);
+
   Response<List<byte[]>> brpop(byte[]... args);
 
-  Response<KeyedTuple> bzpopmax(int timeout, byte[]... keys);
+  Response<List<byte[]>> brpop(double timeout, byte[]... args);
```
gkorland, line 40: @sazzad16 I think you should deprecate the interface methods and not only the implementation methods

src/main/java/redis/clients/jedis/commands/MultiKeyCommands.java
```diff
@@ -35,19 +36,23 @@
...
-  String blmove(String srcKey, String dstKey, ListDirection from, ListDirection to, int timeout);
+  String blmove(String srcKey, String dstKey, ListDirection from, ListDirection to, double timeout);
 
   List<String> blpop(int timeout, String... keys);
 
+  KeyedListElement blpop(double timeout, String... keys);
+
   List<String> brpop(int timeout, String... keys);
 
+  KeyedListElement brpop(double timeout, String... keys);
```
gkorland, line 47: @sazzad16 I think you should deprecate the interface methods and not only the implementation methods

src/main/java/redis/clients/jedis/commands/MultiKeyCommandsPipeline.java
```diff
@@ -32,15 +32,24 @@
...
   Response<List<String>> blpop(String... args);
 
+  Response<List<String>> blpop(int timeout, String... args);
+
+  Response<KeyedListElement> blpop(double timeout, String... args);
+
   Response<List<String>> brpop(String... args);
 
-  Response<KeyedTuple> bzpopmax(int timeout, String... keys);
+  Response<List<String>> brpop(int timeout, String... args);
```
gkorland, line 46: @sazzad16 I think you should deprecate the interface methods and not only the implementation methods

src/main/java/redis/clients/jedis/commands/MultiKeyJedisClusterCommands.java
```diff
@@ -32,15 +32,19 @@
...
+  String blmove(String srcKey, String dstKey, ListDirection from, ListDirection to, double timeout);
 
   List<String> blpop(int timeout, String... keys);
 
+  KeyedListElement blpop(double timeout, String... keys);
+
   List<String> brpop(int timeout, String... keys);
 
-  KeyedTuple bzpopmax(int timeout, String... keys);
+  KeyedListElement brpop(double timeout, String... keys);
```
gkorland, line 43: @sazzad16 I think you should deprecate the interface methods and not only the implementation methods

sazzad16: @gkorland The deprecated methods were not part of any interface.

src/main/java/redis/clients/jedis/commands/MultiKeyCommandsPipeline.java
sazzad16: @gkorland The deprecated methods were not part of any interface.

src/main/java/redis/clients/jedis/commands/MultiKeyCommands.java
sazzad16: @gkorland The deprecated methods were not part of any interface.

src/main/java/redis/clients/jedis/commands/MultiKeyBinaryRedisPipeline.java
sazzad16: @gkorland The deprecated methods were not part of any interface.

src/main/java/redis/clients/jedis/commands/MultiKeyBinaryJedisClusterCommands.java
sazzad16: @gkorland The deprecated methods were not part of any interface.

src/main/java/redis/clients/jedis/BinaryJedis.java
```diff
@@ -2641,25 +2650,33 @@ public Long sort(final byte[] key, final byte[] dstkey) {
     return args;
   }
 
+  private byte[][] getKeysAndTimeout(double timeout, byte[][] keys) {
```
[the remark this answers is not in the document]
sazzad16: In both BinaryJedis and BinaryClient classes I just overloaded existing methods with `int timeout` param for a `double` variant.

src/main/java/redis/clients/jedis/ShardedJedis.java
```diff
@@ -206,20 +207,16 @@ public String psetex(final String key, final long milliseconds, final String val
     return j.psetex(key, milliseconds, value);
   }
 
-  public List<String> blpop(final String arg) {
```
sazzad16, line 209: @gkorland @dengliming I've now removed this method, because it's unusable.

src/main/java/redis/clients/jedis/ShardedJedis.java
```diff
@@ -206,20 +207,16 @@ public String psetex(final String key, final long milliseconds, final String val
...
-    return j.blpop(arg);
-  }
-
   @Override
   public List<String> blpop(final int timeout, final String key) {
     Jedis j = getShard(key);
     return j.blpop(timeout, key);
   }
 
-  public List<String> brpop(final String arg) {
```
sazzad16, line 220: @gkorland @dengliming I've now removed this method, because it's unusable.

src/main/java/redis/clients/jedis/BinaryJedis.java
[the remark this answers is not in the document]
sazzad16: @gkorland Renamed in BinaryClient.

src/main/java/redis/clients/jedis/resps/KeyedZSetElement.java
sazzad16: @gkorland JavaDoc added.

src/main/java/redis/clients/jedis/resps/KeyedZSetElement.java
sazzad16: @dengliming package-info added.

src/main/java/redis/clients/jedis/resps/KeyedListElement.java
sazzad16: @gkorland JavaDoc added.

src/main/java/redis/clients/jedis/ShardedJedis.java
gkorland: @sazzad16 isn't it a backward that will break applications runtime? why not deprecate it?

src/main/java/redis/clients/jedis/ShardedJedis.java
gkorland: @sazzad16 isn't it a backward that will break applications runtime? why not deprecate it?

src/main/java/redis/clients/jedis/resps/KeyedListElement.java
gkorland: @dengliming why do you think it should be generic? when do you see it used not as String String?

src/main/java/redis/clients/jedis/ShardedJedis.java
sazzad16: Programmatically, yes, it's breaking. But functionally, it's broken since the moment it was created. It's unusable. Trying to use it will always generate error from Redis. Because this command requires at least two arguments and there's no way to pass those. I don't believe there's a sane application that has this in their code. That's why I removed it. (Even though I deprecated it first.)

sazzad16: @gkorland Programmatically, yes, it's breaking. But functionally, it's broken since the moment it was created. It's unusable. Trying to use it will always generate error from Redis. Because this command requires at least two arguments and there's no way to pass those. I don't believe there's a sane application that has this in their code. That's why I removed it. (Even though I deprecated it first.)

src/main/java/redis/clients/jedis/ShardedJedis.java
sazzad16: @gkorland Programmatically, yes, it's breaking. But functionally, it's broken since the moment it was created. It's unusable. Trying to use it will always generate error from Redis. Because this command requires at least two arguments and there's no way to pass those. I don't believe there's a sane application that has this in their code. That's why I removed it. (Even though I deprecated it first.)

src/main/java/redis/clients/jedis/ShardedJedis.java
sazzad16: @gkorland Check brpop https://github.com/redis/jedis/pull/2481/files#r602924002

sazzad16: @gkorland check blpop https://github.com/redis/jedis/pull/2481/files#r602924002

dengliming (approved):

PR #2482, opened 2021-03-23
Add support ABSTTL、IDLETIME and FREQ args to RESTORE command

dengliming:
Resolves [https://github.com/redis/jedis/issues/2476](https://github.com/redis/jedis/issues/2476)


sazzad16 (changes requested):
Please deprecate `restoreReplace` methods. Use both annotation and javadoc. In javadoc, suggest the new method(s).

sazzad16 (changes requested):
One deprecation is left in BinaryClient.

sazzad16 (approved):

PR #2490, opened 2021-03-29
Retry with backoff on cluster connection failures (II)

sazzad16:
Original PR #2358 (by @walles)


src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -132,14 +159,60 @@ private T runWithRetries(final int slot) {
     throw maxAttemptsException;
```
yangbodong22011, line 159: If `maxAttempts = 0`, we will get following exception():
```
Exception in thread "main" java.lang.NullPointerException: Cannot suppress a null exception.
	at java.base/java.lang.Throwable.addSuppressed(Throwable.java:1057)
	at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:158)
	at redis.clients.jedis.JedisClusterCommand.run(JedisClusterCommand.java:45)
	at redis.clients.jedis.JedisCluster.set(JedisCluster.java:292)
	at redis.clients.jedis.tests.FancheTest.main(FancheTest.java:27)
```

but in the old code, we will get ` No more cluster attempts left.`, should we define default `lastException = maxAttemptsException`?

sazzad16: @yangbodong22011 
1. This is the current behavior at `master` branch.
2. Why would you send `maxAttempts = 0`?

yangbodong22011: > @yangbodong22011
> 
> 1. This is the current behavior at `master` branch.
> 2. Why would you send `maxAttempts = 0`?

1. ok, ignore that exception
2. just a test, or we should disable user Settings `maxAttempts = 0` (params check may be another PR), this PR can continue.

yangbodong22011 (approved):

src/main/java/redis/clients/jedis/JedisClusterCommand.java
```diff
@@ -85,7 +100,10 @@ public T runWithAnyNode() {
   }
 
   private T runWithRetries(final int slot) {
+    Instant deadline = Instant.now().plus(maxTotalRetriesDuration);
```
gkorland, line 103: It feels like we adding performance impact with calling now() on each command run.

gkorland (approved):

sazzad16:
Everything pushed to the original PR #2358

PR #2501, opened 2021-04-06
JedisDataException should not be wrapped within Pool operations

### LINKED ISSUE #1876: Useless retry when jedicluster configured password and try to connect to a node without password

sazzad16:
Reference https://github.com/redis/jedis/pull/1899#issuecomment-441273778 

> Why are we wrapping the exception into a JedisConnectionException?. Shouldn't just we let it fail and let the caller catch the corresponding exception?

---

Resolves #1876 

Closes #1899


sazzad16:
Reference https://github.com/redis/jedis/pull/1899#issuecomment-441273778 

> Why are we wrapping the exception into a JedisConnectionException?. Shouldn't just we let it fail and let the caller catch the corresponding exception?

dengliming (approved):

gkorland (approved):

Issue #2191, opened 2020-05-19
Improvement of JedisClusterInfoCache#renewClusterSlots


Shawyeok:
Recently I noticed the number of threads of a java application running on production increased significant and then recovered in a short time(~1m), many threads has stacktrace below:
```
"thrift-worker-715" Id=11815 WAITING on java.util.concurrent.locks.ReentrantReadWriteLock$NonfairSync@4fb62711 owned by "thrift-worker-718" Id=11818
        at sun.misc.Unsafe.park(Native Method)
        -  waiting on java.util.concurrent.locks.ReentrantReadWriteLock$NonfairSync@4fb62711
        at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:836)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireShared(AbstractQueuedSynchronizer.java:967)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireShared(AbstractQueuedSynchronizer.java:1283)
        at java.util.concurrent.locks.ReentrantReadWriteLock$ReadLock.lock(ReentrantReadWriteLock.java:727)
        at redis.clients.jedis.JedisClusterInfoCache.getSlotPool(JedisClusterInfoCache.java:234)
        at redis.clients.jedis.JedisSlotBasedConnectionHandler.getConnectionFromSlot(JedisSlotBasedConnectionHandler.java:62)
        at redis.clients.jedis.JedisClusterConnectionHandlerWraper.getConnectionFromSlot(JedisClusterConnectionHandlerWraper.java:103)
        at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:116)
        at redis.clients.jedis.JedisClusterCommand.run(JedisClusterCommand.java:31)
```
and thread **thrift-worker-718** stacktrace:
```
"thrift-worker-718" Id=11818 RUNNABLE (in native)
        at java.net.SocketInputStream.socketRead0(Native Method)
        at java.net.SocketInputStream.socketRead(SocketInputStream.java:116)
        at java.net.SocketInputStream.read(SocketInputStream.java:171)
        at java.net.SocketInputStream.read(SocketInputStream.java:141)
        at java.net.SocketInputStream.read(SocketInputStream.java:127)
[...]

PR #2514, opened 2021-04-18
Reduce lock granularity

Shawyeok:
implements #2191

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
@@ -6,6 +6,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLParameters;
@@ -15,6 +16,7 @@ import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
 
 import redis.clients.jedis.exceptions.JedisConnectionException;
 import redis.clients.jedis.exceptions.JedisException;
+import redis.clients.jedis.util.Pool;
 import redis.clients.jedis.util.SafeEncoder;
 
 public class JedisClusterInfoCache {
@@ -24,7 +26,7 @@ public class JedisClusterInfoCache {
   private final ReentrantReadWriteLock rwl = new ReentrantReadWriteLock();
   private final Lock r = rwl.readLock();
   private final Lock w = rwl.writeLock();
-  private volatile boolean rediscovering;
+  private final Lock rediscoverLock = new ReentrantLock();
 
   private final GenericObjectPoolConfig<Jedis> poolConfig;
   private final JedisClientConfig clientConfig;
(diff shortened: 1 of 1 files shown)


yangbodong22011:
@Shawyeok thanks, I have a concern about "Reduce lock granularity" will bring more `renewClusterSlots`.

In your code, before `discoverClusterSlots->w.lock();`, user requests can also call `getConnectionFromSlot`, but the cluster route is **old** (renewSlotCache not yet complete), so some requests will get `JedisRedirectionException` and call `renewSlotCache` once again.

This is inconsistent with the original Jedis behavior. Do we have a way to block those requests that will return `JedisRedirectionException` again and let others pass?

Shawyeok:
> In your code, before `discoverClusterSlots->w.lock();`, user requests can also call `getConnectionFromSlot`, but the cluster route is **old** (renewSlotCache not yet complete), so some requests will get `JedisRedirectionException` and call `renewSlotCache` once again.

In this scenario, `rediscoverLock.tryLock()` will returns `false` immediately.

yangbodong22011:
> In this scenario, `rediscoverLock.tryLock()` will returns `false` immediately.

oh, nice, so the behavior can be the same as before, and the lock time of `redis.clients.jedis.JedisClusterInfoCache#slots` is reduced.

yangbodong22011 (approved):
LGTM

dengliming (dismissed):

sazzad16 (commented):
LGTM _the core changes_.

src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -3953,6 +3953,10 @@ public void close() {
     }
   }
 
+  public Pool<Jedis> getDataSource() {
+    return dataSource;
+  }
```
sazzad16, line 3958: @Shawyeok @yangbodong22011 @dengliming I'm not so sure about this change. Can we consider not exposing or at least limiting the access of `dataSource`?

dengliming: Maybe `protected` would be better?

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -175,68 +177,72 @@ public void discoverClusterNodesAndSlots(Jedis jedis) {
...
-              }
-            }
+        List<JedisPool> shuffledNodesPool = getShuffledNodesPool();
+        // Lower priority of last node which has issue to retrieve cluster slots
+        if (jedis != null) {
+          Pool<Jedis> dataSource = jedis.getDataSource();
+          if (shuffledNodesPool.remove(dataSource)) {
+            shuffledNodesPool.add((JedisPool) dataSource);
+          }
+        }
```
yangbodong22011, line 198: @Shawyeok It seems unnecessary, because if `jedis != null`, it means that `JedisRedirectionException` was caught before (not `JedisConnectionException)`, then we use it call `discoverClusterSlots` will also success.

src/main/java/redis/clients/jedis/Jedis.java
Shawyeok: Aha, I just keep it as same as setter's access modifier.
```java
  public void setDataSource(Pool<Jedis> jedisPool) {
    this.dataSource = jedisPool;
  }
```

Feel weird that setter is `public`, but getter is `protected` or `package`, maybe we should change both together?

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
Shawyeok: Right, if cut this, then `Jedis#getDataSource()` is no need too.

yangbodong22011 (approved):

sazzad16 (approved):

dengliming (approved):
LGTM

PR #2516, opened 2021-04-20
Limit the access of setDataSource in Jedis

sazzad16:
https://github.com/redis/jedis/blob/10b195f210769a0dca7826d06dafe75f6b99de41/src/main/java/redis/clients/jedis/Jedis.java#L3956-L3958

`setDataSource` is one, if not the most, key part in maintaining thread concurrency in Jedis. We may/should consider limiting access of this method to reduce concurrency related issues.

`protected` modifier seems a reasonable choice.

### DIFF
--- a/src/main/java/redis/clients/jedis/Jedis.java
+++ b/src/main/java/redis/clients/jedis/Jedis.java
@@ -3953,7 +3953,7 @@ public class Jedis extends BinaryJedis implements JedisCommands, MultiKeyCommand
     }
   }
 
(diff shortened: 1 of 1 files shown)


sazzad16:
Reference: the idea is generated from this https://github.com/redis/jedis/pull/2514#discussion_r616445526 thread.

dengliming (approved):

gkorland (approved):

sazzad16:
Reference 2: https://stackoverflow.com/q/67925998/4021802

PR #2519, opened 2021-04-22
Void return while SHUTDOWN

sazzad16:
The return type of `shutdown()` should be `void`.
Reply on error can be thrown as an exception.


dengliming (dismissed):

src/main/java/redis/clients/jedis/BinaryJedis.java
```diff
@@ -3469,19 +3470,18 @@ public Long lastsave() {
...
    * {@link #save() SAVE} and then {@link #quit() QUIT} because other clients may alter the DB data
    * between the two commands.
-   * @return Status code reply on error. On success nothing is returned since the server quits and
-   *         the connection is closed.
+   * @throws JedisException with the status code reply on error. On success nothing is thrown since
+   *         the server quits and the connection is closed.
    */
   @Override
-  public String shutdown() {
+  public void shutdown() throws JedisException {
```
gkorland, line 3477: @sazzad16 shouldn't we deprecate this method first?

sazzad16: @gkorland Why deprecate? We're not removing it.

sazzad16: @gkorland Please check #2545

sazzad16:

PR #2520, opened 2021-04-22
CONFIG SET returns OK status

sazzad16:
So we should make `configSet` to match with all other commands that return `OK` status.


dengliming (dismissed):

src/main/java/redis/clients/jedis/BinaryJedis.java
```diff
@@ -3694,10 +3694,10 @@ public String configRewrite() {
    * @return Status code reply
    */
   @Override
-  public byte[] configSet(final byte[] parameter, final byte[] value) {
+  public String configSet(final byte[] parameter, final byte[] value) {
```
gkorland, line 3697: Shouldn't we deprecate the original method first?

sazzad16: @gkorland We are not removing the method. So I don't support deprecation. My idea is to add a warning message in jdoc. We can do that after we agree on this PR.

gkorland: I know we are not but we change the signature which will break both compile and runtime

sazzad16: Are you suggesting to deprecate without an alternate?

gkorland: no, I suggest we add an new method and deprecate the old one

sazzad16: @gkorland check #2525

src/test/java/redis/clients/jedis/tests/commands/ControlCommandsTest.java
```diff
@@ -131,7 +131,8 @@ public void configGetSetBinary() {
     List<byte[]> info = jedis.configGet(maxmemory);
     assertArrayEquals(maxmemory, info.get(0));
     byte[] memory = info.get(1);
-    assertArrayEquals("OK".getBytes(), jedis.configSet(maxmemory, memory));
+    assertEquals("OK", jedis.configSet(maxmemory, memory));
+    assertEquals("OK", jedis.configSet(maxmemory, memory));
```
gkorland, line 135:
```suggestion
    assertEquals("OK", jedis.configSetBinary(maxmemory, memory));
```
Why do you test the same method twice?

gkorland (changes requested):

src/test/java/redis/clients/jedis/tests/commands/ControlCommandsTest.java
sazzad16: @gkorland first one should be `200`. fixed.

dengliming (approved):

gkorland (approved):

PR #2064, opened 2019-09-18
Expose method to warm connection pool from JedisPool #1828

tmack8001:
I decided to expose a public method on the `AbstractJedisPool` to allow users to then explicitly warm their `JedisPool` after creation instead of adding yet another parameter to the already overwritten set of constructors for the `JedisPool`. This method wraps the call to the internal pool catching exceptions and logging the underlying error that occurred.

Example usage would be the following:
```
JedisPool jedisPool = new JedisPool(new URI("localhost:6380"));
jedisPool.preparePool(); // explicit warming of the internal connection pool for Jedis
```

The logic within preparePool is standard within the `GenericObjectPool` implementation to make sure that there is always a minimum of `minIdle` connections available within the pool. This helps in high throughput applications as then the cost of creating connections are done before hand vs upon demand/use.

PR #2054, opened 2019-09-05
Providing JedisPool to be created with abandon config

venukbh:
There are a lot of discussions and forums indicating connection issues, maxing out over period of time, connections leakage and not returning to pool.

This abandon config will give the client an option to abandon connections which are not returned back in a specified time - making the connection pool steady and available for app all the time

PR #2521, opened 2021-04-22
Pool extends GenericObjectPool

### LINKED ISSUE #1828: Warm the internalPool upon initialization of the JedisPool
### Expected behavior

I'd expect the behavior of creating a GenericConnectionPool to also warm the connection pool based on the configuration of the pool itself (minIdle, etc). This would take place by calling the following method upon initialization of the connection pool:

```
this.internalPool.preparePool();
``` 

This method internally will attempt to ensure that the correct minIdle instances are available for threads to consume and use at the point the pool is first configured. One approach could be to make this optional via a specific pool parameter/argument to say "warmPool: true | false"

### Actual behavior

The GenericObjectPool is created with the correct pool configurations, but the pool is never warmed. The logic within GenericObjectPool itself doesn't warm the pool itself upon creation this is up to the user of the object pool instance to maintain.

### LINKED ISSUE #2197: No way to apply AbandonedConfig in JedisPool
### Expected behavior
Jedis Pool is built over GenericObjectPool from apache commons-pool2. The GenericObjectPool can be configured to detect and remove "abandoned" objects, i.e. objects that have been checked out of the pool but neither used nor returned before the configured removeAbandonedTimeout.

### Actual behavior

There is no way in JedisPool or JedisPoolConfig to configure org.apache.commons.pool2.impl.AbandonedConfig in the underlying jedis pool.

### Steps to reproduce:
We have observed under high load the abandoned connections cause JedisExhaustedPoolException exception and we highly doubt that these are the abandoned connections, as we are releasing all the connections back to pool in a finally block using **jedis.close()** call, evertime we are using it. We get this exception once every 10-12 days in production environment, we are performing around 0.2m operations every minute on a redis setup consisting of 3 servers.

#### Jedis version:
3.1.0

#### Redis version:
redis_version:5.0.2

#### Java version:
java version "1.8.0_51"
Java(TM) SE Runtime Environment (build 1.8.0_51-b16)
Java HotSpot(TM) 64-Bit Server VM (build 25.51-b03, mixed mod

sazzad16:
Resolves #1828 
Closes #2064 
Closes #2054 
Resolves #2197 

By making Pool to extend GenericObjectPool, we can get rid of all the GenericObjectPool specific feature requests. Also, any future GenericObjectPool feature would be readily available in Pool without any sweat.


src/main/java/redis/clients/jedis/JedisSentinelPool.java
```diff
@@ -184,12 +184,12 @@ public JedisSentinelPool(String masterName, Set<HostAndPort> sentinels,
   }
 
   @Override
-  public void destroy() {
+  public void close() {
```
gkorland, line 187: @sazzad16 let's keep the destroy method for backward you can just call close() on destory()

src/main/java/redis/clients/jedis/util/Pool.java
```diff
@@ -12,37 +12,15 @@
...
-
-  private final GenericObjectPool<T> internalPool;
+public class Pool<T> extends GenericObjectPool<T> implements Closeable {
 
   public Pool(final GenericObjectPoolConfig<T> poolConfig, PooledObjectFactory<T> factory) {
-    this.internalPool = new GenericObjectPool<>(factory, poolConfig);
-  }
-
-  @Override
-  public void close() {
```
gkorland, line 24: @sazzad16 can we avoid the backward?

src/main/java/redis/clients/jedis/util/Pool.java
```diff
@@ -12,37 +12,20 @@
...
-    this.internalPool = new GenericObjectPool<>(factory, poolConfig);
+    super(factory, poolConfig);
   }
 
   @Override
   public void close() {
     destroy();
   }
 
-  public boolean isClosed() {
```
gkorland, line 28: @sazzad16 what about isClosed and clearInternalPool? Can't we keep them as delegation to avoid breaking changes too?

sazzad16: @gkorland isClosed is already available in GenericObjectPool.

sazzad16: @gkorland 

isClosed is already available (in super class, GenericObjectPool).

clearInternalPool is `protected`. So I took the chance of removing it in favor of an available public method `clear()` targeting a major release.

src/main/java/redis/clients/jedis/util/Pool.java
```diff
@@ -57,126 +40,41 @@ public T getResource() {
...
     }
   }
-  
-  /**
-   * Returns the number of instances currently borrowed from this pool.
-   *
-   * @return The number of instances currently borrowed from this pool, -1 if
-   * the pool is inactive.
-   */
-  public int getNumActive() {
```
gkorland, line 106: @sazzad16 should we keep these method to avoid backward issues?

sazzad16: @gkorland all of these methods are actually available in the super class GenericObjectPool. related tests are available in src/test/java/redis/clients/jedis/tests/JedisPoolTest.java

gkorland (approved):

Issue #2332, opened 2020-12-29
Why return null while 'nil' in doc ?


jackie88558:
When I call rpoplpush from an empty list, Jedis return null.
While doc tells it should be 'nil' ?
![image](https://user-images.githubusercontent.com/17266285/103284664-c896e400-4a16-11eb-8157-c4eb14f53eba.png)
Is it a doc typo?

sazzad16:
@jackie88558 The command documentation are mostly brought from [Redis documentations](https://redis.io/commands). So the `nil` that you are pointing towards is actually Redis `nil`, equivalent of which is `null` in Java (as in null Object, not `"null"` String).

jackie88558:
> @jackie88558 The command documentation are mostly brought from [Redis documentations](https://redis.io/commands). So the `nil` that you are pointing towards is actually Redis `nil`, equivalent of which is `null` in Java (as in null Object, not `"null"` String).

Thanks for ur explanation. But I think it would be better to add these words to the doc.

sazzad16:
@jackie88558 Jedis is a community driven project and you are very much welcome to craft a PR regarding this. Thanks

PR #2524, opened 2021-04-25
Change nil to null in doc

einverne:
ref #2332

### DIFF
--- a/src/main/java/redis/clients/jedis/BinaryJedis.java
+++ b/src/main/java/redis/clients/jedis/BinaryJedis.java
@@ -423,7 +423,7 @@ public class BinaryJedis implements BasicCommands, BinaryJedisCommands, MultiKey
   }
 
   /**
-   * Get the value of the specified key. If the key does not exist the special value 'nil' is
+   * Get the value of the specified key. If the key does not exist the special value 'null is
    * returned. If the value stored at key is not a string an error is returned because GET can only
    * handle string values.
    * <p>
@@ -857,7 +857,7 @@ public class BinaryJedis implements BasicCommands, BinaryJedisCommands, MultiKey
 
   /**
    * Get the values of all the specified keys. If one or more keys don't exist or is not of type
-   * String, a 'nil' value is returned instead of the value of the specified key, but the operation
+   * String, a 'null value is returned instead of the value of the specified key, but the operation
    * never fails.
    * <p>
    * Time complexity: O(1) for every key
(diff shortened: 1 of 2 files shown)


src/main/java/redis/clients/jedis/BinaryJedis.java
```diff
@@ -857,7 +857,7 @@ public String flushAll(FlushMode flushMode) {
 
   /**
    * Get the values of all the specified keys. If one or more keys don't exist or is not of type
-   * String, a 'nil' value is returned instead of the value of the specified key, but the operation
+   * String, a 'null value is returned instead of the value of the specified key, but the operation
```
dengliming, line 860: I think both are fine. According to the [redis doc](https://redis.io/commands/mget). It should be `nil`. In java, `null` is also make senses. BTW, `'null` should be `'null'`

dengliming (changes requested):

sazzad16:
@dengliming I object to this change. FYI https://github.com/redis/jedis/pull/2268#issuecomment-766351904

einverne:
> @dengliming I object to this change. FYI [#2268 (comment)](https://github.com/redis/jedis/pull/2268#issuecomment-766351904)

Oh, sorry, I misunderstand your intention in #2332, I believe we just need one place to explain this clearly. It's not necessary to update the doc everywhere in the codes. So I will close the PR.

sazzad16:
@einverne There is nothing to be sorry about. And you did not misunderstand. During #2332, I was not against using `null`. But as I was thinking about #2332 and #2268, I was preferring `nil` more and more.

devxzero:
'nil' is just plain wrong from a Java perspective. And since this is Java language, it it should be documented as null.
A lot of people who are experienced with Java, but new to Redis, will get a nullpointerexception, and wonder: _why is it returning null, the doc says "the special value 'nil' is returned"_ The 'nil' is even between quotes, which suggests that it should be taken literary as a literal string. Yes, I did wonder: why didn't they return just null? I thought, maybe Jedis wants to stick as low level as possible, so it doesn't want to translate 'nil' to null.
It causes people who are new to Redis to have to debug the Jedis code to understand why this problem exists. After some googling, I found multiple issues about this problem on this repo.

sazzad16:
@devxzero or anyone else interested, would you like to craft a PR? The PR should cover all 'nil' (as much as possible) and use more modern javadoc styles (e.g. `{@code null}`).

PR #2529, opened 2021-04-28
Remove set methods from socket factory

sazzad16:


src/main/java/redis/clients/jedis/JedisSocketFactory.java
```diff
@@ -29,24 +29,12 @@
   @Deprecated
   String getHost();
 
-  @Deprecated
-  void setHost(String host);
-
   @Deprecated
```
gkorland, line 32: @sazzad16 why did you remove only some of the deprecated methods?

sazzad16: @gkorland I'm 💯 that we don't need `set` methods at least in this interface. (Reference: Connection.java)
But in case of get methods, there's little left to be 💯. (Reference: DefaultJedisSocketFactory.java)

sazzad16: @gkorland I'm 100% convinced that we don't need set methods at least in this interface.
But in case of get methods I'm about 90% convinced. 10% uncertainty is because of codes in DefaultJedisSocketFactory.createSocket().

sazzad16:

PR #2531, opened 2021-04-28
Remove constructors with only host param

sazzad16:


sazzad16:
@gkorland Keeping Client(host) while not allowing Jedis(host), time to time, would rise the question why we are not supporting it? IMHO, we would be better off not having that support.

PR #2532, opened 2021-04-28
Deprecate Remove constructors with only host param #2531

sazzad16:


src/main/java/redis/clients/jedis/JedisShardInfo.java
```diff
@@ -27,9 +27,17 @@
...
-  public JedisShardInfo(String host) {
+  /**
+   * @deprecated This constructor will not support a host string in future. It will accept only a
+   * uri string. {@link JedisURIHelper#isValid(java.net.URI)} can used before this. If this
+   * constructor was being used with a host, it can be replaced with
+   * {@link #JedisShardInfo(java.lang.String, int)} with the host and {@link Protocol#DEFAULT_PORT}.
+   * @param uriString
+   */
+  @Deprecated
+  public JedisShardInfo(String uriString) {
```
gkorland, line 38: @sazzad16 rethinking about it, it's not the best way do it, future users will still be able to compile/upgrade will only discover the behavior change in runntime.

sazzad16: @gkorland Isn't it the exact reason we're deprecating? So that users get warned and take precautions if they need to?

PR #2533, opened 2021-05-02
Unify clustered and non-clustered interfaces

sazzad16:
Closes #2021 
Closes #2400


src/main/java/redis/clients/jedis/BinaryJedisCluster.java
```diff
@@ -327,7 +326,7 @@ public String execute(Jedis connection) {
   }
 
   @Override
-  public Long expire(final byte[] key, final int seconds) {
+  public Long expire(final byte[] key, final long seconds) {
```
mina-asham, line 329: (Not for this PR specifically so not suggesting you deal with it here, but just thining about next major release and if we should change that)

Expire can never return null [according to the docs](https://redis.io/commands/expire), do we know why we have some return values using the primitive wrappers and some not?

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
```diff
@@ -2411,6 +2410,11 @@ public Long execute(Jedis connection) {
     }.runBinary(key);
   }
 
+  @Override
+  public String unwatch() {
+    throw new UnsupportedOperationException();
```
mina-asham, line 2415: Should we put a message here?

```suggestion
    throw new UnsupportedOperationException("UNWATCH is not supported in cluster-mode");
```

src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -4376,11 +4361,8 @@ public Long xlen(final String key) {
     }
   }
 
-  /**
-   * {@inheritDoc}
-   */
   @Override
-  public long xack(final String key, final String group, final StreamEntryID... ids) {
+  public Long xack(final String key, final String group, final StreamEntryID... ids) {
```
mina-asham, line 4365: So my view in Java is to always prefer primitives unless you need the `null` part, can this one return null? (same comment applies to all the other ones below)

mina-asham (changes requested):

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
sazzad16: @mina-asham My guesses:
- At the very core, Jedis uses `Object` (Connection.readProtocolWithCheckingBroken()). So that Object is being wrapped, `Long` came naturally.
- Long long ago, Redis wasn't clear and strict enough about respective responses. `Long` is a safer choice in such cases.

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
sazzad16: @mina-asham We absolutely should. This is somehow missed `TODO`.

sazzad16: @mina-asham We absolutely should. This is somehow missed `TODO` :)

src/main/java/redis/clients/jedis/Jedis.java
sazzad16: @mina-asham I was discussing this with @gkorland just today :)
Please review this PR ignoring this issue.

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
mina-asham: > Redis wasn't clear and strict enough about respective responses

That's what I thought might be the case, I think we should strengthen the API and return `long` instead of `Long` where applicable, also relates to the below comment but per your request I will ignore that part and review the rest for now :)

src/main/java/redis/clients/jedis/commands/BinaryJedisCommands.java
```diff
@@ -482,5 +493,9 @@ default String setex(byte[] key, int seconds, byte[] value) {
   @Deprecated
   List<StreamConsumersInfo> xinfoConsumers(byte[] key, byte[] group);
 
-  List<Object> xinfoConsumersBinary(byte[] key, byte[] group);
+  List<Object> xinfoConsumersBinary(byte[] key, byte[] group);;
```
mina-asham, line 496: Extra semicolon at the end of the line
```suggestion
  List<Object> xinfoConsumersBinary(byte[] key, byte[] group);
```

src/main/java/redis/clients/jedis/commands/MultiKeyBinaryJedisClusterCommands.java
```diff
@@ -1,144 +1,63 @@
...
 
-  Long georadiusStore(byte[] key, double longitude, double latitude, double radius, GeoUnit unit,
-      GeoRadiusParam param, GeoRadiusStoreParam storeParam);
+  /**
+   * @throws UnsupportedOperationException
+   */
+  @Override
+  default byte[] randomBinaryKey() {
+    throw new UnsupportedOperationException();
+  }
```
mina-asham, line 53: Let's add messages to all these unsupported exceptions

src/main/java/redis/clients/jedis/commands/MultiKeyJedisClusterCommands.java
```diff
@@ -1,164 +1,62 @@
...
 
-  List<Map.Entry<String, List<StreamEntry>>> xreadGroup(String groupname, String consumer,
-      XReadGroupParams xReadGroupParams, Map<String, StreamEntryID> streams);
+  /**
+   * @throws UnsupportedOperationException
+   */
+  @Override
+  default String randomKey() {
+    throw new UnsupportedOperationException();
+  }
```
mina-asham, line 53: Messages here too

mina-asham (changes requested):
Since you are attacking duplication here, the main difference between binary and non-binary interfaces is the usage of `String` vs `byte[]`, should we just have 1 interface and use generics here for the implementations?

sazzad16:
@mina-asham 

> Since you are attacking duplication here, the main difference between binary and non-binary interfaces is the usage of `String` vs `byte[]`, should we just have 1 interface and use generics here for the implementations?

I tried that. But Java doesn't allow to implement same interface twice, even with different generics. The situation arises because `BinaryJedis implements Interface<byte[]>` and `Jedis implements Interface<String>`; but `Jedis extends BinaryJedis` as well which is resulting in Jedis implementing Interface twice.

Also, in case of complex queries, Jedis tends to return a custom object while BinaryJedis tends to return raw object so that user can use it their own way. (It's not always true but @dengliming and I are trying to maintain it recently.) It won't be possible by if we use one interface. This is another matter of discussion.

src/main/java/redis/clients/jedis/commands/MultiKeyBinaryJedisClusterCommands.java
sazzad16: @mina-asham I'm temporarily marking the "unsupported exception message" comments are resolved just for easier browsing. Hope you won't mind :)

src/main/java/redis/clients/jedis/BinaryJedisCluster.java
sazzad16: > return long instead of Long where applicable

@mina-asham I tried but it seems to be too complicated to include in this PR. IMHO, we should do this in separate PR. I have created #2543 to remind us.

src/main/java/redis/clients/jedis/commands/BinaryJedisCommands.java
```diff
@@ -325,6 +325,11 @@ default String setex(byte[] key, int seconds, byte[] value) {
 
   byte[] echo(byte[] arg);
 
+  /**
+   * @deprecated This method will be removed from this interface. Use
+   * {@link AdvancedBinaryJedisCommands#move(byte[], int)}.
+   */
+  @Deprecated
   Long move(byte[] key, int dbIndex);
```
gkorland, line 333: Won't it mark the method as deprecated on BinaryJedis too?

src/main/java/redis/clients/jedis/commands/JedisCommands.java
```diff
@@ -361,6 +361,11 @@ default String setex(String key, int seconds, String value) {
 
   String echo(String string);
 
+  /**
+   * @deprecated This method will be removed from this interface. Use
+   * {@link AdvancedJedisCommands#move(java.lang.String, int)}.
+   */
+  @Deprecated
   Long move(String key, int dbIndex);
```
gkorland, line 369: Won't it mark the method as deprecated on Jedis too?

src/main/java/redis/clients/jedis/commands/BinaryJedisCommands.java
sazzad16: @gkorland No. The method will stay. It's just moved to Advanced... interface.

src/main/java/redis/clients/jedis/commands/JedisCommands.java
sazzad16: @gkorland No. The method will stay. It's just moved to Advanced... interface.

src/main/java/redis/clients/jedis/commands/BinaryJedisCommands.java
gkorland: I know it will stay, but BinaryJedis implement them both, which mean the IDEs might think it's deprecated no?

gkorland (approved):

src/main/java/redis/clients/jedis/commands/BinaryJedisCommands.java
sazzad16: @gkorland It shouldn't because there's at least one interface (Advanced...) that doesn't deprecate it.
Note: My IDE doesn't it as deprecated.

PR #2540, opened 2021-05-12
Introduce Endpoint

sazzad16:
and sample implementation for ZCOUNT

My plan is to extend this:
- exclusive range query to XPENDING (#8130)
- exclusive range query to XRANGE and XREVRANGE (#8072)

### DIFF
--- a/src/main/java/redis/clients/jedis/Client.java
+++ b/src/main/java/redis/clients/jedis/Client.java
@@ -10,6 +10,7 @@ import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLParameters;
 import javax.net.ssl.SSLSocketFactory;
 
+import redis.clients.jedis.args.Endpoint;
 import redis.clients.jedis.args.ListDirection;
 import redis.clients.jedis.commands.Commands;
 import redis.clients.jedis.params.*;
@@ -687,6 +688,11 @@ public class Client extends BinaryClient implements Commands {
   @Override
   public void zcount(final String key, final String min, final String max) {
     zcount(SafeEncoder.encode(key), SafeEncoder.encode(min), SafeEncoder.encode(max));
+  }
+
+  @Override
+  public void zcount(final String key, final Endpoint<Double> min, final Endpoint<Double> max) {
+    zcount(SafeEncoder.encode(key), min.getRaw(), max.getRaw());
   }
 
   @Override
--- a/src/main/java/redis/clients/jedis/Client.java
+++ b/src/main/java/redis/clients/jedis/Client.java
@@ -10,8 +10,8 @@ import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLParameters;
 import javax.net.ssl.SSLSocketFactory;
 
-import redis.clients.jedis.args.Endpoint;
 import redis.clients.jedis.args.ListDirection;
+import redis.clients.jedis.args.RangeEndpoint;
 import redis.clients.jedis.commands.Commands;
 import redis.clients.jedis.params.*;
 import redis.clients.jedis.util.SafeEncoder;
@@ -691,7 +691,7 @@ public class Client extends BinaryClient implements Commands {
   }
 
   @Override
-  public void zcount(final String key, final Endpoint<Double> min, final Endpoint<Double> max) {
+  public void zcount(final String key, final RangeEndpoint<Double> min, final RangeEndpoint<Double> max) {
     zcount(SafeEncoder.encode(key), min.getRaw(), max.getRaw());
   }
 
--- a/src/main/java/redis/clients/jedis/args/Endpoint.java
+++ b/src/main/java/redis/clients/jedis/args/Endpoint.java
@@ -10,10 +10,9 @@ public interface Endpoint<T> extends Rawable {
 
   public static abstract class AbstractEndpoint<T> implements Endpoint<T> {
 
-    private final byte[] raw;
-    private boolean exclusive;
+    private byte[] raw;
 
-    public AbstractEndpoint(T data, boolean exclusive) {
+    public AbstractEndpoint(T data) {
       if (data instanceof Rawable) {
         this.raw = ((Rawable) data).getRaw();
       } else if (data instanceof Integer) {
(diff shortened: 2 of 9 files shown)


src/main/java/redis/clients/jedis/args/Endpoint.java
```diff
@@ -0,0 +1,69 @@
...
+    return new AbstractEndpoint<Double>(value) {
+    };
+  }
+
+  public static Endpoint<Double> of(double value, boolean exclusive) {
+    return new AbstractEndpoint<Double>(value, exclusive) {
+    };
+  }
+
+}
```
mina-asham, line 69: 1. I think the name `Endpoint` is really confusing, how about `RangePoint` or `RangeLimit` or anything else (I suck at naming things :D)
2. I think the class is overcomplicated a bit here, can we reduce to just what we need? Specifically
   1. I don't see why we need an interface at all
   2. Not really sure we should support `Rawable` as a data type here, seems a bit risky given the scope of `Rawable` (e.g. I can make an `Endpoint<Command>` but that doesn't make much sense)
   2. The Abstract class is not really abstract here
   3. I don't see the need for the mutability using `inclusive()` and `exclusive()` functions
   4. We don't need a public constructor we can just expose static helper constructors

```suggestion
public final class RangePoint<T> implements Rawable {

  public static <T> RangePoint<T> exclusive(T data) {
    return new RangePoint<>(data, true);
  }

  public static <T> RangePoint<T> inclusive(T data) {
    return new RangePoint<>(data, false);
  }

  private final byte[] raw;

  private RangePoint(T data, boolean exclusive) {
    byte[] temp;

    if (data instanceof Integer) {
      temp = Protocol.toByteArray((Integer) data);
    } else if (data instanceof Double) {
      temp = Protocol.toByteArray((Double) data);
    } else {
      throw new IllegalArgumentException("RangePoint does not support data type: " + data.getClass().getName());
    }

    if (exclusive) {
      this.raw = new byte[1 + temp.length];
      this.raw[0] = '(';
      System.arraycopy(temp, 0, this.raw, 1, temp.length);
    } else {
      this.raw = temp;
    }
  }

  @Override
  public byte[] getRaw() {
    return this.raw;
  }
}
```

src/main/java/redis/clients/jedis/Client.java
```diff
@@ -689,6 +690,11 @@ public void zcount(final String key, final String min, final String max) {
     zcount(SafeEncoder.encode(key), SafeEncoder.encode(min), SafeEncoder.encode(max));
   }
 
+  @Override
+  public void zcount(final String key, final Endpoint<Double> min, final Endpoint<Double> max) {
+    zcount(SafeEncoder.encode(key), min.getRaw(), max.getRaw());
```
mina-asham, line 695: Does this work if I do `Endpoint.of(Double.POSITIVE_INFINITY, true)`? (Specifically I think this will translate it to `(+inf`, does this work in Redis given that `+inf` is a special value?)

mina-asham (changes requested):
I never noticed we are missing exclusive ranges (probably because I never needed to use it), but it seems like it's missing for other commands too beside the ones listed (e.g. `ZRANGEBYSCORE`), can we include these too in the scope of your update?

src/main/java/redis/clients/jedis/Client.java
sazzad16: @mina-asham works in redis-cli
```
ZCOUNT key (-inf (+inf
(integer) 0
```

sazzad16:
@mina-asham 

> I never noticed we are missing exclusive ranges

Some methods have `String/byte[]` variants.

    jedis.zcount("foo", "(0.01", "+inf");

> it seems like it's missing for other commands too beside the ones listed

Yes, it does. (That's why this a draft ;))

> ZRANGEBYSCORE

I was hoping not to do this for all [ZRANGE variants](https://redis.io/commands/zrange) as all other variants are deprecated. So I'm taking some time on how to do it sanely.

src/main/java/redis/clients/jedis/args/Endpoint.java
sazzad16: @mina-asham 

About naming: How about `RangeEndpoint`?

About implementation: I used interface so that users can have their own version. I added abstract class to ease that process.

sazzad16: @mina-asham 

About naming: How about `RangeEndpoint`?

About implementation: I used interface so that users can have their own version. I added sample (abstract) class and public constructor to ease that process.

About abstract class: True. We can rename it to `EndpointBase`/`EndpointImpl` or something like that.

About mutability: I think it'd be better to have that.

mina-asham: > About naming: How about RangeEndpoint?

I am okay with that one :)

> About implementation: I used interface so that users can have their own version. I added sample (abstract) class and public constructor to ease that process.

> About mutability: I think it'd be better to have that.

For both of these, yes in abstract it's nice to have functionality, but why do we really need it? Providing more than needed can the library to get bloated with functionality that no one really needs, making further enhancements updates harder due to having to support some backward comptability issues. If you have a strong use case (e.g. why would I need to mutate a range endpoint into exclusive/inclusive? why would I want to implement my own solution or what is missing from this one?)  to any of that I am all for it but I personally think we shouldn't do it otherwise.

src/main/java/redis/clients/jedis/args/Endpoint.java
```diff
@@ -0,0 +1,69 @@
...
+      }
+      this.exclusive = exclusive;
+    }
+
+    public AbstractEndpoint(T data) {
+      this(data, false);
+    }
+
+    @Override
+    public Endpoint<T> exclusive() {
```
gkorland, line 34: @sazzad16 why won't you update the raw on this call? instead of building each time on `getRaw()`

gkorland (changes requested):

src/main/java/redis/clients/jedis/args/Endpoint.java
sazzad16: @gkorland That's a reasonable request. Is this your only concern?

sazzad16:
@gkorland It'd be also be helpful if you could comment on Mina's change requests as well.

src/main/java/redis/clients/jedis/args/Endpoint.java
gkorland: @sazzad16 I think @mina-asham has a point here, what use case did you image going with the abstract class?

sazzad16: @gkorland Customizability; e.g. for modules (in case something comes up) or user's custom implementations.

src/main/java/redis/clients/jedis/args/Endpoint.java
gkorland: yes, let's fix and merge this PR

sazzad16: @gkorland Please take a look at #2546 to see my current approach to extend this PR.

PR #2545, opened 2021-05-13
Prepare void return while SHUTDOWN

sazzad16:


src/main/java/redis/clients/jedis/BinaryJedis.java
```diff
@@ -3470,18 +3471,21 @@ public Long lastsave() {
...
+  @Deprecated
+  public String shutdown() throws JedisException {
     client.shutdown();
-    String status;
     try {
-      status = client.getStatusCodeReply();
-    } catch (JedisException ex) {
-      status = null;
+      String status = client.getStatusCodeReply();
+      throw new JedisException(status);
```
gkorland, line 3484: @sazzad16 won't it always going to throw a JedisException?

gkorland, line 3484: @sazzad16 I get that you want to deprecate it, but why do you change the logic?

sazzad16: @gkorland I've added throwing exception when there's an error. Other than that, the logic is same. E.g. `null` is returned when everything goes well both in existing code and in new code.

sazzad16:
Closing in favor of #2613

PR #2546, opened 2021-05-14
Introduce Endpoint and new StreamEntryID

sazzad16:
Extended version of #2540 

Resolves https://github.com/redis/jedis/projects/1#card-55852671 
Resolves https://github.com/redis/jedis/projects/1#card-55852669

### DIFF
--- a/src/main/java/redis/clients/jedis/args/StreamEntryID.java
+++ b/src/main/java/redis/clients/jedis/args/StreamEntryID.java
@@ -1,0 +1,8 @@
+package redis.clients.jedis.args;
+
+public interface StreamEntryID extends Rawable {
+
+  long milliseconds();
+
+  long sequence();
+}
--- a/src/main/java/redis/clients/jedis/params/XPendingParams.java
+++ b/src/main/java/redis/clients/jedis/params/XPendingParams.java
@@ -7,6 +7,8 @@ import java.util.List;
 
 import redis.clients.jedis.Protocol;
 import redis.clients.jedis.StreamEntryID;
+import redis.clients.jedis.args.RangeEndpoint;
+import redis.clients.jedis.args.StreamEntryIdFactory;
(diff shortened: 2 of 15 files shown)


src/main/java/redis/clients/jedis/params/XPendingParams.java
```diff
@@ -30,16 +38,34 @@ public XPendingParams idle(long idle) {
     return this;
   }
 
+  /**
+   * @deprecated Use {@link #start(redis.clients.jedis.args.RangeEndpoint)}.
+   */
+  @Deprecated
   public XPendingParams start(StreamEntryID start) {
     this.start = start;
```
gkorland, line 46:
```suggestion
    this.startEndpoint = RangeEndpoint.of(start);
```

Why do you need to deprecate it? it feels like in most cases it's easier to use it

src/main/java/redis/clients/jedis/params/XPendingParams.java
```diff
@@ -52,6 +78,13 @@ public XPendingParams consumer(String consumer) {
 
   @Override
   public byte[][] getByteParams() {
+    if (start != null && startEndpoint == null) {
+      startEndpoint = RangeEndpoint.of(RangeEndpoint.convert(start));
```
gkorland, line 82: If anyway you're going to convert `start` to `startEndpoint` why not do it in the setter?

src/main/java/redis/clients/jedis/args/StreamEntryID.java
```diff
@@ -0,0 +1,8 @@
+package redis.clients.jedis.args;
```
gkorland, line 1: @sazzad16 what about the original StreamEntryID class?

sazzad16: @gkorland Let's discuss...

PR #2548, opened 2021-05-18
Deprecate methods with 'byte[] keyCount' param

sazzad16:


gkorland:
@sazzad16 why do you want to deprecate these methods?

sazzad16:
@gkorland Because these are nonsensical methods. A keyCount can be easily and should be represented by an `int`. There are existing methods for that purpose. So we don't need these methods anymore.

gkorland (approved):

PR #2555, opened 2021-05-20
Allow calling PUBSUB CHANNELS without a pattern

sabbey37:
According to the Redis docs at https://redis.io/commands/PUBSUB#pubsub-channels-pattern PUBSUB CHANNELS should list all channels if no pattern is specified.  Currently Jedis does not support calling `pubSubChannels` without an argument.

### DIFF
--- a/src/test/java/redis/clients/jedis/tests/commands/PublishSubscribeCommandsTest.java
+++ b/src/test/java/redis/clients/jedis/tests/commands/PublishSubscribeCommandsTest.java
@@ -78,8 +78,8 @@ public class PublishSubscribeCommandsTest extends JedisCommandTestBase {
           Jedis otherJedis = createJedis();
           List<String> activeChannels = otherJedis.pubsubChannels();
           // Since we are utilizing sentinel for the tests, there is an additional
-          // '__sentinel__:hello'
-          // channel that has subscribers and will be returned from PUBSUB CHANNELS.
+          // '__sentinel__:hello' channel that has subscribers and will be returned from PUBSUB
+          // CHANNELS.
           assertTrue(activeChannels.containsAll(expectedActiveChannels));
           unsubscribe();
         }
--- a/src/test/java/redis/clients/jedis/tests/commands/PublishSubscribeCommandsTest.java
+++ b/src/test/java/redis/clients/jedis/tests/commands/PublishSubscribeCommandsTest.java
@@ -85,7 +85,7 @@ public class PublishSubscribeCommandsTest extends JedisCommandTestBase {
   }
 
   @Test
-  public void pubSubChannelsWithArgument() {
+  public void pubSubChannelsWithPattern() {
     final List<String> expectedActiveChannels = Arrays
         .asList("testchan1", "testchan2", "testchan3");
     jedis.subscribe(new JedisPubSub() {
--- a/src/test/java/redis/clients/jedis/tests/commands/PublishSubscribeCommandsTest.java
+++ b/src/test/java/redis/clients/jedis/tests/commands/PublishSubscribeCommandsTest.java
@@ -65,10 +65,8 @@ public class PublishSubscribeCommandsTest extends JedisCommandTestBase {
 
   @Test
   public void pubSubChannels() {
-    // Since we are utilizing sentinel for the tests, there is an additional '__sentinel__:hello'
-    // channel that has subscribers.
-    final List<String> expectedActiveChannels = Arrays.asList("testchan1", "testchan2",
-      "testchan3", "__sentinel__:hello");
+    final List<String> expectedActiveChannels = Arrays
+        .asList("testchan1", "testchan2", "testchan3");
     jedis.subscribe(new JedisPubSub() {
       private int count = 0;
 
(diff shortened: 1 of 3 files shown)


src/test/java/redis/clients/jedis/tests/commands/PublishSubscribeCommandsTest.java
```diff
@@ -65,6 +65,27 @@ public void onUnsubscribe(String channel, int subscribedChannels) {
 
   @Test
   public void pubSubChannels() {
+    final List<String> expectedActiveChannels = Arrays.asList("testchan1", "testchan2",
+      "testchan3", "__sentinel__:hello");
```
sabbey37, line 69: Since we are utilizing sentinel for the tests, there is an additional `__sentinel__:hello` channel that has subscribers:
```
Every Sentinel is subscribed to the Pub/Sub channel __sentinel__:hello of every master and replica, looking for unknown sentinels.
```
https://redis.io/topics/sentinel

dengliming (dismissed):

src/test/java/redis/clients/jedis/tests/commands/PublishSubscribeCommandsTest.java
sazzad16: @sabbey37 Nice observation. Please add this in code comment as well.

src/test/java/redis/clients/jedis/tests/commands/PublishSubscribeCommandsTest.java
```diff
@@ -65,6 +65,27 @@ public void onUnsubscribe(String channel, int subscribedChannels) {
...
+      private int count = 0;
+
+      @Override
+      public void onSubscribe(String channel, int subscribedChannels) {
+        count++;
+        // All channels are subscribed
+        if (count == 3) {
+          Jedis otherJedis = createJedis();
+          List<String> activeChannels = otherJedis.pubsubChannels();
+          assertTrue(expectedActiveChannels.containsAll(activeChannels));
```
sazzad16, line 80: Add the comment in the following test as well.

```suggestion
          List<String> activeChannels = otherJedis.pubsubChannels();
          // Can't use 'assertEquals'. Since we are utilizing sentinel for the tests, there is an additional '__sentinel__:hello' channel that has subscribers.
          assertTrue(expectedActiveChannels.containsAll(activeChannels));
```

sazzad16 (changes requested):

src/test/java/redis/clients/jedis/tests/commands/PublishSubscribeCommandsTest.java
sabbey37: I think we'd have to use `containsAll` either way since the list of channels may come back in a different order than the expected list.  We also use `containsAll` in the `pubSubChannelsWithPattern` test below this one.  I added the comment, but put it right above the assignment of the `expectedActiveChannels` variable.  

Alternatively, we could use `contains` instead of `containsAll` and remove the sentinel channel from the expected list.

Let me know what you think!

sabbey37:
Just wanted to clarify that I did not intentionally dismiss @dengliming's review... I think it happened automatically when I pushed an additional commit.

src/test/java/redis/clients/jedis/tests/commands/PublishSubscribeCommandsTest.java
```diff
@@ -65,6 +65,29 @@ public void onUnsubscribe(String channel, int subscribedChannels) {
 
   @Test
   public void pubSubChannels() {
+    // Since we are utilizing sentinel for the tests, there is an additional '__sentinel__:hello'
+    // channel that has subscribers.
+    final List<String> expectedActiveChannels = Arrays.asList("testchan1", "testchan2",
+      "testchan3", "__sentinel__:hello");
```
sazzad16, line 71:
```suggestion
    final List<String> expectedActiveChannels = Arrays.asList("testchan1", "testchan2",
      "testchan3");
```

src/test/java/redis/clients/jedis/tests/commands/PublishSubscribeCommandsTest.java
sazzad16: Between these two, I prefer it as it is.

sazzad16 (changes requested):

sazzad16:
> Just wanted to clarify that I did not intentionally dismiss ...

@sabbey37 Don't worry about it :)

src/test/java/redis/clients/jedis/tests/commands/PublishSubscribeCommandsTest.java
sabbey37: Sounds good, I will leave it as it is.

src/test/java/redis/clients/jedis/tests/commands/PublishSubscribeCommandsTest.java
sabbey37: Not sure if you wanted me to make a change here since you said you prefer it as it is?  If we remove the sentinel channel and still check the list via `containsAll`, the assert will fail.

sabbey37: Ok, that is what we currently do, so I will leave everything as is.  Let me know if you had something different in mind.

sazzad16: No. Look carefully, we're currently currently doing `assertTrue(expectedActiveChannels.containsAll(activeChannels));`. I'm suggesting a reverse check.

sabbey37: Oh my goodness.  I am so sorry.  Yes, I can change that. Should we remove the comment about the sentinel channel altogether or change it to explain why we're doing the reverse check?

sabbey37: I've updated it. Let me know what you think.

sazzad16 (approved):

dengliming (approved):

PR #2391, opened 2021-02-28
Set the GenericObjectPoolConfig type

gkorland:

PR #2557, opened 2021-05-25
Add ShardedJedisPoolConfig

YNedderhoff:
Before, in our project we used to initialise `ShardedJedisPool` with `JedisPoolConfig`, which has become impossible with 3.6.0. This PR fixes that by adding `ShardedJedisPoolConfig`. This became necessary after `Jedis` was added as the type for `GenericObjectPoolConfig` in `JedisPoolConfig`:
 https://github.com/redis/jedis/pull/2391/files#diff-ac2ef340a5aeeb15580f83c0849cf4bb80d86e041a4f3a8797f891af12a8cd4dR5

### DIFF
--- a/src/main/java/redis/clients/jedis/ShardedJedisPoolConfig.java
+++ b/src/main/java/redis/clients/jedis/ShardedJedisPoolConfig.java
@@ -1,0 +1,13 @@
+package redis.clients.jedis;
+
+import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
+
+public class ShardedJedisPoolConfig extends GenericObjectPoolConfig<ShardedJedis> {
+  public ShardedJedisPoolConfig() {
+    // defaults to make your life with connection pool easier :)
+    setTestWhileIdle(true);
+    setMinEvictableIdleTimeMillis(60000);
+    setTimeBetweenEvictionRunsMillis(30000);
+    setNumTestsPerEvictionRun(-1);
+  }
+}
--- a/src/test/java/redis/clients/jedis/tests/ShardedJedisPoolTest.java
+++ b/src/test/java/redis/clients/jedis/tests/ShardedJedisPoolTest.java
@@ -18,6 +18,7 @@ import redis.clients.jedis.Jedis;
 import redis.clients.jedis.JedisShardInfo;
 import redis.clients.jedis.ShardedJedis;
 import redis.clients.jedis.ShardedJedisPool;
+import redis.clients.jedis.ShardedJedisPoolConfig;
 import redis.clients.jedis.exceptions.JedisExhaustedPoolException;
 
 public class ShardedJedisPoolTest {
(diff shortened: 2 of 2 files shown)


dengliming:
I don't know if this change makes sense. @sazzad16

sazzad16:
@dengliming I'm not sure either.

May be we could've done #2391 better? Like `JedisPoolConfig<T> extends GenericObjectPoolConfig<T>` instead of `JedisPoolConfig extends GenericObjectPoolConfig<Jedis>`?

YNedderhoff:
Hey, just following up on this one :)

sazzad16:
@YNedderhoff We are really confused here... :|

YNedderhoff:
What about what you suggested in that last comment?

sazzad16:
A PR would've helped for further thinking. But as you see, even myself couldn't manage enough time for my own idea.

yangbodong22011:
> @dengliming I'm not sure either.
> 
> May be we could've done #2391 better? Like `JedisPoolConfig<T> extends GenericObjectPoolConfig<T>` instead of `JedisPoolConfig extends GenericObjectPoolConfig<Jedis>`?

@sazzad16 We need to reconsider adopting what you said, or revert it back (`public class JedisPoolConfig extends GenericObjectPoolConfig`)

p.s. My old code has a compilation error in master branch, because of type conversion (need pass GenericObjectPoolConfig<Connection> poolConfig)
```
private static final JedisPoolConfig config = new JedisPoolConfig();
JedisCluster jc = new JedisCluster(jedisClusterNode, connectionTimeout, soTimeout,
            maxAttempts, "user", password, "clientName", config);
```

constructor is:
```
  public JedisCluster(Set<HostAndPort> jedisClusterNode, int connectionTimeout,
      int soTimeout, int maxAttempts, String password, String clientName,
      GenericObjectPoolConfig<Connection> poolConfig) {
    this(jedisClusterNode, connectionTimeout, soTimeout, maxAttempts, null, password, clientName,
        poolConfig);
  }
```

sazzad16:
@yangbodong22011 

> ... revert it back ...

We can't. The decision to change GenericObjectPoolConfig to be parameterized was common-pool2's and we just had to adapt to. What I meant was that we could've adapted in a different (better) way.

> ... compilation error in master branch ...

That's a breaking change that I had to go with.

I tried to avoid breaking it by converting `GenericObjectPoolConfig<Jedis>` to `GenericObjectPoolConfig<Connection>` internally. But one of our test was getting stuck. So I had two options: i) use it; knowing that our code would get stuck in a special case. ii) break it, in a major release but our code would be functioning at least on par with Jedis 3.x. I went for the later.

If anyone is using `GenericObjectPoolConfig<Jedis>`, he'd have to switch to `GenericObjectPoolConfig<Connection>`. 
If anyone is using JedisPoolConfig, I've added ConnectionPoolConfig. Hopefully that'll compensate to some extent.

yangbodong22011:
> If anyone is using JedisPoolConfig, I've added ConnectionPoolConfig. Hopefully that'll compensate to some extent.

👍，I can accept this approach, just change `JedisPoolConfig` to `ConnectionPoolConfig`.

YNedderhoff:
Alright 👌 Is there anything to be done about this, maybe documentation or so that makes it easier to find? I could close this PR now, but I don't want to bury the information.

sazzad16:
Closed by #3009

PR #2565, opened 2021-06-14
Support TYPE option in SCAN command

### LINKED ISSUE #2564: Support TYPE option in SCAN command

sazzad16:
Resolves #2564


src/main/java/redis/clients/jedis/BinaryJedis.java
```diff
@@ -4402,10 +4402,12 @@ public long pfcount(final byte[]... keys) {
...
   }
 
+  @Override
   public ScanResult<byte[]> scan(final byte[] cursor) {
     return scan(cursor, new ScanParams());
   }
 
+  @Override
   public ScanResult<byte[]> scan(final byte[] cursor, final ScanParams params) {
     checkIsInMultiOrPipeline();
```
gkorland, line 4412: @sazzad16 shouldn't this method just call the new method with null on Type?

gkorland (changes requested):

src/main/java/redis/clients/jedis/BinaryJedis.java
sazzad16: @gkorland done

dengliming (approved):

src/main/java/redis/clients/jedis/BinaryClient.java
```diff
@@ -1405,9 +1406,17 @@ public void hincrByFloat(final byte[] key, final byte[] field, final double incr
   }
 
   public void scan(final byte[] cursor, final ScanParams params) {
+    scan(cursor, params, (byte[]) null);
+  }
+
+  public void scan(final byte[] cursor, final ScanParams params, final byte[] type) {
```
yangbodong22011, line 1412: Why not put `TYPE` in ScanParams? Just like `MATCH` and `COUNT`

sazzad16: @yangbodong22011 TYPE is supported only in SCAN command but not in the other variants. But ScanParams is used in all the variants.

yangbodong22011: @sazzad16 ohh, i got it.

yangbodong22011 (approved):

PR #2582, opened 2021-07-13
Test Sentinels with ACL

sazzad16:


src/test/java/redis/clients/jedis/tests/JedisSentinelPoolWithCompleteCredentialsTest.java
```diff
@@ -68,67 +60,92 @@ public void repeatedSentinelPoolInitialization() {
...
-      jedis.auth("default", "foobared");
       jedis.set("foo", "bar");
       assertEquals("bar", jedis.get("foo"));
     }
     pool.close();
     assertTrue(pool.isClosed());
   }
 
-  @Test
-  public void ensureSafeTwiceFailover() throws InterruptedException {
```
sazzad16, line 111: We need two more (four in total) Redis+Sentinel instances just to support this single test. But exact test is also present in JedisSentinelPoolTest and JedisSentinelTest. So I took the chance of removing this test. Also note that only this test is removed in this entire PR.

gkorland (approved):

Issue #2583, opened 2021-07-13
please implements stralgo method!


startjava:
please implements stralgo method!

yangbodong22011:
duplicate with https://github.com/redis/jedis/pull/2328

startjava:
@yangbodong22011     this url 2328 , show ljwlc committed on 25 Dec 2020,,,but now now now not release???  :( slow!!!

PR #2328, opened 2020-12-24
Add support for Redis 6.0 command `stralgo`

ljwlc:
- Add strlago command support for Redis 6.0 server

PR #2586, opened 2021-07-14
Support stralgo command

yangbodong22011:
Handle https://github.com/redis/jedis/pull/2328 and Resolves https://github.com/redis/jedis/issues/2583


src/main/java/redis/clients/jedis/BinaryClient.java
```diff
@@ -1093,6 +1094,10 @@ public void strlen(final byte[] key) {
     sendCommand(STRLEN, key);
   }
 
+  public void strAlgoLcs(StrAlgo algorithm, StrAlgoParams params) {
```
dengliming: What about `public void strAlgoLcs(StrAlgoParams params) {` and put `StrAlgo algorithm,` to `StrAlgoParams `? : )

yangbodong22011: @dengliming thanks, agree, I will modify.

sazzad16 (commented):
@yangbodong22011 Sorry, for participating late. But there were several aspects that I wanted to address. So a short comment or review was not an option.

1. If the method name is `strAlgoLcs`, we already know that the algorithm/subcommand is LCS. Adding `StrAlgo algorithm` param is just redundant.
2. Current `StrAlgoParams` works for LCS but if it may not be convenient in future (if any) algorithms for both users and developers. Naming it `(StrAlgo)LCSParams` seems a safer choice.
3. For same reason, `StringMatchResult` may be better to contain `LCS`. Also put that in `resp` package.
4. How about two different methods `strAlgoLcs` and `strAlgoLcsKeys` (and the Params without `keys` option); so that `strAlgoLcsKeys` can be used in Cluster as well?

yangbodong22011:
> @yangbodong22011 Sorry, for participating late. But there were several aspects that I wanted to address. So a short comment or review was not an option.
> 
> 1. If the method name is `strAlgoLcs`, we already know that the algorithm/subcommand is LCS. Adding `StrAlgo algorithm` param is just redundant.
> 2. Current `StrAlgoParams` works for LCS but it may not be convenient in future (if any) algorithms for both users and developers. Naming it `(StrAlgo)LCSParams` seems a safer choice.
> 3. For same reason, `StringMatchResult` may be better to contain `LCS`. Also put that in `resp` package.
> 4. How about two different methods `strAlgoLcs` and `strAlgoLcsKeys` (and the Params without `keys` option); so that `strAlgoLcsKeys` can be used in Cluster as well?

@sazzad16 @dengliming thanks, all done, pls review again when you have time.

src/main/java/redis/clients/jedis/params/StrAlgoLCSParams.java
```diff
@@ -0,0 +1,98 @@
...
+    public StrAlgoLCSParams minMatchLen(long minMatchLen) {
+        addParam(MINMATCHLEN, minMatchLen);
+        return this;
+    }
+
+    /**
+     * Specify the strings to be passed.
+     * @return StrAlgoParams
+     */
+    public StrAlgoLCSParams strings(String... strings) {
```
sazzad16, line 66: Instead of this, strAlgoLcs should have two more (string) params.

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/params/StrAlgoLCSParams.java
yangbodong22011: done

sazzad16: @yangbodong22011 That's not what I meant. It's my mistake, sorry.

`strAlgoLcs` "method" should have two more string params.

yangbodong22011: @sazzad16 you mean this?
```
public LCSMatchResult strAlgoLCS(StrAlgoLCSParams params, String stringA, String stringB);
```

sazzad16: @yangbodong22011 yes

yangbodong22011: @sazzad16 we can pass `String` or `byte[]` args through the following methods:
```
public StrAlgoLCSParams strings(String stringA, String stringB);
public StrAlgoLCSParams strings(byte[] stringA, byte[] stringB); (need to be added)
```

why we need this:
```
public LCSMatchResult strAlgoLCS(StrAlgoLCSParams params, String stringA, String stringB);
```

and when user call `strAlgoLCS(params.strings(a,b), c, d)`, will pass four strings to redis.

sazzad16: @yangbodong22011 I meant removing `strings` option from Params class, just like removing keys option that you did earlier.

yangbodong22011: @sazzad16 ohh, i got it and agree, let me update it.

yangbodong22011: @sazzad16 done, now we have two commands: `strAlgoLCSKeys` and `strAlgoLCSStrings`.

src/main/java/redis/clients/jedis/params/StrAlgoLCSParams.java
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * Specify the minimum match length.
+     * @return StrAlgoParams
+     */
+    public StrAlgoLCSParams minMatchLen(long minMatchLen) {
+        addParam(MINMATCHLEN, minMatchLen);
+        return this;
+    }
+
+    public byte[][] getByteParams(byte[]... args) {
```
sazzad16, line 61: 2. Instead of args: `keyword, a, b`.

sazzad16 (commented):
Almost LGTM!

Sorry to annoy you, but I'm going to nitpick about two things.

1. Order of method parameters should be: `keyA/strA, keyB/strB, params`.

src/main/java/redis/clients/jedis/params/StrAlgoLCSParams.java
yangbodong22011: @sazzad16 It doesn't bother me. Your suggestions is good for code style and strict parameters. I agree and have modified the code.

sazzad16 (commented):
LGTM!

src/main/java/redis/clients/jedis/resps/LCSMatchResult.java
```diff
@@ -0,0 +1,127 @@
+/*
+ * Copyright 2011-2021 the original author or authors.
```
sazzad16, line 2: @dengliming As you are the original author of this (or similar) class, could you please give us your permission to remove the license from here?

src/main/java/redis/clients/jedis/resps/LCSMatchResult.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.List;
+
+/**
+ * Result for STRALGO LCS command.
+ *
+ * @author dengliming
+ * @author Mark Paluch
+ * @author yangbodong
+ * @since 6.0
+ */
```
sazzad16, line 29:
```suggestion
package redis.clients.jedis.resps;

import java.util.Collections;
import java.util.List;

/**
 * Result for STRALGO LCS command.
 */
```

sazzad16 (approved):

dengliming (approved):

PR #2598, opened 2021-07-19
Support EXECABORT Error

sazzad16:


src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -21,6 +21,7 @@
   private static final String NOSCRIPT_PREFIX = "NOSCRIPT ";
   private static final String WRONGPASS_PREFIX = "WRONGPASS";
   private static final String NOPERM_PREFIX = "NOPERM";
+  private static final String EXECABORT_PREFIX = "EXECABORT ";
```
yangbodong22011, line 24: There is a space after `EXECABORT`.

sazzad16: @yangbodong22011 it's deliberate

yangbodong22011: @sazzad16 ohh, i thought it was accidentally written, can you explain why spaces are needed? I am a little puzzled.

sazzad16: @yangbodong22011 just being super safe due to past experience. I don't specifically remember what it was. Probably at some point, Redis had BUSY"SOMETHING" (without space) error which would match BUSY prefix. I'm just trying to avoid that here.

yangbodong22011: @sazzad16 I got it, thank you very much.

yangbodong22011 (approved):

gkorland (approved):

PR #2600, opened 2021-07-19
Support WAIT in Pipeline

sazzad16:


src/main/java/redis/clients/jedis/Pipeline.java
```diff
@@ -166,4 +166,9 @@ public void close() {
     return getResponse(BuilderFactory.STRING);
   }
 
+  public Response<Long> waitReplicas(int replicas, long timeout) {
```
yangbodong22011, line 169: Do we need to declare this method in `RedisPipeline.java` or `BinaryRedisPipeline.java`? And add `@Override` to this method.

gkorland (approved):

PR #2612, opened 2021-08-02
Allow to override sharded jedis pooled object factory

bokshitsky:
Hi,

I want to be able to override default ShardedJedisFactory from ShardedJedisPool.

main reason: current validateObject() method uses PING-PONG mechanism that looks quite expensive when setTestOnReturn(true)/setTestOnBorrow(true) are set and a lot of shards are used.

I want to try something like:
```
@Override
public boolean validateObject(PooledObject<ShardedJedis> pooledShardedJedis) {
  try {
    ShardedJedis jedis = pooledShardedJedis.getObject();
    for (Jedis shard : jedis.getAllShards()) {
      if (shard.isBroken() || !shard.isConnected()) {
        return false;
      }
    }
    return true;
  } catch (RuntimeException ex) {
    return false;
  }
}
```
for our projects at https://github.com/hhru

But I am afraid what merging same code in current implementation is quite dangerous due to possible backward compatibility broke.

So I add required constructor to ShardedJedisPool and make ShardedJedisFactory public to allow in-place override. (which is tested by new unit-test)

Also fixed some code-smells (catch generic Exception instead of just Runtime, make slowlog test more stable by truncating initial state)

Feel free to ask any questions.
Excuse me for my english - I am not native speaker :)

### DIFF
--- a/src/main/java/redis/clients/jedis/ShardedJedisPool.java
+++ b/src/main/java/redis/clients/jedis/ShardedJedisPool.java
@@ -34,7 +34,12 @@ public class ShardedJedisPool extends Pool<ShardedJedis> {
 
   public ShardedJedisPool(final GenericObjectPoolConfig<ShardedJedis> poolConfig,
       List<JedisShardInfo> shards, Hashing algo, Pattern keyTagPattern) {
-    super(poolConfig, new ShardedJedisFactory(shards, algo, keyTagPattern));
+    this(poolConfig, new ShardedJedisFactory(shards, algo, keyTagPattern));
+  }
+
+  public ShardedJedisPool(final GenericObjectPoolConfig<ShardedJedis> poolConfig,
+      PooledObjectFactory<ShardedJedis> shardedJedisPooledObjectFactory) {
+    super(poolConfig, shardedJedisPooledObjectFactory);
   }
 
   @Override
@@ -55,7 +60,7 @@ public class ShardedJedisPool extends Pool<ShardedJedis> {
   /**
    * PoolableObjectFactory custom impl.
    */
-  private static class ShardedJedisFactory implements PooledObjectFactory<ShardedJedis> {
+  public static class ShardedJedisFactory implements PooledObjectFactory<ShardedJedis> {
 
     private final List<JedisShardInfo> shards;
     private final Hashing algo;
--- a/src/test/java/redis/clients/jedis/tests/ShardedJedisPoolTest.java
+++ b/src/test/java/redis/clients/jedis/tests/ShardedJedisPoolTest.java
@@ -8,7 +8,9 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
 
+import org.apache.commons.pool2.PooledObject;
 import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
 import org.junit.Before;
 import org.junit.Test;
@@ -19,6 +21,7 @@ import redis.clients.jedis.JedisShardInfo;
 import redis.clients.jedis.ShardedJedis;
 import redis.clients.jedis.ShardedJedisPool;
 import redis.clients.jedis.exceptions.JedisExhaustedPoolException;
+import redis.clients.jedis.util.Hashing;
 
 public class ShardedJedisPoolTest {
   private static HostAndPort redis1 = HostAndPortUtil.getRedisServers().get(0);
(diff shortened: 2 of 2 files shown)


src/test/java/redis/clients/jedis/tests/commands/SlowlogCommandsTest.java
```diff
@@ -36,6 +36,7 @@ public void tearDown() throws Exception {
 
   @Test
   public void slowlog() {
+    jedis.slowlogReset();
```
sazzad16, line 39: If you think this is necessary, do this in a separate PR.

src/test/java/redis/clients/jedis/tests/ShardedJedisPoolOverrideObjectFactoryTest.java
```diff
@@ -0,0 +1,63 @@
...
+import redis.clients.jedis.*;
+import redis.clients.jedis.util.Hashing;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static org.junit.Assert.assertEquals;
+
+public class ShardedJedisPoolOverrideObjectFactoryTest {
```
sazzad16, line 16: I don't see any necessity of creating a separate test class. Add the test in the existing one.

src/main/java/redis/clients/jedis/ShardedJedisPool.java
```diff
@@ -83,12 +88,12 @@ public void destroyObject(PooledObject<ShardedJedis> pooledShardedJedis) throws
             if (!jedis.isBroken()) {
               jedis.quit();
             }
-          } catch (Exception e) {
+          } catch (RuntimeException e) {
```
sazzad16, line 91: Exception was caught so that every node gets a chance to be destroyed even though there is a bug. This is an operation that MUST be done (as much as possible).

src/main/java/redis/clients/jedis/ShardedJedisPool.java
```diff
@@ -83,12 +88,12 @@ public void destroyObject(PooledObject<ShardedJedis> pooledShardedJedis) throws
...
             logger.warn("Error while QUIT", e);
           }
           try {
             jedis.disconnect();
-          } catch (Exception e) {
+          } catch (RuntimeException e) {
```
sazzad16, line 96: -do-

src/main/java/redis/clients/jedis/ShardedJedisPool.java
```diff
@@ -105,7 +110,7 @@ public boolean validateObject(PooledObject<ShardedJedis> pooledShardedJedis) {
           }
         }
         return true;
-      } catch (Exception ex) {
+      } catch (RuntimeException ex) {
```
sazzad16, line 113: -do-

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/ShardedJedisPool.java
bokshitsky: Right now checked `Exception` is not thrown from the code inside - so my proposal is backward-compatible 
(I change Exception -> RuntimeException, not Throwable -> RutimeExceptio)

but catching generic checked `Exception` is quite "dangerous" if internal contract will change and something like `Interrupted Exception` will be thrown: current code will not handle it properly - but someone who make changes won't get compilation error and may miss this (as a result - thread won't be interrupted)

bokshitsky: > Exception was caught so that every node gets a chance to be destroyed even though there is a bug. This is an operation that MUST be done (as much as possible).

Probably I don't understood something what you mean, 
may you please explain some details how Exception->RuntimeException breaks current execution logic?

src/test/java/redis/clients/jedis/tests/ShardedJedisPoolOverrideObjectFactoryTest.java
bokshitsky: 👍

src/test/java/redis/clients/jedis/tests/commands/SlowlogCommandsTest.java
bokshitsky: 👍

bokshitsky:
> If you think this is necessary, do this in a separate PR.

> I don't see any necessity of creating a separate test class. Add the test in the existing one.

I am quite newbie in contributing - how should I fix these comments?
1 - push some commits to current PR and squash whem in one later?
2 - or just remove/change current commits locally and then force-push them?
 
 I like way 2)

sazzad16:
@bokshitsky 
1. you can push more commits. generally we (maintainers) squash them.
2. force push is okay as well (specially in smaller PRs).

src/test/java/redis/clients/jedis/tests/commands/SlowlogCommandsTest.java
bokshitsky: done - removed this change from PR

src/test/java/redis/clients/jedis/tests/ShardedJedisPoolOverrideObjectFactoryTest.java
bokshitsky: done - moved changes to current test class

src/main/java/redis/clients/jedis/ShardedJedisPool.java
sazzad16: @bokshitsky At lease separate these changes in another PR with explanation.

sazzad16 (changes requested):

bokshitsky:
> At lease separate these changes in another PR with explanation.

done, removed this commit

sazzad16 (approved):

yangbodong22011 (approved):

bokshitsky:
Thank you for merging my branch.

could you please tell something about jedis release policy - when the next release is planned?

I ask because if it will be soon I will wait for version in mvncentral.
If the release is not planned soon i will build version myself locally and put it to company local nexus.

sazzad16:
@bokshitsky There is no recent release plan.

PR #2615, opened 2021-08-04
do not catch checked Exception if is not thrown

bokshitsky:
non significant update - tried to put it in previous pr https://github.com/redis/jedis/pull/2612  but you suggested to put in separate 

catching checked Exception instead of RuntimeException should be a avoided: it is dangerous due to code inside `try` block may add new checked exception with special handling required, but author can miss it due to no compile error (especially dangerous for InterruptedException)
 
commit modifies all production files with this "code smell", do not touch tests files due to less dangerous consequences of possible error in them.


sazzad16:
@bokshitsky any reason for closing this?

ps: tests are now passing.

bokshitsky:
Ouch.

Had problems with local tests run - sentinel tests were falling - I though what my branch was not the reason.
But when I saw CI test falling I decided that I really broke something and was going to fix latter:)

Thanks, reopen the PR

bokshitsky:
@sazzad16 what do you think about these changes?

image

sazzad16:
@bokshitsky My comment about these changes in #2612 still stands. (PS: I get your concern as well.) So I'm just waiting for other reviewers to participate.

dengliming (approved):

Issue #2550, opened 2021-05-19
JedisClusterInfoCache: refresh the nodes map


stillerrr:
### Expected behavior

Jedis would refresh the elements of nodes map  when the cluster nodes changed

![image](https://user-images.githubusercontent.com/35724820/118773816-be875c00-b8b7-11eb-85fc-c8bc237fa56a.png)

### Actual behavior

this nodes parameter just add nodes if not exist in map, but not delete nodes when cluster nodes are deleted
![image](https://user-images.githubusercontent.com/35724820/118774842-d6131480-b8b8-11eb-89a7-e49f00721ea3.png)

### Steps to reproduce:

it would cause a problem in this situation:
1. create a redis cluster A with 10 master nodes  and delete 9 nodes from cluster;
2. use the 9 nodes create a new cluster B
2. delete some slots from the cluster or give a large latency between client and the cluster A， make sure jedis throw JedisConnectionException;

JedisConnectionException would cause a slot refresh in JedisClusterCommand: 

![image](https://user-images.githubusercontent.com/35724820/118774358-52592800-b8b8-11eb-8968-2fe89e37c78e.png)

this method would use JedisClusterInfoCache paramter: nodes map values to refresh slot in JedisClusterInfoCache.renewClusterSlots():
![image](https://user-images.githubusercontent.com/35724820/118773931-d9f26700-b8b7-11eb-901a-5f7cf5a3fd02.png)

![image](https://user-images.githubusercontent.com/35724820/118774608-9c420e00-b8b8-11eb-9e62-e0865685317f.png)

There is a high probability(90%) that the jedis client will connect to cluster B because jedis use the cluster B nodes to refresh slot

#### Jedis version:
3.6.0

sazzad16:
@StillerRen In such massive reconfiguration of Redis Cluster, you should recreate JedisCluster with proper nodes.

StillerRen:
Node adding and deleting is normal operation for cluster and client don't perceive the change.
I think jedis should determine cluster nodes based on the cluster nodes/slots command.

sazzad16:
Jedis is supporting node adding and deleting well enough.
Only issue it doesn't handle (in JedisCluster A) yet is when a node is moved from Redis Cluster A to Redis Cluster B.

StillerRen:
This is a common operation, especially when resources are reused.
but this would cause the client connect to a wrong cluster and hard to find out reason

brunogbv:
[...]

PR #2642, opened 2021-09-09
Refresh nodes when renewSlotCache

yangbodong22011:
This PR solves #2504 #2550, when renewSlotCache, we also remove dead nodes according to the
latest query.


sazzad16:
@yangbodong22011 

1. :+1: for resolving #2550 
2. About #2504, the actual issue would still remain. But this would at least add some _defense_.

Could you please add a reverse test? E.g. start with 4 nodes and end up with 3 nodes after reconfiguration?

yangbodong22011:
> Could you please add a reverse test? E.g. start with 4 nodes and end up with 3 nodes after reconfiguration?

@sazzad16 done

sazzad16 (approved):

dengliming (approved):

Issue #2551, opened 2021-05-19
Add an option to reconfigure JedisCluster


sazzad16:
For simplicity, it can be assumed that saved nodes are either valid or dead but not moved to a new cluster.

Related: #2504 #2550

sazzad16:
Even though #2550 and (somewhat) #2504 are still relevant, closing this in favor of #2643

PR #2643, opened 2021-09-09
Renew cluster slots strategy update

yangbodong22011:
This PR solves #2551, a problem solving scenario is as follows:

1) JedisCluster jc = new JedisCluster(vip);  // now `vip -> ip1`
2) change `vip -> ip2`
3) renewClusterSlots just use getShuffledNodesPool, that is `ip1`
4) connect to `ip1`, and will never connect to `ip2`

The new strategy is as follows:
1. First, if jedis(meet JedisRedirectionException) is available, use jedis renew.
2. Then, we use startNodes to try, as long as startNodes is available,
    whether it is vip, domain, or physical ip, it will succeed.
3. Finally, we go back to the ShuffledNodesPool and try the remaining physical nodes.

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
@@ -31,17 +31,30 @@ public class JedisClusterInfoCache {
 
   private final GenericObjectPoolConfig<Connection> poolConfig;
   private final JedisClientConfig clientConfig;
+  private final Set<HostAndPort> startNodes;
 
   private static final int MASTER_NODE_INDEX = 2;
 
+  @Deprecated
   public JedisClusterInfoCache(final JedisClientConfig clientConfig) {
     this(clientConfig, new GenericObjectPoolConfig<Connection>());
   }
 
+  @Deprecated
   public JedisClusterInfoCache(final JedisClientConfig clientConfig,
       final GenericObjectPoolConfig<Connection> poolConfig) {
+    this(clientConfig, poolConfig, null);
+  }
+
+  public JedisClusterInfoCache(final JedisClientConfig clientConfig, final Set<HostAndPort> startNodes) {
+    this(clientConfig, new GenericObjectPoolConfig<Connection>(), startNodes);
+  }
+
+  public JedisClusterInfoCache(final JedisClientConfig clientConfig,
+      final GenericObjectPoolConfig<Connection> poolConfig, final Set<HostAndPort> startNodes) {
     this.poolConfig = poolConfig;
     this.clientConfig = clientConfig;
+    this.startNodes = startNodes;
   }
 
   public void discoverClusterNodesAndSlots(Connection jedis) {
@@ -82,6 +95,7 @@ public class JedisClusterInfoCache {
     // If rediscovering is already in process - no need to start one more same rediscovering, just return
     if (rediscoverLock.tryLock()) {
       try {
+        // First, if jedis is available, use jedis renew.
         if (jedis != null) {
           try {
             discoverClusterSlots(jedis);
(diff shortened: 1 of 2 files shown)


sazzad16:
@yangbodong22011 Not only this is a breaking change but also backward incompatible.

- (Preferable) Could you find a non-breaking solution?
- If not, at least find a backward compatible solution.
  - Hint: We can handle that `startNodes` can be null.

You are completely disregarding `getShuffledNodesPool()`. IMO we should check those first and if no solution is found, then check among `startNodes`.

yangbodong22011:
> * (Preferable) Could you find a non-breaking solution?
> * If not, at least find a backward compatible solution.
>   
>   * Hint: We can handle that `startNodes` can be null.

This is not actually a behavior visible to the user, so I think we can modify it without having to be compatible with the old (and may cause errors) logic.

If it should be compatible, maybe we can add a `refreshConfigByStartNode` option to control the two logics.

I guess the reason you think it should be compatible is that the performance of the `getShuffledNodesPool()` method will be higher, because operations such as new connections and auth are avoided.


> You are completely disregarding `getShuffledNodesPool()`. IMO we should check those first and if no solution is found, then check among `startNodes`.


This may be a bit difficult, because we did not distinguish whether the type in `NodesPool` is MASTER or SLAVE, (we think that MASTER will always belong to the cluster), but there are two solutions:
- After taking out the Jedis connection, call `jedis.clusterNodes()` to get your own type (master or slave)
- When we add to the NodesPool, use the key to record the type, such as nodes.put(127.0.0.1:6379:**master**, nodePool);

sazzad16:
> This is not actually a behavior visible to the user, so I think we can modify it without having to be compatible with the old (and may cause errors) logic.

Practically, yes. But theoretically no; because JedisClusterInfoCache is a public class.

> I guess the reason you think it should be compatible is that ...

No. If the behavior is changed significantly, we may have to push it for 4.0 instead of 3.x.

sazzad16:
@yangbodong22011 WDYT about #2644 ?

yangbodong22011:
@sazzad16 @dengliming PING, pls have a look when you have time.

sazzad16 (changes requested):
About existing JedisClusterInfoCache constructors, we have to keep those. So, keep these constructors, deprecate these, forward to new constructors (like `this(..., ..., (Set<HostAndPort>) null);` and handle `startNodes == null` in other methods.

yangbodong22011:
> About existing JedisClusterInfoCache constructors, we have to keep those. So, keep these constructors, deprecate these, forward to new constructors (like `this(..., ..., (Set<HostAndPort>) null);` and handle `startNodes == null` in other methods.

done

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -91,20 +105,29 @@ public void renewClusterSlots(Connection jedis) {
           }
         }
 
+        // Then, we use startNodes to try, as long as startNodes is available,
+        // whether it is vip, domain, or physical ip, it will succeed.
+        ArrayList<HostAndPort> startNodeList = new ArrayList<>(startNodes);
```
sazzad16, line 110: `startNodes` can be null. So we should do a safety check.

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
yangbodong22011: 👍，done

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -91,20 +105,31 @@ public void renewClusterSlots(Connection jedis) {
           }
         }
 
+        // Then, we use startNodes to try, as long as startNodes is available,
+        // whether it is vip, domain, or physical ip, it will succeed.
+        if (startNodes != null) {
+          ArrayList<HostAndPort> startNodeList = new ArrayList<>(startNodes);
+          Collections.shuffle(startNodeList);
+          for (HostAndPort hostAndPort : startNodeList) {
```
sazzad16, line 113: What if we put _attempted_ `hostAndPort`s in a Set and skip those when attempting from getShuffledNodesPool()?

sazzad16 (commented):
LGTM so far.

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
yangbodong22011: Agreed, so that during the `getShuffledNodesPool` stage, nodes that have already been tried and failed can be ignored.
I think, instead of using a new variable, we can use `startNodes` directly. Because once it reaches `getShuffledNodesPool`, it means that `startNodes` has been tried all, and none of them have succeeded.

yangbodong22011: already updated code.

sazzad16 (approved):

dengliming (approved):
LGTM

yangbodong22011:
@sazzad16 I would like to backport this PR to version 3.9.x, what do you think?

sazzad16:
@yangbodong22011 I may consider it for 3.10.0

yangbodong22011:
> @yangbodong22011 I may consider it for 3.10.0

That's ok, what should I do next? Submit a PR to the 3.x branch?

PR #2644, opened 2021-09-10
Renew cluster slots depends on startNodes (ii)

sazzad16:


src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -162,6 +170,19 @@ public void renewClusterSlots(Jedis jedis) {
             }
           }
         }
+
+        if (startNodes != null) {
```
yangbodong22011, line 174: If `getShuffledNodesPool` exists, it is almost impossible for the code to come here unless all master and slave nodes in the cluster are unserviceable.

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -95,8 +97,14 @@ public JedisClusterInfoCache(final GenericObjectPoolConfig<Jedis> poolConfig,
 
   public JedisClusterInfoCache(final GenericObjectPoolConfig<Jedis> poolConfig,
       final JedisClientConfig clientConfig) {
+    this(poolConfig, clientConfig, null);
+  }
+
+  public JedisClusterInfoCache(final GenericObjectPoolConfig<Jedis> poolConfig,
```
yangbodong22011, line 103: Do you plan to provide an interface for setting `JedisClusterInfoCache` for`JedisCluster`?

sazzad16: I'm not planning such idea.

But I won't mind to see a PR :)

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
sazzad16: @yangbodong22011 OK, now I have clearer idea about #2643

Issue #2625, opened 2021-08-18
ZRANGE support [BYSCORE|BYLEX] [REV] [LIMIT offset count] [WITHSCORES] option


startjava:
ZRANGE key min max [BYSCORE|BYLEX] [REV] [LIMIT offset count] [WITHSCORES]

or use old api ,replace new that option?

thank you !

yangbodong22011:
@startjava refer : 
```
Set<String> zrangeByScore(String key, String min, String max); // byscore
Set<String> zrangeByLex(String key, String min, String max); // bylex
Set<String> zrangeByScore(String key, double min, double max, int offset, int count); // limit offset count
Set<Tuple> zrangeByScoreWithScores(String key, double min, double max);  // withscores
```

startjava:
杨哥 感谢 &nbsp;还是用原来的api &nbsp;我以为只用一个方法实现呢 &nbsp;这样也好 分散了 &nbsp;减少单一方法的复杂性



发自我的iPhone


------------------ Original ------------------
From: Yang Bodong ***@***.***&gt;
Date: Wed,Sep 15,2021 10:12 AM
To: redis/jedis ***@***.***&gt;
Cc: startjava ***@***.***&gt;, Mention ***@***.***&gt;
Subject: Re: [redis/jedis] has ZAddParam support [BYSCORE|BYLEX] [REV] [LIMIT offset count] [WITHSCORES] option plan?? (#2625)





 
@startjava refer :
 Set<String&gt; zrangeByScore(String key, String min, String max); // byscore Set<String&gt; zrangeByLex(String key, String min, String max); // bylex Set<String&gt; zrangeByScore(String key, double min, double max, int offset, int count); // limit offset count Set<Tuple&gt; zrangeByScoreWithScores(String key, double min, double max);  // withscores  
—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

sazzad16:
@yangbodong22011 Redis has this [ZRANGE](https://redis.io/commands/zrange) command and arguments, so it would be great to include in Jedis as well. Reopening this issue, in case someone wants to give it a shot and create a PR.

Issue #2626, opened 2021-08-19
please implement ZRANGESTORE API


startjava:
please implement ZRANGESTORE API

sazzad16:
Resolves by #2817

PR #2645, opened 2021-09-16
Add ZRangeParams and new zrange command

yangbodong22011:
This PR solves #2625, add ZRangeParams and new zrange command.


yangbodong22011:
After this PR was merged, we can use `ZRangeParams` to solve #2626

sazzad16:
@yangbodong22011 :+1: 

But I'd like to wait before deciding on this PR because I am hoping to have options like
- for zrange (by rank) -> `int star` and `int stop`
- for zrange by score -> `double min` and `double max`
- for zrange by lex -> `String min` and `String max`

yangbodong22011:
> But I'd like to wait before deciding on this PR because I am hoping to have options like
> 
> * for zrange (by rank) -> `int star` and `int stop`
> * for zrange by score -> `double min` and `double max`
> * for zrange by lex -> `String min` and `String max`

You mean we have the following three method signatures:
```
public Set<String> zrange(final String key, final int start, final int stop, final ZRangeParams params);
public Set<String> zrangeByScore(final String key, final double min, final double max, final ZRangeParams params);
public Set<String> zrangeByLex(final String key, final String min, final String max, final ZRangeParams params);
```
And need to remove the `byscore()` and `bylex()` in `ZrangeParams`, right?

sazzad16:
@yangbodong22011 I am not sure :(

It can be three methods or some trick in ZRangeParams. The most complicated part of thinking about it is supporting _exclusive range_ queries. Currently, Jedis does it by `zrangeByScore(String key, String min, String max)` like methods and I am not really a fan of that approach.

Let's see if others have any idea. @dengliming @mina-asham @gkorland

PR #2646, opened 2021-09-18
Added Jedis Cluster Flush All capability.

biprajeet:
Capability :

* Flush all capability for Jedis Cluster clients
* Future capability of running library functions on all nodes capability

Code Change
* JedisCluster.java : Flush all command added
* JedisClusterCommand.java : Added capability to execute commands on all nodes participating in the cluster. This does not throw errors
* JedisClusterTest.java : Test for flush all function
* JedisClusterCommandTest.java : Test for execute command on all nodes function

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisCluster.java
+++ b/src/main/java/redis/clients/jedis/JedisCluster.java
@@ -2966,4 +2966,14 @@ public class JedisCluster extends BinaryJedisCluster implements JedisClusterComm
   private static String[] getKeys(final Map<String, ?> map) {
     return map.keySet().toArray(new String[map.size()]);
   }
+
+  public boolean flushAll() {
+    return new JedisClusterCommand<Boolean>(connectionHandler, maxAttempts, maxTotalRetriesDuration) {
+      @Override
+      public Boolean execute(Jedis connection) {
+        connection.flushAll();
+        return true;
+      }
+    }.runWithAllNodes((a, b) -> a && b, true);
+  }
 }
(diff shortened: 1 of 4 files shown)


yangbodong22011:
@biprajeet thanks for your contribution.
But IMHO, I don’t think this is necessary, and Jedis has not provided such a mechanism (send requests to multiple nodes), user can call `jedisCluster.getClusterNodes()` and then traverse the calls, just like in your code. And this can also support `select`, `flushdb` and other functions.

biprajeet:
Hi @yangbodong22011 , Thanks for the reply,  I understand what you're saying.

Is there a development philosophy this code is breaking, like we should implement code which only redis clients support across libraries and command line tools?  

This PR arised out of my need.

> Standalone Jedis client provides 'flushAll' function but Jedis cluster does not, thus breaking the code structure while changing architecture. 

> This function would be especially helpful while writing unit tests in projects where cleanup is necessary before and after tests (including this project)  

> We should implement the other functions as well - select, flushdb. This would reduce so much effort for people using Jedis cluster client.

sazzad16:
@biprajeet 

> We should implement the other functions as well - select, flushdb. This would reduce so much effort for people using Jedis cluster client.

Redis cluster supports only one DB (`0`). There is not point supporting DB specific commands.

> This function would be especially helpful while writing unit tests in projects where cleanup is necessary before and after tests.

Supporting tests is not strong enough reason to include something of a different *pattern*.

biprajeet:
@yangbodong22011 @sazzad16 Thanks for your replies. I understand your points, we would like users to implement this function manually if they require and also this is a different pattern. I was hoping we could accommodate this into the pattern. Anyway will look forward to contributing in the future if I get anything. Will close the MR. 👍

sazzad16:
@biprajeet Thank you for understanding. Please don't this discourage you from contributing in Jedis whether by issues or PRs.

Happy coding :)

Issue #2659, opened 2021-10-04
Need the jedis params to be serializable


gwdgithubnom:
### Expected behavior

 * for package class: redis.clients.jedis.params, the jedis param super class and implements with serializable </br>
 * for use some env need serializable. eg: spark or flink env.
 * https://github.com/redis/jedis/issues/2659 <br/>

### Actual behavior

redis.clients.jeds.params.*Params could not serializable.

### Steps to reproduce:

1. dev in spark sql, udf function, the prams did not use transient or define in function inner, it would throw a exception about the params did not support serializable.
2. dev in flink, the sink or source function also need the params must be serializable.

### Redis / Jedis Configuration
none

#### Jedis version:
3.6+

#### Redis version:
3.1+

#### Java version:
1.8+

PR #2660, opened 2021-10-04
For the jedis params to be serializable issues#2659

gwdgithubnom:
# Description

refer url: https://github.com/redis/jedis/issues/2659

- for package class: redis.clients.jedis.params, the jedis param super class and implements with serializable
- for use some env need serializable. eg: spark or flink env.
- Need the jedis params to be serializable #2659 

# Motivation
add params serializable  support

# Problems

redis.clients.jeds.params.*Params could not serializable.

Steps to reproduce:
dev in spark sql, udf function, the prams did not use transient or define in function inner, it would throw a exception about the params did not support serializable.
dev in flink, the sink or source function also need the params must be serializable.
Redis / Jedis Configuration
none

Jedis version:
3.6+

Redis version:
3.1+

Java version:
1.8+


gwdgithubnom:
@sazzad16 In some usage scenarios, it is necessary to ensure that the parameters are serializable, so it is necessary to modify the param class to integrate serializable, please check the pull request

sazzad16:
@gwdgithubnom Just adding `implements Serializable` is not enough for a PR. You may have to implement `writeObject` and `readObject` methods. And also write tests.

gwdgithubnom:
> @gwdgithubnom Just adding `implements Serializable` is not enough for a PR. You may have to implement `writeObject` and `readObject` methods. And also write tests.
thank you for your reply. We will do this work later.

sazzad16:
Closed due to inactivity.

PR #2662, opened 2021-10-09
Fix jedis & pipeline has the same client result in thread issue

lionhuo:
### Description
the pipeline has the same client with the jedis which get from jedis pool, if the jedis return to the pool, but the pipeline is still working, then the buf in RedisOutputStream maybe override, it will result in the wrong redis comand flush to redis server.

### DIFF
--- a/src/main/java/redis/clients/jedis/Pipeline.java
+++ b/src/main/java/redis/clients/jedis/Pipeline.java
@@ -99,17 +99,10 @@ public class Pipeline extends MultiKeyPipelineBase implements Closeable {
    * commands you execute.
    */
   public void sync() {
-    try{
-      if (getPipelinedResponseLength() > 0) {
-        List<Object> unformatted = client.getMany(getPipelinedResponseLength());
-        for (Object o : unformatted) {
-          generateResponse(o);
-        }
-      }
-    }finally {
-      if(jedis != null){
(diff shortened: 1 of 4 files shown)


src/main/java/redis/clients/jedis/Pipeline.java
```diff
@@ -94,10 +99,17 @@ public boolean isInMulti() {
...
-        generateResponse(o);
+    try{
+      if (getPipelinedResponseLength() > 0) {
+        List<Object> unformatted = client.getMany(getPipelinedResponseLength());
+        for (Object o : unformatted) {
+          generateResponse(o);
+        }
+      }
+    }finally {
+      if(jedis != null){
```
yangbodong22011, line 110: This means that once `pipeline.sync()`, the pipeline is no longer available, this is incompatible with the previous semantics, and it is not what I approve of.

> It is expected that the user would be careful to avoid these situations.

Agree, the user guarantees that it is correct and recommended to use `Pipeline` before Jedis is returned/closed.

lionhuo:
> the following code is just one more line.

you are right

but i think get `Pipeline` from the pool，It reinforces `Pipeline` is a independent resource of `Jedis` and user don't need careful to use `Jedis` and `Pipeline` if `Pipeline` escapes from `Jedis` block

PR #2681, opened 2021-11-01
Support Json Commands

AvitalFineRedis:


src/main/java/redis/clients/jedis/json/RedisJsonCommands.java
```diff
@@ -1,16 +1,58 @@
...
+
 public interface RedisJsonCommands {
 
   String jsonSet(String key, Object object);
 
+  String jsonSet(String key, Object object, ExistenceModifier flag);
+
   String jsonSet(String key, Path path, Object object);
 
+  String jsonSet(String key, Object object, ExistenceModifier flag, Path path);
```
sazzad16, line 13: Replace ExistenceModifier with an IParams (and name it properly).

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -60,6 +60,35 @@ public String toString() {
     }
   };
 
+  public static final Builder<Class<?>> CLASS = new Builder<Class<?>>() {
```
sazzad16, line 63: 1. rename it more descriptive and unique; i.e. REDIS_JSON_TYPE
2. move it after all core Redis related builders; i.e. after STR_ALGO_LCS_RESULT

sazzad16 (changes requested):

yangbodong22011:
I don’t think it’s a good way to add the Module command directly into the Jedis code. It has the following disadvantages:

- With more and more Redis Modules, adding or modifying module commands may require a Jedis release.
- It is difficult to decide which Module commands need to be supported by Jedis, whether they are driven by the community PR or the Maintainer.

Some of my possible suggestions:
- A plug-in mechanism that allows the community to implement different Module commands by itself. `Jedis + different Module pom` can support different Modules, so as to avoid the modification of RedisJSON, which will cause a passive release of RedisSearch.

chayim:
@yangbodong22011 Interesting! There's another pattern we can consider to properly support this option.  What if there was a loader that made it easy to include (possibly via reflection, ugz) different modules. This could solve that, and we could still go down this path...

PR #2682, opened 2021-11-01
Small fix in JedisClusterInfoCache.discoverClusterNodesAndSlots.

adiamzn:
### Description:
The change calls `Jedis.cluserSlots` before `reset `is called instead of the other way around.

### Why the change?
the `reset` function deletes the `nodes `and `slots `fields.
If the `Jedis.clusterSlots` throws an exception in the try-catch block
the function fails and the data is deleted.
It would be nicer if the data (nodes and slots) were left untouched if jedis.clusterSlots throws an exception.

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
@@ -104,12 +104,10 @@ public class JedisClusterInfoCache {
   }
 
   public void discoverClusterNodesAndSlots(Jedis jedis) {
+    List<Object> slots = jedis.clusterSlots();
     w.lock();
-
     try {
       reset();
-      List<Object> slots = jedis.clusterSlots();
-
       for (Object slotInfoObj : slots) {
         List<Object> slotInfo = (List<Object>) slotInfoObj;
 
--- a/.github/release-drafter-config.yml
+++ b/.github/release-drafter-config.yml
@@ -1,13 +1,34 @@
 name-template: '$NEXT_MAJOR_VERSION'
-tag-template: 'jedis-$NEXT_MAJOR_VERSION'
+tag-template: 'v$NEXT_MAJOR_VERSION'
+autolabeler:
+  - label: 'chore'
+    files:
+      - '*.md'
+      - '.github/*'
+  - label: 'bug'
+    branch:
+      - '/bug-.+'
+  - label: 'chore'
+    branch:
+      - '/chore-.+'
+  - label: 'feature'
+    branch:
+      - '/feature-.+'
 categories:
-  - title: 'Features'
+  - title: 'Breaking Changes'
     labels:
-      - 'ENHANCEMENT'
-  - title: 'Bug Fixes'
+      - 'breakingchange'
+  - title: '🚀 New Features'
     labels:
+      - 'feature'
+      - 'enhancement'
+  - title: '🐛 Bug Fixes'
+    labels:
+      - 'fix'
+      - 'bugfix'
+      - 'bug'
       - 'BUG'
-  - title: 'Maintenance'
+  - title: '🧰 Maintenance'
     label: 'chore'
 change-template: '- $TITLE (#$NUMBER)'
 exclude-labels:
@@ -19,5 +40,6 @@ template: |
 
   ## Contributors
   We'd like to thank all the contributors who worked on this release!
-  
+
   $CONTRIBUTORS
+


nbraun-amazon:
Approved

yangbodong22011:
we just call `discoverClusterNodesAndSlots` at 1:

```
private void initializeSlotsCache(Set<HostAndPort> startNodes, JedisClientConfig clientConfig) {
  ArrayList<HostAndPort> startNodeList = new ArrayList<>(startNodes);
  Collections.shuffle(startNodeList);
    for (HostAndPort hostAndPort: startNodeList) {
    try (Jedis jedis = new Jedis(hostAndPort, clientConfig)) {
      cache.discoverClusterNodesAndSlots(jedis); // 1
      return; // 2
      // try next nodes
    }
  }
}


  public void discoverClusterNodesAndSlots(Jedis jedis) {
    w.lock();

    try {
      reset(); // 3
      List<Object> slots = jedis.clusterSlots(); // 4
      ...
```


When `jedis.clusterSlots()` does not throw an exception, the return at 2 will be executed. On the contrary, if an exception is thrown, enter `discoverClusterNodesAndSlots` again, it does not matter to execute 3 or 4 first, because at this time the variables of `nodes` and `slots` are empty (they were not filled last time, because jedis.clusterSlots() throws an exception)

we just call `discoverClusterNodesAndSlots` at 1:

```
private void initializeSlotsCache(Set<HostAndPort> startNodes, JedisClientConfig clientConfig) {
  ArrayList<HostAndPort> startNodeList = new ArrayList<>(startNodes);
  Collections.shuffle(startNodeList);
    for (HostAndPort hostAndPort: startNodeList) {
    try (Jedis jedis = new Jedis(hostAndPort, clientConfig)) {
      cache.discoverClusterNodesAndSlots(jedis); // 1
      return; // 2
      // try next nodes
    }
  }
}


  public void discoverClusterNodesAndSlots(Jedis jedis) {
    w.lock();

    try {
      reset(); // 3
      List<Object> slots = jedis.clusterSlots(); // 4
      ...
```


When `jedis.clusterSlots()` does not throw an exception, the return at 2 will be executed. On the contrary, if an exception is thrown, enter `discoverClusterNodesAndSlots` again, it does not matter to execute 3 or 4 first, because at this time the variables of `nodes` and `slots` are empty (they were not filled last time, because jedis.clusterSlots() throws an exception NS)

I would like to ask, does this PR fix the actual problems encountered in the production environment, or is the idea obtained by reading the code?

adiamzn:
Hi Yang,
Agree that this does not cause any issues the way the code is currently written.
As you mentioned this function is currently only called from [1] and does not cause issues the way it is currently being called.

Since the function `discoverClusterNodesandSlots `is part of the API for the class `JedisClusterInfoCache`, I thought it was a good idea to add a better safety guarantee. 
The idea came up just from reading the code. 
All the best,
Adi

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -107,8 +107,8 @@ public void discoverClusterNodesAndSlots(Jedis jedis) {
     w.lock();
```
yangbodong22011, line 107:
```suggestions
List<Object> slots = jedis.clusterSlots();
w.lock();
```

yangbodong22011: > Since the function discoverClusterNodesandSlots is part of the API for the class JedisClusterInfoCache, I think it is a good idea to add a better safety guarantee.

@adiamzn Agree, I think we can move `jedis.clusterSlots()` to the top of `w.lock()`, just like the `discoverClusterSlots` function.

adiamzn: agreed, made the change

yangbodong22011: @adiamzn `jedis.clusterSlots()` to the `top` of `w.lock()`, the lock time can be shortened.

adiamzn: true, fixing

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -104,12 +104,10 @@ public JedisClusterInfoCache(final GenericObjectPoolConfig<Jedis> poolConfig,
   }
 
   public void discoverClusterNodesAndSlots(Jedis jedis) {
+    List<Object> slots = jedis.clusterSlots();
     w.lock();
-
     try {
       reset();
-      List<Object> slots = jedis.clusterSlots();
-
```
sazzad16, line 112: @yangbodong22011 @adiamzn If you want this change to be similar to `discoverClusterSlots` which is a private method and all the calling places are within try-catch block. So there is safeguard for `clusterSlots()`. But this `discoverClusterNodesAndSlots` method is a public one and this change would remove existing safeguard.

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
adiamzn: @sazzad16 @yangbodong22011 both options seem fine with me. The critical part here is the lost data in case of failure. 
@sazzad16 From what I see here the try catch block is just for the lock, so we release it even if there is an exception.
I don't mind put the `clusterSlots()` inside the try catch block.
Please let me know what you prefer and I will make the change

sazzad16: @adiamzn Ohh, thanks. And, sorry, I missed there isn't any `catch`.

sazzad16 (approved):

yangbodong22011 (approved):

dengliming (approved):

PR #2683, opened 2021-11-01
XRANGE support exclusive range

### LINKED ISSUE #2631: please implements XRANGE ( Exclusive ranges param !
please implements XRANGE ( Exclusive ranges param !

AvitalFineRedis:
Closes #2631

### DIFF
--- a/src/main/java/redis/clients/jedis/Client.java
+++ b/src/main/java/redis/clients/jedis/Client.java
@@ -1508,6 +1508,12 @@ public class Client extends BinaryClient implements Commands {
   }
 
   @Override
+  public void xrange(String key, String start, String end, int count) {
+    xrange(SafeEncoder.encode(key), SafeEncoder.encode(start == null ? "-" : start),
+            SafeEncoder.encode(end == null ? "+" : end), count);
+  }
+
+  @Override
   public void xrange(final String key, final StreamEntryID start, final StreamEntryID end,
       final long count) {
     xrange(SafeEncoder.encode(key), SafeEncoder.encode(start == null ? "-" : start.toString()),
@@ -1524,6 +1530,12 @@ public class Client extends BinaryClient implements Commands {
   public void xrevrange(String key, StreamEntryID end, StreamEntryID start, int count) {
     xrevrange(SafeEncoder.encode(key), SafeEncoder.encode(end == null ? "+" : end.toString()),
       SafeEncoder.encode(start == null ? "-" : start.toString()), count);
+  }
+
+  @Override
+  public void xrevrange(String key, String end, String start, int count) {
+    xrevrange(SafeEncoder.encode(key), SafeEncoder.encode(end == null ? "+" : end),
+            SafeEncoder.encode(start == null ? "-" : start), count);
   }
 
   @Override
--- a/src/main/java/redis/clients/jedis/Jedis.java
+++ b/src/main/java/redis/clients/jedis/Jedis.java
@@ -4396,6 +4396,13 @@ public class Jedis extends BinaryJedis implements JedisCommands, MultiKeyCommand
   }
 
   @Override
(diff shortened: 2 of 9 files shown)


src/main/java/redis/clients/jedis/Client.java
```diff
@@ -1507,6 +1507,12 @@ public void xrange(final String key, final StreamEntryID start, final StreamEntr
       SafeEncoder.encode(end == null ? "+" : end.toString()), count);
   }
 
+  @Override
+  public void xrange(String key, String start, String end, int count) {
```
yangbodong22011, line 1511: we should also support binary commands?
```
public void xrange(byte[] key, byte[] start, byte[] end, int count) {
```

sazzad16: @yangbodong22011 those are already available in `Binary...` classes.

yangbodong22011: @sazzad16 ohh...yes, ignore my comment

sazzad16:
Please check that this PR, as of now, is a breaking change. I.e. any invocation of 

    jedis.xrange(streamName, null, null, count);

must be changed to any of following (or something else)

    xrange(streamName, (StreamEntryID) null, (StreamEntryID) null, count);
    xrange(streamName, (String) null, (String) null, count);

I can think of two different options

1. target this PR for 4.0.0 and mention the breaking behavior in release notes
2. rename methods (e.g. xrange2/xrevrange2 or xrangeV2/xrevrangeV2)

Please share your thoughts.

@yangbodong22011 @dengliming @mina-asham

yangbodong22011:
> I can think of two different options
> 
> 1. target this PR for 4.0.0 and mention the breaking behavior in release notes
> 2. rename methods (e.g. xrange2/xrevrange2 or xrangeV2/xrevrangeV2)

vote for 1，rename is not a good way, because the binary `xrange` already exists and may be used by the user.
```
public void xrange(final byte[] key, final byte[] start, final byte[] end, final int count);
```

AvitalFineRedis:
> Please check that this PR, as of now, is a breaking change. I.e. any invocation of
> 
> ```
> jedis.xrange(streamName, null, null, count);
> ```
> 
> must be changed to any of following (or something else)
> 
> ```
> xrange(streamName, (StreamEntryID) null, (StreamEntryID) null, count);
> xrange(streamName, (String) null, (String) null, count);
> ```
> 
> I can think of two different options
> 
> 1. target this PR for 4.0.0 and mention the breaking behavior in release notes
> 2. rename methods (e.g. xrange2/xrevrange2 or xrangeV2/xrevrangeV2)
> 
> Please share your thoughts.
> 
> @yangbodong22011 @dengliming @mina-asham

vote for 1 also

AvitalFineRedis:
Vote for 1 also

dengliming:
> I can think of two different options
> 
> 1. target this PR for 4.0.0 and mention the breaking behavior in release notes
> 2. rename methods (e.g. xrange2/xrevrange2 or xrangeV2/xrevrangeV2)

Vote for 1, agree with @yangbodong22011

sazzad16:
@yangbodong22011 @dengliming We are done voting. I have set this for 4.0.0. Please review the code 🙂

dengliming:
Actually, I'm thinking about introducing class `Range` ( refer to [https://github.com/lettuce-io/lettuce-core/blob/main/src/main/java/io/lettuce/core/Range.java](https://github.com/lettuce-io/lettuce-core/blob/main/src/main/java/io/lettuce/core/Range.java), and the method name like the following 

`jedis.xrange(streamName, range, count);`

Would it be better?

sazzad16:
@dengliming It'd be great! 

But this would bring the feature early and it won't break your plan either.

PR #2697, opened 2021-11-24
Improving ACL Command Support

chayim:
- Adding missing ACL commands
- Started javadocs

PR #2699, opened 2021-11-24
Support for BGSAVE SCHEDULE command

chayim:
- BGSAVE SCHEDULE support

Accidentally branched off #2697 so we should merge after that, back into this branch.


sazzad16:
@chayim `ControlCommandsTest.bgsave_schedule:54 expected:<[OK]> but was:<[Background saving started]>`

src/test/java/redis/clients/jedis/tests/commands/ControlCommandsTest.java
```diff
@@ -47,6 +47,17 @@ public void bgsave() {
     }
   }
 
+  @Test
+  public void bgsave_schedule() {
```
sazzad16, line 51: Let's keep it Java (as much as we can).
```suggestion
  public void bgsaveSchedule() {
```

sazzad16 (changes requested):
Also, please fix conflicts.

PR #2700, opened 2021-11-25
Add jedis cluster pipeline tests

yangbodong22011:


sazzad16:
Thank you @yangbodong22011

PR #2701, opened 2021-11-25
deal possible StringIndexOutOfBoundsException and NumberFormatException

suoyi123wang:


yangbodong22011:
@suoyi123wang hello, We believe that it is the caller's responsibility to catch the `NumberFormatException`, not to handle it inside Jedis.

suoyi123wang:
> @suoyi123wang Hello, we believe that it is the caller's responsibility to catch the `NumberFormatException`, not to handle it inside Jedis.

PR #2707, opened 2021-11-30
Implemented JedisSharding

sazzad16:


yangbodong22011:
Suppose a `JedisSharding` has two `shards`:

The old `ShardedJedis` will create 160*2 = 320 connections by default, see [JedisShardInfo#createResource](https://github.com/redis/jedis/blob/master/src/main/java/redis/clients/jedis/JedisShardInfo.java#L262).

The new `JedisSharding` will create two `JedisPool(maxTotal/maxIdle/minIdle = 8/8/0)` by default (360-16=344 fewer connections than before). I am not sure whether the user may encounter the error of `Could not get a resource from the pool`.

I just raised this question, but I still agree with the current approach. After all, we also reserve an interface for configuring JedisPool parameters.
```
public JedisSharding(List<HostAndPort> shards, JedisClientConfig clientConfig,
   GenericObjectPoolConfig<Connection> poolConfig) {
 this(new ShardedConnectionProvider(shards, clientConfig, poolConfig));
}
```

Issue #2628, opened 2021-08-19
zrandmemberWithScores api has bug!!!!  warn !!!   warn !!!!


startjava:
```
package com.ghy.www.test7;

import com.ghy.www.f.F;
import redis.clients.jedis.Jedis;
import redis.clients.jedis.JedisPool;
import redis.clients.jedis.JedisPoolConfig;
import redis.clients.jedis.Tuple;

import java.util.Iterator;
import java.util.Set;

public class ZRANDMEMBER_BUG {
    private static JedisPool pool = new JedisPool(new JedisPoolConfig(), F.IP_1, F.PORT_1, F.TIMEOUT_1, F.PASSWORD_1);

    public static void main(String[] args) {
        Jedis jedis = null;
        try {
            jedis = pool.getResource();
            jedis.flushDB();

            jedis.zadd("zset1", 1, "a");
            jedis.zadd("zset1", 2, "b");
            jedis.zadd("zset1", 3, "c");
            jedis.zadd("zset1", 4, "d");
            jedis.zadd("zset1", 5, "e");
            
            Set<Tuple> set = jedis.zrandmemberWithScores("zset1", -10);
            Iterator<Tuple> iterator = set.iterator();
            while (iterator.hasNext()) {
                Tuple tuple = iterator.next();
                System.out.println(tuple.getElement() + "       " + tuple.getScore());
            }
        } catch (Exception e) {
            e.printStackTrace();
        } finally {
            if (jedis != null) {
                jedis.close();
            }
        }
    }
}

```
more runn this program :

```
c       3.0
e       5.0
b       2.0
a       1.0
```

---------------------------
```
c       3.0
a       1.0
d       4.0
b       2.0
```
---------------------------
```
a       1.0
b       2.0
c       3.0
e       5.0
```

so where is count -10 absolute value count 10??

sazzad16:
Redis returns repeated elements in this case. It is not possible to support repetition while return Java `Set`.

sazzad16:
Resolved by #2709 and #2693

Issue #2579, opened 2021-07-08
Consistent return among commands


sazzad16:
There are commands that are (can be considered) similar. But there are inconsistencies among the return types of those commands. We may/should make those similar.

Some such methods are:
- `Set<String> zrandmember`, `List<String> hrandfield`

yangbodong22011:
I have not seen the benefit of using Set (wrapping List) instead of using List directly. If so, please point out.

sazzad16:
@yangbodong22011 IMO, your comment is more aligned with #2278

yangbodong22011:
update in https://github.com/redis/jedis/issues/2278#issuecomment-878946050

sazzad16:
Resolved by https://github.com/redis/jedis/commit/858c8051301854e8343eef66aa68c6b9dab1c328

Issue #2278, opened 2020-11-23
Return List instead of Set from smembers method


mudit-saxena:
### Expected behavior

Redis `SMEMBERS` returns an array of values for the given set. While in Jedis `smembers` method returns a `SET` of values going by interface signature. So, as a result, the client of the library expects that `smembers` will honor all the `SET` related constraints in terms of time-complexity i.e `contains`, `add`, etc will be O(1).

### Actual behavior

On the contrary, looking at the implementation of `smemebers()` method & `SetFromList` class it becomes clear that the above constraints won't hold as `smemebers()` actually returns a List wrapped in a Set interface. This use of decorator pattern is quite misleading as the operations `contains`, `add`, etc on the resultant `Set` from  `smemebers()` will be O(n). Also, this violates the general Java Set constraints in terms of time-complexity expectation 

#### Jedis version:
3.0.1
#### Redis version:
6.0.5
#### Java version:
11

sazzad16:
@gkorland PING!

dengliming:
Agree with @mudit-saxena . The native `SET` is more suitable for this scenario.

sazzad16:
@dengliming But @mudit-saxena is asking for native `List`, not Set. And from my _Jedis user_ experience, I agree with returning a List.

yangbodong22011:
Some API need to ensure the return order, such as `zrange`, can only use `SetFromList`. some do not need, such as `zrandmember`, `spop`, can use Java Set to get O(1) `contains`.

I think we need to distinguish between the two cases, return to Java's `Set` or `List`, and then delete `SetFromList`.

Or we can use `List` directly and avoid introducing `Set`.

mudit-saxena:
Given that under the hood implementation for `SetFromList ` is `List`, directly using `List` looks to be a better choice without impact on existing usage from performance POV. Also, `List` is in line with Redis `smembers` https://redis.io/commands/smembers

sazzad16:
Related #2628

PR #2709, opened 2021-11-30
Return List instead of Set where it makes sense

sazzad16:
References:
- #2278 
- #2628 
- #2579


yangbodong22011:
I agree with this PR:
- Redis's [RESP](https://redis.io/topics/protocol) protocol itself does not include the SET type.
- We can solve https://github.com/redis/jedis/issues/2628

PR #2717, opened 2021-12-06
support XINFO STREAM key FULL [COUNT count]

zeekling:
support XINFO STREAM key FULL [COUNT count] like:
```bash
xinfo stream xadd-stream1 full
```


zeekling:
conficted, i close and recommit it

PR #2719, opened 2021-12-08
Support exclusive range in XRANGE command

### LINKED ISSUE #2631: please implements XRANGE ( Exclusive ranges param !

sazzad16:
Resolves #2631 
Closes #2683

### DIFF
--- a/src/main/java/redis/clients/jedis/Jedis.java
+++ b/src/main/java/redis/clients/jedis/Jedis.java
@@ -8183,9 +8183,6 @@ public class Jedis implements ServerCommands, DatabaseCommands, JedisCommands, J
     return connection.executeCommand(commandObjects.xrange(key, start, end));
   }
 
-  /**
-   * {@inheritDoc}
-   */
   @Override
   public List<StreamEntry> xrange(final String key, final StreamEntryID start,
       final StreamEntryID end, final int count) {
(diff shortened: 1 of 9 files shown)


src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -8200,16 +8197,37 @@ public long xlen(final String key) {
     return connection.executeCommand(commandObjects.xrevrange(key, end, start));
   }
 
-  /**
-   * {@inheritDoc}
-   */
   @Override
   public List<StreamEntry> xrevrange(final String key, final StreamEntryID end,
```
chayim, line 8201: According to the [XRANGE docs](https://redis.io/commands/XRANGE) this should be start, end, count instead of end, start, count.

This applies to all signatures.

chayim (changes requested):
Recommend we change the signature to match the docs.

src/main/java/redis/clients/jedis/Jedis.java
sazzad16: @chayim Please check, XRANGE is fine.

This is [XREVRANGE](https://redis.io/commands/xrevrange). So that's fine as well.

chayim: DERP You are right. Eye blead.

chayim (approved):

PR #2721, opened 2021-12-08
throw exception if JedisCluster cannot connect to any host-port

sazzad16:
Original PR: #2275 
Original author: @chinafzy 

If none of the `startNodes`, given in JedisCluster constructor, works:
- JedisCluster does not work.
- No message/exception comes up.
- JedisCluster cannot not retry later.

User would be instantly aware if an exception is thrown.

Cause of new PR: I don't have write permission in the repo of original PR.

Closes #2275


src/test/java/redis/clients/jedis/JedisClusterWithoutSetupTest.java
```diff
@@ -0,0 +1,13 @@
+package redis.clients.jedis;
+
+import org.junit.Test;
+
+import redis.clients.jedis.exceptions.JedisClusterOperationException;
+
+public class JedisClusterWithoutSetupTest {
+
+  @Test(expected = JedisClusterOperationException.class)
+  public void test() {
```
sazzad16, line 10:
```suggestion
  public void uselessStartNodes() {
```

yangbodong22011 (approved):

PR #2722, opened 2021-12-08
Tries all hosts resolved for DNS name

adiamzn:
### Problem
By default the default DNS resolver in [InetAddress](https://docs.oracle.com/javase/7/docs/api/java/net/InetAddress.html) class may cache ip addresses after successfully connecting for the first time. 

This is can be problematic if we expect to be able to connect to more than one IP address behind a DNS name.
For example multiple read replicas behind a common dns name, or a failed host that was replaced and now has a new ip.

### Proposed Fixed
When creating connecting to a host using a dns name, try all addresses the DNS server was able to resolve.


src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
```diff
@@ -64,6 +70,25 @@ public DefaultJedisSocketFactory(HostAndPort hostAndPort, JedisClientConfig conf
     }
   }
 
+  private void connectToFirstSuccsefulHost(Socket socket, HostAndPort hostAndPort) throws IOException {
+    List<InetAddress> hosts = Arrays.asList(InetAddress.getAllByName(hostAndPort.getHost()));
+    Collections.shuffle(hosts);
+    List<ConnectException> connectExceptions = new ArrayList<ConnectException>();
```
yangbodong22011, line 76: `ConnectException` can be use `Exception` ? because connect will throw: IOException, SocketTimeoutException, java.nio.channels.IllegalBlockingModeException, IllegalArgumentException

yangbodong22011 (commented):
Thanks for you contribution.

This brings almost no performance loss, because `java.net.InetAddress#getByName(java.lang.String)` also calls `java.net.InetAddress#getAllByName(java.lang.String)`, but gets [0 ] Element, just add the overhead of `Collections.shuffle(hosts)`, but it can be ignored.
```
     public static InetAddress getByName(String host)
         throws UnknownHostException {
         return InetAddress.getAllByName(host)[0];
     }
```
Some comments are detailed in the code section.

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
```diff
@@ -64,6 +70,25 @@ public DefaultJedisSocketFactory(HostAndPort hostAndPort, JedisClientConfig conf
...
+    for (InetAddress host : hosts) {
+      try{
+        socket.connect(new InetSocketAddress(host.getHostAddress(), hostAndPort.getPort()), connectionTimeout);
+        connected = true;
+        break;
+      } catch (ConnectException e) {
+        connectExceptions.add(e);
+      }
+    }
+    if(connected == false) {
```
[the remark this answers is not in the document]
adiamzn: agreed

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
```diff
@@ -75,7 +100,7 @@ public Socket createSocket() throws JedisConnectionException {
       socket.setSoLinger(true, 0); // Control calls close () method, the underlying socket is closed immediately
 
       HostAndPort _hostAndPort = getSocketHostAndPort();
-      socket.connect(new InetSocketAddress(_hostAndPort.getHost(), _hostAndPort.getPort()), connectionTimeout);
+      connectToFirstSuccsefulHost(socket, _hostAndPort);;
```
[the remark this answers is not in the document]
adiamzn: will do

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
adiamzn: sure np

yangbodong22011: > sure np

@adiamzn What is that mean? 

Should we replace `ConnectException` with `Exception` ?

zeekling (approved):

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
adiamzn: ohh sorrry :)

it means sure no problem

yangbodong22011: @adiamzn Does `no problem` mean that it is correct to use `ConnectException`?

But if `socket.connect()` throw `SocketTimeoutException`, it will not be caught here.

adiamzn: I apologize for not being clear.

I changed it to Exception instead of ConnectException.
I agree with your comment

yangbodong22011: @adiamzn https://github.com/redis/jedis/blob/master/src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java#L102

here, I think we should also change to `Exception`, WDYT？ if you agree, pls also update it.

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
```diff
@@ -64,6 +70,25 @@ public DefaultJedisSocketFactory(HostAndPort hostAndPort, JedisClientConfig conf
...
+      try {
+        socket.connect(new InetSocketAddress(host.getHostAddress(), hostAndPort.getPort()), connectionTimeout);
+        connected = true;
+        break;
+      } catch (Exception e) {
+        exceptions.add(e);
+      }
+    }
+    if (!connected) {
+      throw new ConnectException(exceptions.toString());
```
yangbodong22011, line 88:
```suggestion
      throw new Exception(exceptions.toString());
```

adiamzn: @yangbodong22011 I disagree with this change for 2 reasons:
1. Some unit tests depend on the exception thrown here to be a ConnectException
2. I think its better to be specific when we can.
WDYT?

sazzad16: IMO, it should be 
```
JedisConnectionException jce = JedisConnectionException("PROPER_MESSAGE");
for each exceptions => jce.addSuppressed(e);
throw jce;
```
then in createSocket()
```
if (ex instanceof JedisConnectionException) throw ex;
else throw ...;
```

yangbodong22011: @adiamzn Agree with @sazzad16, pls refer to his opinion.

adiamzn: @yangbodong22011 @sazzad16 
Thanks for the feedback!
Agreed, making the changes

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
```diff
@@ -64,6 +70,25 @@ public DefaultJedisSocketFactory(HostAndPort hostAndPort, JedisClientConfig conf
     }
   }
 
+  private void connectToFirstSuccsefulHost(Socket socket, HostAndPort hostAndPort) throws IOException {
+    List<InetAddress> hosts = Arrays.asList(InetAddress.getAllByName(hostAndPort.getHost()));
+    Collections.shuffle(hosts);
+    JedisConnectionException jce = new JedisConnectionException("Failed to connect all host resolved for DNS name.");
```
sazzad16, line 76:
```suggestion
    JedisConnectionException jce = new JedisConnectionException("Failed to connect to any host resolved for DNS name.");
```

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
```diff
@@ -64,6 +70,25 @@ public DefaultJedisSocketFactory(HostAndPort hostAndPort, JedisClientConfig conf
...
+    boolean connected = false;
```
sazzad16, line 77:
```suggestion
```

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
```diff
@@ -64,6 +70,25 @@ public DefaultJedisSocketFactory(HostAndPort hostAndPort, JedisClientConfig conf
...
+    for (InetAddress host : hosts) {
+      try {
+        socket.connect(new InetSocketAddress(host.getHostAddress(), hostAndPort.getPort()), connectionTimeout);
+        connected = true;
+        break;
```
sazzad16, line 82:
```suggestion
        return;
```

sazzad16 (changes requested):

src/test/java/redis/clients/jedis/UnavailableConnectionTest.java
```diff
@@ -62,9 +59,6 @@ public void testAvoidQuitInDestroyObjectForBrokenConnection() throws Interrupted
       fail("Should not get connection from pool");
     } catch (Exception ex) {
       assertSame(JedisConnectionException.class, ex.getClass());
-//      assertSame(JedisConnectionException.class, ex.getCause().getClass());
-//      assertSame(java.net.ConnectException.class, ex.getCause().getCause().getClass());
-      assertSame(java.net.ConnectException.class, ex.getCause().getClass());
```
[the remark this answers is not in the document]
adiamzn: will do

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
```diff
@@ -64,6 +70,25 @@ public DefaultJedisSocketFactory(HostAndPort hostAndPort, JedisClientConfig conf
     }
   }
 
+  private void connectToFirstSuccsefulHost(Socket socket, HostAndPort hostAndPort) throws IOException {
```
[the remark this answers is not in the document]
adiamzn: getAllbyName throws unkownHostException,
we can catch it and throw JedisConnectionException instead but I dont see the advantage
WDYT?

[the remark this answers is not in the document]
sazzad16: in that case, `Exception`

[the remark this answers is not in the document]
adiamzn: fair enough :)

yangbodong22011 (approved):

adiamzn:
@sazzad16 are there any further changes required?

sazzad16:
> are there any further changes required?

@adiamzn No. We're good for now. Thanks!

We'll ping you in case anything comes up. Have a nice day!

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
```diff
@@ -51,6 +57,21 @@ public DefaultJedisSocketFactory(HostAndPort hostAndPort, JedisClientConfig conf
     }
   }
 
+  private void connectToFirstSuccsefulHost(Socket socket, HostAndPort hostAndPort) throws Exception {
+    List<InetAddress> hosts = Arrays.asList(InetAddress.getAllByName(hostAndPort.getHost()));
+    Collections.shuffle(hosts);
```
dengliming: Not sure whether the size of hosts is always greater than 1? If not, maybe we can do better. e.g.
```
if (hosts.size() > 1) {
  Collections.shuffle(hosts);
}
```

sazzad16, line 62:
```suggestion
    if (hosts.size() > 1) {
      Collections.shuffle(hosts);
    }
```

dengliming (approved):

sazzad16 (approved):

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
```diff
@@ -1,8 +1,14 @@
 package redis.clients.jedis;
 
 import java.io.IOException;
+import java.net.ConnectException;
```
sazzad16, line 4:
```suggestion
```

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
```diff
@@ -1,8 +1,14 @@
...
+import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.Socket;
+import java.util.ArrayList;
```
sazzad16, line 8:
```suggestion
```

PR #2731, opened 2021-12-11
Easier constructors for pipeline classes

sazzad16:


src/main/java/redis/clients/jedis/ClusterPipeline.java
```diff
@@ -1,16 +1,40 @@
...
 
+  @Override
+  public void close() {
+    try {
+      super.close();
+    } finally {
+      IOUtils.closeQuietly(closeable);
+    }
+  }
+
```
dolavb, line 37: Does this close the cache refresher whenever we close a pipeline? Does this mean that the cache refresh is stopped whenever we close a pipeline? A pipeline does not have the same lifecyle as the connection provider, I do not understand this close here.

ggivo: @dolavb 
Looking at the newly introduced constructor they are creating a fresh  
`new ClusterConnectionProvider(clusterNodes, clientConfig)`
and only in this case preserve it for clean up later.

I think the idea is  since we create it inside the constructor, we clean it

PR #2743, opened 2021-12-15
Support TairHash commands

yangbodong22011:
This PR supports the commands of [TairHash](https://github.com/alibaba/TairHash) (A redis module, similar to redis hash, but you can set expiration and version for field).

Hash that supports field expiration has always been needed by Redis users. There have been many discussions on Redis Google Groups([this](https://groups.google.com/g/redis-db/c/GYo2XTPY03Q/m/BRnKA6YHAgAJ)) and GitHub (https://github.com/redis/redis/issues/6620, https://github.com/redis/redis/issues/1042, https://github.com/redis/redis/issues/167). 

Jedis4 already supports `RediSearch` and `RedisJSON`, hope more Modules can be supported.

Any comments or comments are welcome, thanks.


src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -1341,6 +1342,63 @@ public AggregationResult build(Object data) {
...
+    @SuppressWarnings("unchecked")
+    public ExhgetWithVerResult<String> build(Object data) {
+      if (data == null) {
+        return null;
+      }
+      List<Object> result = (List<Object>)data;
+      if (result.size() == 0) {
+        return null;
+      } else {
+        String value = SafeEncoder.encode((byte[])result.get(0));
```
zeekling, line 1357: check result size

src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -1341,6 +1342,63 @@ public AggregationResult build(Object data) {
...
+        long version = (Long)result.get(1);
+        return new ExhgetWithVerResult<>(value, version);
+      }
+    }
+
+    @Override
+    public String toString() {
+      return "exhgetwithver";
```
zeekling, line 1365:
```suggestion
      return "ExhgetWithVerResult<String>";
```

src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -1341,6 +1342,63 @@ public AggregationResult build(Object data) {
...
+      if (result.size() == 0) {
+        return null;
+      } else {
+        List<ExhgetWithVerResult<String>> results = new ArrayList<ExhgetWithVerResult<String>>();
+        for (Object o : result) {
+          if (o == null) {
+            results.add(null);
+          } else {
+            List<Object> lo = (List<Object>)o;
+            String value = SafeEncoder.encode((byte[])lo.get(0));
```
zeekling, line 1387: check list zise

src/main/java/redis/clients/jedis/tairhash/TairHashCommands.java
```diff
@@ -0,0 +1,82 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+import redis.clients.jedis.params.ScanParams;
+import redis.clients.jedis.resps.ScanResult;
+
+public interface TairHashCommands {
+
+  long exhset(String key, String field, String value);
```
zeekling, line 11: please add javadoc for redis module

src/test/java/redis/clients/jedis/modules/tairhash/TairHashTest.java
```diff
@@ -0,0 +1,200 @@
...
+import org.junit.BeforeClass;
+import org.junit.Test;
+import redis.clients.jedis.modules.RedisModuleCommandsTestBase;
+import redis.clients.jedis.params.ScanParams;
+import redis.clients.jedis.resps.ScanResult;
+import redis.clients.jedis.tairhash.ExhgetWithVerResult;
+import redis.clients.jedis.tairhash.ExhincrByParams;
+import redis.clients.jedis.tairhash.ExhsetParams;
+
+import static org.junit.Assert.*;
```
zeekling, line 19: do not use .* when import class

src/main/java/redis/clients/jedis/BuilderFactory.java
yangbodong22011: As long as the return value is not empty, it must be two elements, and the engine guarantees, so there is no need to do check

src/main/java/redis/clients/jedis/BuilderFactory.java
yangbodong22011: As long as the return value is not empty, it must be two elements, and the engine guarantees, so there is no need to do check.

dengliming:
@yangbodong22011 Hey, That‘s a great support 👍 . But Jedis seems to be mainly supporting the redis `official` Module for now. WDYT? @sazzad16

sazzad16:
@dengliming I asked and got neither positive nor negative response.

sazzad16:
Apparently, the current plan is to support [Redis Stack](https://redis.io/docs/stack/) commands in Jedis. And, TairHash is not part of RS.

PR #2748, opened 2021-12-16
Maven deploy (snapshot) on specific branches in CI

chayim:
Notice that the CI job name changed to ci/circleci: build. Effectively I removed a matrix that wasn't being used and simplified things. However, if we would like to build and test in parallel across multiple JDKs this can be restored. I would argue that's outside the scope of this PR.


.circleci/config.yml
```diff
@@ -55,16 +60,36 @@ jobs:
...
-          matrix:
-            parameters:
-              os: [linux-8-jdk]
+      - build_and_deploy:
+          <<: *on-integ-branches
+          context: common
+          requires:
+            - build-linux-8-jdk
+      - build-linux-8-jdk:
+        <<: *on-all-branches
```
sazzad16, line 95:
```suggestion
          <<: *on-all-branches
```

sazzad16 (approved):

PR #2762, opened 2021-12-23
Fix tests

AvitalFineRedis:


chayim:
@AvitalFineRedis rather than delete the flaky tests what do you think of disabling them somehow? This way, they can be analyzed and possibly reused in the future.

AvitalFineRedis:
> @AvitalFineRedis rather than delete the flaky tests what do you think of disabling them somehow? This way, they can be analyzed and possibly reused in the future.

fixed

sazzad16:
@AvitalFineRedis JUnit has `@Ignore` annotation. You can put that before each test class declaration.

chayim (approved):

PR #2768, opened 2021-12-26
Added the BYTE|BIT option for bitcount (binary)

zeekling:
Added the BYTE|BIT option for bitcount

image


sazzad16 (commented):
Return type for both bitcountByte and bitcountBit is same. IMO, it's better to create a BitCountOption in `args` package and just do `bitcount(key, start, end, bitcountOption)`.

zeekling:
done

src/main/java/redis/clients/jedis/CommandObjects.java
```diff
@@ -516,6 +516,10 @@ protected CommandArguments commandArguments(ProtocolCommand command) {
     return new CommandObject<>(commandArguments(BITCOUNT).key(key).add(start).add(end), BuilderFactory.LONG);
   }
 
+  public final CommandObject<Long> bitcount(byte[] key, long start, long end, BitCountOption option) {
+    return new CommandObject<>(commandArguments(BITCOUNT).key(key).add(start).add(end).add(option.getRaw()), BuilderFactory.LONG);
```
sazzad16, line 520:
```suggestion
    return new CommandObject<>(commandArguments(BITCOUNT).key(key).add(start).add(end).add(option), BuilderFactory.LONG);
```

sazzad16 (changes requested):
Also, add `bitcount(String key, long start, long end, BitCountOption option)` variants.

yangbodong22011 (approved):

src/main/java/redis/clients/jedis/CommandObjects.java
zeekling: done

sazzad16 (approved):

Issue #2791, opened 2022-01-01
support LMPOP numkeys key


zeekling:
https://redis.io/commands/lmpop

zeekling:
supported in https://github.com/redis/jedis/pull/2796

Razi007:
@zeekling   @sazzad16  interested to work here.Please suggest if i need to refer any code.

Razi007:
@zeekling @sazzad16 Please assign it to me.Will raise a PR in a while code is completed almost.Working fine just writing test cases.

sazzad16:
> Please assign it to me.

@Razi007 Not necessary. You can work on your own without assignment.

Akaame:
@Razi007 have you started working on this?

Razi007:
@Akaame Yes i am working on it.Will raise asap.

Razi007:
@sazzad16 raise a PR https://github.com/redis/jedis/pull/2796/files. Please review once and let me know  is approach  correct.Please suggest if any improvement needed.Currently Working on test cases and checking if we should add lmpop in other interfaces as well.

PR #2796, opened 2022-01-03
support LMPOP command

zeekling:
support LMPOP command

https://github.com/redis/jedis/issues/2791


src/main/java/redis/clients/jedis/CommandObjects.java
```diff
@@ -800,6 +800,14 @@ protected CommandArguments commandArguments(ProtocolCommand command) {
         .key(dstKey).add(from).add(to).add(timeout), BuilderFactory.STRING);
   }
 
+  public final CommandObject<Map<String, List<String>>> lmpop(ListDirection from, String... keys) {
+    return new CommandObject<>(commandArguments(LMPOP).add(keys.length).keys((Object[]) keys).add(from), BuilderFactory.SEARCH_SYNONYM_GROUPS);
+  }
+
+  public final CommandObject<Map<String, List<String>>> lmpop(ListDirection from, int count, String... keys) {
+    return new CommandObject<>(commandArguments(LMPOP).add(keys.length).keys((Object[]) keys).add(from).add(COUNT).add(count), BuilderFactory.SEARCH_SYNONYM_GROUPS);
+  }
```
sazzad16, line 809: IMO, we can create a new `KeyedListElements` similar to `KeyedListElement` by using using `List<String> elements` instead of `String element` and use that as return type here. [1]

src/main/java/redis/clients/jedis/CommandObjects.java
```diff
@@ -809,6 +817,15 @@ protected CommandArguments commandArguments(ProtocolCommand command) {
...
         .key(dstKey).add(from).add(to).add(timeout), BuilderFactory.BINARY);
   }
+
+  public final CommandObject<Map<byte[], List<byte[]>>> lmpop(ListDirection from, byte[]... keys) {
+    return new CommandObject<>(commandArguments(LMPOP).add(keys.length).keys((Object[]) keys).add(from), BuilderFactory.GROUPS_LIST_BINARY);
+  }
+
+  public final CommandObject<Map<byte[], List<byte[]>>> lmpop(ListDirection from, int count, byte[]... keys) {
+    return new CommandObject<>(commandArguments(LMPOP).add(keys.length).keys((Object[]) keys).add(from).add(COUNT).add(count), BuilderFactory.GROUPS_LIST_BINARY);
+  }
```
sazzad16, line 827: If we agree on [1], we can return `List<Object>` here.

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/CommandObjects.java
zeekling: added

src/main/java/redis/clients/jedis/CommandObjects.java
zeekling: why not `List<KeyedListElements>` ?

zeekling: > @zeekling To leave a variant so that users can make their own implementation.

ok, i see.

zeekling: i modified it @sazzad16

Avital-Fine:
closes #2791

sazzad16:
Closed by a08fa3e

PR #2814, opened 2022-01-11
Support `ZRANGESTORE` and REV, BYLEX and BYSCORE arguments to `ZRANGE`

AvitalFineRedis:
This PR covers 2 misses:
1. Support REV, BYLEX and BYSCORE arguments to ZRANGE by adding **ZRangeParams** Class
2. Support ZRANGESTORE command (also using ZRangeParams Class)

Covers #2668 and #2645
Closes #2625

PR #2668, opened 2021-10-18
Support ZRANGESTORE

AvitalFineRedis:

PR #2817, opened 2022-01-13
Support new ZRANGE(STORE)

### LINKED ISSUE #2625: ZRANGE support [BYSCORE|BYLEX] [REV] [LIMIT offset count] [WITHSCORES] option

sazzad16:
Based on #2814 by @AvitalFineRedis 

Resolves #2625 
Closes #2645 
Closes #2668 
Closes #2814


sazzad16:
@yangbodong22011 @dengliming @AvitalFineRedis 

So this is what I could think at this moment.
Let me know WDYT about this.

dengliming (approved):

dengliming:
LGTM

src/main/java/redis/clients/jedis/params/ZRangeParams.java
```diff
@@ -0,0 +1,108 @@
...
+  private final Keyword by;
+  private final Rawable min;
+  private final Rawable max;
+  private boolean rev = false;
+
+  private boolean limit = false;
+  private int offset;
+  private int count;
+
+  private ZRangeParams() {
```
yangbodong22011, line 24: Why we need a `private and empty args` constructor?

sazzad16: @yangbodong22011 It's not uncommon to use Jedis with reflection. I've added this contributor so that to reduce issues when used with reflection because two of mandatory arguments of zrange command are included in this class. Our other params classes are fine because they don't contain mandatory arguments.

yangbodong22011: @sazzad16 ok, thanks for your explain.

yangbodong22011 (approved):

PR #2818, opened 2022-01-13
Support XPENDING with exclusive range

AvitalFineRedis:


AvitalFineRedis:
One of the problems I noticed here is that it's breaking the previous use of the `xpending` function when both (start and end parameters) are **null**. Maybe we can rename the new function?

src/main/java/redis/clients/jedis/CommandObjects.java
```diff
@@ -2195,6 +2195,14 @@ protected CommandArguments commandArguments(ProtocolCommand command) {
     return new CommandObject<>(args, BuilderFactory.STREAM_PENDING_ENTRY_LIST);
   }
 
+  public final CommandObject<List<StreamPendingEntry>> xpending(String key, String groupname,
+      String start, String end, int count, String consumername) {
+    CommandArguments args = commandArguments(XPENDING).key(key).add(groupname)
+            .add(start == null ? "-" : start).add(end == null ? "+" : end).add(count);
+    if (consumername != null) args.add(consumername);
+    return new CommandObject<>(args, BuilderFactory.STREAM_PENDING_ENTRY_LIST);
```
sazzad16, line 2203: Don't support `null` in methods for exclusive support.

sazzad16 (commented):
> Maybe we need to rename the new function?

Ignore the conflict. Assume that users have to type-cast themselves.

@yangbodong22011 @dengliming I don't think we can afford delaying it for Jedis 5.0.0 (similar to how we delayed #2683 for 4.0.0). So I'm thinking about including this in next minor release.

src/main/java/redis/clients/jedis/CommandObjects.java
```diff
@@ -2195,6 +2195,14 @@ protected CommandArguments commandArguments(ProtocolCommand command) {
     return new CommandObject<>(args, BuilderFactory.STREAM_PENDING_ENTRY_LIST);
   }
 
+  public final CommandObject<List<StreamPendingEntry>> xpending(String key, String groupname,
+      String start, String end, int count, String consumername) {
```
zeekling, line 2199: add `XpendingOption` in args

src/test/java/redis/clients/jedis/commands/jedis/StreamsCommandsTest.java
```diff
@@ -442,6 +441,28 @@ public void xack() {
       jedis.xack("xack-stream", "xack-group", range.get(0).getValue().get(0).getID()));
   }
 
+  @Test
+  public void xpendingRange() {
+    Map<String, String> map = new HashMap<>();
+    map.put("foo", "bar");
+    StreamEntryID m1 = jedis.xadd("xpendeing-stream", (StreamEntryID) null, map);
```
zeekling, line 448: null can not convert to `StreamEntryID`

src/test/java/redis/clients/jedis/commands/jedis/StreamsCommandsTest.java
```diff
@@ -647,7 +668,7 @@ public void xautoclaimJustIdBinary() {
 
     // Get the pending event
     List<StreamPendingEntry> pendingRange = jedis.xpending("xpending-stream", "xpending-group",
-            null, null, 3, "xpending-consumer");
+            (StreamEntryID) null, null, 3, "xpending-consumer");
```
zeekling, line 671: null can not convert to StreamEntryID

zeekling (changes requested):

src/test/java/redis/clients/jedis/commands/jedis/StreamsCommandsTest.java
AvitalFineRedis: Not in this case - there are two different signatures for `xadd`, so It has to be explicitly converted.

src/main/java/redis/clients/jedis/CommandObjects.java
AvitalFineRedis: It's problematic - because if the user not supplying `start` and `end` params, the `addParams` function adds the defaults '-' and '+'. But if the user does supply `start` and `end` as Strings, he can't use exclusive range.

Any ideas what to do?

zeekling: Is this OK？
```java
 /**
  * Should be used only with XGROUP CREATE  
  * 
  * <code>
   * XGROUP CREATE mystream consumer-group-name $
   * </code>
   */
  public static final StreamEntryID LAST_ENTRY = new StreamEntryID() {

    private static final long serialVersionUID = 1L;

    @Override
    public String toString() {
      return "$";
    }
  };
```

PR #2821, opened 2022-01-17
Support SINTERCARD Command

AvitalFineRedis:


sazzad16 (approved):

src/main/java/redis/clients/jedis/commands/SetCommands.java
```diff
@@ -42,6 +42,10 @@
 
   long sinterstore(String dstkey, String... keys);
 
+  long sintercard(String... keys);
```
zeekling, line 45: We'd better add Javadoc when adding new interfaces

zeekling (dismissed):

src/main/java/redis/clients/jedis/commands/SetCommands.java
AvitalFineRedis: I added Javadoc in Jedis.py, and non of the commands here are documented 
Do you want me to add it here also?

dengliming (dismissed):

src/main/java/redis/clients/jedis/commands/SetCommands.java
sazzad16: @AvitalFineRedis It's better to add javadoc in interfaces. Then you may or may not add javadoc in classes.

sazzad16 (approved):

Issue #2795, opened 2022-01-01
support CLUSTER DELSLOTSRANGE start-slot end-slot


zeekling:
https://redis.io/commands/cluster-delslotsrange

sazzad16:
Resolved by #2823

sazzad16:
Resolved by #2823

Issue #2792, opened 2022-01-01
support CLUSTER ADDSLOTSRANGE start-slot end-slot


zeekling:
https://redis.io/commands/cluster-addslotsrange

sazzad16:
Resolved by #2823

PR #2823, opened 2022-01-17
support cluster addslotsrange and cluster delslotsrange

zeekling:
support cluster addslotsrange and cluster delslotsrange


zeekling:
issue:https://github.com/redis/jedis/issues/2795  https://github.com/redis/jedis/issues/2792

src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -8297,6 +8297,22 @@ public String clusterMyId() {
             .map(BuilderFactory.ENCODED_OBJECT_MAP::build).collect(Collectors.toList());
   }
 
+  @Override
+  public String clusterAddSlotsRange(int... ranges) {
```
dengliming, line 8301: I don't think it's a good idea to use `int... ranges` as args here. Because the start and end slots are required. My suggestion is 
You can introduce new class `SlotRange` like the below code.
```
class SlotRange {
   int startSlot;
   int endSlot;
}
```
then you can do in this way  clusterAddSlotsRange(SlotRange... slotRanges) {}

dengliming (changes requested):

src/main/java/redis/clients/jedis/Jedis.java
sazzad16: @dengliming I think this is okay. Many users prefer this way. Also, we have similar pattern in other commands, e. g. MSET.

We can add your suggestion alongside these methods.

dengliming: You were right. There is no doubt that this is the easiest way. but users must avoid passing only one parameter. e.g. 
clusterAddSlotsRange(1). Or we can check the params. `if (ranges.length % 2 != 0) { throw err }`

sazzad16: You're right but Jedis doesn't do that, at least not yet :laughing:

PR #2829, opened 2022-01-18
Add javadoc to interfaces

AvitalFineRedis:


sazzad16 (commented):
If you're changing the variable names in interfaces (e.g. member -> members, string -> strings, arg -> args), you should also rename those in implemented methods (in classes).

sazzad16 (changes requested):
1. Don't use `@see` tags for URLs. You can add URLs in description if you want.
2. When using `@see` tag (except above [1]), Clarify how respective method is different from the method in `@see`. (Or, simply avoid `@see` and add whole description.)

src/main/java/redis/clients/jedis/commands/DatabaseCommands.java
```diff
@@ -35,22 +35,65 @@
...
+   * @param timeout the maximum idle time in any moment of the communication with the
+   *               destination instance in milliseconds.
+   * @param params {@link MigrateParams}
+   * @param keys
+   * @return OK on success, or NOKEY if no keys were found in the source instance.
+   */
   String migrate(String host, int port, int destinationDB, int timeout, MigrateParams params,
       String... keys);
 
+  String migrate(String host, int port, int destinationDB, int timeout, MigrateParams params,
```
zeekling, line 97: Please add javadoc for this method

src/main/java/redis/clients/jedis/commands/GeoCommands.java
```diff
@@ -13,67 +13,439 @@
...
+   */
   List<GeoRadiusResponse> georadius(String key, double longitude, double latitude, double radius,
       GeoUnit unit);
 
+  /**
+   * Readonly version of {@link GeoCommands#georadius(String, double, double, double, GeoUnit) GEORADIUS},
+   * <p>
+   * Time complexity: O(N+log(M)) where N is the number of elements inside the bounding box of
+   * the circular area delimited by center and radius and M is the number of items inside the index.
+   * @param key
```
zeekling, line 120: Please add description for params.

zeekling (changes requested):

src/main/java/redis/clients/jedis/commands/KeyCommands.java
```diff
@@ -11,74 +11,500 @@
...
+   * Test if the specified key exist.
+   * <p>
+   * Time complexity: O(N)
+   * @param key
+   * @return True is the key exists, False otherwise
+   */
   boolean exists(String key);
 
+  /**
+   * Similar to {@link KeyCommands#exists(String) EXISTS} with multiple keys.
```
sazzad16, line 25: Elaborate; return type is different. What does the return mean?

src/main/java/redis/clients/jedis/commands/SortedSetCommands.java
```diff
@@ -11,112 +11,608 @@
...
 
+  /**
+   * Similar to {@link SortedSetCommands#zadd(String, double, String) ZADD} for multiple members.
+   * @param key
+   * @param scoreMembers
+   */
   long zadd(String key, Map<String, Double> scoreMembers);
 
+  /**
+   * Similar to {@link SortedSetCommands#zadd(String, Map<String, Double>) ZADD} with optional params.
```
sazzad16, line 49: Method you're referring to does not contain any detail but refers to another one. Don't do that.

src/main/java/redis/clients/jedis/commands/SortedSetCommands.java
```diff
@@ -11,112 +11,608 @@
...
+   * @param min
+   * @param max
+   * @param offset
+   * @param count
+   * @return A List of elements in the specified score range
+   */
   List<String> zrangeByScore(String key, double min, double max, int offset, int count);
 
+  /**
+   * Similar to {@link SortedSetCommands#zrevrangeByScore(String, double, double) ZREVRANGEBYSCORE} with string range.
```
sazzad16, line 387:
```suggestion
   * Similar to {@link SortedSetCommands#zrevrangeByScore(String, double, double) ZREVRANGEBYSCORE} with <i>exclusive</i> range.
```

src/main/java/redis/clients/jedis/commands/ListCommands.java
```diff
@@ -2,48 +2,291 @@
 
 import java.util.List;
 
+import redis.clients.jedis.args.GeoUnit;
```
sazzad16, line 5:
```suggestion
```

sazzad16 (changes requested):
Please re-check all your `Similar ...`. It would be better to use a method to use as Similar which has actual description.

Also, `Similar...`s are very good place to use `@see` tag.

Furthermore, use full method name. E.g. `KeyCommands#sort(...)` instead of just `#sort(...)`.

Avital-Fine:
Can be closed due to #2842

PR #2837, opened 2022-01-24
support the SCAN TYPE option

alessandro-modolo:
Added method to provide an easy support to the [TYPE](https://redis.io/commands/scan#the-type-option) option of the SCAN command


sazzad16:
@alessandro-modolo Jedis already has `ScanResult<String> scan(String cursor, ScanParams params, String type)` method. We've avoided updating ScanParams because it is also used in hscan/sscan/zscan methods which do not have type option.

alessandro-modolo:
@sazzad16 thanks for the very quick response. I had not seen that overload. In that case I would say you can close this PR.

Thank you 1000

PR #2862, opened 2022-02-01
support arg ABORT for shutdown command

zeekling:
support arg ABORT for shutdown command


sazzad16:
@zeekling Jedis already supports SHUTDOWN ABORT

https://github.com/redis/jedis/blob/71162fce391a100f6698ae45618b0f0b6e21c0a0/src/main/java/redis/clients/jedis/Jedis.java#L3264

PR #2870, opened 2022-02-05
LMPOP

Razi007:


sazzad16 (commented):
From https://redis.io/commands/lmpop

> A two-element array with the first element being the name of the key from which elements were popped, and the second element is an array of elements.

In summary, this is a single list along with respective key which makes just a single map entry.

@Razi007 Why do want to return an entire `Map` with just one entry?

Razi007:
> From https://redis.io/commands/lmpop
> 
> > A two-element array with the first element being the name of the key from which elements were popped, and the second element is an array of elements.
> 
> In summary, this is a single list along with respective key which makes just a single map entry.
> 
> @Razi007 Why do want to return an entire `Map` with just one entry?

@sazzad16 Got your point.I was assuming it might return multiple keys with respective elements.Changing it and pushing the code for the same.

Razi007:
> > From https://redis.io/commands/lmpop
> > > A two-element array with the first element being the name of the key from which elements were popped, and the second element is an array of elements.
> > 
> > 
> > In summary, this is a single list along with respective key which makes just a single map entry.
> > @Razi007 Why do want to return an entire `Map` with just one entry?
> 
> @sazzad16 Got your point.I was assuming it might return multiple keys with respective elements.Changing it and pushing the code for the same.

src/main/java/redis/clients/jedis/resps/KeyedListElement.java
```diff
@@ -2,13 +2,16 @@
...
+import java.awt.*;
+
 /**
  * This class is used to represent a List element when it is returned with respective key name.
  */
 public class KeyedListElement {
 
   private final String key;
   private final String element;
+  private final List elementList;
```
Razi007, line 14: @sazzad16 added a new list value type here.Please suggest can we proceed this way.

Razi007:
closed by mistake

src/main/java/redis/clients/jedis/resps/KeyedListElement.java
sazzad16: @Razi007 I don't like how it has been handled later, like `this.elementList = new List();` and `this.element = "";`.

sazzad16 (commented):
@Razi007 Your methods for binary variants have issues which are not apparent due to lack of tests. A personal suggestion, Implement String variants (without Pipeline) first. After this is done properly, then implement binary ones. And after that Pipeline variants.

src/main/java/redis/clients/jedis/resps/KeyedListValueElements.java
```diff
@@ -0,0 +1,26 @@
...
+
+
+import java.util.List;
+
+
+public class KeyedListValueElements<K,V> {
+
+  private final K key;
+  private final List<V> elements;
+
```
Razi007, line 11: @sazzad16 implemented a new class here.Using generic key values.

src/test/java/redis/clients/jedis/commands/jedis/ListCommandsTest.java
```diff
@@ -875,4 +876,25 @@ public void blmove() {
     assertByteArrayListEquals(Collections.singletonList(b3), jedis.lrange(bbar, 0, -1));
     assertByteArrayListEquals(Arrays.asList(b1, b2), jedis.lrange(bfoo, 0, -1));
   }
+
+  @Test
+  public void lmpop(){
+    KeyedListValueElements keyedListValueElements;
```
Razi007, line 882: @sazzad16 added all test cases here.String args are working fine.Please check the approach

sazzad16:
Closed by a08fa3e

PR #2924, opened 2022-02-24
Keyed util class and (B)(L/Z)MPOP commands

sazzad16:


sazzad16:
@dengliming @yangbodong22011 Please review. After we agree on (non-pipeline) String variants, I'll add binary variants and then Pipelines.

dengliming (dismissed):
`Keyed` or `KeyValue` both good to me. : )

sazzad16:
> `Keyed` or `KeyValue` both good to me. : )

@dengliming I thought of having a dedicated class/interface for `List`s and KeyedList seemed much better than KeyList. So, instead of KeyValue, I opted for KeyedValue or simply Keyed.

Now I'm thinking, may be we don't need Key(ed)List at all?

Avital-Fine (dismissed):

PR #2930, opened 2022-02-28
Add RFC6901 json pointer support

leonchen83:
Now Path can convert RFC6901 style path to json path
for example
Path.ofJosnPointer("/abc/0") can convert to Path.of(".[\"abc\"].[0]")

the limitation of JsonPointer path is /abc/0 always convert to .[\"abc\"].[0] and not .[\"abc\"].[\"0\"]
if users want to use .abc.0, they should use Path.of(".abc.0") instead of Path.ofJosnPointer("/abc/0")


sazzad16:
@leonchen83 We generally avoid adding utilities. Perhaps RedisJSON itself can start supporting RFC6901?

leonchen83:
@sazzad16 
yes. but since 2017 to present. there is not an implement from RedisJSON.
related issue [#3](https://github.com/RedisJSON/RedisJSON/issues/3) , [#18](https://github.com/RedisJSON/RedisJSON/issues/18)

uglide:

PR #2934, opened 2022-03-05
add CRC32Hashing

SeaSand1024:


gkorland:
@SeaSand1024 can you please extend about this PR? why is it needed?

SeaSand1024:
> @SeaSand1024 can you please extend about this PR? why is it needed?
When we use RedisShardAPI in python with crc32hash, we can't compatible with Jedis,In addition, we will also add sharding algorithms,because it is different from python.

PR #2936, opened 2022-03-07
add crc32hash and compatible with RedisShardAPI

SeaSand1024:
When we use RedisShardAPI in python with crc32hash, we can't compatible with Jedis,In addition, we will also add sharding algorithms,because it is different from python.


sazzad16:
@SeaSand1024 Have you checked if this implementation is actually compatible with RedisShardAPI?

SeaSand1024:
> @SeaSand1024 Have you checked if this implementation is actually compatible with RedisShardAPI?
Yes, I did. You'd better specify the format of HostAndPort the same as python if you use ShardedConnectionProvider.The default format is "host:port".

SeaSand1024:
> @SeaSand1024 Have you checked if this implementation is actually compatible with RedisShardAPI?



> @SeaSand1024 Have you checked if this implementation is actually compatible with RedisShardAPI?

Have you checked if this implementation is actually compatible with RedisShardAPI?

SeaSand1024:
> @SeaSand1024 Have you checked if this implementation is actually compatible with RedisShardAPI?

Yes, I did. You'd better specify the format of HostAndPort the same as python if you use ShardedConnectionProvider.The default format is "host:port".

sazzad16:
Sharding feature will be removed.

PR #2951, opened 2022-03-18
Support CF.MEXISTS command

doumdoum:


src/main/java/redis/clients/jedis/UnifiedJedis.java
```diff
@@ -3922,6 +3923,9 @@ public boolean bfExists(String key, String item) {
 
   @Override
   public List<Boolean> bfMExists(String key, String... items) {
+    if (items.length == 0) {
+      return Collections.emptyList();
+    }
```
sazzad16, line 3928:
```suggestion
```
We don't do this in Jedis.

src/main/java/redis/clients/jedis/UnifiedJedis.java
```diff
@@ -3975,6 +3979,15 @@ public boolean cfExists(String key, String item) {
     return executeCommand(commandObjects.cfExists(key, item));
   }
 
+  @Override
+  public List<Boolean> cfMExists(String key, String... items) {
+    if (items.length == 0) {
+      return Collections.emptyList();
+    }
```
sazzad16, line 3986:
```suggestion
```
We don't do this in Jedis.

src/test/java/redis/clients/jedis/modules/bloom/CuckooTest.java
```diff
@@ -180,12 +181,25 @@ public void testInsertNxNoCreate() {
...
+
+  @Test
+  public void testNoItemMExists() {
+    assertEquals(Collections.emptyList(), client.cfMExists("cuckoo17"));
+  }
+
+  @Test
+  public void testMixedItemsMExists() {
+    client.cfInsert("cuckoo17", "foo");
+    assertEquals(Arrays.asList(true, false), client.cfMExists("cuckoo17", "foo", "bar"));
```
sazzad16, line 202:
```suggestion
    client.cfInsert("cuckoo18", "foo");
    assertEquals(Arrays.asList(true, false), client.cfMExists("cuckoo18", "foo", "bar"));
```

src/main/java/redis/clients/jedis/CommandObjects.java
```diff
@@ -3567,6 +3567,11 @@ protected CommandArguments commandArguments(ProtocolCommand command) {
     return new CommandObject<>(commandArguments(CuckooFilterCommand.EXISTS).key(key).add(item), BuilderFactory.BOOLEAN);
   }
 
+  public final CommandObject<List<Boolean>> cfMExists(String key, String... items) {
+    return new CommandObject<>(commandArguments(CuckooFilterCommand.MEXISTS).key(key).
+            addObjects((Object[]) items), BuilderFactory.BOOLEAN_LIST);
```
sazzad16, line 3572:
```suggestion
    return new CommandObject<>(commandArguments(CuckooFilterCommand.MEXISTS).key(key)
        .addObjects((Object[]) items), BuilderFactory.BOOLEAN_LIST);
```

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/UnifiedJedis.java
doumdoum: Hi @sazzad16, varargs for items in function signature implicitly assumes 0 or more but redis expects 1 or more so we have to account for the 0 length (or we have to tell in documentation that 1 or more items are expected but i'd rather stick to regular understanding of function signature).

sazzad16: @doumdoum `bfMExists` is not a simple straightforward word. The user who is calling bfMExists(), we assume that s/he has some idea about `BF.MEXISTS` command. We also take that documentation provided by Redis, e.g. https://oss.redis.com/redisbloom/Bloom_Commands/#bfmexists, is our prime documentation. Finally, we try to keep it as close to plain command as possible.

doumdoum: @sazzad16 i can't agree with you: as a user of jedis i should not need to know redis protocol details, i should only rely on jedis API. Anyway i've remove the checks from the PR according to your proposal.

sazzad16 (approved):

PR #2959, opened 2022-03-27
support command CLIENT REPLY ON|OFF|SKIP

zeekling:
support command CLIENT REPLY ON|OFF|SKIP


sazzad16 (commented):
What would a user do with the `jedis` object after calling clientReplyOff() or clientReplySkip() methods? Because all other methods (commands) would be broken as those would try to read their responses. We need a better idea.

zeekling:
Below is the explanation given in redis.io，and I'm not sure anyone would use this

image

sazzad16:
@zeekling Yes, exactly. If we introduce something supporting fire & forget operations, we can use this command internally. It may be better to avoid adding this command to Jedis class.

Closing this PR for now.

Issue #2961, opened 2022-03-29
Why JSON support (gson)?


sebdehne:
I really appreciate the good work which has good into this excellent library, but why do we need JSON capabilities? Should a redis-client be pure? 
For example there is a security issue with GSON (https://github.com/google/gson/pull/1991) which now becomes everybodies problem who uses Jedis. 
Would it be possible to at least make JSON support optional such that the library still functions even without gson on the classpath?

Thanks a lot. If you need help - I can submit a change which makes gson optional.

sazzad16:
> can submit a change which makes gson optional.

@sebdehne You're always welcome to craft a PR.

sazzad16:
@sebdehne Sorry about slow progress on this.

WDYT about replacing Gson completely with Jackson?

sazzad16:
Thanks to #3223, GSON can now be excluded.

Keeping this issue open as the dependency is not exactly `optional`.

PR #2962, opened 2022-03-30
Made GSON dependency optional 

sebdehne:
As suggested in https://github.com/redis/jedis/issues/2961

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -66,10 +66,13 @@
 			<artifactId>json</artifactId>
 			<version>20211205</version>
 		</dependency>
+
+		<!-- optional GSON, used for JSON-commands -->
 		<dependency>
 			<groupId>com.google.code.gson</groupId>
 			<artifactId>gson</artifactId>
-			<version>2.8.9</version>
+			<version>2.9.0</version>
+			<scope>provided</scope>
 		</dependency>
 
 		<dependency>
--- a/src/main/java/redis/clients/jedis/json/util/JsonUtils.java
+++ b/src/main/java/redis/clients/jedis/json/util/JsonUtils.java
@@ -1,0 +1,53 @@
+package redis.clients.jedis.json.util;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import redis.clients.jedis.exceptions.JedisException;
+
+public final class JsonUtils {
+
+    private static final Logger logger = LoggerFactory.getLogger(JsonUtils.class);
+
+    private static volatile JsonTools JSON_TOOLS;
+
+    static {
+        JSON_TOOLS = loadGsonTools();
+    }
+
+    private JsonUtils() {
+    }
+
+    public static JsonTools getJsonTools() {
+        return JSON_TOOLS;
+    }
+
+    public static void setJsonTools(JsonTools jsonTools) {
+        JSON_TOOLS = jsonTools;
+    }
+
+    public static String toJson(Object src) {
+        if (JSON_TOOLS == null) {
+            throw new JedisException("GSON not found on classpath, please add com.google.code.gson:gson to your classpath or inject your own impl. of JsonTools");
+        }
+        return JSON_TOOLS.toJson(src);
+    }
+
+    public static <T> T fromJson(String json, Class<T> classOfT) {
+        if (JSON_TOOLS == null) {
+            throw new JedisException("GSON not found on classpath, please add com.google.code.gson:gson to your classpath or inject your own impl. of JsonTools");
+        }
+        return JSON_TOOLS.fromJson(json, classOfT);
+    }
+
+    private static JsonTools loadGsonTools() {
+        try {
+            Class jsonUtilsGson =
+                    Thread.currentThread().getContextClassLoader().loadClass("redis.clients.jedis.json.util.JsonUtilsGson");
+            return (JsonTools) jsonUtilsGson.getDeclaredConstructor().newInstance();
+        } catch (Exception e) {
+            logger.debug("Could not load optional GSON: ", e);
+            return null;
+        }
+    }
+
+}
--- a/src/main/java/redis/clients/jedis/CommandObjects.java
+++ b/src/main/java/redis/clients/jedis/CommandObjects.java
@@ -3,7 +3,6 @@ package redis.clients.jedis;
 import static redis.clients.jedis.Protocol.Command.*;
 import static redis.clients.jedis.Protocol.Keyword.*;
 
-import com.google.gson.Gson;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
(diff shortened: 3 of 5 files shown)


dengliming (commented):
It make sense to expose the interface for setting the JSON serialization method. Allows users to customize the serialization implementation. Not just only `Gson/Jackson`

sebdehne:
> It make sense to expose the interface for setting the JSON serialization method. Allows users to customize the serialization implementation. Not just only `Gson/Jackson`

Added public getter/setter. 

I am not a huge fan of this static stuff (JsonTools should preferably be configured as part of JedisConfiguration), but that would require a larger change. As the PR is now, it is more aligned with the current way of doing it.

pom.xml
```diff
@@ -66,13 +67,35 @@
...
 			<artifactId>gson</artifactId>
-			<version>2.8.9</version>
+			<version>2.9.0</version>
+      <scope>provided</scope>
 		</dependency>
+    <!-- optional Jackson, used for JSON-commands -->
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+      <version>${jackson.version}</version>
```
zeekling, line 82: suggest 2.13.2.2

image

src/main/java/redis/clients/jedis/util/json/JsonUtils.java
```diff
@@ -0,0 +1,71 @@
+package redis.clients.jedis.util.json;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import redis.clients.jedis.exceptions.JedisException;
+
+public class JsonUtils {
```
zeekling, line 8:
```suggestion
public final class JsonUtils {
```

src/main/java/redis/clients/jedis/util/json/JsonUtils.java
```diff
@@ -0,0 +1,71 @@
...
+        try {
+            Class jsonUtilsJackson =
+                Thread.currentThread().getContextClassLoader().loadClass("redis.clients.jedis.util.json.JsonUtilsGson");
+            return (JsonTools) jsonUtilsJackson.getDeclaredConstructor().newInstance();
+        } catch (Exception e) {
+            logger.debug("Could load optional GSON: ", e);
+            return null;
+        }
+    }
+
```
zeekling, line 64:
```suggestion
private JsonUtils(){
}
```

zeekling (changes requested):

pom.xml
gokhanoner: @sebdehne you need to add a separate `jackson.databind.version` param, since for now, there's a specific patch version of jackson-databind for a vulnerability fix.

sazzad16 (commented):
I like the concept. But few points I'd like to add/discuss:

- Ignore Jackson from source. We can add examples in test about how to use Jackson.
  - This way we can avoid adding more dependencies.
  - This is something [similar](https://github.com/redis/jedis/blob/master/src/test/java/redis/clients/jedis/UdsTest.java) we have done for `UDS` connections.
- Simply use `jackson-bom` (in test).
- Package should be `.../jedis/json/util/`.
- `JsonTools` doesn't need to be an inner interface.
- We should code in such a way that in future it's easier to change `CommandObjects` to have one `JsonTools`.

pom.xml
sebdehne: Thanks for the heads-up, but as soon as I change it to the latest, a new version will be release. The whole point of this change is that Jedis shall NOT require/dictate any version. Therefore the dependency is marked as <scope>provided</scope>. So there is no need really to keep up with the latest version.

sebdehne:
> 

@sazzad16 I moved the code to the package you suggested and remove Jackson. I have not spent the time to setup a complete test environment (yet). Maybe if you need further changes with regards to tests, you could supplement those? Thanks
Would be nice to get this through, as I am currently unable to upgrade to 4.x

src/main/java/redis/clients/jedis/json/util/JsonUtils.java
```diff
@@ -0,0 +1,53 @@
...
+        return JSON_TOOLS;
+    }
+
+    public static void setJsonTools(JsonTools jsonTools) {
+        JSON_TOOLS = jsonTools;
+    }
+
+    public static String toJson(Object src) {
+        if (JSON_TOOLS == null) {
+            throw new JedisException("GSON found on classpath, please add com.google.code.gson:gson to your classpath or inject your own impl. of JsonTools");
```
Marcono1234, line 30: Shouldn't this say "GSON <ins>not</ins> found on classpath"?

src/main/java/redis/clients/jedis/json/util/JsonUtils.java
```diff
@@ -0,0 +1,53 @@
...
+        }
+        return JSON_TOOLS.toJson(src);
+    }
+
+    public static <T> T fromJson(String json, Class<T> classOfT) {
+        if (JSON_TOOLS == null) {
+            throw new JedisException("GSON found on classpath, please add com.google.code.gson:gson to your classpath or inject your own impl. of JsonTools");
```
Marcono1234, line 37: Shouldn't this say "GSON <ins>not</ins> found on classpath"?

src/main/java/redis/clients/jedis/json/util/JsonUtils.java
```diff
@@ -0,0 +1,53 @@
...
+        return JSON_TOOLS.fromJson(json, classOfT);
+    }
+
+    private static JsonTools loadGsonTools() {
+        try {
+            Class jsonUtilsGson =
+                    Thread.currentThread().getContextClassLoader().loadClass("redis.clients.jedis.json.util.JsonUtilsGson");
+            return (JsonTools) jsonUtilsGson.getDeclaredConstructor().newInstance();
+        } catch (Exception e) {
+            logger.debug("Could load optional GSON: ", e);
```
Marcono1234, line 48: Shouldn't this say "Could <ins>not</ins> load optional GSON"?

Marcono1234 (commented):
Sorry for this "drive-by" review; hopefully the comments are helpful.

src/main/java/redis/clients/jedis/json/util/JsonUtils.java
sebdehne: thanks!

src/main/java/redis/clients/jedis/json/util/JsonUtils.java
sebdehne: thanks!

src/main/java/redis/clients/jedis/json/util/JsonUtils.java
sebdehne: thanks!

sazzad16:
Closing in favor of #3223

Issue #3017, opened 2022-06-09
Fix return value of HRANDFIELD With Values when count is negative


BalaKaza:
`###` Expected behavior

`hrandfieldWithValues(String key, long count)` should return the count equivalent entries in case of a negative (absolute value of count) or a positive count value.

### Actual behavior

`hrandfieldWithValues(String key, long count)` currently returns a `Map<String, String>` which does not account for the duplicates that can be returned(https://redis.io/commands/hrandfield/) when a negative count is given. This results in an inaccurate number (less than the count) of random entries being returned when there is a negative count.

### Steps to reproduce:
   
1. Create a Hash Set with size 5
2. Call `hrandfieldWithValues(String key, long count)` with a count of -3.
3. Assert for the size of the result.

```
Map<String, String> hash = createHash(5); 
jedis.hset(KEY, hash); 
long count = -3L; 
Map<String, String> result = jedis.hrandfieldWithValues(KEY, count); 
assertThat(result.size()).isEqualTo(Math.abs(count));
```

s-sathish:
As per Redis documentation -

> When the count is a negative value, the behavior changes as follows:
> 
> Repeating fields are possible.
> Exactly count fields, or an empty array if the hash is empty (non-existing key), are always returned.
> The order of fields in the reply is truly random.

So output may contain repeating fields when the count is a negative value.
Current implementation uses "_Map<String, String>_" for response builder which will not account for repeating fields.

@sazzad16 A map which stores collection of values for a key will work good here ?
Something like this -
```
public static final Builder<Map<String, List<String>>> STRING_MULTI_MAP = new Builder<Map<String, List<String>>>() {
    @Override
    @SuppressWarnings("unchecked")
    public Map<String, List<String>> build(Object data) {
      final List<byte[]> flatHash = (List<byte[]>) data;
      final Map<String, List<String>> hash = new HashMap<>(flatHash.size() / 2, 1);
      final Iterator<byte[]> iterator = flatHash.iterator();
      while (iterator.hasNext()) {
        String key = STRING.build(iterator.next());
        String value = STRING.build(iterator.next());
        List<String> list = hash.containsKey(key) ? hash.get(key) : new ArrayList<>();
        list.add(value);
        hash.put(key, list);
      }

      return hash;
    }

    @Override
    public String toString() {
      return "Map<String, List<String>>";
    }

};
```
The new return type will be "Map<String, List<String>>". Count of collections for all keys will be equivalent to the absolute value of input count.

[...]

PR #3027, opened 2022-06-15
Fix return value of HRANDFIELD With Values when count is negative #3017

s-sathish:
Fix return value of HRANDFIELD With Values when count is negative #3017

### DIFF
--- a/src/main/java/redis/clients/jedis/BuilderFactory.java
+++ b/src/main/java/redis/clients/jedis/BuilderFactory.java
@@ -302,6 +302,26 @@ public final class BuilderFactory {
     }
   };
 
+  public static final Builder<List<Map.Entry<byte[], byte[]>>> BINARY_PAIR_LIST = new Builder<List<Map.Entry<byte[], byte[]>>>() {
+    @Override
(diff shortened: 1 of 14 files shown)


sazzad16:
@s-sathish Thank you for your effort. Please note that this is a breaking change and we cannot include this change before Jedis 5.0. So merging will be delayed for quite sometime.

s-sathish:
@sazzad16 Agreed. It's a breaking change and should be part of next major release.

Just asking as I'm unaware of the process - If someone needs this fix immediately, what is the ideal way to do it ? It's like we need to create a fork from latest master and patch this commit on top of that ? Build it and then use it ?

sazzad16:
@s-sathish Yes.

sazzad16:
@s-sathish Yes, that's one option.

Or, you can simply use the `sendCommand()`/`executeCommand()` methods.

s-sathish:
Sorry. Accidentally deleted the fork and pull request was closed.
Raised a new one - https://github.com/redis/jedis/pull/3049
@sazzad16 Please check and add milestone, labels

Issue #3044, opened 2022-06-20
Extend support to all arguments of FT.AGGREGATE command


sazzad16:
https://redis.io/commands/ft.aggregate/

s-sathish:
Hi @sazzad16 Can this be closed ? I see all the arguments are now supported for FT.AGGREGATE command.
The extended one's are -
```
VERBATIM
TIMEOUT
PARAMS
DIALECT
```

sazzad16:
@s-sathish We will.

Issue #3043, opened 2022-06-20
Extend support to all arguments of FT.SEARCH command


sazzad16:
https://redis.io/commands/ft.search/

PR #3054, opened 2022-07-03
Extend support to all arguments of FT.SEARCH and FT.AGGREGATE commands

s-sathish:
Extend support to all arguments of FT.SEARCH, FT.AGGREGATE commands. https://github.com/redis/jedis/issues/3043 & https://github.com/redis/jedis/issues/3044


src/main/java/redis/clients/jedis/search/SearchResult.java
```diff
@@ -33,12 +33,16 @@ public static class SearchResultBuilder extends Builder<SearchResult> {
     private final boolean hasContent;
     private final boolean hasScores;
     private final boolean hasPayloads;
+    private final boolean hasSortKeys;
+    private final boolean hasExplainScore;
     private final boolean decode;
 
-    public SearchResultBuilder(boolean hasContent, boolean hasScores, boolean hasPayloads, boolean decode) {
+    public SearchResultBuilder(boolean hasContent, boolean hasScores, boolean hasPayloads, boolean hasSortKeys, boolean hasExplainScore, boolean decode) {
```
sazzad16, line 40: Breaking change. We can't _modify_ this now.

src/main/java/redis/clients/jedis/search/aggr/Params.java
```diff
@@ -0,0 +1,20 @@
+package redis.clients.jedis.search.aggr;
+
+public class Params {
```
sazzad16, line 3: We can find other alternatives of adding this new class.

src/test/java/redis/clients/jedis/modules/search/AggregationBuilderTest.java
```diff
@@ -83,6 +84,17 @@ public void testAggregations() {
...
+    r = new AggregationBuilder()
+        .groupBy("@name", Reducers.sum("@count").as("sum"))
+        .verbatim()
+        .timeout(1000)
+        .params(new Params("@name", "abc"))
+        .dialect(2);
+
+    // actual search
+    res = client.ftAggregate(index, r);
+    assertEquals(2, res.totalResults);
```
sazzad16, line 97: It'd be nicer to actually test the effects of the arguments.
Also, can be considered a separate test method.

src/main/java/redis/clients/jedis/search/Query.java
```diff
@@ -192,6 +198,9 @@ public void addParams(CommandArguments args) {
     if (_withPayloads) {
       args.add(SearchKeyword.WITHPAYLOADS.getRaw());
     }
+    if (_withSortKeys) {
+      args.add(SearchKeyword.WITHSORTKEYS.getRaw());
```
sazzad16, line 202: The doc says:

> This is usually not needed by users, and exists for distributed search coordination purposes.

I would prefer to ignore this for now.

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/search/aggr/Params.java
angelo147: Could a simple Map be a good alternative?

src/main/java/redis/clients/jedis/search/Query.java
s-sathish: Ok. Removed the changes related to sortKeys.
Will create a separate PR for this to be included in 5.0.0 release. Anyways we can just give the option for user to use it or not.

src/test/java/redis/clients/jedis/modules/search/AggregationBuilderTest.java
s-sathish: Wrote individual tests for each arguments to test their effects.

src/main/java/redis/clients/jedis/search/aggr/Params.java
s-sathish: Thanks for the suggestion. Replaced with Map<String, Object>

src/main/java/redis/clients/jedis/search/SearchResult.java
s-sathish: Ok. Removed the changes related to sortKeys & explainScore from this PR.
Will create a separate PR for those to be included in 5.0.0 release.

src/main/java/redis/clients/jedis/search/Query.java
```diff
@@ -309,6 +313,25 @@ public void addParams(CommandArguments args) {
       args.add(SearchKeyword.DIALECT.getRaw());
       args.add(_dialect);
     }
+
+    if (_slop >= 0) {
+      args.add(SearchKeyword.SLOP.getRaw());
+      args.add(_slop);
+    }
+
+    if (_timeout > 0) {
```
sazzad16, line 322:
```suggestion
    if (_timeout >= 0) {
```

src/main/java/redis/clients/jedis/search/aggr/AggregationBuilder.java
```diff
@@ -149,4 +149,36 @@ public String getArgsString() {
   public boolean isWithCursor() {
     return isWithCursor;
   }
+
+  public AggregationBuilder verbatim() {
+    args.add(1, "VERBATIM");
```
sazzad16, line 154: Let the user call in proper order.
```suggestion
    args.add("VERBATIM");
```

src/test/java/redis/clients/jedis/modules/search/SearchTest.java
```diff
@@ -1360,4 +1360,60 @@ public void synonym() {
     expected.put("child", Arrays.asList(group1_str, group2_str));
     assertEquals(expected, dump);
   }
+
+  @Test
```
sazzad16, line 1364: I see a separate dialect test in JsonSearchTest. Can we have one here as well?

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/search/aggr/AggregationBuilder.java
s-sathish: Updated it. But just curious on why we don't want the library to handle it ?
Because we know that this argument should be the first for this command if it is used.

src/main/java/redis/clients/jedis/search/aggr/AggregationBuilder.java
```diff
@@ -149,4 +149,36 @@ public String getArgsString() {
...
+    }
+
+    return this;
+  }
+
+  public AggregationBuilder dialect(int dialect) {
+    args.add("DIALECT");
+    args.add(Integer.toString(dialect));
+    return this;
+  }
```
sazzad16, line 183: Move this block above `getArgs()`.

sazzad16 (changes requested):

PR #3056, opened 2022-07-05
Adding link to the OM client

chayim:


sazzad16 (changes requested):
I don't agree with the placement. My suggestion:
- Either at the end of **Using Redis modules** block (my preference)
- Or at the end of **What is Jedis?** block

PR #3061, opened 2022-07-07
[WIP] Re-work RediSearch with Jedis-style params

sazzad16:


src/main/java/redis/clients/jedis/args/LazyRaw.java
```diff
@@ -0,0 +1,18 @@
+package redis.clients.jedis.args;
+
+public class LazyRaw implements Rawable {
```
bsbodden, line 3: Can you explain the name of the class I see the "raw" part as having a `byte[]` payload but I don't get the "lazy" part?

src/main/java/redis/clients/jedis/params/GeoRadiusParam.java
```diff
@@ -18,8 +17,7 @@ public class GeoRadiusParam implements IParams {
 
   private Integer count = null;
   private boolean any = false;
-  private boolean asc = false;
```
bsbodden, line 21: Would it be better to use an Enum for this case, so we don't get into a case where `asc` and `desc` have conflicting values?

src/main/java/redis/clients/jedis/params/GeoRadiusParam.java
```diff
@@ -91,10 +92,8 @@ public void addParams(CommandArguments args) {
       }
     }
 
-    if (asc) {
```
bsbodden, line 94: Exactly this... Maybe just an ASC/DESC that can be used everywhere in Jedis?

src/test/java/redis/clients/jedis/modules/search/AggregationWithParamsTest.java
```diff
@@ -0,0 +1,297 @@
...
+import java.util.Map;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import redis.clients.jedis.exceptions.JedisDataException;
+import redis.clients.jedis.search.*;
+import redis.clients.jedis.search.aggr.*;
+import redis.clients.jedis.modules.RedisModuleCommandsTestBase;
+
+public class AggregationWithParamsTest extends RedisModuleCommandsTestBase {
```
bsbodden, line 16: I'm assuming the commented-out tests are placeholders from JRediSearch?

bsbodden (dismissed):
The only things that smelled was the ASC/DESC but you might have deeper reason for it being two boolean values so I'll defer to you.

src/main/java/redis/clients/jedis/params/GeoRadiusParam.java
sazzad16: @bsbodden That's what I did. These are deleted lines. Please check added line 20.

src/main/java/redis/clients/jedis/params/GeoRadiusParam.java
sazzad16: @bsbodden Check added line 95-96.

src/test/java/redis/clients/jedis/modules/search/AggregationWithParamsTest.java
sazzad16: @bsbodden Kinda yes but these are also available in src/test/java/redis/clients/jedis/modules/search/AggregationBuilderTest.java. Things are still in progress :smiley:

src/main/java/redis/clients/jedis/args/LazyRaw.java
sazzad16: @bsbodden All our existing implementation of `Rawable` are such that `getRaw()` can be called immediately to get the proper value. But here I needed to create an object but I would know the actual value later on. And, among all the names (something ending in Raw) that I could think of LazyRaw seemed least bad. Do you have any suggestion?

PR #3082, opened 2022-07-20
Support new TimeSeries arguments

sazzad16:
- LATEST
- EMPTY
- TWA
- BUCKETTIMESTAMP
- ALIGNTIMESTAMP


chayim (changes requested):
Perhaps docstrings? Sure - we don't have these everywhere.. but it's good to start

PR #3097, opened 2022-08-07
Support TDIGEST commands

sazzad16:


src/test/java/redis/clients/jedis/modules/bloom/TDigestTest.java
```diff
@@ -0,0 +1,171 @@
...
+  private void assertMergedUnmergedNodes(String key, int mergedNodes, int unmergedNodes) {
+    Map<String, Object> info = client.tdigestInfo(key);
+    assertEquals(Long.valueOf(mergedNodes), info.get("Merged nodes"));
+    assertEquals(Long.valueOf(unmergedNodes), info.get("Unmerged nodes"));
+  }
+
+  private void assertTotalWeight(String key, double totalWeight) {
+    Map<String, Object> info = client.tdigestInfo(key);
+    assertEquals(totalWeight, Double.parseDouble((String) info.get("Merged weight"))
+        + Double.parseDouble((String) info.get("Unmerged weight")), 0.01);
```
shacharPash, line 37: Why there is 0.01 in the assertEquals?
the total weight should not always be always equal to 0.01, no?

sazzad16: @shacharPash I don't get what your exact question is but here is my first attempt to answer:

JUnit, the library we based our tests upon, deprecated [assertEquals(double, double)](https://junit.org/junit4/javadoc/4.13/org/junit/Assert.html#assertEquals(double,%20double)) and suggested to use [assertEquals(double, double, double)](https://junit.org/junit4/javadoc/4.13/org/junit/Assert.html#assertEquals(double,%20double,%20double)).

chayim (approved):

PR #3118, opened 2022-08-25
Makes getBackoffSleepMillis in ClusterCommandExecutor nondeterministic

adiamzn:
Currently sleepBackOffMillis is deterministic.

As can be seen by the benchmarks taken in this [blog post](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/), this out performed by backoff methods which add Jitter.

In the terminology of the blog post referenced, this PR introduces the "Full Jitter" method, with minimal code changes.


sazzad16:
@yangbodong22011 @dengliming Your reviews are always appreciated.

@walles @jensgreen You were first to introduce backoff. WDYT about this change?

yangbodong22011:
@adiamzn Hi, thanks for you PR, we don't have lock contention, just network queuing, so I'm not sure what your PR advantage is? Specifically:

1. Suppose there are 10 JedisCluster client. At t1, Redis is down or ha, so the client access fails and starts to retry.
2. t1+0.4 s starts for the first time, and still all fails (redis has not recovered yet)
3. t1 + 0.6s start the second time, same as above.
4. t1 + 0.8s Redis Cluster returns to normal.
5. t1 + 1.0s start the third time, all successful.

Your PR may only be useful in eliminating traffic spikes, but not in speeding up the time to success. Is that what you're trying to do with this PR?

adiamzn:
@yangbodong22011 
Thanks for the comment,
Yes this PR is to mitigate traffic spikes, especially spikes in new connections which can be caused my multiple clients retrying commands at the same time. Spikes in new connections can be especially problematic for Redis clusters which serve a large number of connections.

yangbodong22011:
> @yangbodong22011 Thanks for the comment, Yes this PR is to mitigate traffic spikes, especially spikes in new connections which can be caused my multiple clients retrying commands at the same time. Spikes in new connections can be especially problematic for Redis clusters which serve a large number of connections.

Can you share some diagrams or specific issues you encountered in production?

adiamzn:
@yangbodong22011 
It is common in large production environments to see spikes in new connection requests following an event in which the cluster was unresponsive.
This scenario is most easily solved by adding some backoff with jitter to the retry policy in the client.
Unfortunately I don't think I can share any diagrams or specific scenarios from production.
I think this [blog post](https://aws.amazon.com/builders-library/timeouts-retries-and-backoff-with-jitter/) summarizes the benefit of adding jitter quite well.

Thanks!

yangbodong22011:
> Unfortunately I don't think I can share any diagrams or specific scenarios from production.

@adiamzn OK, but you can test some data and compare the new algorithm you proposed with the old one to make everyone understand it better？

adiamzn:
@yangbodong22011 
Ideally yes, but this could take me a few weeks.
I think it would make sense to compare CPU usage and traffic spikes in a high client scenarios using this algorithm and the previous one.
Since this can take me a some time to get to, would you prefer I close this PR and open a new one when I have some results?
Thanks

yangbodong22011:
@adiamzn Thanks, I don't think a new PR is needed, we can just keep working on this PR and keep the stack for others to see.

adiamzn:
@yangbodong22011 @sazzad16 
I think its worth mentioning that similar jitter mechanisms exists in some other popular client libraries
redispy: https://github.com/redis/redis-py/pull/1494
phpredis: https://github.com/phpredis/phpredis/pull/1986
envoy proxy: https://github.com/envoyproxy/envoy/pull/19869/files

sazzad16:
@yangbodong22011 @adiamzn I am thinking about going forward with this change.

Before saying further, two equations:
A = `millisLeft / (attemptsLeft * attemptsLeft)`
B = `millisLeft / (attemptsLeft * (attemptsLeft + 1))`

In original PR where backOff time was added, the time was calculated as A. But in my testing I found that few tests here and there would get timeout exceeded. So I was looking for something **definitely smaller than A**, and I went with B.

Now in this PR, total sleep time is _exptected to be_ **half** of total calculated backOff time. Contrary to current code where total sleep time is _equals to_ total calculated time.

WDYT about considering A again but with this change (jitter) as sleep time is expected to be half of A which is smaller than A.

BTW, I have added tests of ClusterCommandExecutor in #3139

src/main/java/redis/clients/jedis/executors/ClusterCommandExecutor.java
```diff
@@ -148,7 +149,8 @@ private static long getBackoffSleepMillis(int attemptsLeft, Instant deadline) {
       throw new JedisClusterOperationException("Cluster retry deadline exceeded.");
     }
 
-    return millisLeft / (attemptsLeft * (attemptsLeft + 1));
+    long maxBackOff = millisLeft / (attemptsLeft * (attemptsLeft + 1));
```
sazzad16, line 152:
```suggestion
    long maxBackOff = millisLeft / (attemptsLeft * attemptsLeft);
```

adiamzn: @sazzad16 
Glad to hear :)
Regarding the two possible equations A and B.
I am OK with either

PR #3162, opened 2022-10-10
Add dependabot configuration

yeikel:
See https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "maven"
+    directory: "/"
+    schedule:
+      interval: "daily"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "maven"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
chayim, line 6: IMHO let's do this weekly. Daily is far too much noise with dependabot.

yeikel: Fail enough. I addressed that comment with 8ebcdd5ba72a8e387b6bcbf4dc9f249d2a83e16d

PR #3180, opened 2022-10-27
Handle Non-ANSI chars in error message

### LINKED ISSUE #3179: Garbled message when ERR message contains Non-ANSI chars
### Expected behavior

We use redis-proxy-server to access redis clusters, and then use jedis to access redis-proxy-server.
The proxy server hold some HA and Consistency functions based on our business cases, and will return some customized ERR messages with Non-ANSI chars(such as Chinese Char like ”ERR 配置错误。“) to our client users who use jedis to connect and operate redis servers indirectly accross our proxy servers.
And then we should expected the client get the err message correctly.

### Actual behavior

It will be garbled message.

### Steps to reproduce:

Redis server return a response message like:
`“-ERR 配置错误。”+CRLF`

#### Jedis version:
all current branches

#### Redis version:
all

#### Java version:
JDK 1.8/11

#### Cause Reason
In `RedisInputStream` class, the method `readLine` read a byte and put it as a char into StringBuilder.
But a Non-ANSI char here will be 2-4 bytes, generally a Chinese char will be 3 bytes.
And then readLine will get 3 char instead of a Non-ANSI char, the garbled message occurs.

#### Suggestion
So we guess there are at least two ways to avoid this case:
1. Use `readLineBytes` method instead of `readLine`, when call `Protocal.processError()`
2. Add an option like "is_support_non_ansi_char" to turn on/off it to switch readLine and readLineBytes

kimmking:
fixed #3179 to avoid garbled message when use Non-ANSI chars in err message.


zeekling:
Why redis-proxy-server return chinese error message ? 

And Can redis-proxy-server return english message?

@kimmking

kimmking:
> Why redis-proxy-server return chinese error message ?
> 
> And Can redis-proxy-server return english message?
> 
> @kimmking

That’s a choice. Cause the proxy has some biz logic and return some biz exceptions that contains Chinese chars.
Now  we use a URLencode to transform non ANSI chars to ANSI chars before response to jedis client to avoid this case.

PR #3194, opened 2022-11-06
Introduce JedisBroadcast to broadcast commands

sazzad16:


zeekling (dismissed):

yangbodong22011 (commented):
I prefer not to introduce `JedisBroadcast`, but directly modify the default behavior of commands such as `cluster.scriptLoad` and `cluster.ftCreate` (because the previous behavior is actually wrong, so we don't need to keep it).

Adding `JedisBroadcast` certainly won't cause a bug, but I think Jedis should keep it simple and quiet to fix the behavior.

sazzad16:
@yangbodong22011 Are you okay with other things? I.e. changes in ConnectionProvider interface and implementations, method signatures, etc?

yangbodong22011:
> @yangbodong22011 Are you okay with other things? I.e. changes in ConnectionProvider interface and implementations, method signatures, etc?

That's OK, but I tend to avoid having `PooledConnectionProvider` provide a `getConnectionMap()` method.

chayim:
> > @yangbodong22011 Are you okay with other things? I.e. changes in ConnectionProvider interface and implementations, method signatures, etc?
> 
> That's OK, but I tend to avoid having `PooledConnectionProvider` provide a `getConnectionMap()` method.

@yangbodong22011 why?

src/main/java/redis/clients/jedis/BroadcastResponse.java
```diff
@@ -0,0 +1,25 @@
+package redis.clients.jedis;
+
+import java.util.function.Supplier;
+
+public class BroadcastResponse<T> implements Supplier<T> {
```
chayim, line 5: I know we don't have a lot of them, but docstrings here would be a great start. Let's help users understand why these exist.

src/main/java/redis/clients/jedis/JedisBroadcast.java
```diff
@@ -0,0 +1,130 @@
...
+
+  private final ConnectionProvider provider;
+
+  public JedisBroadcast(UnifiedJedis jedis) {
+    this(jedis.provider);
+  }
+
+  public JedisBroadcast(ConnectionProvider provider) {
+    if (provider == null) {
+      throw new NullPointerException("ConnectionProvider is null.");
```
chayim, line 30:
```
      throw new NullPointerException("ConnectionProvider cannot be null.");
```

src/main/java/redis/clients/jedis/JedisBroadcast.java
```diff
@@ -0,0 +1,130 @@
...
+  private <T> T executeCommand(Object connection, CommandObject<T> commandObject) {
+    if (connection instanceof Connection) {
+      return ((Connection) connection).executeCommand(commandObject);
+    }
+    if (connection instanceof Pool) {
+      try (Connection _conn = ((Pool<Connection>) connection).getResource()) {
+        return _conn.executeCommand(commandObject);
+      }
+    }
+    throw new IllegalStateException(connection.getClass() + "is not supported.");
```
chayim, line 59: Any reason to not:

if (connection instanceof Connection)
else if (connection instanceof Pool)
// throw here

Reducing multiple ifs to a single if else one makes thing function cheaper.

yangbodong22011:
> > > @yangbodong22011 Are you okay with other things? I.e. changes in ConnectionProvider interface and implementations, method signatures, etc?
> > 
> > 
> > That's OK, but I tend to avoid having `PooledConnectionProvider` provide a `getConnectionMap()` method.
> 
> @yangbodong22011 why?

`PooledConnectionProvider` usually corresponds to one node in the backend rather than multiple nodes. Exposing the interface(getConnectionMap) is only for the neatness of the code, but it will confuse the user.

chayim:
> `PooledConnectionProvider` usually corresponds to one node in the backend rather than multiple nodes. Exposing the interface(getConnectionMap) is only for the neatness of the code, but it will confuse the user.

I think that in the interest of consistency and quality, we should do this. If anything, a docstring can always be added in order to help users out.

uglide:
Continuing the 

> I prefer not to introduce `JedisBroadcast`, but directly modify the default behavior of commands such as `cluster.scriptLoad` and `cluster.ftCreate` (because the previous behavior is actually wrong, so we don't need to keep it).

Continuing the @yangbodong22011 idea. @sazzad16 Why not override commands and make them broadcasted under the hood in the `JedisCluster`?

sazzad16:
@uglide Please check #3211, it addresses your concerns. We can discuss if you have further concerns.

uglide:
@sazzad16 I like the plug-in approach in this PR and don't like the new `xxxBroadcast` commands in the https://github.com/redis/jedis/pull/3211 :)

Could you explain why we can't add broadcast commands to the JedisCluster instead? From my understanding, users are interested in "broadcasting" using JedisCluster in their codebases. Am I missing something?

yangbodong22011:
@sazzad16 Now I vote 3194 instead of 3211 because:
1, `scriptExists` v.s. `scriptExistsBroadcast` I chose `scriptExists`.
2. I think the code of the 3194 plug-in is cleaner.

src/test/java/redis/clients/jedis/JedisClusterTest.java
```diff
@@ -780,4 +781,17 @@ private boolean isAnyNodeHandshaking(Jedis node) {
...
     return false;
   }
+
+  @Test
+  public void broadcast() {
+    try (JedisCluster cluster = new JedisCluster(Collections.singleton(new HostAndPort(LOCAL_IP, 7379)),
+        DefaultJedisClientConfig.builder().password("cluster").build())) {
+      JedisBroadcast broadcast = new JedisBroadcast(cluster);
+      
+      Map<?, Supplier<String>> replies = broadcast.broadcastCommand(new CommandObject<>(
```
yangbodong22011, line 791: may be we should add `PING`JedisBoardcast.java ?

sazzad16:
> Could you explain why we can't add broadcast commands to the `JedisCluster` instead? From my understanding, users are interested in "broadcasting" using `JedisCluster` in their codebases. Am I missing something?

@uglide We can, but then...

We would miss these methods from other classes (UnifiedJedis, JedisPooled, JedisSharding). Let's discuss only with JedisPooled. We can add these methods in JedisPooled separately but then concerned methods cannot be (easily) switched between Redis standalone mode and Redis cluster mode. This was long ask from users which we just added in Jedis 4.0.0 and I am trying to keep common base between JedisCluster and JedisPooled.

src/test/java/redis/clients/jedis/JedisClusterTest.java
sazzad16: @yangbodong22011 After we decided on a design, we can add many commands :slightly_smiling_face:

uglide:
@sazzad16 Thanks for explaining the context. This PR suits user needs better because it offers a "plug-in" approach, just like @yangbodong22011 mentioned in the comment above. If there are no objections, we should proceed with this PR and add more "broadcasted" commands.

sazzad16:
@yangbodong22011 @uglide 

- Added `broadcasted()` method, so the code can now be fluent. Naming is such for similarity with existing `pipelined()`.
- Added PING

Please review.

src/main/java/redis/clients/jedis/JedisBroadcast.java
```diff
@@ -0,0 +1,172 @@
...
+    return broadcastCommand(new CommandObject<>(new CommandArguments(Command.PING),
+        BuilderFactory.STRING));
+  }
+
+  public Map<?, Supplier<String>> configSet(final String... parameterValues) {
+    if (parameterValues.length > 0 && parameterValues.length % 2 == 0) {
+      // ok
+    } else {
+      throw new IllegalArgumentException("It requires 'pair's of config parameter-values.");
+    }
```
yangbodong22011, line 72:
```suggestion
    if (parameterValues.length > 0 && parameterValues.length % 2 != 0) {
      throw new IllegalArgumentException("It requires 'pair's of config parameter-values.");
    }
```

PR #3195, opened 2022-11-12
Address empty cluster slots list when discovering slots

jgaoIXL:
This is a bug that I encountered where `renewClusterSlots` was called when one node was removed from the cluster. The removed node happened to get selected to do the `discoverClusterSlots`. Because the removed node return no cluster slots, we have an empty slots cache, and there is no recovery possible from there since the nodes and slots map are all empty. 

I think it makes sense to return if slotsInfo is size 0 because under no normal circumstances, the cluster should have 0 cluster info provided, it can only mean one thing, which is you are pointing to a node that is not in the cluster.


zeekling (dismissed):
It seems reasonable.

sazzad16 (changes requested):
Let's use `isEmpty()` method.

jgaoIXL:
> Let's use `isEmpty()` method.

Fixed, also `discoverClusterNodesAndSlots` should probably throw an exception when this happens.

yangbodong22011:
@jgaoIXL Jedis assumes that the `initial node` of the cluster is `constant` (that is, it always belongs to this cluster)，whether it is vip, domain, or physical ip.

But I agree to merge this PR, as a defensive protection.

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -139,6 +143,9 @@ public void renewClusterSlots(Connection jedis) {
 
   private void discoverClusterSlots(Connection jedis) {
     List<Object> slotsInfo = executeClusterSlots(jedis);
+    if (slotsInfo.isEmpty()) {
+      return;
```
yangbodong22011, line 147:
```suggestion
      throw new JedisClusterException("Cluster slots list is empty");
```

yangbodong22011: throw JedisClusterException, catch and ignore it, and try next nodes.

jgaoIXL: Fixed

yangbodong22011:
[here](https://github.com/redis/jedis/pull/3195/files#diff-4a02ef8d8a92f62e06be741dac2e739e37916ea33c7f0cb5b82ab040361adbcbR106) need catch `JedisClusterException`.

src/main/java/redis/clients/jedis/providers/ClusterConnectionProvider.java
```diff
@@ -41,7 +42,7 @@ private void initializeSlotsCache(Set<HostAndPort> startNodes, JedisClientConfig
       try (Connection jedis = new Connection(hostAndPort, clientConfig)) {
         cache.discoverClusterNodesAndSlots(jedis);
         return;
-      } catch (JedisConnectionException e) {
+      } catch (JedisConnectionException | JedisClusterException e) {
```
yangbodong22011, line 45:
```suggestion
      } catch (JedisException e) {
```

src/main/java/redis/clients/jedis/providers/ClusterConnectionProvider.java
```diff
@@ -14,6 +14,7 @@
 import redis.clients.jedis.Connection;
 import redis.clients.jedis.ConnectionPool;
 import redis.clients.jedis.JedisClusterInfoCache;
+import redis.clients.jedis.exceptions.JedisClusterException;
```
yangbodong22011, line 17:
```suggestion
```

yangbodong22011: Consistent with the rest.

src/main/java/redis/clients/jedis/providers/ClusterConnectionProvider.java
yangbodong22011: Consistent with the rest.

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -58,6 +59,9 @@ public JedisClusterInfoCache(final JedisClientConfig clientConfig,
 
   public void discoverClusterNodesAndSlots(Connection jedis) {
     List<Object> slotsInfo = executeClusterSlots(jedis);
+    if (slotsInfo.isEmpty()) {
+      throw new JedisClusterException("Cluster slots list is empty");
```
sazzad16, line 63:
```suggestion
      throw new JedisClusterOperationException("Cluster slots list is empty.");
```

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -14,6 +14,7 @@
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
 
+import redis.clients.jedis.exceptions.JedisClusterException;
```
sazzad16, line 17:
```suggestion
import redis.clients.jedis.exceptions.JedisClusterOperationException;
```

src/main/java/redis/clients/jedis/providers/ClusterConnectionProvider.java
sazzad16, line 17:
```suggestion
```

src/main/java/redis/clients/jedis/providers/ClusterConnectionProvider.java
sazzad16, line 45:
```suggestion
      } catch (JedisConnectionException | JedisClusterOperationException e) {
```

sazzad16 (changes requested):
The exception should be `JedisClusterOperationException`. WIP #3204

sazzad16 (approved):

yangbodong22011 (approved):

jgaoIXL:
Thank you. When are we supposed to see a stable release with this fix? @yangbodong22011

sazzad16:
@jgaoIXL We don't have an exact date yet but this change is available in 4.4.0 snapshot.

PR #3198, opened 2022-11-14
Test: Ensure test case supports repeated execution

dormanze:


sazzad16:
FLUSHALL is executed in all master nodes before each test.

https://github.com/redis/jedis/blob/ae29bce70fb9d3b3d3f2fad4f164b533e6f7c2d0/src/test/java/redis/clients/jedis/commands/jedis/ClusterBinaryValuesCommandsTest.java#L23 

https://github.com/redis/jedis/blob/ae29bce70fb9d3b3d3f2fad4f164b533e6f7c2d0/src/test/java/redis/clients/jedis/commands/jedis/ClusterJedisCommandsTestBase.java#L32-L42 

Proposed change, that too in only one test, is not necessary.

PR #3205, opened 2022-11-16
Replace org.json:json with jackson-databind

sazzad16:


dengliming:
@sazzad16 Hello~ May I ask if there is any particular purpose in this? Is it possible we could consider gson? : )

dengliming (approved):

Issue #3189, opened 2022-10-31
Consider switching to org.json implementation without "don't be evil" license restriction


philwebb:
### Expected behavior
It should be possible to use Jedis without needing to have lawyers accept the "don't be evil" clause of `org.json`.

### Actual behavior
As of commit 858c8051301854e8343eef66aa68c6b9dab1c328 Jedis has a mandatory dependency on `org.json:json`. This library has an unfortunate "don't be evil" clause added to its MIT license which causes legal problems for some organizations.

See https://wiki.debian.org/qa.debian.org/jsonevil and https://www.apache.org/legal/resolved.html#json for additional background.

Perhaps the library could be optional (similar to #2961) or replaced with `com.vaadin.external.google:android-json` which is a clean room implementation of the API without any license issues.

### Steps to reproduce:

n/a

### Redis / Jedis Configuration

n/a

#### Jedis version:

4.0 and above

#### Redis version:

n/a

#### Java version:

n/a

sazzad16:
@philwebb Thank you for creating this issue including possible alternates.

`org.json:json` cannot be made optional because the integrated APIs are exposed. It is possible with Gson because it is used only internally.

It can only be replaced. The idea of `com.vaadin.external.google:android-json` is interesting as we may not have to break our APIs.

sazzad16:
@philwebb WDYT about breaking Jedis in release 4.4.0 to resolve this issue?

References:
- #3205 
- #3206

vpavic:
I just learned that `org.json:json` recently changed their license to remove the _The Software shall be used for Good, not Evil_ clause. Unfortunately, the replacement is something that appears to be ambiguous. See here for more details:

- https://github.com/stleary/JSON-java/pull/688
- https://github.com/stleary/JSON-java/issues/706

chayim:
In the ideal world, we would not make any changes - because changes like the ones contemplated, mean API changes. API changes like this break users, and I'll err on the side of not doing this.

Prior to choosing a path, it would be great to at least see what the response to [the JSON issue][https://github.com/stleary/JSON-java/issues/706] will be.

vpavic:
I agree, waiting to see what's the outcome of `org.json` licensing discussion seems to be like the most reasonable approach.

Until then, users that are affected by `org.json`'s present license can exclude that dependency and pull in `com.vaadin.external.google:android-json` as a workaround.

vpavic:
[...]

PR #3206, opened 2022-11-16
Replace both org.json and gson with jackson-databind

sazzad16:


zeekling:
Is it the same as the pr below：
https://github.com/redis/jedis/pull/3205

sazzad16:
@zeekling This PR is an extension of #3205 which replaces only org.json:json. This also replaces gson.

sazzad16:
@chayim @dengliming @yangbodong22011 

Do you have like or dislike for this or #3205 ?

sazzad16:
@chayim @dengliming @yangbodong22011 

Do you have like or dislike for this or #3205 ?

Either of these is a breaking change (affects only RedisJSON commands) but we are considering for Jedis 4.4.0.

yangbodong22011:
> @chayim @dengliming @yangbodong22011
> 
> Do you have like or dislike for this or #3205 ?
> 
> Either of these is a breaking change (affects only RedisJSON commands) but we are considering for Jedis 4.4.0.

I read the related issue, is there an option for us to wait for the dust to settle on the license changes in this [comment](https://github.com/redis/jedis/issues/3189#issuecomment-1319824273) before making a decision?

Another question is why we rely on both `org.json` and `google.gson` now, can only relying on `google.gson` not meet our needs?

PR #3211, opened 2022-11-21
Broadcast commands from UnifiedJedis

sazzad16:


sazzad16:
@chayim @yangbodong22011 @dengliming 

Do you have like or dislike for this or #3194 ?

Here we are lengthening already huge UnifiedJedis. But it is arguably easier user experience.

zeekling:
I think this is better.

chayim:
> I think this is better.

@zeekling can you qualify why?

zeekling:
> > I think this is better.
> 
> @zeekling can you qualify why?

The new code calling method has not undergone breakthrough changes and is easy to use.

sazzad16:
Closed by #3194

PR #3223, opened 2022-12-06
Allow exclusion of GSON dependency

agavrilov76:
Move creation of the GSON instance to a separate class which is initialized only when any of the JSON methods are being used. This allows `gson` transitive Maven dependency to be excluded for the applications which don't use JSON methods.

See #2961 and  #2962 for more information and rationale


sazzad16 (changes requested):
This PR addresses only Gson. Please update title and description accordingly.

agavrilov76:
> This PR addresses only Gson. Please update title and description accordingly.

Done. The title and description are updated

chayim (approved):

sazzad16 (approved):

PR #3224, opened 2022-12-06
Introduce credentials provider

### LINKED ISSUE #632: Security Flaw In How Jedis Stores Server Password
Jedis stores Redis server password as a String. Both  ["Java Cryptography Architecture guide" ](http://docs.oracle.com/javase/6/docs/technotes/guides/security/crypto/CryptoSpec.html#PBEEx) and ["Secure Coding Guidelines for Java SE"](http://www.oracle.com/technetwork/java/seccodeguide-139067.html#2) recommend that sensitive information be stored in char array instead of a String.

sazzad16:
References:

1. https://github.com/redis/jedis/issues/1602 and related PRs. Current PR is probably better than handling in JedisFactory 
2. https://github.com/redis/redis-py/pull/2261 - main reason of this PR 
3. https://github.com/lettuce-io/lettuce-core/issues/1774 
4. https://github.com/redis/jedis/issues/632 

---

Resolves #632


dengliming (approved):

yangbodong22011:
@sazzad16 Great design, can you add background to this PR to give me a fuller understanding.

sazzad16:
References:

1. https://github.com/redis/redis-py/pull/2261 - this is what triggered current PR 
2. https://github.com/lettuce-io/lettuce-core/issues/1774 
3. https://github.com/redis/jedis/issues/1602 and related PRs. Current PR is probably better than handling in JedisFactory 
4. https://github.com/redis/jedis/issues/632 

@yangbodong22011

yangbodong22011 (commented):
the current code LGTM。
but it seems we could not new single Jedis with credentialsProvider. 🤔

sazzad16:
@yangbodong22011 We can with JedisClientConfig. https://github.com/redis/jedis/blob/0cf061ca34a39283f0a64269724a0f4ad708a6ff/src/main/java/redis/clients/jedis/Jedis.java#L74

yangbodong22011:
> @yangbodong22011 We can with JedisClientConfig.
> 
> https://github.com/redis/jedis/blob/0cf061ca34a39283f0a64269724a0f4ad708a6ff/src/main/java/redis/clients/jedis/Jedis.java#L74

ohh, I missed.

yangbodong22011 (dismissed):

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -367,14 +371,38 @@ private void initializeFromClientConfig(JedisClientConfig config) {
...
-    sendCommand(Protocol.Command.AUTH, user, password);
-    return getStatusCodeReply();
+  private void auth(final Supplier<RedisCredentials> credentialsProvider) {
+    RedisCredentials credentials = credentialsProvider.get();
+    if (credentials == null) return;
+    char[] password = credentials.getPassword();
+    if (password != null) {
+      String user = credentials.getUser();
+      byte[] binaryPassword = toBytes(password);
+      if (user != null) {
```
chayim, line 387: WDYT ```if user != null && user != ""```

Looking at redis.clients.jedis.RedisCredentials, isn't it possible for an implementor to pass in an empty username?

src/main/java/redis/clients/jedis/ConnectionFactory.java
```diff
@@ -35,6 +35,11 @@ public ConnectionFactory(final JedisSocketFactory jedisSocketFactory, final Jedi
     this.jedisSocketFactory = jedisSocketFactory;
   }
 
+  /**
+   * @deprecated Use {@link RedisCredentialsProvider} through
+   * {@link JedisClientConfig#getCredentialsProvider()}.
+   */
+  @Deprecated
```
chayim, line 42: Why do we have to deprecate this? IMHO we can still keep the interface.

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -49,7 +49,7 @@ public final class Protocol {
   private static final String NOPERM_PREFIX = "NOPERM";
 
   private Protocol() {
-    // this prevent the class from instantiation
+    throw new InstantiationError("Must not instantiate this class");
```
chayim, line 52: ```    throw new InstantiationError("Do not instantiate this class directly");```

chayim:
@sazzad16 this is a big feature - in line with the redis-py one. WDYT about providing a README or [example](https://github.com/redis/redis-py/blob/master/docs/examples/asyncio_examples.ipynb), as opposed to just tests?

src/main/java/redis/clients/jedis/Connection.java
sazzad16: We don't do such checks in Jedis and leave it for the users to do their own thing.

(If I recall correctly,) Once there was such check for password which _had to be_ withdrawn due to issues.

src/main/java/redis/clients/jedis/ConnectionFactory.java
sazzad16: This method<sup>[1]</sup> has been a headache for a very long time. We attempted several times to remove it but couldn't only to support Sentinel mode. Now handling it through RedisCredentialsProvider is much more saner way.

sazzad16: This method has been a headache for a very long time. We attempted several times to remove it but couldn't only to support Sentinel mode. Now handling it through RedisCredentialsProvider is much more saner way.

src/main/java/redis/clients/jedis/Protocol.java
sazzad16: 99.9999% users won't generate this exception. So the string acts more like a code comment. IMO, as a code comment `Must not ...` is better.

Also, we have this same string in several other places.

sazzad16:
> this is a big feature - in line with the redis-py one. WDYT about providing a README or [example](https://github.com/redis/redis-py/blob/master/docs/examples/asyncio_examples.ipynb), as opposed to just tests?

@chayim Sure. At first let's make the implementation fixed. Otherwise the example would also play musical chairs along with the implementation.

src/main/java/redis/clients/jedis/DefaultRedisCredentials.java
```diff
@@ -0,0 +1,27 @@
...
+
+  private final String user;
+  private final char[] password;
+
+  DefaultRedisCredentials(String user, char[] password) {
+    this.user = user;
+    this.password = password;
+  }
+
+  DefaultRedisCredentials(String user, String password) {
```
mp911de, line 13: Instead of accepting `String password`, you can also accept `CharSequence`.

src/main/java/redis/clients/jedis/RedisCredentialsProvider.java
```diff
@@ -0,0 +1,10 @@
+package redis.clients.jedis;
+
+import java.util.function.Supplier;
+
+public interface RedisCredentialsProvider extends Supplier<RedisCredentials> {
+
+  void prepare();
```
mp911de, line 7: Probably fine to have both methods as `default`. Also, you need to provide documentation on both methods what these should do and what the library expectation is that can happen or must not happen.

sazzad16:
> this is a big feature - in line with the redis-py one. WDYT about providing a README or example, as opposed to just tests?

@chayim added an example

vladvildanov (dismissed):

shiranyosef1:
@sazzad16
Thanks for that! Quick question about the behavior when the connection is disconnected. 
We want to use it with AWS IAM for Elasticache with cluster mode enabled.

According to AWS documentation (https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/auth-iam.html#auth-iam-limits), an IAM authenticated connection will automatically be disconnected after 12 hours ("...to extend the connection for another 12 hours, an AUTH or HELLO command with a new IAM authentication token must be sent"). 
How does the client handle the disconnection from the server after 12 hours connection using the dynamic credentials provider?
In the context of idle connections in the pool, is the expectation that these idle connections will be broken, and the client will fetch a new password from the credentials provider (credentials.getPassword())? 
What is the anticipated behavior for idle connections in the pool in this scenario? 
I want to verify that there is a connection renewal with new credentials

shiranyo:
@sazzad16
Thanks for that! Quick question about the behavior when the connection is disconnected. 
We want to use it with AWS IAM for Elasticache with cluster mode enabled.

According to AWS documentation (https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/auth-iam.html#auth-iam-limits), an IAM authenticated connection will automatically be disconnected after 12 hours ("...to extend the connection for another 12 hours, an AUTH or HELLO command with a new IAM authentication token must be sent"). 
How does the client handle the disconnection from the server after 12 hours connection using the dynamic credentials provider?
In the context of idle connections in the pool, is the expectation that these idle connections will be broken, and the client will fetch a new password from the credentials provider (credentials.getPassword())? 
What is the anticipated behavior for idle connections in the pool in this scenario? 
I want to verify that there is a connection renewal with new credentials

sazzad16:
> In the context of idle connections in the pool, is the expectation that these idle connections will be broken, and the client will fetch a new password from the credentials provider (credentials.getPassword())?

@shiranyo Yes.

Note: In such case, enablilng setTest..., i.e. `setTestWhileIdle(true)` or `setTestOnBorrow(true)`, might be a good idea.

PR #3226, opened 2022-12-07
Support intermediate sync() calls in multi node pipelines

sazzad16:


yangbodong22011:
@sazzad16 Pipelines have always been temporary and one-shot, are we sure we want to change this default behavior?

sazzad16:
> Pipelines have always been temporary and one-shot, are we sure we want to change this default behavior?

@yangbodong22011 It is possible in classic (single connection) pipeline. So it's not exactly the _default_ behavior.

Also, if it is really one-shot, we don't need separate `sync()` and `close()`. Perhaps that's a question we should discuss?

chayim (approved):

yangbodong22011:
> It is possible in classic (single connection) pipeline. So it's not exactly the default behavior.

@sazzad16 yes, my mistake.

PR #3235, opened 2022-12-08
Catch JedisException in executeCommand

jgaoIXL:
Hey, I wonder why not catch JedisException in executeCommand?  If the node died, the `connection.getResource()`will possibly throw a `JedisException` not `JedisConnectionException`, right?


sazzad16:
> If the node died, the `connection.getResource()`will possibly throw a `JedisException` not `JedisConnectionException`, right?

No.

1. If there is an already created connection available, there won't be any exception in getResource(). We'll get JedisConnectionException in executeCommand();
2. Otherwise there will be JedisConnectionException while creating/initiating connection. We'll still get JedisConnectionException due to following because it is a subclass of JedisException.

https://github.com/redis/jedis/blob/23d4ab0c3586da4823a73411dce2b2e67a48848f/src/main/java/redis/clients/jedis/util/Pool.java#L36-L44

PR #3242, opened 2022-12-12
Make org.json safely replace-able with android-json

sazzad16:
Source #3191

### DIFF
--- a/src/main/java/redis/clients/jedis/BuilderFactory.java
+++ b/src/main/java/redis/clients/jedis/BuilderFactory.java
@@ -1616,6 +1616,8 @@ public final class BuilderFactory {
       try {
         return new JSONArray(str);
       } catch (JSONException ex) {
+        // This is not necessary but we are doing this just to make is safer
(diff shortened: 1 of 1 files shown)


src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -1611,7 +1612,12 @@ public Object build(Object data) {
     @Override
     public JSONArray build(Object data) {
       if (data == null) return null;
-      return new JSONArray(STRING.build(data));
+      String str = STRING.build(data);
+      try {
+        return new JSONArray(str);
+      } catch (JSONException ex) {
+        throw new JedisException(ex);
```
sazzad16, line 1619:
```suggestion
      } catch (JSONException ex) {
        // This is not necessary but we are doing this just to make is safer
        // for com.vaadin.external.google:android-json library
        throw new JedisException(ex);
```

chayim (approved):

PR #3249, opened 2022-12-16
Allow exclusion of org.json dependency

agavrilov76:
Move creation of the JSON command builders to a separate class which is initialized only when any of the JSON methods are being used. This allows org.json transitive Maven dependency to be excluded for the applications which don't use JSON methods.

See #2961 and #2962 for more information and rationale. Similar to #3223

### DIFF
--- a/src/main/java/redis/clients/jedis/BuilderFactory.java
+++ b/src/main/java/redis/clients/jedis/BuilderFactory.java
@@ -3,9 +3,6 @@ package redis.clients.jedis;
 import java.io.Serializable;
 import java.util.*;
 import java.util.stream.Collectors;
-import org.json.JSONArray;
-import org.json.JSONException;
-import org.json.JSONObject;
 
 import redis.clients.jedis.exceptions.JedisException;
 import redis.clients.jedis.resps.StreamConsumerFullInfo;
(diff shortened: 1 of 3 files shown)


agavrilov76:
@sazzad16 Can you review this one please? It is similar to the one for GSON.

sazzad16:
@agavrilov76 org.json classes (JSONObject, JSONArray) are present in CommandObject class. So this change won't work like it did for Gson.

agavrilov76:
@sazzad16 I've successfully tested this change. I think it is going to work.

JVM will load org.json classes only when the JSON command objects are actually being created, that is, the first time  jsonXXX methods are called.
 
I can  create a repo to demo.

sazzad16:
@agavrilov76 You're right. Just tested myself. Seems to be working. Thanks!

agavrilov76:
@sazzad16 do you plan to release 4.4.0 soon ?

sazzad16:
@agavrilov76 4.4.0 seems to take some time. Please use latest milestone release (if possible). Those can be considered as good as the full release (with rare breaking changes).

agavrilov76:
@sazzad16 I'm waiting for the release with an ability to exclude unwanted JSON dependencies (the one I've contributed with :)) so I can migrate to 4.X.

I hope you can find time to release 4.4.0 some time soon (thanks for maintaining the project, btw!)

PR #3254, opened 2022-12-22
Remove JedisBroadcast

sazzad16:


yangbodong22011:
@sazzad16 can you explain the reason?

sazzad16:
@yangbodong22011 Product managers within Redis are not satisfied how error handling is looking with this approach (even though they are satisfied with actual broadcast). I feel there is a big chance of a future breaking request if we release this in 4.4.0 release and I'd prefer to avoid that. So I've created this PR as a reminder to remove this feature just in case we have to release 4.4.0 before the _green signal_ of this broadcast approach.

yangbodong22011:
> Product managers within Redis are not satisfied how error handling is looking with this approach (even though they are satisfied with actual broadcast).

Did they provide their own ideas or plans? Directly modify the command of JedisCluster itself or other?

>  I feel there is a big chance of a future breaking request if we release this in 4.4.0 release and I'd prefer to avoid that.

It will not bring breaking change, it's a new feature and we did not change the default behavior of original Commands.

sazzad16:
> Did they provide their own ideas or plans? Directly modify the command of JedisCluster itself or other?

I'm waiting for that.

> It will not bring breaking change, it's a new feature and we did not change the default behavior of original Commands.

Introducing it is not the breaking change. I meant releasing JedisBroadcast in 4.4.0 and then possibly breaking it in immediate/later release.

sazzad16:
Resolved by #3285

PR #3262, opened 2023-01-02
Improved Transaction classes documentation

sazzad16:


src/main/java/redis/clients/jedis/ReliableTransaction.java
```diff
@@ -3,14 +3,25 @@
 import java.util.List;
 import redis.clients.jedis.exceptions.JedisException;
 
+/**
+ * A transaction where each command will be immediately sent to Redis server and checked the 'QUEUED' reply.
```
chayim, line 7:
```
/* 
ReliableTransaction is a transaction whereby commands are immediately sent to Redis and the 'QUEUED' reply checked.
*/
```

src/main/java/redis/clients/jedis/ReliableTransaction.java
```diff
@@ -3,14 +3,25 @@
...
+ */
 public class ReliableTransaction extends TransactionBase {
 
+  private static final String QUEUED_STR = "QUEUED";
+
   public ReliableTransaction(Connection connection) {
     super(connection);
   }
 
   /**
```
chayim, line 17:
```
  /**
   * Creates a new transaction.
   *
   * A user wanting to WATCH/UNWATCH keys followed by a call to MULTI ({@link #multi()}) it should
   * be {@code doMulti=false}.
   *
   * @param connection connection
   * @param doMulti {@code false} should be set to enable manual WATCH, UNWATCH and MULTI
   */
```

src/main/java/redis/clients/jedis/Transaction.java
```diff
@@ -17,7 +20,13 @@ public Transaction(Connection connection) {
   }
 
   /**
-   * If you want to WATCH/UNWATCH keys before MULTI command you should do {@code doMulti = true}.
+   * Creates a transaction.
+   *
+   * If user wants to WATCH/UNWATCH keys and then call MULTI ({@link #multi()}) ownself, it should
```
chayim, line 25: Similar comments to the previous trasaction

chayim (approved):

PR #3278, opened 2023-01-11
Make GSON and org.json dependencies optional

agavrilov76:
Stop requiring GSON and org.json dependencies by declaring them 'optional' in the pom file as described in https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html

This change will break existing projects that use the JSON jedis feature and don't have explicit dependencies to GSON and org.json. Projects that don't use JSON would not longer have unwanted dependencies in their class-path

See https://github.com/redis/jedis/issues/2961 and https://github.com/redis/jedis/pull/2962 for more information and rationale.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,11 +65,13 @@
 			<groupId>org.json</groupId>
 			<artifactId>json</artifactId>
 			<version>20220924</version>
+			<optional>true</optional>
 		</dependency>
 		<dependency>
 			<groupId>com.google.code.gson</groupId>
 			<artifactId>gson</artifactId>
 			<version>2.8.9</version>
+			<optional>true</optional>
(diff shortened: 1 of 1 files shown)


sazzad16:
This is unlikely to be accepted, for the reasons mentioned in the description. Advanced and concerned users can now exclude those dependencies.

agavrilov76:
> This is unlikely to be accepted, for the reasons mentioned in the description. Advanced and concerned users can now exclude those dependencies.

Alright. Perhaps it can be accepted for the next major version if the JSON support remains implemented in the same way.

uglide:
Closing as stale. Related PR https://github.com/redis/jedis/pull/3971

PR #3282, opened 2023-01-19
Arielber/added json encoder decoder interface and injection

ArielBerkovich:
added JsonEncoderDecoder Interface and use it as a constructor parameter for relevant classes.
this enables the injection of different encoderDecoder rather than the default gson.

for exmple, this would allow the encoding and decoding of objects that require special type adapter(ie: InstantTypeAdapter)


sazzad16:
@bsbodden You have just worked with Jedis 4. WDYT about this PR?

I understand that such feature would be helpful. My question is mostly about the approach.

Also, do prefer adding several new constructors, or just a `set...` method in necessary classes?

sazzad16:
@ArielBerkovich Thank you for your interest in Jedis. Please don't feel discouraged to contribute improving Jedis.

bsbodden:
> @bsbodden You have just worked with Jedis 4. WDYT about this PR?
> 
> 
> 
> I understand that such feature would be helpful. My question is mostly about the approach.
> 
> 
> 
> Also, do prefer adding several new constructors, or just a `set...` method in necessary classes?

Somehow I missed this! Yes, it's a needed feature, pluggability in general. Constructor-based would be fine!

sazzad16:
@bsbodden We went for only set based approach. PR #3319, method: `setJsonObjectMapper(JsonObjectMapper jsonObjectMapper)`

PR #3285, opened 2023-02-01
OSS Broadcast support

sazzad16:


sazzad16:
@uglide

chayim:
@sazzad16 WDYT about adding examples (and a jedis-examples pom) with this PR? This would be the first one, but IMHO for a PR like this, it's nice, because it's easier to see the usage of the implementation?

src/main/java/redis/clients/jedis/exceptions/JedisBroadcastException.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Map;
+import redis.clients.jedis.HostAndPort;
+
+/**
+ * Note: This exception extends {@link JedisDataException} just so existing applications catching
+ * JedisDataException do not get broken.
+ */
+public class JedisBroadcastException extends JedisDataException {
+
+  private static final String BROADCAST_ERROR_MESSAGE = "<a generic broadcast error message goes here.>";
```
chayim, line 13: ```"<a generic broadcast error message goes here.>"```

I think we need one

src/main/java/redis/clients/jedis/executors/ClusterCommandExecutor.java
```diff
@@ -34,6 +37,40 @@ public void close() {
...
+    JedisBroadcastException holder = new JedisBroadcastException();
+    for (Map.Entry<String, ConnectionPool> entry : connectionMap.entrySet()) {
+      HostAndPort node = HostAndPort.from(entry.getKey());
+      ConnectionPool pool = entry.getValue();
+      try (Connection connection = pool.getResource()) {
+        try {
+          T aReply = execute(connection, commandObject);
+          holder.addReply(node, aReply);
+          if (isErrored) { // already errored
+          } else if (reply == null) {
```
chayim, line 55: nit: the non okay cases should be together

src/main/java/redis/clients/jedis/exceptions/JedisBroadcastException.java
sazzad16: @chayim Please give me one

src/main/java/redis/clients/jedis/executors/ClusterCommandExecutor.java
sazzad16: @chayim how else?

There's only one non-okay case (the `else` block), so all non okay are already together. Once non okay, I'm ignoring everything by first `if`.

sazzad16:
> @sazzad16 WDYT about adding examples (and a jedis-examples pom) with this PR? This would be the first one, but IMHO for a PR like this, it's nice, because it's easier to see the usage of the implementation?

@chayim It's not possible due to the current structure of this repo. We'd have to restructure it with a main parent directory which would contain several directories, i.e. jedis-examples, as well as `jedis` the current main directory.

src/main/java/redis/clients/jedis/exceptions/JedisBroadcastException.java
chayim: @uglide will care more than me about text. I suggest
```A failure occurred broadcasting the command```

File: src/main/java/redis/clients/jedis/executors/ClusterCommandExecutor.java (hunk 2 above)
Comment (chayim) [reply src/main/java/redis/clients/jedis/executors/ClusterCommandExecutor.java:55]: The reploy.equals first (because it's ok), then the replyy == null

**Comment by chayim** [comment]
> @chayim It's not possible due to the current structure of this repo. We'd have to restructure it with a main parent directory which would contain several directories, i.e. jedis-examples, as well as `jedis` the current main directory.

I mean we have src/main and src/tests, why not src/examples, and exclude it from the package...? This can also be out of scope. Alternatively we add an examples tree at the top, with at least a single file end-to-ending it as part of this.

**Comment by sazzad16** [comment]
> we have src/main and src/tests

Those directories are following maven standard.

> why not src/examples

That's not part of maven standard. You still can if you want but that wouldn't be following any _standard directory structure_.

> Alternatively we add an examples tree at the top, with at least a single file end-to-ending it as part of this.

I don't understand.

**Comment by sazzad16** [comment]
@chayim 

> we have src/main and src/tests

Those directories are following maven standard.

> why not src/examples

That's not part of maven standard. You still can if you want but that wouldn't be following any _standard directory structure_.

> Alternatively we add an examples tree at the top, with at least a single file end-to-ending it as part of this.

I don't understand.

File: src/main/java/redis/clients/jedis/executors/ClusterCommandExecutor.java (hunk 2 above)
Comment (sazzad16) [reply src/main/java/redis/clients/jedis/executors/ClusterCommandExecutor.java:55]: It'll cause a NullPointerException during the first iteration.

**Comment by sazzad16** [comment]
@uglide @chayim PING

File: docs/cluster-broadcast.md [hunk 3]
```diff
@@ -0,0 +1,25 @@
+# Broadcast commands to all Redis Cluster nodes
```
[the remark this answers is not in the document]
sazzad16: Oh, I forgot to write... I am waiting for @uglide's opinion and so I kept the doc format just in case he prefers it over the example format.

Yes, we should definitely keep only of one these two formats.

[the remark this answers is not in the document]
chayim: I'll weight. Example only because we can at least compile at test it. There's a docgen plan - no need to support two. Approving.!

chayim (dismissed):

docs/cluster-broadcast.md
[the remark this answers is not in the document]
sazzad16: It'll be compiled, yes, but it won't be tested.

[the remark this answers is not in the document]
chayim: That's a first step. Examples aren't currently tested in general.

src/test/java/redis/clients/jedis/examples/BroadcastCommandsToAllClusterNodes.java
```diff
@@ -0,0 +1,23 @@
+package redis.clients.jedis.examples;
+
+import redis.clients.jedis.HostAndPort;
+import redis.clients.jedis.JedisCluster;
+
+/**
+ * When using <a href="https://redis.io/docs/reference/cluster-spec/">Open Source Redis Cluster API</a>, some commands
```
chayim, line 7: /**
 * When using the <a href="https://redis.io/docs/reference/cluster-spec/">Open Source Redis Cluster API</a>, some commands
 * must be executed against all primary nodes. To simplify this task, Jedis provides an easy way to
 * broadcast commands.
 * 
 * For example, to update the server configuration of all nodes in the Redis Cluster, we broadcast
 * the command [CONFIG SET](https://redis.io/commands/config-set/) to all nodes.
 */

chayim:
@uglide after everything we discussed, need your feedback here. I've already had discussions with @sazzad16

uglide:
@sazzad16 @chayim LGTM. I can't wait to see support for `JSON.MGET` support in broadcasted mode and round-robin for `TS.MGET` and `FT.SEARCH`.

uglide (approved):

PR #3290, opened 2023-02-10
Fix jedis#hdel docs

keeferrourke:
This is a non-breaking change that simply updates the documentation for `Jedis#hdel` and adds a test that proves the functionality is what is specified.

Previously, Jedis's docs suggested that the `hdel` operation always returned 1L for a successful deletion, or 0L for an unsuccessful deletion.

Redis [official docs](https://redis.io/commands/hdel/) state:

> **Return**
> [Integer reply](https://redis.io/docs/reference/protocol-spec#resp-integers): the number of fields that were removed from the hash, not including specified but non existing fields.

### DIFF
--- a/src/main/java/redis/clients/jedis/Jedis.java
+++ b/src/main/java/redis/clients/jedis/Jedis.java
@@ -5687,8 +5687,8 @@ public class Jedis implements ServerCommands, DatabaseCommands, JedisCommands, J
   }
 
   /**
-   * Remove the specified field(s) from a hash stored at key.
-   * Specified fields that do not exist within this hash are ignored.
+   * Remove the specified field(s) from a hash stored at key. Specified fields that do not exist
+   * within this hash are ignored.
    * <p>
    * <b>Time complexity:</b> O(1)
    * @param key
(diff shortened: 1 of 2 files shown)


src/test/java/redis/clients/jedis/commands/jedis/HashesCommandsTest.java
```diff
@@ -200,7 +200,7 @@ public void hexists() {
 
   @Test
   public void hdel() {
-    Map<String, String> hash = new HashMap<String, String>();
```
zeekling, line 203: Do we need to consider the jdk 7 version? @sazzad16

sazzad16: @zeekling No.

src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -5687,13 +5687,15 @@ public boolean hexists(final String key, final String field) {
   }
 
   /**
-   * Remove the specified field from an hash stored at key.
+   * Remove the specified field(s) from a hash stored at key.
+   * Specified fields that do not exist within this hash are ignored.
```
sazzad16, line 5691:
```suggestion
   * Remove the specified field(s) from a hash stored at key. Specified fields that do not exist
   * within this hash are ignored.
```

sazzad16 (changes requested):

sazzad16 (approved):

PR #3310, opened 2023-03-03
Cross cluster failover

uglide:


sazzad16:
@uglide I may take some time (days) to do proper review.

BTW, are we going to do this without tests?

@yangbodong22011 @dengliming @mina-asham Please participate if you could. Thanks!

pom.xml
```diff
@@ -103,6 +104,21 @@
 			<version>3.12.4</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>io.github.resilience4j</groupId>
+			<artifactId>resilience4j-all</artifactId>
+			<version>${resilience4j.version}</version>
+		</dependency>
```
dengliming: If you only need to introduce partial modules you want to use, should we remove `resilience4j-all`?

dengliming (commented):
Just a quick look, I think we also need to write unit tests to cover these new feature changes.

pom.xml
sazzad16: +1

pom.xml
```diff
@@ -103,6 +104,21 @@
...
+		<dependency>
+			<groupId>io.github.resilience4j</groupId>
+			<artifactId>resilience4j-circuitbreaker</artifactId>
+			<version>${resilience4j.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>io.github.resilience4j</groupId>
+			<artifactId>resilience4j-retry</artifactId>
+			<version>${resilience4j.version}</version>
+		</dependency>
```
sazzad16, line 121: These should be optional (`<optional>true</optional>`).

sazzad16 (changes requested):

pom.xml
aterleto: Done, thanks for the suggestion

yangbodong22011:
Is there an issue discussion that already exists before your PR that gave me the background for this PR.

sazzad16:
> Is there an issue discussion that already exists before your PR that gave me the background for this PR.

@yangbodong22011 Sadly, no. 

I'm writing from my own: This PR is to communicate with multiple Redis servers which are set in multiple geo-locations and are eventually consistent. The aim is to choose the fastest and most reliable server for the (java) application.

sazzad16:
@aterleto @uglide Please resolve the conflicts with the master branch. I have created https://github.com/aterleto/jedis/pull/2 for your convenience.

sazzad16:
@aterleto Sorry but my previous PR mistakenly targeted for your `master` branch. Our target is your `cross-cluster-failover`. I have now created https://github.com/aterleto/jedis/pull/4. Hopefully it'll be okay this time :crossed_fingers:

chayim:
CI failure here is a red herring.. ping @sazzad16

sazzad16:
```
Error:  Failures: 
Error:    MultiClusterPooledConnectionProviderTest.testRunClusterFailoverPostProcessor:112 expected:<true> but was:<false>
```

src/test/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProviderTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+import static org.junit.Assert.*;
+
+/**
+ * @see MultiClusterPooledConnectionProvider
+ */
+public class MultiClusterPooledConnectionProviderTest {
+
+    private static final HostAndPort hostAndPort1 = HostAndPorts.getRedisServers().get(0);
+    private static final HostAndPort hostAndPort2 = HostAndPorts.getRedisServers().get(0);
```
sazzad16, line 21: These are same server!

sazzad16 (changes requested):

docs/failover.md
```diff
@@ -0,0 +1,225 @@
...
+
+When your application starts, Jedis will attempt to connect to each cluster in the order that the clusters appear
+in your client configuration. It's important to understand this, especially in the case where Jedis has failed over.
+If Jedis has failed over to a new cluster, then restarting the application may result in an inadvertent failback.
+This can happen only if a failed cluster comes back online and the application subsequently restarts. 
+
+If you need to avoid this scenario, consider using a failover callback, as described above, to externally record
+the name of the cluster that your application was most recently connected to. You can then check this state on startup
+to ensure that you application only connects to the most recently used cluster. For assistance with this technique,
+consider [opening an issue](https://github.com/redis/jedis/issues).
```
sazzad16, line 225: Should be: starting a Q & A discussion

src/main/java/redis/clients/jedis/executors/CircuitBreakerCommandExecutor.java
```diff
@@ -0,0 +1,95 @@
...
+            // to use the next cluster's connection pool - according to the configuration's prioritization/order
+            int activeMultiClusterIndex = provider.incrementActiveMultiClusterIndex();
+
+            // Implementation is optionally provided during configuration. Typically, used for activeMultiClusterIndex persistence or custom logging
+            provider.runClusterFailoverPostProcessor(activeMultiClusterIndex);
+        }
+
+        // Once the priority list is exhausted only a manual failback can open the circuit breaker so all subsequent operations will fail
+        else if (provider.isLastClusterCircuitBreakerForcedOpen()) {
+            throw new JedisConnectionException("Cluster/database endpoint could not failover since the MultiClusterJedisClientConfig was not " +
```
sazzad16, line 86: Is this a definite connection error? If so, it'd be a good idea to include that in this exception either as a cause or suppressed. Otherwise, use other exception.

pom.xml
```diff
@@ -48,6 +48,7 @@
 		<github.global.server>github</github.global.server>
 		<slf4j.version>1.7.36</slf4j.version>
 		<jedis.module.name>redis.clients.jedis</jedis.module.name>
+		<resilience4j.version>1.7.1</resilience4j.version>
```
sazzad16, line 51: Can we use Resilience4j 2.0?

aterleto: It would introduce breaking changes. IIRC there was a Decorator object that was removed even though its the most efficient way that I found to get failover to work for our use-case. So I decided to stick with 1.7 until they added a capability that worked around this deprecated way to configure the supplier on every call.

src/main/java/redis/clients/jedis/exceptions/JedisValidationException.java
```diff
@@ -0,0 +1,21 @@
+package redis.clients.jedis.exceptions;
+
+/**
+ * A validation error.
+ */
+public class JedisValidationException extends JedisException {
```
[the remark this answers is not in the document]
aterleto: I thought it was odd that one didn't already exist. Validation exceptions are normal. I don't think making it MultiCluster specific makes much sense since its not in anyway related to multi-clusters, it just so happens that I am the first to validate instead of allowing things to just blow up due to incorrect configuration. I think you should reconsider this ask but if you feel strongly I can rename it.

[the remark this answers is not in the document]
aterleto: I rarely see exceptions added as inner classes. Especially in this case where there is no direct dependency. It feels like future technical debt.

src/main/java/redis/clients/jedis/executors/CircuitBreakerCommandExecutor.java
aterleto: I'm not sure I understand your meaning. This condition is best handled with a checked exception so instead of creating yet another exception type I returned a connection exception because the problem is related to not being able to connect. I'm not sure what is at issue?

pom.xml
```diff
@@ -115,6 +116,24 @@
...
+			<artifactId>resilience4j-circuitbreaker</artifactId>
+			<version>${resilience4j.version}</version>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>io.github.resilience4j</groupId>
+			<artifactId>resilience4j-retry</artifactId>
+			<version>${resilience4j.version}</version>
+			<optional>true</optional>
+		</dependency>
```
sazzad16, line 136: @aterleto Could you please give 1 try and see if defining these two only as test dependency works? (Remove `<optional>true</optional>` and add `<scope>test</scope>`).

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
```diff
@@ -0,0 +1,321 @@
...
+    public void runClusterFailoverPostProcessor(Integer multiClusterIndex) {
+        if (clusterFailoverPostProcessor != null)
+            clusterFailoverPostProcessor.accept(getClusterCircuitBreaker(multiClusterIndex).getName());
+    }
+
+    public void setClusterFailoverPostProcessor(Consumer<String> clusterFailoverPostProcessor) {
+        this.clusterFailoverPostProcessor = clusterFailoverPostProcessor;
+    }
+
+    public static class Cluster {
```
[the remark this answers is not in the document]
sazzad16: ^^^

sazzad16 (changes requested):

docs/failover.md
sazzad16, line 225:
```suggestion
consider [starting a Q&A discussion](https://github.com/redis/jedis/discussions/new?category=q-a).
```

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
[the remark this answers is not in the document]
sazzad16: Made resolved just to hide.

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
```diff
@@ -0,0 +1,321 @@
...
```
sazzad16, line 292: In our documentation we are saying _Redis deployments_. Even more so this PR is done in a way this supports standalone Redis instances. So each of the deployments wouldn't necessarily be a _Cluster_ deployment. (Am I correct so far?)

This inner `Cluster` class is actually represents a single deployment, right? If so, the word _Cluster_ is not appropriate name for this class.

@aterleto Please this class name.

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
sazzad16: May I suggest a set of renamings?
```
            MultiClusterJedisClientConfig => MultiDeploymentConfig / MultiDeploymentClientConfig
(ic)             ClusterJedisClientConfig => DeploymentConfig / SingleDeploymentConfig
     MultiClusterPooledConnectionProvider => MultiDeploymentConnectionProvider
(ic)                              Cluster => Deployment / SingleDeployment
```
(ic = inner class)

src/main/java/redis/clients/jedis/MultiClusterJedisClientConfig.java
```diff
@@ -0,0 +1,332 @@
...
+
+    public List<Class> getCircuitBreakerIgnoreExceptionList() {
+        return circuitBreakerIgnoreExceptionList;
+    }
+
+    public SlidingWindowType getCircuitBreakerSlidingWindowType() {
+        return circuitBreakerSlidingWindowType;
+    }
+
+    public static class ClusterJedisClientConfig {
```
sazzad16, line 163:
```suggestion
    public static class ClusterConfig {
```

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/MultiClusterJedisClientConfig.java
```diff
@@ -0,0 +1,332 @@
...
+ * that all share retry and circuit breaker configuration settings.
+ * <p>
+ * With this Config users can seamlessly failover to Disaster Recovery (DR), Backup, and Active-Active cluster(s)
+ * by using simple configuration which is passed through from Resilience4j - https://resilience4j.readme.io/docs
+ * <p>
+ * Configuration options related to automatic failback (e.g. HALF_OPENED state) are not supported and therefore
+ * not passed through to Jedis users.
+ * <p>
+ */
+public final class MultiClusterJedisClientConfig {
```
sazzad16, line 25:
```suggestion
public final class MultiClusterClientConfig {
```

src/main/java/redis/clients/jedis/MultiClusterJedisClientConfig.java
sazzad16, line 163:
```suggestion
    public static class ClusterConfig {
```

sazzad16 (changes requested):
@aterleto Please at least accept these change requests and help us bring less confusion between this implementation and long time existing JedisCluster.

aterleto:
> @aterleto Please at least accept these change requests and help us bring less confusion between this implementation and long time existing JedisCluster.

Completed as requested

Issue #3315, opened 2023-03-10
Build tests failed from master branch


clebersonp:
### Expected behavior

All unit tests pass without code change from a new branch created from the master branch.

### Actual behavior

After created 'my new branch' from the `master branch` i tried to build the project but some unit tests failed when i run `make test` command.
1. `AccessControlListCommandsTest.basicPermissionsTest:292 expected:<NOPERM [No permissions to access a key]> but was:<NOPERM [this user has no permissions to access one of the keys used as arguments]>`. The two strings are different. Maybe because of the redis-server version that has changed the `error message`. I don't really know if checking the equality of exception message is the best approach in this case.
2. `AccessControlListCommandsTest.basicPermissionsTest:307 expected:<NOPERM [No permissions to access a key]> but was:<NOPERM [this user has no permissions to access one of the keys used as arguments]>`. The same issue as item 1.
3. `DocumentTest.serialize:45 expected:<id:9f, score: 10[,]0, payload:1a, prope...> but was:<id:9f, score: 10[.]0, payload:1a, prope...>`. My default locale is `pt_BR`. The `java.lang.String.format()` statement of the test is not passing any Locale, so it's considering the system default.
4. `DocumentTest.serialize:62 expected:<id:9f, score: 10[,]0, payload:1a, prope...> but was:<id:9f, score: 10[.]0, payload:1a, prope...>`. The same issue as item 3.
5. `DocumentTest.serialize:76 expected:<id:9f, score: 10[,]0, payload:1a, prope...> but was:<id:9f, score: 10[.]0, payload:1a, prope...>`. The same issue as item 4.

### Screenshot
![Screenshot from 2023-03-09 22-50-15](https://user-images.githubusercontent.com/9445673/224202718-8b9654c0-bf27-4d62-b900-18a40ecb002b.png)

### Steps to reproduce:

Please create a reproducible case of your problem. Make sure
that case repeats consistently and it's not random
1. Make sure the default Locale is not `en_US` or anything else that formats the decimal different of a `dot`.
2. Run `make test` command.

#### Jedis version:
`4.4.0-SNAPSHOT from master branch`

#### Redis version:
`Redis server v=7.0.9 sha=00000000:0 malloc=jemalloc-5.2.1 bits=64 build=448f00a942bc1f77`

#### Java version:
`jdk1.8.0_221`

sazzad16:
@clebersonp You are welcome to craft PRs. For 1 & 2, we should check if the error message is about key(s).

[...]

PR #3318, opened 2023-03-14
Message and format locale junit adjustments

clebersonp:
1. Done some adjustments related at the error message '**noperm**' for **keys** adding the new one to the assertion test;
2. Forced the string's '**format**' method with a **default locale** to ensure the decimal number formated will have the same separator decimal as the assertion's message.

### Result
Screenshot from 2023-03-13 21-41-02

_PR_ related to issue [#3315](https://github.com/redis/jedis/issues/3315).


sazzad16 (changes requested):
Does `Locale.ENGLISH` work? If so, please use that.

Otherwise, if `Locale.US` is required, mention it as code comment.

clebersonp:
Thanks @sazzad16  for your suggestion. `Locale.ENGLISH` works fine. I've already done the adjustments

sazzad16 (approved):

PR #3319, opened 2023-03-14
Abstraction for JSON object mapper

clebersonp:
Framework created for redis json parser commands using abstraction for different json parser custom engines to achieve decoupling and programming code for interface instead of implementation.

Junit tests were written for the custom gson and jackson parser engine.


pom.xml
```diff
@@ -70,6 +70,7 @@
 			<groupId>com.google.code.gson</groupId>
 			<artifactId>gson</artifactId>
 			<version>2.10.1</version>
+			<optional>true</optional>
```
sazzad16, line 73: Remove this. Because:
1. This will be a breaking change.
2. This will disrupt the kick-starting for novice users.
```suggestion
```

src/main/java/redis/clients/jedis/exceptions/JedisConfigException.java
```diff
@@ -0,0 +1,21 @@
+package redis.clients.jedis.exceptions;
+
+/**
+ * A config error.
+ */
+public class JedisConfigException extends JedisException {
```
sazzad16, line 6: Remove this exception. A simple `JedisException` with proper exception message is enough to serve the purpose ATM.

src/main/java/redis/clients/jedis/json/parser/RedisJsonParser.java
```diff
@@ -0,0 +1,19 @@
+package redis.clients.jedis.json.parser;
+
+/**
+ * The parser contract for all the <em>Redis Command</em> that perform JSON operations.
+ *
+ * @see redis.clients.jedis.json.RedisJsonCommands
+ * @see redis.clients.jedis.json.RedisJsonPipelineCommands
+ */
+public interface RedisJsonParser {
```
sazzad16, line 9: I'm failing to see the _added benefit_ of this separate interface. Remove this.

PS: Of course we need `setJsonParser` method in necessary classes.

src/test/java/redis/clients/jedis/modules/RedisModuleCommandsTestBase.java
```diff
@@ -38,6 +39,7 @@ public void setUp() {
       jedis.flushAll();
     }
     client = new UnifiedJedis(hnp);
+    client.setJsonParser(new DefaultGsonParser());
```
sazzad16, line 42: By default, it should be able to use the default parser without needing to set it by user.

src/main/java/redis/clients/jedis/json/parser/JsonParser.java
```diff
@@ -0,0 +1,26 @@
+package redis.clients.jedis.json.parser;
+
+/**
+ * Represents the ability of serialize an object to JSON format string
+ * and deserialize it to the typed object.
+ *
+ * @see DefaultGsonParser Default implementation for JSON parser engine with com.google.gson.Gson
+ */
+public interface JsonParser {
```
sazzad16, line 9: This represents both a serializer and deserializer. The word parser generally refers to deserializer. Is it possible to have another name?

@chayim @uglide Any suggestion?

sazzad16 (changes requested):

pom.xml
clebersonp: Sure. Are you right. I'll provide the adjustment for that. Thanks

src/main/java/redis/clients/jedis/exceptions/JedisConfigException.java
clebersonp: All right. I'll provided this correction. Thanks.

src/test/java/redis/clients/jedis/modules/RedisModuleCommandsTestBase.java
clebersonp: Perfect. I'll make gson the default serializer/deserializer like it was before. Thanks.

src/main/java/redis/clients/jedis/json/parser/RedisJsonParser.java
clebersonp: That's ok. I will make the adjustment based on your suggestion. Thanks.

src/main/java/redis/clients/jedis/json/parser/JsonParser.java
clebersonp: Yes. You are right.
I've some suggestions.

1. RedisJsonMapper;
2. JsonMapper;
3. RedisObjectMapper;
4. ObjectMapper;
5. RedisParseGenerateJson;
6. RedisConvertObjectJson;
7. RedisJson;
8. Redison;
9. Rson;

sazzad16: 1-4 are reasonable. One more to consider: `JsonObjectMapper`.

src/test/java/redis/clients/jedis/modules/RedisModuleCommandsTestBase.java
```diff
@@ -11,6 +11,7 @@
 import redis.clients.jedis.Protocol;
 import redis.clients.jedis.UnifiedJedis;
 import redis.clients.jedis.exceptions.JedisConnectionException;
+import redis.clients.jedis.json.parser.DefaultGsonParser;
```
sazzad16, line 14:
```suggestion
```

src/test/java/redis/clients/jedis/modules/RedisModulesPipelineTest.java
```diff
@@ -23,6 +23,7 @@
 import redis.clients.jedis.json.JsonSetParams;
 import redis.clients.jedis.json.Path;
 import redis.clients.jedis.json.Path2;
+import redis.clients.jedis.json.parser.DefaultGsonParser;
```
sazzad16, line 26:
```suggestion
```

src/test/java/redis/clients/jedis/util/JsonParserTestUtil.java
```diff
@@ -0,0 +1,81 @@
...
+      this.om = om;
+    }
+    @Override
+    public <T> T fromJson(String value, Class<T> valueType) {
+      try {
+        return om.readValue(value, valueType);
+      } catch (JsonProcessingException e) {
+        e.printStackTrace();
+      }
+      return null;
```
sazzad16, line 34: Don't just print. Perhaps:
```suggestion
      try {
        return om.readValue(value, valueType);
      } catch (JsonProcessingException e) {
        throw new JedisException(e);
      }
```

src/test/java/redis/clients/jedis/util/JsonParserTestUtil.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+    @Override
+    public String toJson(Object value) {
+      try {
+        return om.writeValueAsString(value);
+      } catch (JsonProcessingException e) {
+        e.printStackTrace();
+      }
+      return null;
```
sazzad16, line 43: Perhaps:
```suggestion
      try {
        return om.writeValueAsString(value);
      } catch (JsonProcessingException e) {
        throw new JedisException(e);
      }
```

src/main/java/redis/clients/jedis/json/parser/package-info.java
```diff
@@ -0,0 +1,4 @@
+/**
+ * This package contains the classes and interfaces related to JSON processing module.
+ */
+package redis.clients.jedis.json.parser;
```
sazzad16, line 4: After update, now it seems we don't need a separate package. `util` package could be an option to put the interface and default class; but I'm leaning more towards `json` package.

src/main/java/redis/clients/jedis/CommandObjects.java
```diff
@@ -34,6 +34,9 @@
 
 public class CommandObjects {
 
+  // initialize the parser with simple gson configuration as the default.
+  private Optional<JsonParser> jsonParser = Optional.of(new DefaultGsonParser());
```
sazzad16, line 38: This invalidates #3223 where people may want to exclude Gson when they are not using RedisJSON commands.

We can do this in `getJsonParser()`. When it is called we can be sure that a RedisJSON command is requested and as no parser/mapper is set, we can go ahead for the default one. Note: We will need [Double-checked locking](https://en.wikipedia.org/wiki/Double-checked_locking).



image

sazzad16 (changes requested):
Thank you for addressing previous change requests. Here are more :slightly_smiling_face:

src/test/java/redis/clients/jedis/modules/RedisModuleCommandsTestBase.java
clebersonp: Clear. I just forgot to format the code. Thanks.

src/main/java/redis/clients/jedis/json/parser/package-info.java
clebersonp: I also think that the `json` package it's a better place.

sazzad16 (approved):

sazzad16:
Merged. Thank you @clebersonp for your contribution!

clebersonp:
> 

@sazzad16, Thanks for your time and great suggestions. :+1:  :smile:

PR #3331, opened 2023-03-23
Obtain multiple pipelines concurrently. high performance improvement

chenshi5012:
multiple pipelines sync concurrently will improvement the performance  under  redis cluster model
A 5-node cluster, vertically representing the results from the same batch
Extraordinary Suggestion: Convert sync to multi-threaded execution, which can improve performance by about 80%

image

### DIFF
--- a/src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
+++ b/src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
@@ -8,12 +8,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.Queue;
 import java.util.Set;
-import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import java.util.concurrent.ThreadPoolExecutor;
-import java.util.concurrent.TimeUnit;
 
 import org.json.JSONArray;
 import org.slf4j.Logger;
--- a/src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
+++ b/src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
@@ -44,6 +44,13 @@ public abstract class MultiNodePipelineBase implements PipelineCommands, Pipelin
 
   private final Logger log = LoggerFactory.getLogger(getClass());
 
+  /**
+   * default number of processes for sync, if you got enough cores for client
+   * or your cluster nodes more than 3 nodes, you may increase this workers number.
+   * suggest <= cluster nodes
+   */
+  public static volatile int MULTI_NODE_PIPELINE_SYNC_WORKERS = 3;
+
   private final Map<HostAndPort, Queue<Response<?>>> pipelinedResponses;
   private final Map<HostAndPort, Connection> connections;
   private volatile boolean syncing = false;
@@ -51,7 +58,7 @@ public abstract class MultiNodePipelineBase implements PipelineCommands, Pipelin
   private final CommandObjects commandObjects;
   private GraphCommandObjects graphCommandObjects;
 
-  private final ExecutorService executorService = Executors.newFixedThreadPool(Protocol.CLUSTER_PIPELINE_SYNC_WORKERS);
+  private final ExecutorService executorService = Executors.newFixedThreadPool(MULTI_NODE_PIPELINE_SYNC_WORKERS);
 
   public MultiNodePipelineBase(CommandObjects commandObjects) {
     pipelinedResponses = new LinkedHashMap<>();
--- a/src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
+++ b/src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
@@ -117,7 +117,7 @@ public abstract class MultiNodePipelineBase implements PipelineCommands, Pipelin
 
     CountDownLatch countDownLatch = new CountDownLatch(pipelinedResponses.size());
     Iterator<Map.Entry<HostAndPort, Queue<Response<?>>>> pipelinedResponsesIterator
-            = pipelinedResponses.entrySet().iterator();
+        = pipelinedResponses.entrySet().iterator();
     while (pipelinedResponsesIterator.hasNext()) {
       Map.Entry<HostAndPort, Queue<Response<?>>> entry = pipelinedResponsesIterator.next();
       HostAndPort nodeKey = entry.getKey();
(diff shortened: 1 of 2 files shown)


src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
```diff
@@ -114,10 +123,13 @@ public final void sync() {
...
-        for (Object o : unformatted) {
-          queue.poll().set(o);
-        }
+        executor.submit(() -> {
+          List<Object> unformatted = connection.getMany(queue.size());
+          for (Object o : unformatted) {
+            queue.poll().set(o);
+          }
+          countDownLatch.countDown();
+        });
```
yangbodong22011, line 132:
```suggestion
      executor.submit(() -> {
        try {
          List<Object> unformatted = connection.getMany(queue.size());
          for (Object o : unformatted) {
            queue.poll().set(o);
          }
        } catch (JedisConnectionException jce) {
          log.error("Error with connection to " + nodeKey, jce);
          // cleanup the connection
          pipelinedResponsesIterator.remove();
          connections.remove(nodeKey);
          IOUtils.closeQuietly(connection);
        } finally {
          countDownLatch.countDown();
        }
      });
```
1. Need to place `countDownLatch.countDown()` in `finally` to prevent it from being broken by exceptions.
2. It is necessary to catch `RejectedExecutionException` to prevent the thread pool from rejecting tasks.

src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
```diff
@@ -127,6 +139,12 @@ public final void sync() {
       }
     }
 
+    try {
+      countDownLatch.await();
+    } catch (InterruptedException e) {
+      e.printStackTrace();
```
yangbodong22011, line 145: logger error info

yangbodong22011 (commented):
In general, I agree with this PR and also submitted some comments.

src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
chenshi5012: It is crucial to execute countDown regardless of whether the above program is abnormal, otherwise the await cannot be released

yangbodong22011: > It is crucial to execute countDown regardless of whether the above program is abnormal, otherwise the await cannot be released

So I moved it into `finally` to ensure release.

src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
```diff
@@ -47,6 +53,9 @@ public abstract class MultiNodePipelineBase implements PipelineCommands, Pipelin
   private final CommandObjects commandObjects;
   private GraphCommandObjects graphCommandObjects;
 
+  private final ExecutorService executorService = new ThreadPoolExecutor(3, Integer.MAX_VALUE, 60L, TimeUnit.SECONDS,
+          new ArrayBlockingQueue<>(3));
```
sazzad16, line 57: - `Integer.MAX_VALUE` for maximumPoolSize seems risky.
- `60L, TimeUnit.SECONDS` - I understand it's a lot for most applications but still small enough. We shouldn't restrict it in smaller time as we are not allowing users to modify yet.

src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
```diff
@@ -106,25 +115,36 @@ public final void sync() {
     }
     syncing = true;
 
+    CountDownLatch countDownLatch = new CountDownLatch(pipelinedResponses.size());
     Iterator<Map.Entry<HostAndPort, Queue<Response<?>>>> pipelinedResponsesIterator
-        = pipelinedResponses.entrySet().iterator();
+            = pipelinedResponses.entrySet().iterator();
```
sazzad16, line 120:
```suggestion
        = pipelinedResponses.entrySet().iterator();
```

src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
sazzad16: How about simply [Executors#newFixedThreadPool](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Executors.html#newFixedThreadPool-int-)

chenshi5012: Theoretically, the maximum number of nodes is 16384.redis. The author recommends that it be within 1000. We set it to 1024. It is more suitable. What do you think?

chenshi5012: > * `Integer.MAX_VALUE` for maximumPoolSize seems risky.
> * `60L, TimeUnit.SECONDS` - I understand it's a lot for most applications but still small enough. We shouldn't restrict it in smaller time as we are not allowing users to modify yet.

provide default values that can be customized by users， what do you think?
image

src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
```diff
@@ -47,6 +52,9 @@ public abstract class MultiNodePipelineBase implements PipelineCommands, Pipelin
   private final CommandObjects commandObjects;
   private GraphCommandObjects graphCommandObjects;
 
+  private final ThreadPoolExecutor executor = new ThreadPoolExecutor(3, Integer.MAX_VALUE, 60L, TimeUnit.SECONDS,
```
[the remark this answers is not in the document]
chenshi5012: as sazzad16 said , we just adjust the thread poll size fix to MULTI_NODE_PIPELINE_SYNC_WORKERS  , thanks for your advice

src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
yangbodong22011: Our (alibaba) programming specification does not recommend using `Executors.*` to directly create a thread pool. The correct way is to use `ThreadPoolExecutor`, see [here](https://github.com/alibaba/Alibaba-Java-Coding-Guidelines#concurrency) for details.

Again, I think we have to provide `JedisThreadFactory` to name the threads (Jedis may introduce other ThreadPools in the future, we can reuse JedisThreadFactory).

Also, I think the executors in `MultiNodePipelineBase` need to have modification interfaces (we don't have to create them when users want to pass their own thread pool in).

sazzad16:
@yangbodong22011 WDYT about merging this PR before your commit and start a new PR from your commit?

yangbodong22011:
> @yangbodong22011 WDYT about merging this PR before your commit and start a new PR from your commit?

sure, everything is back to the way it was.

sazzad16:
@yangbodong22011 Thank you

sazzad16 (approved):

sazzad16:
Merged. Thank you @chenshi5012 for your contribution!

PR #3333, opened 2023-03-29
Introduce thread pool for multi node pipeline

yangbodong22011:
This PR mainly includes the following changes：
1. Introduce JedisThreadFactoryBuilder and JedisThreadPoolBuilder to standardize the creation of thread pools (this is beneficial for future debugging).
2. make executorService as a static object, created by default (even if the user does not use MultiNodePipeline), this can be recycled by multiple pipelines instead of frequent resource creation and destruction.
3. Users can customize the executor through `setExecutorService`.
4. submit method catch reject exception and record.

### DIFF
--- a/src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
+++ b/src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
@@ -7,9 +7,10 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Queue;
+import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
+import java.util.concurrent.RejectedExecutionException;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


sazzad16:
> make executorService as a static object, created by default (even if the user does not use MultiNodePipeline)

1. From my understanding, if MultiNodePipelineBase class isn't accessed, those thread pool classes won't be accessed and so initialized. Am I missing something?

2. WDYT about adding an option to disable this common thread pool? I.e. using system property?

yangbodong22011:
> * From my understanding, if MultiNodePipelineBase class isn't accessed, those thread pool classes won't be accessed and so initialized. Am I missing something?

Previously, every time a new MultiNodePipelineBase was created, an Executor would be created and destroyed immediately after use (very resource-intensive)

So I made it a static object, created by default, but if not used, no thread will be created (just a partial memory footprint)

> * WDYT about adding an option to disable this common thread pool? I.e. using system property?

I don't tend to add such options. Another way is:

We only add the setExecutorService interface, and the executorService is null by default. When the user does not set it through setExecutorService, it will be executed serially, as before; when the user sets the executor, it will be executed through the executor.

src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
```diff
@@ -58,14 +52,39 @@ public abstract class MultiNodePipelineBase implements PipelineCommands, Pipelin
...
     pipelinedResponses = new LinkedHashMap<>();
     connections = new LinkedHashMap<>();
     this.commandObjects = commandObjects;
   }
 
+  /**
+   * Provide an interface for users to set executors themselves.
+   * @param executor the executor
+   */
+  public static void setExecutorService(ExecutorService executor) {
```
dengliming, line 60: IMO, I suggest that it is better to supply this parameter directly to the constructor method rather than the setter method.  : )

sazzad16: @dengliming This is set once, works every time approach :)

yangbodong22011: > IMO, I suggest that it is better to supply this parameter directly to the constructor method rather than the setter method. : )

We need to pass the executor when `new JedisCluster` or `new JedisSharding`, and need a clear name to avoid misunderstanding by users, and there are many modifications(JedisCluster->ClusterPipeline->MultiNodePipelineBase).

But the benefit is that we can create the executor lazily, if the user passes it, we use it; otherwise, we can create a static object.

But I personally prefer the current code, and it is one-time and clear to set directly through the `MultiNodePipelineBase#setExecutorService` interface (About cost: there is only a part of memory usage, if the user is not using it, no thread will be created).

sazzad16:
@yangbodong22011 https://github.com/yangbodong22011/jedis/pull/1

sazzad16:
@yangbodong22011 https://github.com/yangbodong22011/jedis/pull/2

yangbodong22011:
> @yangbodong22011 [yangbodong22011#2](https://github.com/yangbodong22011/jedis/pull/2)

@sazzad16 commented.

asolimando:
@yangbodong22011 @sazzad16, we have a down-stream fork of Jedis exactly for the missing capability of re-using the thread pool. 

The PR seems in good shape and there are no pending comments from what I see, can you share what is preventing it from being merged?

Thanks!

sazzad16:
@asolimando We both agreed on the solution but couldn't agree on what should be the default behavior. There were no one else for the tie-break. So this PR got stuck.

asolimando:
> @asolimando We both agreed on the solution but couldn't agree on what should be the default behavior. There were no one else for the tie-break. So this PR got stuck.

Thanks for your reply, when you say default behavior you mean if we need to add a new constructor so that we always pass the executor, or if we rely on the `MultiNodePipelineBase#setExecutorService` method (defaulting to the current implementation if).

In our downstream version of this patch we went for the additional parameter in the constructor wherever needed, but honestly I wouldn't mind if the current patch goes in as-is and we rework our code to set the executor via `MultiNodePipelineBase#setExecutorService`, so if that helps breaking the tie, for me the patch could go in as it is now.

yangbodong22011:
> In our downstream version of this patch we went for the additional parameter in the constructor wherever needed

@sazzad16 @asolimando if you all agree with this change, I can revise this PR.

killergerbah:
I would love to see this functionality eventually merged upstream because I noticed performance issues with `MultiNodePipelineBase.sync`. However it does seem like a code smell to use a static setter that affects the behavior of all instances of the class. Why not implement setter on `JedisCluster` instead and supply the executor through the constructor of `MultiNodePipelineBase`?

killergerbah:
Apologies for the continued unsolicited feedback but because the choice of default behavior here does seem to have major performance implications, so I am going to offer it anyway:

I would argue that the most reasonable default behavior is to not try to perform the connection reads in parallel AT ALL. The original code attempts to do this by creating a thread pool for every call to `sync`, which is very expensive, but this PR attempts to give all the `MultiNodePipelinebase` instances the same thread pool, which runs the risk of thread pool starvation. Meanwhile, reading from the connections synchronously does not seem to make `sync` significantly slower - in fact, in my own tests locally it seems to go slightly faster most of the time.

kolulu23:
Got the same problem recently, I can see that thread pool creation is clearly the bottleneck when you use pipeline in a loop.
Previously in e193365 the `executorService` is a field of `MultiNodePipelineBase`, and it was changed to create anew everytime `sync` gets called (c33bdf7). It seems to me that being able to setup a dedicated thread pool just for `sync` is perfectly fine. 
Hope to see this PR get merged soon.

ggivo:
Closing as superseded. 

This was fixed by #4479: you can now pass your own shared executor with `cluster.pipelined(ExecutorService)`, and `pipelined()` keeps the old behavior.

 Available since 7.4.1 (backport #4480) and 8.0.0. If your scenario is still not covered, feel free to open an issue.

PR #3334, opened 2023-03-29
Introduce round-robining for scan and ft.search commands

sazzad16:


sazzad16:
@bsbodden

src/main/java/redis/clients/jedis/util/JedisRoundRobinBase.java
```diff
@@ -0,0 +1,81 @@
...
+      }
+    } else {
+      throw new IllegalArgumentException(connection.getValue().getClass() + "is not supported.");
+    }
+
+    lastReply = builder.build(rawReply);
+    iterationCompleted = isIterationCompleted(lastReply);
+    if (iterationCompleted) {
+      if (connections.isEmpty()) {
+        roundRobinCompleted = true;
```
bsbodden, line 76: Do we have any potential issues with threading / re-entrant code here?

bsbodden (commented):
LGTM! Nice job!

src/main/java/redis/clients/jedis/util/JedisRoundRobinBase.java
sazzad16: @bsbodden I thought it is not necessary to make these round robin objects thread safe. (Initiating them; e.g. calling scan, ftSearch, are thread safe.)

src/main/java/redis/clients/jedis/ScanRoundRobin.java
```diff
@@ -0,0 +1,43 @@
...
+import redis.clients.jedis.providers.ConnectionProvider;
+import redis.clients.jedis.resps.ScanResult;
+import redis.clients.jedis.util.JedisRoundRobinBase;
+
+public class ScanRoundRobin extends JedisRoundRobinBase<ScanResult<String>> {
+
+  private final int count;
+  private final Function<String, CommandArguments> args;
+
+  public ScanRoundRobin(ConnectionProvider connectionProvider, int batchCount, String match) {
```
chayim, line 15: IMHO having the function being named the same as the class feels clunky. The class looks like it should be named RoundRobinScanner - or something, just to differentiate.

I recognize it's the norm (plenty of socket.socket in the universe), but  IMHO purpose based function/class combinations are more clear

dengliming (approved):

yangbodong22011 (approved):
LGTM

PR #3336, opened 2023-03-30
Adding javadoc generation to CI

chayim:


.github/workflows/integration.yml
```diff
@@ -45,6 +45,9 @@ jobs:
       - name: Maven offline
         run: |
           mvn -q dependency:go-offline
+      - name: Build docs
+        run: |
+          mvn javadoc:jar
```
sazzad16, line 50: Move these after all test runs?

sazzad16: I said "move" not "remove". However, never mind.

sazzad16 (approved):

PR #3353, opened 2023-04-03
Avoid using QUIT command

yangbodong22011:
Please refer to: https://github.com/redis/redis/issues/11420

This PR cancels calling the quit command when the Jedis internal link is abnormal, but the user interface `Jedis.quit()` is not affected. This optimizes Jedis connection pool performance and addresses https://github.com/redis/jedis/issues/2105 and https://github.com/redis/jedis/issues/2108

See comments for some code details.


src/main/java/redis/clients/jedis/ConnectionFactory.java
```diff
@@ -74,21 +66,9 @@ public PooledObject<Connection> makeObject() throws Exception {
     Connection jedis = null;
     try {
       jedis = new Connection(jedisSocketFactory, clientConfig);
-      jedis.connect();
```
yangbodong22011, line 77: In the previous line of `new Connection`, connect will be called actively, so delete this line.

src/main/java/redis/clients/jedis/ConnectionFactory.java
```diff
@@ -74,21 +66,9 @@ public PooledObject<Connection> makeObject() throws Exception {
...
       return new DefaultPooledObject<>(jedis);
     } catch (JedisException je) {
-      if (jedis != null) {
```
yangbodong22011, line 80: If an exception is caught in new Connection, the following operations will be completed automatically and do not need to be executed again, see initializeFromClientConfig.

sazzad16 (changes requested):
Both Connection and Jedis has `quit()` methods. Deprecate those and suggest users to use `disconnect()` in javadoc.

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -406,6 +407,11 @@ private String clientSetname(final String name) {
     return getStatusCodeReply();
   }
 
+  /**
+   * @deprecated The quit command is deprecated and disconnect is recommended, see
+   * <a href="https://github.com/redis/redis/issues/11420">#11420</a>
```
sazzad16, line 412:
```suggestion
   * @deprecated The QUIT command is deprecated, see <a href="https://github.com/redis/redis/issues/11420">#11420</a>.
   * {@link Connection#disconnect()} can be used instead.
```

src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -410,8 +412,11 @@ public String flushAll(FlushMode flushMode) {
 
   /**
    * Ask the server to silently close the connection.
+   * @deprecated The quit command is deprecated and disconnect is recommended, see
+   * <a href="https://github.com/redis/redis/issues/11420">#11420</a>
```
sazzad16, line 416:
```suggestion
   * @deprecated The QUIT command is deprecated, see <a href="https://github.com/redis/redis/issues/11420">#11420</a>.
   * {@link Jedis#disconnect()} can be used instead.
```

src/main/java/redis/clients/jedis/commands/ServerCommands.java
```diff
@@ -24,7 +24,10 @@ public interface ServerCommands {
    * Ask the server to close the connection. The connection is closed as soon as all pending replies
    * have been written to the client.
    * @return OK
+   * @deprecated The quit command is deprecated and disconnect is recommended, see
+   * <a href="https://github.com/redis/redis/issues/11420">#11420</a>
```
sazzad16, line 28:
```suggestion
   * @deprecated The QUIT command is deprecated, see <a href="https://github.com/redis/redis/issues/11420">#11420</a>.
   * If available, {@code disconnect()} method in the concerned class can be used instead.
```

src/main/java/redis/clients/jedis/Connection.java
yangbodong22011: @sazzad16 How about recommending users to use 'Jedis. close()'? (Its also to call disconnect())

sazzad16: @yangbodong22011 If the object (Connection/Jedis) is received from a Pool, close() would just return it to the pool, not disconnect.

dengliming (approved):

PR #3356, opened 2023-04-04
Support report libname and libver to Redis

### LINKED ISSUE #3338: Support client name and version report
refer: https://github.com/redis/redis/pull/11758

yangbodong22011:
Reference: https://github.com/redis/redis/pull/11758 
Resolves #3338


src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -350,16 +354,41 @@ private void initializeFromClientConfig(JedisClientConfig config) {
...
-      if (clientName != null) {
-        // TODO: need to figure out something without encoding
-        clientSetname(clientName);
-      }
+      try {
+        List<CommandArguments> fireAndForgetMsg = new ArrayList<>();
+
+        int dbIndex = config.getDatabase();
+        if (dbIndex > 0) {
+          fireAndForgetMsg.add(new CommandArguments(Command.SELECT).add(Protocol.toByteArray(dbIndex)));
```
sazzad16, line 362: IMO, we should not ignore error of SELECT command.

src/main/java/redis/clients/jedis/args/ClientAttrOption.java
```diff
@@ -0,0 +1,23 @@
...
+
+import redis.clients.jedis.util.SafeEncoder;
+
+/**
+ * CLIENT SETINFO command attr option
+ * since redis 7.2
+ */
+public enum ClientAttrOption implements Rawable {
+    LIBNAME("LIB-NAME"),
+    LIBVER("LIB-VER");
```
sazzad16, line 11:
```suggestion
    LIB_NAME("LIB-NAME"),
    LIB_VER("LIB-VER");
```
WDYT?

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -350,16 +354,41 @@ private void initializeFromClientConfig(JedisClientConfig config) {
...
+
+        for (CommandArguments arg : fireAndForgetMsg) {
+          sendCommand(arg);
+        }
 
+        // getMany and ignore result
+        getMany(fireAndForgetMsg.size());
+
+      } catch (JedisDataException e) {
+        // just ignore dataException, such as unknown command
```
sazzad16, line 390: We can ignore unknown command but we should throw errors like illegal characters, etc.

sazzad16 (changes requested):
1.
```
Error:  testCredentialsProvider(redis.clients.jedis.JedisPooledTest)  Time elapsed: 0.01 s  <<< ERROR!
redis.clients.jedis.exceptions.JedisDataException: NOAUTH Authentication required.
	at redis.clients.jedis.JedisPooledTest.testCredentialsProvider(JedisPooledTest.java:246)
```

src/main/java/redis/clients/jedis/Connection.java
sazzad16: that's fine to me

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -350,16 +354,47 @@ private void initializeFromClientConfig(JedisClientConfig config) {
...
+        sendCommand(arg);
       }
 
+      List<Object> objects = getMany(fireAndForgetMsg.size());
+      for (Object obj : objects) {
+        if (obj instanceof JedisDataException) {
+          JedisDataException e = (JedisDataException)obj;
+          String errorMsg = e.getMessage().toUpperCase();
+          if (errorMsg.contains("UNKNOWN") ||
+              errorMsg.contains("NOAUTH")) {
```
yangbodong22011, line 391: @sazzad16 Here, we need to discuss: Jedis allows the following usage

```
Jedis jedis = new Jedis(host, port);
jedis.auth("xxx");
```

Since `auth` is run in the second step, the `client setinfo` command here will necessarily return a `NOAUTH` error, so I ignored `NOAUTH` here, and I modified the `testCredentialsProvider` test.

src/test/java/redis/clients/jedis/JedisPooledTest.java
```diff
@@ -243,6 +243,9 @@ public void cleanUp() {
       assertEquals(1, prepareCount.get());
       assertEquals(1, cleanupCount.get());
 
+      // clear existing `noauth` links
+      pool.getPool().clear();
+
```
sazzad16, line 248: We don't need this. The issue is with test itself but hidden due to `.clientName("my_shiny_client_name")`. Please check #3357 
```suggestion
```

src/main/java/redis/clients/jedis/args/ClientAttrOption.java
```diff
@@ -0,0 +1,23 @@
+package redis.clients.jedis.args;
+
+import redis.clients.jedis.util.SafeEncoder;
+
+/**
+ * CLIENT SETINFO command attr option
+ * since redis 7.2
+ */
+public enum ClientAttrOption implements Rawable {
```
sazzad16, line 9: Sorry but I promise I suggested it before. My browser or internet or something else didn't register it :sweat: 
```suggestion
public enum ClientAttributeOption implements Rawable {
```

src/main/java/redis/clients/jedis/Connection.java
sazzad16: @yangbodong22011 Agree. One idea is to enforce the constructor(s) to be sufficient for a connection. So NOAUTH won't be tolerated here. We can target this change for Jedis 5.

src/main/java/redis/clients/jedis/args/ClientAttrOption.java
yangbodong22011: done

src/main/java/redis/clients/jedis/Connection.java
yangbodong22011: OK，so here we keep filtering NOAUTH.

sazzad16: Yes, but add a comment maybe?
```suggestion
              errorMsg.contains("NOAUTH")) { // TODO: not filter out NOAUTH
```

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -397,16 +432,6 @@ private void auth(final Supplier<RedisCredentials> credentialsProvider) {
     getStatusCodeReply(); // OK
   }
 
-  public String select(final int index) {
-    sendCommand(Protocol.Command.SELECT, Protocol.toByteArray(index));
-    return getStatusCodeReply();
-  }
-
```
sazzad16, line 404: Keep this.

src/test/java/redis/clients/jedis/commands/jedis/ClientCommandsTest.java
```diff
@@ -62,6 +64,17 @@ public void nameBinary() {
     assertArrayEquals(name, client.clientGetnameBinary());
   }
 
+  @Test
+  public void clientSetInfoDefault() {
+    String libName = "jedis";
+    String libVersion = "999.999.999";
+    Assert.assertEquals("OK", client.clientSetInfo(ClientAttributeOption.LIB_NAME, libName));
+    Assert.assertEquals("OK", client.clientSetInfo(ClientAttributeOption.LIB_VER, libVersion));
```
sazzad16, line 72:
```suggestion
    assertEquals("OK", client.clientSetInfo(ClientAttributeOption.LIB_NAME, libName));
    assertEquals("OK", client.clientSetInfo(ClientAttributeOption.LIB_VER, libVersion));
```

src/test/java/redis/clients/jedis/commands/jedis/ClientCommandsTest.java
```diff
@@ -16,10 +16,12 @@
 import java.util.regex.Pattern;
 
 import org.junit.After;
+import org.junit.Assert;
```
sazzad16, line 19:
```suggestion
```

yangbodong22011:
`client list` output is 
```
id=109 addr=127.0.0.1:64276 laddr=127.0.0.1:6379 fd=10 name= age=26 idle=2 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 qbuf=0 qbuf-free=16890 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=18704 events=r cmd=get user=default redir=-1 resp=2 lib-name=jedis lib-ver=4.5.0-SNAPSHOT
```

src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -8514,6 +8521,13 @@ public String clientInfo() {
     return connection.getBulkReply();
   }
 
+  @Override
+  public String clientSetInfo(ClientAttributeOption attr, String value) {
+    checkIsInMultiOrPipeline();
+    connection.sendCommand(CLIENT, SETINFO.getRaw(), attr.getRaw(), encode(value));
+    return connection.getStatusCodeReply();
+  }
+
```
sazzad16, line 8530: [Discussion]

@yangbodong22011 WDYT about removing this support from Jedis class? Meaning to keep the usages of SETINFO only internally.

https://github.com/redis/redis/pull/11758 mentions it is for _Client libraries_, not necessarily for users.

yangbodong22011: I hope to keep the status quo. Although users do not need to actively set it in most cases, if users want to force it, we can still support it. And `redis-cli` or `sendCommand` can also accomplish the same function.

sazzad16 (approved):

sazzad16:
@pvarga88 

**Please state which Redis version you are using.** It'd be more helpful with the error, stacktrace, etc.

The doc just says _ignore failures_ but doesn't specify those failures. We tried to identify failures as much as we could.

According to my testing, this works fine from Redis 6.2. According to my knowledge, this should work fine from Redis 5.0.
@yangbodong22011 may be able to confirm further.

That leaves us pre-5.0. Frankly, that's not [_officially supported_](https://github.com/redis/jedis#supported-redis-versions) anymore.

sazzad16:
@pvarga88 No problem. Sharing information is how we can improve. FYI, there _can be_ a fix coming for Redis 4.0

PR #3357, opened 2023-04-06
Fix test with issue hidden due to optional client name

sazzad16:


src/test/java/redis/clients/jedis/JedisPooledTest.java
```diff
@@ -231,20 +231,23 @@ public void cleanUp() {
       }
     };
 
+    // TODO: do it without the help of pool config; from Connection constructor? (configurable) force ping?
```
yangbodong22011, line 234: > TODO: do it without the help of pool config; from Connection constructor? (configurable) force ping?

I don't think we need to do much about it.

If the user's JedisPoolConfig does not configure the correct password, it is normal to return NOAUTH in subsequent requests.

yangbodong22011 (approved):

PR #3360, opened 2023-04-08
Add round robin command example

sazzad16:


yangbodong22011 (approved):
It would be perfect if the README.md had a short introduction.
or we can update [wiki](https://github.com/redis/jedis/wiki/AdvancedUsage)

Issue #3350, opened 2023-04-02
Support ZREVRANK: Added the optional `WITHSCORE` argument.


chayim:
Documented [here](https://redis.io/commands/ZREVRANK).

Issue #3349, opened 2023-04-02
Support ZRANK: Added the optional `WITHSCORE` argument.


chayim:
Documented [here](https://redis.io/commands/ZRANK).

PR #3364, opened 2023-04-10
Support Z[REV]RANK command with score option

yangbodong22011:
fixes https://github.com/redis/jedis/issues/3349
fixes https://github.com/redis/jedis/issues/3350 

Since our `Tuple` class doesn't support `<Long, Double>` pairs, I chose `KeyValue`, but it's worth discussing.

Or we can consider changing the current `Tuple` to `Tuple<K,V>`.


sazzad16:
> we can consider changing the current `Tuple` to `Tuple<K,V>`.

This is one of the best thing we can do. But sadly it'd be a breaking change :disappointed: 

> Since our `Tuple` class doesn't support `<Long, Double>` pairs, I chose `KeyValue`, but it's worth discussing.

All I can think of is `RankTuple` which is exactly Tuple but with rank instead of element. Given the circumstances, I'm okay with KeyValue for Jedis 4.

@dengliming WDYT?

src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -389,6 +389,22 @@ public String toString() {
     }
   };
 
+  public static final Builder<KeyValue<Long, Double>> RANK_WITHSCORE_PAIR = new Builder<KeyValue<Long, Double>>() {
```
sazzad16, line 392:
```suggestion
  public static final Builder<KeyValue<Long, Double>> ZRANK_WITHSCORE_PAIR = new Builder<KeyValue<Long, Double>>() {
```

sazzad16 (changes requested):

Issue #3220, opened 2022-11-30
Introduce a common interface or base class among Pipeline and multi node pipeline classes


sazzad16:
Concerned methods:
- **sync()**
- close()

sazzad16:
Currently, Pipeline already extends `Queable` which is common with Transaction classes. Because we can't extend more than once, right now it is not possible to simply create PipelineBase class and extend from Pipeline. Now, perhaps, it is more important to share a common PipelineBase among Pipeline and multi node pipeline classes than to share Queable among Pipeline and Transaction classes.

PR #3369, opened 2023-04-12
Standalone mode do not implements samplekey commands in pipeline

yangbodong22011:
In standalone mode, pipeline commands do not need to implement samplekeyed commands. This pull request modifies the inheritance relationship of PipelineCommands and removes the relevant implementation commands.

### DIFF
--- a/src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
+++ b/src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
@@ -21,6 +21,8 @@ import redis.clients.jedis.bloom.*;
 import redis.clients.jedis.commands.PipelineBinaryCommands;
 import redis.clients.jedis.commands.PipelineCommands;
 import redis.clients.jedis.commands.RedisModulePipelineCommands;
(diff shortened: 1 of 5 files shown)


sazzad16:
@yangbodong22011 The objective is to allow users same interface (PipelineCommands "or" PipelineBinaryCommands) for both standalone and cluster mode. So even though sampleKey is not necessary for standalone mode, the methods are there so that the users may choose to use same to for both standalone and cluster mode.

I am even "hoping" to extend the idea (see #3220) by allowing both PipelineCommands "and" PipelineBinaryCommands.

yangbodong22011:
> The objective is to allow users same interface (PipelineCommands "or" PipelineBinaryCommands) for both standalone and cluster mode.

Because we have `UnifiedJedis.pipelined()` right?

sazzad16:
Not exactly because of `UnifiedJedis.pipelined()`; those common interfaces were introduced before that. But, yes, the common interfaces and `UnifiedJedis.pipelined()` are towards similar goal.

PR #3376, opened 2023-04-19
Simple documentation examples as test cases

dmaier-redislabs:
Hello,

this pull request adds the additional test folder `doc` to the existing set of JUnit tests.

The idea is that the files in this folder will be pulled for code examples within the documentation on redis.io. It would be good to run those tests on a regular basis for new versions of Jedis in order to ensure that the documentation examples are functional.

In the sake of the example, we tried to minimize the surrounding code as much as possible. This includes, for instance:

* A simple but not too abstract way of establishing the connection
* A neutral method name (e.g., `run`)
* Minimal error handling
* Blocks of code that are 'by default hidden' (`HIDE_START` ... `HIDE_END`) or entirely hidden (`REMOVE_START` ... `REMOVE_END`) within the documentation.
* The `@Test` test annotation will not be visible in the documentation.

Regards,
David


src/test/java/redis/clients/jedis/doc/example.json
```diff
@@ -0,0 +1,4 @@
+{
+    "id": "auto",
+    "file": "*Example.java"
+}
```
sazzad16, line 4: What is the purpose of this file?

src/test/java/redis/clients/jedis/doc/SetGetExample.java
```diff
@@ -0,0 +1,35 @@
...
+import org.junit.Test;
+import static org.junit.Assert.assertEquals;
+//REMOVE_END
+
+public class SetGetExample {
+
+    @Test
+    public void run() {
+
+        Jedis jedis = new Jedis("localhost", 6379);
```
sazzad16, line 16: Shouldn't we encourage users to use UnifiedJedis and variants?

sazzad16, line 16: This test will fail as Redis server running on `localhost:6379` is password protected in our test suite.

src/test/java/redis/clients/jedis/doc/SetGetExample.java
```diff
@@ -0,0 +1,35 @@
+//EXAMPLE: set_and_get
+//HIDE_START
+package redis.clients.jedis.doc;
```
sazzad16, line 3: Please add a file to define this package. Something like https://github.com/redis/jedis/blob/master/src/main/java/redis/clients/jedis/args/package-info.java

sazzad16 (changes requested):

src/test/java/redis/clients/jedis/doc/SetGetExample.java
dmaier-redislabs: Would it be possible to run your default test against another Redis instance with a password, but those tests here without password protection. I initially extended your BaseTest class, but removed it then again to keep the example as clean as possible. However, let me discuss with Igor. Those are multi-language examples. So if we decide for a password, then all languages (client libraries) would need to use the same password.

src/test/java/redis/clients/jedis/doc/SetGetExample.java
dmaier-redislabs: Will do.

src/test/java/redis/clients/jedis/doc/SetGetExample.java
dmaier-redislabs: Good question. The basic examples here (e.g., the README in this repo) use Jedis instead of UnifiedJedis. It would be clearly good to use the UnifiedJedis class for any Redis Stack related example. Let me check with Igor.

src/test/java/redis/clients/jedis/doc/example.json
dmaier-redislabs: It was initially used by the tool/script that builds the documentation site to discover examples. The latest version just looks for all files in this folder and checks for the `//EXAMPLE: $id` comment in the first line. So, I think that this can be removed in the upstream repo (your repo).

src/test/java/redis/clients/jedis/doc/SetGetExample.java
uglide: I vote for using UnifiedJedis to make it consistent.

uglide: Let's keep it as simple as possible and set up a separate GH Actions workflow to run "doctests" similar to what we have in redis-py https://github.com/redis/redis-py/blob/emb-examples/.github/workflows/doctests.yml

sazzad16: tagged @chayim

src/test/java/redis/clients/jedis/doc/SetGetExample.java
sazzad16: @dmaier-redislabs I've added one. Feel free add comments, instructions there.

src/test/java/redis/clients/jedis/doc/SetGetExample.java
bsbodden: For tests like this, I suggest we use TestContainers, particularly the Redis Version that Julien created and currently maintains. That way, we can specify the right version of Redis and configure it in a self-contained JUnit test. See https://github.com/redis-developer/testcontainers-redis

dmaier-redislabs: @bsbodden We decided by intention against such an abstraction because the test needs to act as an example for the documentation. So a reader of doc would be more interested in the vanilla way of using Redis instead of a simplified way of testing.

dmaier-redislabs:
> > It would be good to run those tests on a regular basis
> 
> To achieve that, the file name must end with `Test.java`

Hi @sazzad16,

as agreed with @chayim, we moved our test cases under a new test source directory `doctests`. I think it would make sense to create an additional profile in your Maven config to be able to switch the `testSourceDirectory` dependent on if we want to execute tests under `doctests` or `tests`. I was a bit shy to edit your maven build file. So your help is appreciated with that.

I think that we should allow letting test classes under `doctests` end with `*Example` because this will let the example code look cleaner.  We did put quite a bit effort it to ensure that other references to the surrounding test stuff (annotations, imports) aren't visible to the reader of the code example. Also, my local test runner didn't seem to have any issues with this extension. If you are using the Maven Surefire plugin for testing, then you can include test files with any prefix by configuring the includes/excudes. Please find a link to the documentation here:

* https://maven.apache.org/surefire/maven-surefire-plugin/examples/inclusion-exclusion.html

Regards,
David

sazzad16:
@dmaier-redislabs 

[My comment](https://github.com/redis/jedis/pull/3376#issuecomment-1514900767) 

> > It would be good to run those tests on a regular basis
> 
> To achieve that, the file name must end with `Test.java`

was based on running examples alongside core test suite 
and was made before @uglide['s comment](https://github.com/redis/jedis/pull/3376#discussion_r1172611536)

> Let's keep it as simple as possible and set up a separate GH Actions workflow to run "doctests" similar to what we have in redis-py https://github.com/redis/redis-py/blob/emb-examples/.github/workflows/doctests.yml

So, that comment of mine is not applicable when we are going to run examples separately.

sazzad16:
@chayim The rest is yours buddy :)

dmaier-redislabs:
@sazzad16 OK, cool. Should we anyway extend the Maven file to make running the `doctests` easier, or would you prefer to keep it as it is?

sazzad16:
@dmaier-redislabs AFAIU, extending Maven file is not enough; CI config file(s) are needed to be modified as well. That's why I pinged @chayim.

chayim:
Testing for Jedis, within CI currently does two things:

1.[ make test ](https://github.com/redis/jedis/blob/52a10e1ec21e3465e7ef38a4c207ff2978a8a0d7/.github/workflows/integration.yml#L53)- which runs mvn -Dtest=<stuff> clean compile test. In CI we run this with _TEST=""_

2. The second set of [CI tests](https://github.com/redis/jedis/blob/52a10e1ec21e3465e7ef38a4c207ff2978a8a0d7/.github/workflows/integration.yml#L57) are run after starting the docker and running specific tests against a known port.

As @uglide side, we should be taking a similar, _lazy_ approach to [redis-py](https://github.com/redis/redis-py/blob/emb-examples/.github/workflows/doctests.yml).  IMHO for Jedis this means:

1. A new job (doctests.yml) that can run in parallel with what we have.
2. It starts a docker
3. Triggers a mvn compile 
4. Runs only these tests, from this tree - against the stock redis running on 6379.

I'd be asking for a pom file we can trigger. I've pushed into this PR the proposed doctest run - and would like an assist with the pom trigger. WDYT?

uglide:
@dmaier-redislabs @chayim @sazzad16  AFAIK, it's not possible to change tests location for JUnit in Maven profile, so both types of tests should reside under the `src/test` folder. I've added required changes to make doctests runnable in my PR #3418

dmaier-redislabs:
@uglide Hi Igor,

I think that Maven allows you to override the test source folder for unit test executions. This config parameter needs to be changed globally. One way of doing this would be to have two different profiles with different values being set for this folder. I can look into it later.

PR #3381, opened 2023-04-27
Newly disconnected connections are set as broken

sazzad16:
Address 'Avoid using QUIT command (#3353)'

Imported from #3380


src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -190,13 +190,21 @@ public void connect() throws JedisConnectionException {
 
         outputStream = new RedisOutputStream(socket.getOutputStream());
         inputStream = new RedisInputStream(socket.getInputStream());
+
+        broken = false;
```
sazzad16, line 194: unset broken status when connection (re)initialized.

PR #3402, opened 2023-05-09
Fix binary value in comment in test code src/test/java/redis/clients/jedis/commands/jedis/BitCommandsTest.java

vector090:
When first setting "0" (string) to redis, the bits are actually 0011 0000, instead of number 0 with bits 0000 0000.

Goods new is: all the code logic are correct.   The problem only exists in the comment.

### DIFF
--- a/src/test/java/redis/clients/jedis/commands/jedis/BitCommandsTest.java
+++ b/src/test/java/redis/clients/jedis/commands/jedis/BitCommandsTest.java
@@ -35,6 +35,7 @@ public class BitCommandsTest extends JedisCommandsTestBase {
     String foo = "foo";
 
     jedis.set(foo, String.valueOf(0));
+    //  string "0" with bits: 0011 0000
 
     jedis.setbit(foo, 3, true);
     jedis.setbit(foo, 7, true);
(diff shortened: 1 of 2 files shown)


vector090:
For convenience, I've also written some code to print bits, which makes bit navigation easier.
e.g. The final bits are:

    /*
0011 0001  0000 0100  0000 0000  0000 0000  0000 0001  
0123 4567  8901 2345  6789 0123  4567 8901  2345 6789  
     */

sazzad16 (changes requested):
Same state and comment is also present in following `bitposBinary()` test method.

Again, both of these tests are available in src/test/java/redis/clients/jedis/commands/unified/BitCommandsTestBase.java as well.

@vector090 Could you please change in all of these places? 3 more places to be exact.

vector090:
> Same state and comment is also present in following `bitposBinary()` test method.
> 
> Again, both of these tests are available in src/test/java/redis/clients/jedis/commands/unified/BitCommandsTestBase.java as well.
> 
> @vector090 Could you please change in all of these places? 3 more places to be exact.

WIP

sazzad16:
@vector090 It seems to be the first one.

vector090:
> @vector090 It seems to be the first one.

Code updated.
Please review.

sazzad16 (approved):

Issue #3410, opened 2023-05-12
CI for different Redis versions (4.0, 5.0, 6.0, 7.0...) 


yangbodong22011:
We need CIs for different Redis versions (4.0, 5.0, 6.0, 7.0...) to guarantee Jedis backwards compatibility. Before that, we need to mark the test cases differently. I created a new issue to track this issue.

uglide:
This was introduced long time ago.

PR #3411, opened 2023-05-12
Fix: client setname need ignore syntax error in redis4

yangbodong22011:
1. Redis [4.0](https://github.com/redis/redis/blob/4.0/src/networking.c#L1744) and before, we need to ignore `Syntax error`. 
2. Redis [5.0](https://github.com/redis/redis/blob/5.0/src/networking.c#L1933) and later, we need to ignore `Unknown subcommand error`.
3. Because Jedis allows Jedis jedis = new Jedis() in advance, and jedis.auth(password) later, we need to ignore `NOAUTH errors`.

We need CIs for different Redis versions (4.0, 5.0, 6.0, 7.0...) to guarantee Jedis backwards compatibility. Before that, we need to mark the test cases differently. I created a new issue https://github.com/redis/jedis/issues/3410 to track this issue.

### DIFF
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -398,7 +398,14 @@ public class Connection implements Closeable {
         if (obj instanceof JedisDataException) {
           JedisDataException e = (JedisDataException)obj;
           String errorMsg = e.getMessage().toUpperCase();
-          if (errorMsg.contains("UNKNOWN") ||
+          /**
+           * 1. Redis 4.0 and before, we need to ignore `Syntax error`.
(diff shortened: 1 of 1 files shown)


sazzad16 (approved):

chayim:
Always happy to see PRs supporting older versions of redis, but they shouldn't be a focus point. This is the [support matrix](https://github.com/redis/jedis#supported-redis-versions)

yangbodong22011:
> Always happy to see PRs supporting older versions of redis, but they shouldn't be a focus point. This is the [support matrix](https://github.com/redis/jedis#supported-redis-versions)

ohh, thanks for your notice, but for this PR, I tend to merge it.

PR #3413, opened 2023-05-12
fearture: Implemented FILTER function at groupby lvl and added tests.

rasrov:
Implemented FILTER function at groupby level.

### DIFF
--- a/src/main/java/redis/clients/jedis/search/aggr/AggregationBuilder.java
+++ b/src/main/java/redis/clients/jedis/search/aggr/AggregationBuilder.java
@@ -95,18 +95,21 @@ public class AggregationBuilder {
     return this;
   }
 
-  public AggregationBuilder groupBy(Collection<String> fields, Collection<Reducer> reducers) {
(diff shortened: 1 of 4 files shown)


sazzad16:
@rasrov Thank you for your PR. We're trying hard but it may take some time to review your PR. Thanks again for your patience.

sazzad16:
FILTER is not a sub-argument under GROUPBY argument. Rather it is a direct argument under FT.AGGREGATE command itself.

What you're trying to achieve is already possible. Check following portion of your test case and where I've made the adjustments:

```java
    Group firstGroup = new Group("@eventId", "@type", "@name", "@createdTime");
    //firstGroup.filter("@createdTime < 1683714000"); // <-- this what you're trying to implement

    Group secondGroup = new Group( "@type", "@name");
    secondGroup.reduce(Reducers.count().setAlias("count"));

    AggregationBuilder r = new AggregationBuilder("*")
        .groupBy(firstGroup)
        .filter("@createdTime < 1683714000") // <-- this is what you have to do
        .groupBy(secondGroup)
        .filter(("@count > 1"));

    AggregationResult res = client.ftAggregate(index, r);
    assertEquals(2, res.getResults().size());
```

PR #3415, opened 2023-05-14
List Jedis 5.0 breaking changes

sazzad16:


docs/4to5.md
```diff
@@ -0,0 +1,82 @@
+# Jedis 4 to Jedis 5 Breaking Changes
+
+- `StreamConsumersInfo` has been renamed to `StreamConsumerInfo`.
+
+- Return type of `getAgeSeconds()` in `AccessControlLogEntry` has been changed from `String` to `Double`.
```
chayim, line 5: For all of these I think you should change to 
```getAgeSeconds() in AccessControlLogEntry``` now returns `Double` instead of `String` .

Then, we can sort alphabetically too - and make it easier for users.

File: docs/4to5.md [hunk 2]
```diff
@@ -0,0 +1,82 @@
+# Jedis 4 to Jedis 5 Breaking Changes
```
chayim, line 1: ```# Jedis 5 Breaking Changes```

No need to declare vs. version 4. Should be self-evident.

docs/4to5.md
```diff
@@ -0,0 +1,82 @@
...
+- `STREAM_CONSUMERS_INFO_LIST` in BuilderFactory has been renamed to `STREAM_CONSUMER_INFO_LIST`.
+- `STREAM_AUTO_CLAIM_ID_RESPONSE` in BuilderFactory has been renamed to `STREAM_AUTO_CLAIM_JUSTID_RESPONSE`.
+
+- Following BuilderFactory implementations have been removed:
+  - `BYTE_ARRAY` (use `BINARY`)
+  - `BYTE_ARRAY_LIST` (use `BINARY_LIST`)
+
+<!--- Deprecated in Jedis 4 --->
+
+- `quit()` method has been removed from `ServerCommands` and implementations.
```
chayim, line 28: `quit() method has been removed from `Connection` and ServerCommands implementations`

PR #3416, opened 2023-05-15
CI improvements for RESP3 and version validation

chayim:


chayim:
@sazzad16 This now fails on a per redis basis. Can we limit the tests it runs? This is the part I'd love your help on...

sazzad16:
@chayim Sure. Let's discuss _how_ we want to limit...

chayim:
Supplanted in existing release.

Issue #3438, opened 2023-05-25
Jedis 4.4.1 fails on CLIENT SETINFO for Redis 6.x if the user does not have all +CLIENT permissions


XA21X:
### Expected behavior

Jedis should successfully connect to Redis 6.x clusters using a user without _all_ `+CLIENT` permissions.

### Actual behavior

Jedis fails to perform slots renewal on connection due to a `NOPERM` error. 

```
...
Caused by: redis.clients.jedis.exceptions.JedisClusterOperationException: Could not initialize cluster slots cache.
...
Suppressed: redis.clients.jedis.exceptions.JedisAccessControlException: NOPERM this user has no permissions to run the 'client' command or its subcommand
```

### Steps to reproduce:

1. Setup a Redis 6.2.11 server.
2. Create a Redis user without _the_ `+client` permission, e.g. `+@all -@dangerous`, or more specifically:
  ```
  ACL SETUSER test +@all allkeys -@dangerous +client|id +client|getname +client|setname +config|get +cluster|info +cluster|keyslot +cluster|myid +cluster|nodes +cluster|slots >test on
  ```
4. Connect to Redis using Jedis 4.4.1 with the limited user - which does not work.

This error can be reproduced using `redis-cli` by running `CLIENT SETINFO LIB-NAME test`.

Adding specific `client` subpermissions to the user does not affect the result unless it's `+client|setinfo`. The command doesn't exist in Redis 6.x, but its permission can still be added. Having `+client|setinfo` is a sufficient workaround, because after passing the permissions check, the command correctly fails with `ERR Unknown` instead of `NOPERM`.

#### Jedis version:

Jedis **4.4.1** introduced the "fire and forget" `CLIENT SETINFO` on connection: redis/jedis#3356 but does not ignore `NOPERM` errors.

#### Redis version:

Redis **6.2.11**, but it is likely that any 6.x is affected.

Redis 7.x seems to have fixed this as part of wider work surrounding redis/redis#9504.

Most of the `CLIENT` commands are no longer in the `@dangerous` group, so my example above _includes_ the required permission. It also makes the command existence check able to detect the missing command before evaluating permissions.

#### Java version:

```
openjdk version "11.0.19" 2023-04-18 LTS
OpenJDK Runtime Environment Corretto-11.0.19.7.1 (build 11.0.19+7-LTS)
OpenJDK 64-Bit Server VM Corretto-11.0.19.7.1 (build 11.0.19+7-LTS, mixed mode)
```

[...]

PR #3449, opened 2023-06-01
Client setinfo ignore all exception

yangbodong22011:
This PR contains the following changes (fixes https://github.com/redis/jedis/issues/3438):

1. Check the character legality of clientName in advance through `validateClientInfo`, so that character exceptions are ignored in `initializeFromClientConfig`.
2. Move the `select` command out of the fire-and-forget message (we care about its return value).
3. For the test of RedisCredentials, we gave it "invalidPass" for the first time to make it fail the verification instead of skipping the verification.

### DIFF
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -20,6 +20,7 @@ import redis.clients.jedis.commands.ProtocolCommand;
 import redis.clients.jedis.exceptions.JedisConnectionException;
 import redis.clients.jedis.exceptions.JedisDataException;
 import redis.clients.jedis.exceptions.JedisException;
+import redis.clients.jedis.exceptions.JedisValidationException;
 import redis.clients.jedis.util.IOUtils;
 import redis.clients.jedis.util.JedisMetaInfo;
 import redis.clients.jedis.util.RedisInputStream;
@@ -370,6 +371,23 @@ public class Connection implements Closeable {
     return responses;
   }
 
+  /**
+   * Check if the client name libname, libver, characters are legal
+   * @param info the name
+   * @return Returns true if legal, false throws exception
+   * @throws JedisException if characters illegal
+   */
+  private static boolean validateClientInfo(String info) {
+    for (int i = 0; i < info.length(); i++) {
+      char c = info.charAt(i);
+      if (c < '!' || c > '~') {
+        throw new JedisValidationException("client info cannot contain spaces, "
+            + "newlines or special characters.");
+      }
+    }
+    return true;
+  }
+
   private void initializeFromClientConfig(JedisClientConfig config) {
     try {
       connect();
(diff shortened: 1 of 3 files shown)


src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -427,30 +426,9 @@ private void initializeFromClientConfig(JedisClientConfig config) {
...
-          } else {
-            throw e;
-          }
-        }
-      }
+      getMany(fireAndForgetMsg.size());
     } catch (JedisException je) {
       try {
-        setBroken();
         disconnect();
```
yangbodong22011, line 432: `disconnect()` will call `setBroken()` in `finally`, so we remove this line.

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -387,9 +387,12 @@ private void initializeFromClientConfig(JedisClientConfig config) {
...
         Supplier<RedisCredentials> credentialsProvider = config.getCredentialsProvider();
         if (credentialsProvider instanceof RedisCredentialsProvider) {
-          ((RedisCredentialsProvider) credentialsProvider).prepare();
-          auth(credentialsProvider);
-          ((RedisCredentialsProvider) credentialsProvider).cleanUp();
+          try {
+            ((RedisCredentialsProvider) credentialsProvider).prepare();
+            auth(credentialsProvider);
+          } finally {
+            ((RedisCredentialsProvider) credentialsProvider).cleanUp();
```
sazzad16, line 394: :+1: Good job!

src/main/java/redis/clients/jedis/DefaultJedisClientConfig.java
```diff
@@ -113,6 +113,25 @@ public HostAndPortMapper getHostAndPortMapper() {
     return hostAndPortMapper;
   }
 
+  /**
+   * Check if the client name characters are legal
+   * @param name the name
+   * @return Returns true if legal, false otherwise
+   */
+  public static boolean validateClientName(String name) {
```
sazzad16, line 121: This validation should be done in Connection. Otherwise, any other implementation of JedisClientConfig will not go through this validation.

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -370,6 +370,23 @@ public List<Object> getMany(final int count) {
     return responses;
   }
 
+  /**
+   * Check if the client name libname, libver, characters are legal
+   * @param info the name
+   * @return Returns true if legal, false throws exception
+   * @throws JedisException if characters illegal
+   */
+  public static boolean validateClientInfo(String info) {
```
sazzad16, line 379: It should be private.
If you really want it to be public, util package is a better place.

yangbodong22011: I will change it to public.
I currently haven't found a suitable class in the util package to store it (I should need to create a new class, but I haven't come up with a good name). If you prefer to move it to util, I can also accept it.

sazzad16: > but I haven't come up with a good name

I understand :smile:

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -370,6 +370,23 @@ public List<Object> getMany(final int count) {
...
+   * Check if the client name libname, libver, characters are legal
+   * @param info the name
+   * @return Returns true if legal, false throws exception
+   * @throws JedisException if characters illegal
+   */
+  private static boolean validateClientInfo(String info) {
+    for (int i = 0; i < info.length(); i++) {
+      char c = info.charAt(i);
+      if (c < '!' || c > '~') {
+        throw new JedisException("client info cannot contain spaces, "
```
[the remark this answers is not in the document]
yangbodong22011: Thanks for the reminder, I remember we had a specific class for syntax exceptions, but I couldn't remember the name.
already changed.

sazzad16 (approved):

PR #3452, opened 2023-06-01
Allow setting default dialect for search module

### LINKED ISSUE #3447: Add RediSearch Dialect Version Configuration Option
In the current implementation of the client, users are required to specify the RediSearch dialect version for every `FT.SEARCH` or `FT.AGGREGATE` command. This repetitive configuration process might be prone to errors and can lead to inconsistency issues across different parts of a project where different dialect versions might be unintentionally used.

In order to make our client more user-friendly and maintain consistency throughout the usage, we need to provide a configuration option that allows users to set the RediSearch dialect version at the connection level. This will ensure the specified version is used across all `FT.SEARCH` and `FT.AGGREGATE` commands, removing the need to set it individually each time.

**Acceptance Criteria:**

- A new configuration option should be added to the client that allows users to specify the RediSearch dialect version at the connection level.
- This configuration option should be optional. If it is not specified, the client should fall back to the current behavior.
- If the configuration option is set, all `FT.SEARCH`, `FT.AGGREGATE`, `FT.EXPLAIN`, `FT.EXPLAINCLI`, and `FT.SPELLCHECK` commands should use the specified RediSearch dialect version by default.

sazzad16:
Resolves #3447

### DIFF
--- a/src/main/java/redis/clients/jedis/CommandObjects.java
+++ b/src/main/java/redis/clients/jedis/CommandObjects.java
@@ -4172,9 +4172,6 @@ public class CommandObjects {
     this.jsonObjectMapper = jsonObjectMapper;
   }
 
-  /**
-   * If this is set, dialect set in respective params/builders will be ignored.
-   */
   public void setDefaultSearchDialect(int dialect) {
     if (dialect == 0) throw new IllegalArgumentException("DIALECT=0 cannot be set.");
     this.searchDialect.set(dialect);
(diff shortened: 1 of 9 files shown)


uglide:
@sazzad16 Could you add the same for FT.AGGREGATE? I've updated the issue description.

src/main/java/redis/clients/jedis/CommandObjects.java
```diff
@@ -4168,6 +4172,11 @@ public void setJsonObjectMapper(JsonObjectMapper jsonObjectMapper) {
     this.jsonObjectMapper = jsonObjectMapper;
   }
 
+  public void setDefaultSearchDialect(int dialect) {
```
chayim, line 4175: The issue says ```This configuration option should be optional. If it is not specified, the client should fall back to the current behavior.```

Raising this argument isn't that.

sazzad16: Firstly, users would be using the one in UnifiedJedis. I guess we're concerned about that and so check my answer in test file.

Issue #3392, opened 2023-05-05
Spring Boot 3.x.x update throws JedisClusterOperationException Could not initialize cluster slots cache.


kableprime:
### Actual behavior

We have a spring boot application using jedis for connecting to redis in some of the services but since we updated to spring boot 3.x.x all unit tests using jedis fail with the following stack:
(I have shortened the entire stack, because the redis config and redis method is written in a our ownlibrary that we re-use across multiple apps)

```
Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'redisUtils' defined in URL [jar:file:/Users/I343767/.m2/repository/com/sap/etd/cloud/streaming-commons/3.2.5/streaming-commons-3.2.5.jar!/com/sap/etd/cloud/commons/utils/RedisUtils.class]: Unsatisfied dependency expressed through constructor parameter 1: Error creating bean with name 'getRedisTemplate' defined in class path resource [com/sap/etd/cloud/commons/config/redis/RedisConfig.class]: Failed to instantiate [org.springframework.data.redis.core.StringRedisTemplate]: Factory method 'getRedisTemplate' threw exception with message: Error creating bean with name 'getClusterRedisConnectionFactory' defined in class path resource [com/sap/etd/cloud/commons/config/redis/RedisConfig.class]: Could not initialize cluster slots cache.
	at org.springframework.beans.factory.support.ConstructorResolver.createArgumentArray(ConstructorResolver.java:800)
	at org.springframework.beans.factory.support.ConstructorResolver.autowireConstructor(ConstructorResolver.java:245)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.autowireConstructor(AbstractAutowireCapableBeanFactory.java:1352)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBeanInstance(AbstractAutowireCapableBeanFactory.java:1189)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:560)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:520)
	at org.springframework.beans.factory.support.AbstractBeanFactory.lambda$doGetBean$0(AbstractBeanFactory.java:326)
[...]

PR #3455, opened 2023-06-03
Add an option to ignore cluster init error

sazzad16:
References of inspiration:
- https://github.com/redis/jedis/issues/3392 
- https://stackoverflow.com/questions/76207822/updating-to-spring-boot-3-0-6-jedis-fails-unit-tests-with-jedisclusteroperatio 
- https://stackoverflow.com/questions/76207822/updating-to-spring-boot-3-0-6-jedis-fails-unit-tests-with-jedisclusteroperatio/76224927#comment134405538_76207822 
  > I suggest also filing a ticket with the Jedis driver to have the exception handling improved during slot initialization as the actual error doesn't bubble up. – **mp911de** 
- https://github.com/redis/jedis/issues/3392#issuecomment-1573925414 
  > In my case whole application is not able to start as I am configuring everything on startup, so this is creating tight coupling – @nikhilgurjar 

---

A `JedisClusterOperationException` with message `Could not initialize cluster slots cache.` is thrown during failed cluster initialization. This behavior is introduced since Jedis 4.0.0. To avoid this error, just set `jedis.cluster.initNoError` (any value) system property.

### DIFF
--- a/src/main/java/redis/clients/jedis/providers/ClusterConnectionProvider.java
+++ b/src/main/java/redis/clients/jedis/providers/ClusterConnectionProvider.java
@@ -19,6 +19,8 @@ import redis.clients.jedis.exceptions.JedisException;
 
 public class ClusterConnectionProvider implements ConnectionProvider {
 
+  private static final String INIT_NO_ERROR_PROPERTY = "jedis.cluster.initNoError";
+
   protected final JedisClusterInfoCache cache;
 
   public ClusterConnectionProvider(Set<HostAndPort> clusterNodes, JedisClientConfig clientConfig) {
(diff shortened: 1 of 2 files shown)


yangbodong22011:
I don't particularly want to add `INIT_NO_ERROR_PROPERTY `, unless we wait for more user feedback.

yangbodong22011:
It's really not easy to deal with, let's hear @kableprime opinion on this PR.

chayim:
I agree with @yangbodong22011 that we shouldn't **have** **to** handle this. Let's let it bake for a bit... It sounds like fixing this library do to an issue there - ideally the fix would be in the downstream library

chayim (approved):

sazzad16:
@chayim I'm thinking here https://github.com/redis/jedis/wiki/FAQ. WDYT?

fzheng2022:
The init error is suppressed by the option. But will it work normally afterwards? what is meaning by "failed cluster initialization". My use case: All tries had the error while the Cluster seems up and stable.

sazzad16:
@fzheng2022 

> The init error is suppressed by the option.

To be exact, the init error _can be_ suppressed by the option.

> will it work normally afterwards?

No.

> what is meaning by "failed cluster initialization".

(Mostly) Not being able to connect to any of the nodes and fetch cluster information.

PR #3458, opened 2023-06-05
Client setinfo ignore all exception (#3449)

sazzad16:
This PR contains the following changes:

1. Check the character legality of clientName in advance through `validateClientInfo`, so that character exceptions are ignored in `initializeFromClientConfig`.
2. Move the `select` command out of the fire-and-forget message (we care about its return value).
3. For the test of RedisCredentials, we gave it "invalidPass" for the first time to make it fail the verification instead of skipping the verification.

### DIFF
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -352,39 +352,59 @@ public class Connection implements Closeable {
     return responses;
   }
 
+  /**
+   * Check if the client name libname, libver, characters are legal
+   * @param info the name
+   * @return Returns true if legal, false throws exception
+   * @throws JedisException if characters illegal
+   */
+  private static boolean validateClientInfo(String info) {
+    for (int i = 0; i < info.length(); i++) {
+      char c = info.charAt(i);
+      if (c < '!' || c > '~') {
+        throw new JedisException("client info cannot contain spaces, "
+            + "newlines or special characters.");
+      }
+    }
+    return true;
+  }
+
   private void initializeFromClientConfig(JedisClientConfig config) {
     try {
       connect();
 
       Supplier<RedisCredentials> credentialsProvider = config.getCredentialsProvider();
       if (credentialsProvider instanceof RedisCredentialsProvider) {
-        ((RedisCredentialsProvider) credentialsProvider).prepare();
-        auth(credentialsProvider);
-        ((RedisCredentialsProvider) credentialsProvider).cleanUp();
+        try {
+            ((RedisCredentialsProvider) credentialsProvider).prepare();
+            auth(credentialsProvider);
+          } finally {
+            ((RedisCredentialsProvider) credentialsProvider).cleanUp();
+          }
       } else {
         auth(credentialsProvider);
       }
 
-      List<CommandArguments> fireAndForgetMsg = new ArrayList<>();
-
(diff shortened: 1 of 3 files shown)


src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -352,70 +352,69 @@ public List<Object> getMany(final int count) {
...
+  /**
+   * Check if the client name libname, libver, characters are legal
+   * @param info the name
+   * @return Returns true if legal, false throws exception
+   * @throws JedisException if characters illegal
+   */
+  private static boolean validateClientInfo(String info) {
+    for (int i = 0; i < info.length(); i++) {
+      char c = info.charAt(i);
+      if (c < '!' || c > '~') {
```
chayim, line 364: Since INFO should just contain [a-zA-z0-9] why not validate with a regular expression, instead of looping through a string? _java.util.regex._ for the win?

sazzad16: https://github.com/redis/redis/pull/11758 

> The values cannot contain spaces, newlines and any wild ASCII characters, but all other normal chars are accepted, e.g `.`, `=` etc (same as CLIENT NAME).

@chayim

yangbodong22011 (approved):

src/main/java/redis/clients/jedis/Connection.java
yangbodong22011: Interesting suggestion, in order to find out which performance is better between `regex` and `for range`, I wrote a test code and did an experiment to get the result:

|cost time/length(byte)|10|1000|1000|10000|
|-|-|-|-|-|
|for range(ms)|1|1|2|5|
|regex(ms)|14|13|14|15|

conclusion: clientname is usually a token of no more than 20 bytes, so we don't have to modify it to regex.

test code
```
import java.util.ArrayList;
import java.util.List;
import java.util.Random;

public class Main {
    /**
     * generate Random String between 0 ~ 127
     * @param length
     * @return
     */
    public static String generateRandomString(int length) {
        Random random = new Random();
        StringBuilder sb = new StringBuilder(length);
        for (int i = 0; i < length; i++) {
            sb.append((char) (random.nextInt(128)));
        }
        return sb.toString();
    }

    /**
     * regex validate
     * @param info
     * @return
     */
    public static boolean validateClientInfoRegex(String info) {
        String regex = "[!-~]+";
        return info.matches(regex);
    }

    /**
     * for range validate
     * @param info
     * @return
     */
    private static boolean validateClientInfo(String info) {
        for (int i = 0; i < info.length(); i++) {
            char c = info.charAt(i);
            if (c < '!' || c > '~') {
                return false;
            }
        }
        return true;
    }

        List<String> infoList = new ArrayList<>();
        for (int i = 0; i < 10000; i++) {
            infoList.add(generateRandomString(100000));
        }

        long begin = System.currentTimeMillis();
        for (String info : infoList) {
            validateClientInfo(info);
        }
        long end = System.currentTimeMillis();
        System.out.println("for range use time: " + (end-begin));

        begin = System.currentTimeMillis();
        for (String info : infoList) {
            validateClientInfoRegex(info);
        }
        end = System.currentTimeMillis();
        System.out.println("regex use time: " + (end-begin));
    }
}
```

chayim: Great analysis.

PR #3460, opened 2023-06-06
Support Json commands after RESP3 update

sazzad16:


chayim (dismissed):

chayim:
Cross: Let's pull out the JSONV1 universe. They can't pass, everything has been V2 for a while now. IMHO V1 is completely unsupported.

PR #3475, opened 2023-06-27
Discard transaction if any command failed to queued

### LINKED ISSUE #1473: Writing to read-only slave fails silently inside of MULTI block

mikolajstaszkiewicz:
Fixes #1473

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
 	<packaging>jar</packaging>
 	<groupId>redis.clients</groupId>
 	<artifactId>jedis</artifactId>
-	<version>3.8.0-SNAPSHOT</version>
+	<version>3.11.0-SNAPSHOT</version>
 	<name>Jedis</name>
 	<description>Jedis is a blazingly small and sane Redis java client.</description>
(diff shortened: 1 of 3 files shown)


sazzad16:
@mikolajstaszkiewicz Thank you for your interest in Jedis and improving it.

Unless something severe or urgent comes up, we are not considering anything for Jedis 3.x. IMO, this does not falls into either category.

You may look into ReliableTransaction (since 4.x) if that suits your purpose.

PR #3480, opened 2023-07-13
Replace `synchronized` with `j.u.c.l.ReentrantLock` for Loom

babanin:
## Goal

In Sep 2023, JDK 21 will be released. One of the most anticipated feature - Project Loom aka Virtual threads.

One of the problems with the current implementation, are `synchronized` blocks:
> However, not all Java constructs have been retrofitted that way. Such operations include synchronized methods and code blocks. Using them causes the virtual thread to become pinned to the carrier thread. When a thread is pinned, blocking operations will block the underlying carrier thread—precisely as it would happen in pre-Loom times.

From here: https://softwaremill.com/what-is-blocking-in-loom/#synchronization-problems

The main goal of this PR is to replace `synchronized` blocks with `ReentrantLock`s who are not susceptible to this problem.

## Approach

My approach was quite simple: replace `synchronized` blocks with `ReentrantLock`'s. However, in a few cases, though, I believe we can get rid of synchronization at all:

1. This code was written by @clebersonp:
```java
 private JsonObjectMapper getJsonObjectMapper() {
    JsonObjectMapper localRef = this.jsonObjectMapper;
    if (Objects.isNull(localRef)) {
      synchronized (this) {
        localRef = this.jsonObjectMapper;
        if (Objects.isNull(localRef)) {
          this.jsonObjectMapper = localRef = new DefaultGsonObjectMapper();
        }
      }
    }
    return localRef;
  }
```
I replaced it with:
```java
  private JsonObjectMapper getJsonObjectMapper() {
    if (Objects.isNull(this.jsonObjectMapper)) {
        this.jsonObjectMapper = new DefaultGsonObjectMapper();
    }
    
    return this.jsonObjectMapper;
  }
```
The field `jsonObjectMapper` marked as `volatile`, thus we establish connection (acquire-release) between thread which writes to this field and other threads which will read this field.

However, there is still a scenario when multiple threads will enter this method at the same time and see `null`: all of them will create an object and write it to a variable. This won't cause any problems since the `DefaultGsonObjectMapper` is stateless, but the object creation operation itself may be more costly than lock acquisition. Do you think it is worthwhile to return the lock here?

2. This code was written by @sazzad16: 

```java
  private void createBuilder(String graphName) {
    synchronized (builders) {
      builders.putIfAbsent(graphName, new ResultSetBuilder(new GraphCacheImpl(graphName)));
    }
  }
```
Since `builders` is `ConcurrentHashMap`, we can avoid synchronization and either use `putIfAbsent` or, if creation of `ResultSetBuilder` and `GraphCacheImpl` is expensive, use `computeIfAbsent`:
```java
builders.computeIfAbsent(graphName, graphNameKey -> new ResultSetBuilder(new GraphCacheImpl(graphNameKey)));
```

## Alternative
It worth mentioning a [comment](https://www.reddit.com/r/java/comments/14vrfim/comment/jrgkeja/?utm_source=share&utm_medium=web2x&context=3) on Reddit left by Ron Pressler (Project Loom lead) that the work on making `synchronized` not pin is underway. 

Leave `synchronized` blocks as is and wait for next version of JDK (22+).

### DIFF
--- a/src/main/java/redis/clients/jedis/CommandObjects.java
+++ b/src/main/java/redis/clients/jedis/CommandObjects.java
@@ -4432,20 +4432,20 @@ public class CommandObjects {
    * @see DefaultGsonObjectMapper
    */
   private JsonObjectMapper getJsonObjectMapper() {
-      JsonObjectMapper localRef = this.jsonObjectMapper;
-      if (Objects.isNull(localRef)) {
-          mapperLock.lock();
-          
-          try {
-              localRef = this.jsonObjectMapper;
-              if (Objects.isNull(localRef)) {
-                  this.jsonObjectMapper = localRef = new DefaultGsonObjectMapper();
-              }
-          } finally {
-              mapperLock.unlock();
-          }
+    JsonObjectMapper localRef = this.jsonObjectMapper;
+    if (Objects.isNull(localRef)) {
+      mapperLock.lock();
+
+      try {
+        localRef = this.jsonObjectMapper;
+        if (Objects.isNull(localRef)) {
+          this.jsonObjectMapper = localRef = new DefaultGsonObjectMapper();
+        }
+      } finally {
+        mapperLock.unlock();
       }
-      return localRef;
+      }
+    return localRef;
   }
 
   public void setJsonObjectMapper(JsonObjectMapper jsonObjectMapper) {
--- a/src/main/java/redis/clients/jedis/CommandObjects.java
+++ b/src/main/java/redis/clients/jedis/CommandObjects.java
@@ -4427,7 +4427,11 @@ public class CommandObjects {
   /**
    * Get the instance for JsonObjectMapper if not null, otherwise a new instance reference with
    * default implementation will be created and returned.
-   *
+   * <p>This process of checking whether or not
+   * the instance reference exists follows <a
+   * href="https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_Java"
+   * target="_blank">'double-checked lock optimization'</a> approach to reduce the overhead of
+   * acquiring a lock by testing the lock criteria (the "lock hint") before acquiring the lock.</p>
    * @return the JsonObjectMapper instance reference
    * @see DefaultGsonObjectMapper
    */
--- a/src/main/java/redis/clients/jedis/CommandObjects.java
+++ b/src/main/java/redis/clients/jedis/CommandObjects.java
@@ -5,6 +5,8 @@ import static redis.clients.jedis.Protocol.Keyword.*;
 
 import java.util.*;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import org.json.JSONArray;
@@ -50,6 +52,7 @@ public class CommandObjects {
     return protocol;
   }
 
+  private Lock mapperLock = new ReentrantLock(true);    
   private volatile JsonObjectMapper jsonObjectMapper;
   private final AtomicInteger searchDialect = new AtomicInteger(0);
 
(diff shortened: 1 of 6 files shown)


stevenzyang:
@babanin do you know if the work done in these synchronized blocks is large enough to cause significant impact?

pmverma:
I would love to see this get merged!

sazzad16:
> 1. ... ... ... Do you think it is worthwhile to return the lock here?

Yes.

babanin:
> > 1. ... ... ... Do you think it is worthwhile to return the lock here?
> 
> Yes.

Thanks @sazzad16 - I've updated my PR.

src/main/java/redis/clients/jedis/graph/GraphCommandObjects.java
```diff
@@ -106,9 +108,7 @@ private Builder<ResultSet> getBuilder(String graphName) {
   }
 
   private void createBuilder(String graphName) {
-    synchronized (builders) {
-      builders.putIfAbsent(graphName, new ResultSetBuilder(new GraphCacheImpl(graphName)));
-    }
+      builders.computeIfAbsent(graphName, graphNameKey -> new ResultSetBuilder(new GraphCacheImpl(graphNameKey)));
```
sazzad16, line 111:
```suggestion
    builders.computeIfAbsent(graphName, graphNameKey -> new ResultSetBuilder(new GraphCacheImpl(graphNameKey)));
```

src/main/java/redis/clients/jedis/CommandObjects.java
```diff
@@ -4424,25 +4427,25 @@ public final CommandObject<Object> tFunctionCallAsync(String library, String fun
...
+          mapperLock.lock();
+          
+          try {
+              localRef = this.jsonObjectMapper;
+              if (Objects.isNull(localRef)) {
+                  this.jsonObjectMapper = localRef = new DefaultGsonObjectMapper();
+              }
+          } finally {
+              mapperLock.unlock();
+          }
```
sazzad16, line 4446:
```suggestion
    JsonObjectMapper localRef = this.jsonObjectMapper;
    if (Objects.isNull(localRef)) {
      mapperLock.lock();

      try {
        localRef = this.jsonObjectMapper;
        if (Objects.isNull(localRef)) {
          this.jsonObjectMapper = localRef = new DefaultGsonObjectMapper();
        }
      } finally {
        mapperLock.unlock();
      }
```

src/main/java/redis/clients/jedis/CommandObjects.java
```diff
@@ -4424,25 +4427,25 @@ public final CommandObject<Object> tFunctionCallAsync(String library, String fun
...
       }
-    }
-    return localRef;
+      return localRef;
```
sazzad16, line 4448:
```suggestion
    return localRef;
```

src/main/java/redis/clients/jedis/CommandObjects.java
```diff
@@ -4424,22 +4427,22 @@ public final CommandObject<Object> tFunctionCallAsync(String library, String fun
   /**
    * Get the instance for JsonObjectMapper if not null, otherwise a new instance reference with
    * default implementation will be created and returned.
-   * <p>This process of checking whether or not
-   * the instance reference exists follows <a
-   * href="https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_Java"
-   * target="_blank">'double-checked lock optimization'</a> approach to reduce the overhead of
-   * acquiring a lock by testing the lock criteria (the "lock hint") before acquiring the lock.</p>
+   *
```
sazzad16, line 4430: Double check lock is back.
```suggestion
   * <p>This process of checking whether or not
   * the instance reference exists follows <a
   * href="https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_Java"
   * target="_blank">'double-checked lock optimization'</a> approach to reduce the overhead of
   * acquiring a lock by testing the lock criteria (the "lock hint") before acquiring the lock.</p>
```

sazzad16 (approved):

PR #3495, opened 2023-08-08
Support TOPK.LIST with WITHCOUNT option

### LINKED ISSUE #3488: Feature Request: topkList with counts
### Expected behavior

Hi folks, the TOPK.LIST command in Redis Bloom supports a WITHCOUNT parameter, which will return the counts along with the approximate list of items.

https://redis.io/commands/topk.list/

It would be great if this were supported by Jedis.

Expected API

```java
public interface TopKFilterPipelineCommands {
  ...
  Response<List<String>> topkList(String key);

  Response<List<Tuple<String, Long>>> topkListWithCount(String key);
  ...
}
```

### Actual behavior

There is no such functionality, so I think that I'll need to use eval and a Lua script.

### Steps to reproduce:

N/A

N/A

#### Jedis version:

4.4.0-m1

#### Redis version:

redis_version:6.2.6

#### Java version:

8

sazzad16:
Resolves #3488

### DIFF
--- a/src/main/java/redis/clients/jedis/BuilderFactory.java
+++ b/src/main/java/redis/clients/jedis/BuilderFactory.java
@@ -442,7 +442,7 @@ public final class BuilderFactory {
 
     @Override
     public String toString() {
-      return "Map<String, String>";
+      return "Map<String, Long>";
     }
   };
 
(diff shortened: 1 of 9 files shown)


kchro3:
Awesome, thanks @sazzad16 !

PR #3496, opened 2023-08-09
Updating client license to clear, MIT

chayim:
This pull request exists to unify all of our clients until the MIT license. In the case of some repositories this is more permissive, than the prior BSD-3-Clause.


LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2022-2023, Redis, inc.
```
sazzad16, line 3:
```suggestion
Copyright (c) 2021-2023, Redis, inc.
```
WDYT?

sazzad16 (approved):

PR #3507, opened 2023-08-19
Use RedisCredentials with HELLO command

sazzad16:


src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -391,44 +392,37 @@ private static boolean validateClientInfo(String info) {
...
-
-      boolean doClientName = true;
 
-      /// HELLO and AUTH --> 
-      if (protocol == RedisProtocol.RESP3 && config.getUser() != null) {
+      protocol = config.getRedisProtocol();
 
-        hello(protocol, config.getUser(), config.getPassword(), config.getClientName());
-        doClientName = false;
+      AtomicBoolean clientNamePending = new AtomicBoolean(false);
```
yangbodong22011, line 398: Since we now have fire-and-forgot messages, I don't think it is necessary to pass `SETNAME` in the `Hello` command. We can put it in the fire-and-forgot message (previously put hello to prevent sending more commands, and now fire-and-forgot will not bring additional overhead, because we use pipeline).

In this way, our code will become more clear and concise.

yangbodong22011 (approved):
LGTM

PR #3520, opened 2023-09-02
Helper methods to use a Jedis object

oscar-besga-panel:
A couple of simple methods to operate with a Jedis object with JedisPool and JedisPooled without worrying about closing resources.


src/main/java/redis/clients/jedis/JedisPooled.java
```diff
@@ -401,4 +403,21 @@ public final Pool<Connection> getPool() {
   public Pipeline pipelined() {
     return (Pipeline) super.pipelined();
   }
+
+  public Jedis newJedis() {
+    return new Jedis(getPool().getResource());
```
sazzad16, line 408: This bears the similar issue stated in https://github.com/redis/jedis/pull/3533#discussion_r1330045696

oscar-besga-panel: Ok, I didn't know this was there.
I'll close this PR and make one that limits helpers to JedisPool
I think now that JedisPooled doesn't need helpers like I've done

Thanks

PR #3533, opened 2023-09-12
From pool to pooled

oscar-besga-panel:


src/main/java/redis/clients/jedis/JedisPool.java
```diff
@@ -392,4 +395,39 @@ public void returnResource(final Jedis resource) {
...
+
+  ConnectionProvider asConnectionProvider(){
+    return new JedisPoolConnectionProvider();
+  }
+
+  private class JedisPoolConnectionProvider implements ConnectionProvider {
+
+    @Override
+    public Connection getConnection() {
+      return JedisPool.this.getResource().getConnection();
```
sazzad16, line 419: You're loosing the reference of the Jedis object; so you can't call `close()` on it and so it can't be returned to the pool.
The pool will be exhausted in a multi-threaded environment too quickly.

sazzad16 (commented):
@oscar-besga-panel Please add some description to why we should consider this PR.

src/main/java/redis/clients/jedis/JedisPool.java
oscar-besga-panel: Yep, I checked with a test. 
I'm trying to fix around, but it seems impossible.
I'll try but if I don't find a way, I will close and discard this PR

oscar-besga-panel: Yep, after reviewing the code and trying to make it work in local, I see there's no use (for now)

I wanted to to this because my library uses JedisPool and anyone using JedisPooled in a project could have problems integrating it; so for a future release migth be a good idea. 

But now better close this PR

PR #3536, opened 2023-09-14
Address further CLIENT SETINFO suffix rules

### LINKED ISSUE #3535: Update CLIENT-SETINFO to support suffixes
CLIENT SETINFO should be extended to support community members inserting their own suffixes, for Redis usage.

1. Clients should by default set-lib during connect with their name and version.
2. Clients should allow users to not send this
3. We should allow a user do send a suffix to the LIB-NAME portion - this allows us to learn about how things are used.
4. Our clients continue to send their version upstream for LIB-VER

Strings should have the client name, and an optional suffix. They should be formatted ClientName(<suffix>). For example:

1. If you're NRedisStack - LIB-NAME: NRedisStack(StackExchange.Redis-v2.6.8;.NET-6.2.18) LIB-VERSION: 0.9.1
2. If you're a project using NRedisStack (say SemtanticKernel): LIB-NAME (NRedisStack(SemanticKernel-v0.1.7;.NET-6.2.18) LIB-VERSION: 0.9.1

Considerations:
1. Users should not send brackets, or braces, and we should error accordingly.
2. We should replace spaces a user sends with hyphens

sazzad16:
Completes and resolves #3535


src/main/java/redis/clients/jedis/ClientSetInfoConfig.java
```diff
@@ -1,18 +1,70 @@
...
+  private static final HashSet<Character> BRACES = new HashSet<>(Arrays.asList('(', ')', '[', ']', '{', '}'));
+
+  private static String validateLibNameSuffix(String suffix) {
+    if (suffix == null || suffix.trim().isEmpty()) {
+      return null;
+    }
+
+    for (int i = 0; i < suffix.length(); i++) {
+      char c = suffix.charAt(i);
+      if (c < ' ' || c > '~' || BRACES.contains(c)) {
```
chayim, line 42: IMHO why not use java.util.regex.Matcher and java.util.regex.Pattern; Rather than roll your own via a for loop, regular expressions are pretty performant. Is this faster?

sazzad16: https://github.com/redis/jedis/pull/3458#discussion_r1218063512 
Please check full discussion.

chayim: Awesome - hadn't seen that. Love it.

chayim (approved):

sazzad16:
Released on Jedis versions 4.4.5 and 5.0.1

PR #3537, opened 2023-09-14
Extend CLIENT SETINFO support and address suffix rules

sazzad16:


sazzad16:
Released on Jedis versions 4.4.5 and 5.0.1

PR #3541, opened 2023-09-19
Add active detection in sentinel mode,Reconstruct sentinel mode Listener

chenshi5012:
[sentinel mode， lost "+switch-master" message, should client add watchdog thread #3525](https://github.com/redis/jedis/discussions/3525)
please check this, JedisSentinelPool / SentineledConnectionProvider  provider monitor master-slave failover by subscribe the message from “+switch-master”。

this pr modify  master-slave failover monitor，Two listening mechanisms are provided, one is the default subscription mechanism and the other is the active detection mechanism; 
When due to network reasons, the client loses the subscription message and the main node connection of the jedis factory is still not switched, it will cause the connection to be unavailable.

1. SentinelMasterSubscribeListener the default listener
2. SentinelMasterActiveDetectListener the active detect listener

you can specify which listener to use;  or open both.  default is SubscribeListener, 
`

    SentinelPoolConfig config = new SentinelPoolConfig();
    config.setEnableActiveDetectListener(true);  // default off 
    config.setEnableDefaultSubscribeListener(true); // default on 
    config.setActiveDetectIntervalTimeMillis(5*1000); // default 5s 
    config.setSubscribeRetryWaitTimeMillis(5*1000); // default  5s 

    JedisSentinelPool pool = new JedisSentinelPool(MASTER_NAME, sentinels, config, 1000,
            "foobared", 2);
`

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisSentinelPool.java
+++ b/src/main/java/redis/clients/jedis/JedisSentinelPool.java
@@ -1,11 +1,9 @@
 package redis.clients.jedis;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.Set;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 
 import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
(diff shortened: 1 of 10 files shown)


sazzad16:
@chenshi5012 your tests are failing

PR #3547, opened 2023-09-20
Add active detection in sentinel mode,Reconstruct sentinel mode Listener

chenshi5012:
1. Add active detection in sentinel mode,
2. Reconstruct sentinel mode Listener
3. Add test case for master listener for sentinel

[sentinel mode， lost "+switch-master" message, should client add watchdog thread #3525](https://github.com/redis/jedis/discussions/3525)
please check this, JedisSentinelPool / SentineledConnectionProvider  provider monitor master-slave failover by subscribe the message from “+switch-master”。

this pr modify  master-slave failover monitor，Two listening mechanisms are provided, one is the default subscription mechanism and the other is the active detection mechanism; 
When due to network reasons, the client loses the subscription message and the main node connection of the jedis factory is still not switched, it will cause the connection to be unavailable.

1. SentinelMasterSubscribeListener the default listener
2. SentinelMasterActiveDetectListener the active detect listener

you can specify which listener to use;  or open both.  default is SubscribeListener, 
`

    SentinelPoolConfig config = new SentinelPoolConfig();
    config.setEnableActiveDetectListener(true);   // default off 
    config.setEnableDefaultSubscribeListener(true); // default on 
    config.setActiveDetectIntervalTimeMillis(5*1000); // default 5s 
    config.setSubscribeRetryWaitTimeMillis(5*1000); // default  5s 

    JedisSentinelPool pool = new JedisSentinelPool(MASTER_NAME, sentinels, config, 1000,
            "foobared", 2);
`


chenshi5012:
@sazzad16   thanks for your suggest .  please take a look  for this PR.

chenshi5012:
> @chenshi5012 Have you made any change after your previous PR?
> 
> Moreover, your tests are still `@Ignore`.

ignore has beenremove  and the test case move to JedisSentinelTest.java

src/main/java/redis/clients/jedis/SentinelPoolConfig.java
```diff
@@ -0,0 +1,44 @@
+package redis.clients.jedis;
+
+import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
+
+public class SentinelPoolConfig extends GenericObjectPoolConfig {
```
sazzad16, line 5: GenericObjectPoolConfig takes a type argument. Declaring a sub-class without the type argument is anti-pattern.

Idea: Just declare this interface separately without sub-classing it; with a more appropriate name.

src/main/java/redis/clients/jedis/SentinelMasterActiveDetectListener.java
```diff
@@ -0,0 +1,84 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * active detect master node .in case of the subscribe message lost
+ * @see SentinelMasterSubscribeListener subscribe failover message from "+switch-master" channel
+ */
+public abstract class SentinelMasterActiveDetectListener extends Thread
```
sazzad16, line 14: This is listening to only Sentinel nodes. No need to add `Master` in name.

src/main/java/redis/clients/jedis/SentinelMasterSubscribeListener.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * subscribe failover message from "+switch-master" channel , the default listener mode use this
+ * @see SentinelMasterActiveDetectListener active detect master node .in case of the subscribe
+ *      message lost
+ */
+public abstract class SentinelMasterSubscribeListener extends Thread
```
sazzad16, line 16: This is listening to only Sentinel nodes. No need to add `Master` in name.

src/main/java/redis/clients/jedis/SentinelMasterListener.java
```diff
@@ -0,0 +1,14 @@
+package redis.clients.jedis;
+
+/**
+ * interface for monitor the master failover under sentinel mode We offer two implementation options
+ * @see SentinelMasterSubscribeListener Passive subscription
+ * @see SentinelMasterActiveDetectListener Active detection
+ */
+public interface SentinelMasterListener {
```
sazzad16, line 8: Check comment in implementing classes.

src/test/java/redis/clients/jedis/JedisSentinelTest.java
```diff
@@ -87,13 +93,26 @@ public void sentinelFailover() throws InterruptedException {
...
           Integer.parseInt(masterHostAndPort.get(1)));
 
+      // get master node from pool connection
+      SentinelPoolConfig config = new SentinelPoolConfig();
+      config.setEnableActiveDetectListener(true);
+      config.setEnableDefaultSubscribeListener(false);
+      JedisSentinelPool pool = new JedisSentinelPool(FAILOVER_MASTER_NAME, sentinels, config, 1000,
+        "foobared", 2);
+      HostAndPort masterGetFromPoolBefore = pool.getResource().connection.getHostAndPort();
+
```
sazzad16, line 103: Keep the existing test and add a new test?

sazzad16 (changes requested):
Many new classes only for sentinel. May consider a new package.

sazzad16:
@chenshi5012 This is a _not so simple_ PR. I already have too many things in my regular job. So, would take time to properly review this. However, added some quick comments.

src/main/java/redis/clients/jedis/SentinelMasterListener.java
chenshi5012: fix

src/main/java/redis/clients/jedis/SentinelPoolConfig.java
chenshi5012: may be  sub-class is the best way to avoid add more Constructor  method .  currently threse are 24  Constructor  method for JedisSentinelPool. 
use sub-class will not break the Constructor  parameter type GenericObjectPoolConfig.

chenshi5012:
> Many new classes only for sentinel. May consider a new package.

thanks for  your pantinet ,  new package "redis.clients.jedis.sentinel.listener"

src/main/java/redis/clients/jedis/SentinelPoolConfig.java
sazzad16: If that's your concern, you can introduce a Builder.

yangbodong22011:
As far as I know, [SentinelListener#run](https://github.com/redis/jedis/blob/master/src/main/java/redis/clients/jedis/providers/SentineledConnectionProvider.java#L201) will always run and possibly reinitialize sentinelMaster:

1. Receive "+switch-master"
2. After receiving JedisException, it will be reinitialized after sleep subscribeRetryWaitTimeMillis.

If your client does not recover in either case, what could be the reason?

chenshi5012:
> As far as I know, [SentinelListener#run](https://github.com/redis/jedis/blob/master/src/main/java/redis/clients/jedis/providers/SentineledConnectionProvider.java#L201) will always run and possibly reinitialize sentinelMaster:
> 
> 1. Receive "+switch-master"
> 2. After receiving JedisException, it will be reinitialized after sleep subscribeRetryWaitTimeMillis.
> 
> If your client does not recover in either case, what could be the reason?

actually ，active detect will  enhanced stability.   while client lost  subscribe message 

image

yangbodong22011:
@chenshi5012 Yes, I know Pubsub is unreliable, and I'd like to know what specific issues you're experiencing in production?

pubsub messages lost? Or is the client always in a silent state and unable to sense that the link is disconnected?

chenshi5012:
Yes, in Sentinel mode, we've had this problem many times in the production environment.
1. The data load is heavy. (80G+,  flush every 30mins )
2.  the network not stable  (client in other vpc.   ping-latancy Occasionally very high )
4. The node breaks down unexpectedly (the server and sentinel are together in one node ), triggering the active/standby switchover.

PR #3567, opened 2023-10-05
Add possibility to create text/tag fields with option WITHSUFFIXTRIE

AntonGabov:
This change allows to create Text and Tag fields in schema with option WITHSUFFIXTRIE.

The change will allow to extend annotations `@Indexed` and `@Searchable` with new option in the project: https://github.com/redis/redis-om-spring

### DIFF
--- a/src/main/java/redis/clients/jedis/search/Schema.java
+++ b/src/main/java/redis/clients/jedis/search/Schema.java
@@ -226,6 +226,7 @@ public class Schema {
     private final double weight;
     private final boolean nostem;
     private final String phonetic;
+    private final boolean withSuffixTrie;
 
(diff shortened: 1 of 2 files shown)


sazzad16:
WITHSUFFIXTRIE is available in [TextField](https://github.com/redis/jedis/blob/master/src/main/java/redis/clients/jedis/search/schemafields/TextField.java) and [TagField](https://github.com/redis/jedis/blob/master/src/main/java/redis/clients/jedis/search/schemafields/TagField.java) which can be used with [FTCreateParams](https://github.com/redis/jedis/blob/master/src/main/java/redis/clients/jedis/search/FTCreateParams.java)

PR #3571, opened 2023-10-09
Fix binary variants of XRANGE and XREAD commands

sazzad16:
where each individual element in the list is actually a complex object rather than a simple byte array.


yangbodong22011 (approved):
LGTM

PR #3589, opened 2023-10-18
Move pom.properties file into redis.clients.jedis package

### LINKED ISSUE #3556: Namespace `pom.properties`
### Expected behavior
Jedis JAR contents should be namespaced to avoid conflicts when compiled together with other JARs in an application.

I think the more conventional namespace would be `META-INF/maven/redis.clients/jedis/pom.properties`

### Actual behavior
Jedis JAR contains pom.properties at the root which could conflict with other libraries at compile time.

### Steps to reproduce:
1. Unzip the Jedis 5.0 JAR 
2. See pom.properties in the root

5.0

sazzad16:
Resolves #3556


yangbodong22011:
@sazzad16 Thanks for your fix, I packaged it and verified it, and it is as expected.
image

yangbodong22011 (approved):

PR #3594, opened 2023-10-23
Pipeline with failover to multi cluster

### LINKED ISSUE #3568: Add support for automatic failover in Pipeline
### Expected behavior

The `Pipeline` class supports automatic failover.

### Actual behavior

The `Pipeline` class uses the `Connection` class directly instead of utilizing the Command executor with Circuit breaker support.

### Steps to reproduce:

1. Follow the https://github.com/redis/jedis/blob/master/docs/failover.md instructions
2. Use `pipelined()` to send commands as a command
3. Simulate the Redis connection problem by shutting down the first cluster.

https://github.com/redis/jedis/blob/master/docs/failover.md

#### Jedis version:
5.0

#### Redis version:
Any

#### Java version:
Any

sazzad16:
Resolves #3568


src/main/java/redis/clients/jedis/UnifiedJedis.java
```diff
@@ -4822,8 +4822,11 @@ public Object tFunctionCallAsync(String library, String function, List<String> k
...
       throw new IllegalStateException("It is not allowed to create Pipeline from this " + getClass());
+    } else if (provider instanceof MultiClusterPooledConnectionProvider) {
+      return new MultiClusterFailoverPipeline((MultiClusterPooledConnectionProvider) provider);
+    } else {
+      return new Pipeline(provider.getConnection(), true);
     }
-    return new Pipeline(provider.getConnection(), true);
   }
 
   public Transaction multi() {
```
banker, line 4832: @sazzad16 See my comment in https://github.com/redis/jedis/issues/3568#issuecomment-1750799211.

This is also a problems for multi(), subscribe(), and any methods that use JedisCommandIterationBase. Shouldn't we fix these as well?

banker:
We need to ensure that we've tested these changes. Would you like assistance with that @sazzad16 ?

sazzad16:
> We need to ensure that we've tested these changes.

@banker I haven't found tests from our original failover implementation to imitate here for pipeline implementation. The only closest thing I can see is https://github.com/redis/jedis/blob/d2f6712cb633a5cba84ef0c555be7caeec7a8abb/src/test/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProviderTest.java#L102-L114

src/main/java/redis/clients/jedis/UnifiedJedis.java
chayim: Current plan: Those will be in a different issue, and estimated there.

chayim (approved):

src/main/java/redis/clients/jedis/UnifiedJedis.java
chayim: #3602 contains multi and supersedes this subscribe

PR #3596, opened 2023-10-25
Introducing periodic topology mechanism for JedisCluster

yangbodong22011:
solves #3595

EDIT: The main changes in this PR are as follows:

1. Add `topologyRefreshEnabled` and `topologyRefreshPeriod` to control the periodic topology refresh mechanism.
2. `topologyRefreshExecutor` is a single-threaded executor responsible for running `TopologyRefreshTask`.
3. Add `checkClusterSlotSequence` to check whether the cluster slots returned by the server are consecutive and there are no duplicates.
4. [Test] In JedisClusterTestBase, `CLUSTER RESET SOFT` was changed to `HARD`, because SOFT cannot clean up the Redis Cluster configuration and may cause a crash. Please refer to https://github.com/redis/redis/issues/12701
5. [Test] Adjust the cluster-node-timeout in the Makefile to 15s, consistent with the default configuration of Redis.

PR #3597, opened 2023-10-26
Fix connect socket InetSocketAddress

yangbodong22011:
Passing `host` directly will avoid another call to `InetAddress.getByName` inside the `InetSocketAddress` function.

For machines with ipv4 and ipv6, but the startNode uses ipv4 to connect, the ipv6 connection may fail.

### DIFF
--- a/src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
+++ b/src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
@@ -70,7 +70,7 @@ public class DefaultJedisSocketFactory implements JedisSocketFactory {
         socket.setTcpNoDelay(true); // Socket buffer Whetherclosed, to ensure timely delivery of data
         socket.setSoLinger(true, 0); // Control calls close () method, the underlying socket is closed immediately
 
-        socket.connect(new InetSocketAddress(host.getHostAddress(), hostAndPort.getPort()), connectionTimeout);
+        socket.connect(new InetSocketAddress(host, hostAndPort.getPort()), connectionTimeout);
         return socket;
       } catch (Exception e) {
         jce.addSuppressed(e);


sazzad16:
@yangbodong22011 The code makes sense.

But to make sure your reasoning, do you understand that due to 
```java
    List<InetAddress> hosts = Arrays.asList(InetAddress.getAllByName(hostAndPort.getHost()));
    if (hosts.size() > 1) {
      Collections.shuffle(hosts);
    }
```
you can't ensure ipv4 comes before ipv6?

yangbodong22011:
> @yangbodong22011 The code makes sense.
> 
> But to make sure your reasoning, do you understand that due to
> 
> ```java
>     List<InetAddress> hosts = Arrays.asList(InetAddress.getAllByName(hostAndPort.getHost()));
>     if (hosts.size() > 1) {
>       Collections.shuffle(hosts);
>     }
> ```
> 
> you can't ensure ipv4 comes before ipv6?

Yes, I noticed the shuffle.

My scenario is that startNode is localhost, and two `InetAddress` are obtained through `InetAddress.getAllByName`, one is ipv4 and the other is ipv6 (my machine has two IPs at the same time).

During the subsequent connection process, if it is an ipv6 address, after socket.connect, an error "Unexpected end of stream." will be reported when performing read.

Because of the old code, the ipv6 address will still perform `getAllByName` in `new InetSocketAddress`. At this time, the host obtained is Null, which may cause read to return eof.

image
(The left is the new code, the right is the old code)

yangbodong22011:
@sazzad16 I'd prefer to merge this PR and let https://github.com/redis/jedis/pull/3596 rebase it.

This will provide more information for people looking at the code in the future, at least https://github.com/redis/jedis/pull/3597 this link.

sazzad16 (approved):

sazzad16:
> I'd prefer to merge this PR and let #3596 rebase it.
> 
> This will provide more information for people looking at the code in the future, at least #3597 this link.

@yangbodong22011 I'm not sure. This change was not necessary until #3596.
Moreover if we separate the changes, #3596 won't be self sufficient. I'd assume we'd have a 4.5.0 release including that PR. In that case there would be an added liability to downmerge this PR in `4.x` branch while downmerging that #3596.
IMO, adding code comment there is lesser hectic way.

sazzad16:
Resolved in #3596

PR #3606, opened 2023-11-06
Execute Redis commands using Netty

sazzad16:


src/main/java/redis/clients/jedis/netty/Connection.java
```diff
@@ -0,0 +1,80 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(Connection.class);
+
+//    private final String host;
+//    private final int port;
+
+    private final EventLoopGroup group;
+    private final Channel channel;
+
+    public Connection(String host, int port) throws InterruptedException {
+//        this.host = host;
```
chayim, line 33: clean

src/main/java/redis/clients/jedis/netty/Connection.java
```diff
@@ -0,0 +1,80 @@
...
+import java.io.Closeable;
+import java.io.IOException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class Connection implements Closeable {
+
+    private static final Logger logger = LoggerFactory.getLogger(Connection.class);
+
+//    private final String host;
```
chayim, line 26: clean

src/main/java/redis/clients/jedis/netty/Connection.java
```diff
@@ -0,0 +1,80 @@
+package redis.clients.jedis.netty;
```
chayim, line 1: I think we should not name this under .netty, but as .async or asyncio. The library implementation isn't important in the namespace,  IMHO - and we should do everything we can, to ensure it isn't.

chayim (approved):

PR #3658, opened 2023-12-19
Initial support for client-side caching

sazzad16:


src/main/java/redis/clients/jedis/JedisClientSideCache.java
```diff
@@ -0,0 +1,45 @@
...
+  private final ClientSideCache cache;
+
+  public JedisClientSideCache(final HostAndPort hostPort, final JedisClientConfig config) {
+    this(hostPort, config, new ClientSideCache());
+  }
+
+  public JedisClientSideCache(final HostAndPort hostPort, final JedisClientConfig config,
+      ClientSideCache cache) {
+    super(hostPort, config);
+    if (config.getRedisProtocol() != RedisProtocol.RESP3) {
```
chayim, line 16: @vladvildanov @dvora-h IMHO we should do this in other clients if not already there.

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -133,7 +137,7 @@ private static String[] parseTargetHostAndSlot(String clusterRedirectResponse) {
 
   private static Object process(final RedisInputStream is) {
     final byte b = is.readByte();
-    //System.out.println((char) b);
+    //System.out.println("BYTE: " + (char) b);
```
chayim, line 140: remove?

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -167,6 +171,15 @@ private static Object process(final RedisInputStream is) {
     }
   }
 
+  private static void processPush(final RedisInputStream is, ClientSideCache cache) {
+    List<Object> list = processMultiBulkReply(is);
+    //System.out.println("PUSH: " + SafeEncoder.encodeObject(list));
```
chayim, line 176: x2

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -193,11 +206,13 @@ private static byte[] processBulkReply(final RedisInputStream is) {
   private static List<Object> processMultiBulkReply(final RedisInputStream is) {
   // private static List<Object> processMultiBulkReply(final int num, final RedisInputStream is) {
     final int num = is.readIntCrLf();
+    //System.out.println("MULTI BULK: " + num);
```
chayim, line 209: etc

src/main/java/redis/clients/jedis/util/RedisInputStream.java
```diff
@@ -43,6 +43,11 @@ public RedisInputStream(InputStream in) {
     this(in, INPUT_BUFFER_SIZE);
   }
 
+  public Byte peekByte() {
```
chayim, line 46: Can we start adding javadocs here

src/main/java/redis/clients/jedis/util/RedisInputStream.java
```diff
@@ -252,4 +257,18 @@ private void ensureFill() throws JedisConnectionException {
       }
     }
   }
+
+  private void ensureFillSafe() {
```
chayim, line 261: and here

src/main/java/redis/clients/jedis/util/RedisInputStream.java
```diff
@@ -252,4 +257,18 @@ private void ensureFill() throws JedisConnectionException {
...
+    if (count >= limit) {
+      try {
+        limit = in.read(buf);
+        count = 0;
```
chayim, line 265: count doesn't need to be within the try, can be outside

src/main/java/redis/clients/jedis/util/RedisInputStream.java
```diff
@@ -252,4 +257,18 @@ private void ensureFill() throws JedisConnectionException {
       }
     }
   }
+
+  private void ensureFillSafe() {
+    if (count >= limit) {
```
chayim, line 262: possibly silly but `if count >= limit && limit > 0`?

src/test/java/redis/clients/jedis/JedisClientSideCacheTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+public class JedisClientSideCacheTest {
+
+  protected static final HostAndPort hnp = HostAndPorts.getRedisServers().get(1);
+
+  protected Jedis jedis;
+
+  @Before
+  public void setUp() throws Exception {
+    jedis = new Jedis(hnp, DefaultJedisClientConfig.builder().timeoutMillis(500).password("foobared").build());
```
chayim, line 20: glad to see this test

src/test/java/redis/clients/jedis/JedisClientSideCacheTest.java
```diff
@@ -0,0 +1,70 @@
...
+    try (JedisClientSideCache jCache = new JedisClientSideCache(hnp, configForCache, cache)) {
+      jedis.set("foo", "bar");
+      assertEquals("bar", jCache.get("foo"));
+      jedis.del("foo");
+      assertNull(jCache.get("foo"));
+    }
+
+    InOrder inOrder = Mockito.inOrder(cache);
+    inOrder.verify(cache).invalidateKeys(Mockito.notNull());
+    inOrder.verify(cache).getValue("foo");
```
chayim, line 54: nit: store the key name in a variable and reuse

src/main/java/redis/clients/jedis/ClientSideCache.java
```diff
@@ -0,0 +1,58 @@
...
+import redis.clients.jedis.util.SafeEncoder;
+
+public class ClientSideCache {
+
+  private final Map<ByteBuffer, Object> cache = new HashMap<>();
+
+  protected ClientSideCache() {
+  }
+
+  protected void invalidateKeys(List list) {
```
vladvildanov, line 18: Design discussion:

As I understood `ClientSideCache` object represents a DAO/persistence layer, so it's actually just a storage and you can perform CRUD operations against it. From this perspective this method should be rather named as `delete()` or `batchDelete()`.

Invalidation is a more complex process built around cache that involves business logic and belong to another layer, something like this:
1. Get invalid key (keys).
2. Find all command responses associated with this key (keys).
3. Delete them from cache.

src/main/java/redis/clients/jedis/ClientSideCache.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.nio.ByteBuffer;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import redis.clients.jedis.exceptions.JedisException;
+import redis.clients.jedis.util.SafeEncoder;
+
+public class ClientSideCache {
```
vladvildanov, line 11: Do you really need to implement your own cache storage? Don't you have something ready to use, maybe some lightweight library that provides a cache storage API?

I'm pretty much sure it will be quite time consuming to implement search mechanism in cache to search for all the data associated with key, as well as provide metadata for telemetry (cache hits, cache misses). Also, you have to think about TTL mechanism.

src/main/java/redis/clients/jedis/JedisClientSideCache.java
vladvildanov: @chayim It's already there

But, we had a discussion with you and @uglide and we agreed that we don't throw an exception in case if user do not explicitly set RESP3, but instead we're forcing RESP3 connection.

src/main/java/redis/clients/jedis/JedisClientSideCache.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    this.cache = cache;
+    this.connection.setClientSideCache(cache);
+    clientTrackingOn();
+  }
+
+  private void clientTrackingOn() {
+    String reply = connection.executeCommand(new CommandObject<>(
+        new CommandArguments(Protocol.Command.CLIENT).add("TRACKING").add("ON").add("BCAST"),
```
vladvildanov, line 27: @chayim @uglide As I remember BCAST is not an option here

src/main/java/redis/clients/jedis/JedisClientSideCache.java
```diff
@@ -0,0 +1,45 @@
+package redis.clients.jedis;
+
+import redis.clients.jedis.exceptions.JedisException;
+
+public class JedisClientSideCache extends Jedis {
+
+  private final ClientSideCache cache;
+
+  public JedisClientSideCache(final HostAndPort hostPort, final JedisClientConfig config) {
```
vladvildanov, line 9: How about to use Proxy pattern here? Instead of introducing new connection type you can introduce a proxy object that decides whether you need to read against cache or delegate it to another "real" connection against Redis.

Aggregation is always preferable against inheritance :)

src/main/java/redis/clients/jedis/Protocol.java
sazzad16: as we are still in development, these things help me to debug

src/main/java/redis/clients/jedis/Protocol.java
sazzad16: x2

src/main/java/redis/clients/jedis/Protocol.java
sazzad16: likewise

src/test/java/redis/clients/jedis/JedisClientSideCacheTest.java
sazzad16: this is our regular Jedis object, to verify and cleanup stuff

src/main/java/redis/clients/jedis/ClientSideCache.java
sazzad16: We should define our cache APIs. I don't think we'll find anything that perfectly matches all our use cases. With a defined set of APIs, users may choose to provide their own implementation based on their preferred library.

At some point, this will be based on a library. But we have started simple.

src/main/java/redis/clients/jedis/ClientSideCache.java
sazzad16: We have used Redis terms for easier recognition. Things could be changed later but not now.

src/main/java/redis/clients/jedis/JedisClientSideCache.java
sazzad16: I've found it easier to **start**.

src/main/java/redis/clients/jedis/ClientSideCache.java
chayim: Internal separation of concern is a good idea

src/main/java/redis/clients/jedis/JedisClientSideCache.java
chayim: Correct. @sazzad16

chayim (dismissed):
Alpha ready

PR #3673, opened 2024-01-01
Support for client-side caching - phase 2

sazzad16:


src/main/java/redis/clients/jedis/JedisClientSideCache.java
```diff
@@ -24,7 +24,7 @@ public JedisClientSideCache(final HostAndPort hostPort, final JedisClientConfig
 
   private void clientTrackingOn() {
     String reply = connection.executeCommand(new CommandObject<>(
-        new CommandArguments(Protocol.Command.CLIENT).add("TRACKING").add("ON").add("BCAST"),
+        new CommandArguments(Protocol.Command.CLIENT).add("TRACKING").add("ON"),
```
chayim, line 27: correct

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -232,20 +222,35 @@ private static List<KeyValue> processMapKeyValueReply(final RedisInputStream is)
     return ret;
   }
 
+  @Deprecated
```
chayim, line 225: Fair. This has been around since 2014

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -232,20 +222,35 @@ private static List<KeyValue> processMapKeyValueReply(final RedisInputStream is)
...
         processPush(is, cache);
       }
     }
   }
 
+  private static void processPush(final RedisInputStream is, ClientSideCache cache) {
+    List<Object> list = processMultiBulkReply(is);
+    //System.out.println("PUSH: " + SafeEncoder.encodeObject(list));
+    if (list.size() == 2 && list.get(0) instanceof byte[]
+        && Arrays.equals(INVALIDATE_BYTES, (byte[]) list.get(0))) {
```
chayim, line 249: Glad this is predefined

src/main/java/redis/clients/jedis/util/RedisInputStream.java
```diff
@@ -43,9 +43,9 @@ public RedisInputStream(InputStream in) {
     this(in, INPUT_BUFFER_SIZE);
   }
 
-  public Byte peekByte() {
-    ensureFillSafe();
-    return buf[count];
+  public boolean peek(byte b) throws JedisConnectionException {
+    ensureFill(); // in current design, at least one reply is expected. so ensureFillSafe() is not necessary.
```
chayim, line 47: But, what happens if there is no reply? It shouldn't occur, but maybe there's something we can do for resiliency here. At the very least raise a data specific exception so that we can know this happened?

src/test/java/redis/clients/jedis/JedisClientSideCacheTest.java
```diff
@@ -26,45 +28,83 @@ public void tearDown() throws Exception {
...
 
   @Test
   public void simple() {
-    try (JedisClientSideCache jCache = new JedisClientSideCache(hnp, configForCache)) {
+    try (JedisClientSideCache jCache = new JedisClientSideCache(hnp, clientConfig)) {
       jedis.set("foo", "bar");
       assertEquals("bar", jCache.get("foo"));
       jedis.del("foo");
-      assertNull(jCache.get("foo"));
+      assertThat(jCache.get("foo"), Matchers.oneOf("bar", null)); // ?
```
chayim, line 39: Can this be achieved without adding this import?

src/main/java/redis/clients/jedis/util/RedisInputStream.java
sazzad16: @chayim A JedisConnectionException wrapping a ReadTimeoutException will be thrown.

src/test/java/redis/clients/jedis/JedisClientSideCacheTest.java
sazzad16: @chayim We'd have to write a "utility" method.
But why bother? We already have these as our test dependency and being used in some other tests.

chayim: General desire to have fewer dependencies.. even if it's just for test. Not a hard requirement.

chayim (approved):

Issue #3624, opened 2023-11-27
Create Java code snippet for sorted sets


sav-norem:
The redis.io webpage for [sorted sets](https://redis.io/docs/data-types/sorted-sets/) needs the Java example as well. The python equivalent can be used to copy the setup and can be found [here](https://github.com/redis/redis-py/pull/3002/files#diff-196240727c71b03fd8fc6bda16e52eaf85b85838c246d8874d4bd10ea05ce184)

The first line of the file needs to be the ID for the example - # EXAMPLE: ss_tutorial

Each separate section is delimited with a tag that you will need to copy exactly for it to appear in the correct place, ie.
```
# STEP_START zadd 
{code}                                    
# STEP_END

# HIDE_START, # HIDE_END, # REMOVE_START, and # REMOVE_END can be used for debugging and deleting keys. Hide will hide until a user expands the example, remove will only be seen in the source code and will never appear on the webpage.
```

@dwdougherty and I will be reviewing and tracking these issues - feel free to tag us with any questions

sazzad16:
Resolved by #3680

PR #3680, opened 2024-01-09
Create Java code snippet for sorted sets 

Harsh4902:
- added code snippet for sorted sets in Java
- Fixes: issue #3624

### DIFF
--- a/src/test/java/io/redis/examples/SortedSetsExample.java
+++ b/src/test/java/io/redis/examples/SortedSetsExample.java
@@ -1,4 +1,4 @@
-//ExAMPLE: ss_tutorial
+//EXAMPLE: ss_tutorial
 //HIDE_START
 package io.redis.examples;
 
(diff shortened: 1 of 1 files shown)


src/test/java/io/redis/examples/SortedSetsExample.java
```diff
@@ -0,0 +1,111 @@
+//ExAMPLE: ss_tutorial
```
dwdougherty, line 1: Turns out, case matters. Please commit this change and then it's good to go.

```suggestion
//EXAMPLE: ss_tutorial
```

sazzad16 (changes requested):
1. For single entry, Use `zadd(String key, double score, String member)` method.
2. Format HashMap. Hint: https://github.com/redis/jedis/blob/fbed1665c9941a6009b14b8d8ef46ee7e41ef769/src/test/java/io/redis/examples/CMSExample.java#L35-L38 
3. Format spaces.

dwdougherty (approved):
@sazzad16 This PR builds correctly and it's good to go.

PR #3697, opened 2024-01-26
Avoid NPE in MultiNodePipelineBase.java

stillerrr:
should get connection first and then create new pipeline queue, otherwise it would cause NPE when timeout for getting connection and call sync() method

if there is timeout for getting connection, the pipeline queue would be create for node, but connections map has no connection for this node

once executing sync() method, it would throw NPE on connection.getMany() (line 112, connection is null)


sazzad16 (approved):

sazzad16:
@stillerrr Thank you for your contribution!

PR #3698, opened 2024-01-28
Client-side caching by command arguments

sazzad16:
TODO: Compute hash code of command arguments.

---

Currently superseded by #3700

### DIFF
--- a/src/main/java/redis/clients/jedis/ClientSideCache.java
+++ b/src/main/java/redis/clients/jedis/ClientSideCache.java
@@ -35,7 +35,7 @@ public abstract class ClientSideCache {
 
   protected abstract Object get(long hash);
 
-  final void invalidateKeys(List list) {
+  final void invalidate(List list) {
     if (list == null) {
       invalidateAll();
       return;
--- a/src/main/java/redis/clients/jedis/ClientSideCache.java
+++ b/src/main/java/redis/clients/jedis/ClientSideCache.java
@@ -14,7 +14,7 @@ import redis.clients.jedis.util.SafeEncoder;
 
 public abstract class ClientSideCache {
 
-  private final Map<ByteBuffer, Set<Long>> keyHashes;
+  private final Map<ByteBuffer, Set<Long>> keyHashes; // TODO: clean-up
   private final ReentrantReadWriteLock rwl = new ReentrantReadWriteLock();
   private final Lock readLock = rwl.readLock();
   private final Lock writeLock = rwl.writeLock();
@@ -44,6 +44,7 @@ public abstract class ClientSideCache {
     Set<Long> hashes = new HashSet<>();
     list.forEach(key -> hashes.addAll(getHashes(key)));
     invalidateAll(hashes);
+    // TODO: clean-up keyHashes
   }
 
   private Set<Long> getHashes(Object key) {
--- a/src/main/java/redis/clients/jedis/ClientSideCache.java
+++ b/src/main/java/redis/clients/jedis/ClientSideCache.java
@@ -1,55 +1,108 @@
 package redis.clients.jedis;
 
 import java.nio.ByteBuffer;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
-import redis.clients.jedis.exceptions.JedisException;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.function.Function;
 import redis.clients.jedis.util.SafeEncoder;
 
 public abstract class ClientSideCache {
 
-  protected ClientSideCache() { }
+  private final Map<ByteBuffer, Set<Long>> keyHashes;
+  private final ReentrantReadWriteLock rwl = new ReentrantReadWriteLock();
+  private final Lock readLock = rwl.readLock();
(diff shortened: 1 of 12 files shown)


src/main/java/redis/clients/jedis/ClientSideCache.java
```diff
@@ -1,71 +1,108 @@
...
+          } else {
+            Set<Long> set = new HashSet<>();
+            set.add(hash);
+            keyHashes.put(mapKey, set);
+          }
+        }
+      } finally {
+        writeLock.unlock();
+      }
+    }
```
gerzse, line 91: What if there are cached values for this key? Should we remove them, given that the key was removed on the server, most probably? Should there be an `else` for the `if (value != null)`?

src/main/java/redis/clients/jedis/ClientSideCache.java
```diff
@@ -1,71 +1,108 @@
...
+    } finally {
+      readLock.unlock();
+    }
   }
 
-  protected <T> T getValue(Object key) {
-    return (T) getMapValue(key);
+  final <T> T getValue(Function<CommandObject<T>, T> loader, CommandObject<T> command, String... keys) {
+
+    final long hash = getHash(command);
```
gerzse, line 66: Just to confirm that I understood right, we are doing this trick with command hashes in order to cover situations when different commands get differently formatted responses for the same Redis key, right? So we want to store all the different types of responses, hence we can't use a global cache based only on the Redis key.

src/main/java/redis/clients/jedis/ClientSideCache.java
```diff
@@ -1,71 +1,108 @@
...
-  private final Map<ByteBuffer, Object> cache;
+  private final Map<ByteBuffer, Set<Long>> keyHashes;
+  private final ReentrantReadWriteLock rwl = new ReentrantReadWriteLock();
+  private final Lock readLock = rwl.readLock();
+  private final Lock writeLock = rwl.writeLock();
 
-  public ClientSideCache() {
-    this.cache = new HashMap<>();
+  protected ClientSideCache() {
+    this.keyHashes = new ConcurrentHashMap<>();
```
gerzse, line 23: It seems to me that we are only adding stuff to this map. Some more housekeeping will be needed, otherwise the map will only grow, we need to remove unused keys and stuff, at some point. Could this be in itself a `Cache`, either Caffeine, or Guava?

src/main/java/redis/clients/jedis/util/CaffeineCSC.java
```diff
@@ -0,0 +1,50 @@
+package redis.clients.jedis.util;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import java.util.concurrent.TimeUnit;
+import redis.clients.jedis.ClientSideCache;
+
+public class CaffeineCSC extends ClientSideCache {
```
gerzse, line 8: Just a thought that comes to my mind when I read this: the general advice is to favor composition over inheritance. Maybe we could model this via composition, put an interface on top of `CaffeineCSC` and `GuavaCSC` and inject that interface into `ClientSideCache`?

gerzse (commented):
LGTM, I guess some details will need attention

src/main/java/redis/clients/jedis/util/CaffeineCSC.java
sazzad16: ClientSideCache itself is that interface (abstract class). It is an optional parameter for our actual client classes.

src/main/java/redis/clients/jedis/ClientSideCache.java
```diff
@@ -1,71 +1,108 @@
...
+  public abstract void invalidateAll();
+
+  protected abstract void invalidateAll(Iterable<Long> hashes);
+
+  protected abstract void put(long hash, Object value);
 
-  public final void invalidateKeys(List list) {
+  protected abstract Object get(long hash);
+
+  final void invalidateKeys(List list) {
```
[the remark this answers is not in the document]
sazzad16: Its only usage is in https://github.com/redis/jedis/blob/3827e7d7dc98480d28f0e1b67559d288a94bf31a/src/main/java/redis/clients/jedis/Protocol.java#L251 where it is programmatically possible to ensure generics. The other things we can consider are `Object` or `List<Object>`. So I have been a bit less thoughtful here.

src/main/java/redis/clients/jedis/ClientSideCache.java
sazzad16: It's literally taken from the job description. Slacked you the link.

src/main/java/redis/clients/jedis/ClientSideCache.java
sazzad16: Do you mean, what if there are cached values for this "command"? In that case we have already returned the result in Line#70.

src/main/java/redis/clients/jedis/ClientSideCache.java
sazzad16: Adding TODO comments for now.

src/main/java/redis/clients/jedis/ClientSideCache.java
sazzad16: It seems redundant as it should be covered by `void invalidate(List list)` (previously `void invalidateKeys(List list)`).

PR #3700, opened 2024-01-30
Client-side caching by hashing command arguments

sazzad16:


src/main/java/redis/clients/jedis/ClientSideCache.java
```diff
@@ -1,71 +1,99 @@
...
-  }
+  public abstract void invalidateAll();
 
-  public final void clear() {
-    cache.clear();
-  }
+  protected abstract void invalidateAll(Iterable<Long> hashes);
 
-  public final void invalidateKeys(List list) {
+  final void invalidate(List list) {
```
gerzse, line 26: There is a lot of invalidation going on in this class. I preferred the original name for this method, i.e. `invalidateKeys`, because it points out what is being invalidated. Along the same lines, the method `invalidateAll(Iterable<Long>)` could be named `invalidateHashes`. I'm not sure about `invalidateAll()`, maybe it could be `invalidateAllHashes()`? `invalidate0` could stay `invalidateKey`.

You know the joke with the two difficult problems in computer science: cache invalidation and naming/

src/main/java/redis/clients/jedis/ClientSideCache.java
```diff
@@ -1,71 +1,99 @@
...
+        }
+      } finally {
+        writeLock.unlock();
+      }
     }
+
+    return value;
+  }
+
+  protected abstract long getHash(CommandObject command);
```
gerzse, line 90: Why have different ways of hashing a command? Also, why bind the hashing algorithm to the cache library that we use? Why not allow using Caffeine as a cache store with Guava's hashing routine? That would mean another interface for hash computation.

gerzse:
Add unit tests for `GuavaCSC` and `CaffeineCSC`?

src/main/java/redis/clients/jedis/ClientSideCache.java
sazzad16: @gerzse This `invalidate(List)` is called only when the server pushes with the header `invalidate`. So it seemed fit.

As for `invalidateAll()` and `invalidateAll(Iterable)`, the names are aligned with the respective methods in both Guava and Caffeine libraries.

src/main/java/redis/clients/jedis/ClientSideCache.java
sazzad16: > Why not allow using Caffeine as a cache store with Guava's hashing routine?

@gerzse This is exactly what I'm trying to achieve.

Guava has a nice hash solution where we can do `newHasher().put(a).put(b).put(c).hash()`.
But most other library only allows `function.hash(a)`. So we have to do `function.hash(array[function.hash(a), function.hash(b), function.hash(c)])`.
We can add an interface for later approach. But it makes the append style approach of Guava unusable (hashing is still possible in second way).
On the other hand, if we create an interface to support the first approach, it makes most hashing libraries unusable.
Now, what is best approach we can do?

src/main/java/redis/clients/jedis/ClientSideCache.java
```diff
@@ -1,71 +1,99 @@
...
-
-import redis.clients.jedis.exceptions.JedisException;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Function;
 import redis.clients.jedis.util.SafeEncoder;
 
-public class ClientSideCache {
+public abstract class ClientSideCache {
```
chayim, line 13: I'd really like us to javadoc these so that others can bring their own implementations

src/main/java/redis/clients/jedis/ClientSideCache.java
```diff
@@ -1,71 +1,99 @@
...
-    list.forEach(this::invalidateKey);
+    list.forEach(this::invalidate0);
   }
 
-  private void invalidateKey(Object key) {
-    if (key instanceof byte[]) {
-      cache.remove(convertKey((byte[]) key));
-    } else {
-      throw new JedisException("" + key.getClass().getSimpleName() + " is not supported. Value: " + String.valueOf(key));
+  private void invalidate0(Object key) {
```
chayim, line 35: Why this name?

src/main/java/redis/clients/jedis/ClientSideCache.java
```diff
@@ -1,71 +1,99 @@
...
-  protected void setKey(Object key, Object value) {
-    cache.put(getMapKey(key), value);
-  }
+    final ByteBuffer mapKey = makeKey((byte[]) key);
 
-  protected <T> T getValue(Object key) {
-    return (T) getMapValue(key);
+    Set<Long> hashes = keyHashes.get(mapKey);
+    if (hashes != null) {
+      writeLock.lock();
```
chayim, line 44: Not sure what I think about locking to invalidate. You're aligning the cache **but** at the cost of sharing.

src/main/java/redis/clients/jedis/ClientSideCache.java
```diff
@@ -1,71 +1,99 @@
...
+  }
+
+  protected abstract long getHash(CommandObject command);
+
+  private ByteBuffer makeKey(String key) {
+    return makeKey(SafeEncoder.encode(key));
   }
 
-  private static ByteBuffer convertKey(byte[] b) {
+  private static ByteBuffer makeKey(byte[] b) {
```
chayim, line 96: Github ate my comment :(

WDYT about make instead of `makeKey`? Or, were you thinking that there would be a different type of make (makeEntry, makeCommandKey, etc)?

src/main/java/redis/clients/jedis/ClientSideCache.java
```diff
@@ -1,71 +1,99 @@
...
-   */
-  ClientSideCache(Map<ByteBuffer, Object> map) {
-    this.cache = map;
-  }
+  public abstract void invalidateAll();
 
-  public final void clear() {
-    cache.clear();
-  }
+  protected abstract void invalidateAll(Iterable<Long> hashes);
```
chayim, line 24: Why should this be protected and the other public?

src/main/java/redis/clients/jedis/util/CaffeineCSC.java
```diff
@@ -0,0 +1,95 @@
+package redis.clients.jedis.util;
```
chayim, line 1: Can we subpackage the caching bits to `jedis.util.caching` or similar?

src/main/java/redis/clients/jedis/util/CaffeineCSC.java
```diff
@@ -0,0 +1,95 @@
...
+import net.openhft.hashing.LongHashFunction;
+import redis.clients.jedis.ClientSideCache;
+import redis.clients.jedis.CommandObject;
+import redis.clients.jedis.args.Rawable;
+
+public class CaffeineCSC extends ClientSideCache {
+
+  private static final int DEFAULT_MAXIMUM_SIZE = 10_000;
+  private static final int DEFAULT_EXPIRE_SECONDS = 100;
+  private static final LongHashFunction DEFAULT_HASH_FUNCTION = LongHashFunction.xx3();
```
chayim, line 15: IMHO let's share with users a comment on why this function. Be kind to our future selves and community.

src/main/java/redis/clients/jedis/util/GuavaCSC.java
```diff
@@ -0,0 +1,92 @@
...
+import com.google.common.cache.CacheBuilder;
+import com.google.common.hash.HashFunction;
+import com.google.common.hash.Hasher;
+import java.util.concurrent.TimeUnit;
+import redis.clients.jedis.ClientSideCache;
+import redis.clients.jedis.CommandObject;
+
+public class GuavaCSC extends ClientSideCache {
+
+  private static final int DEFAULT_MAXIMUM_SIZE = 10_000;
```
chayim, line 13: I'm more convinced now that we need some layer between these two cache types, within the package. Note that they share the same finals,  but different functions. It sounds like we should set the general rules in a parent, then the override (is HashFunction here) in the child.

Or.. is this overkill for two classes?

src/main/java/redis/clients/jedis/ClientSideCache.java
sazzad16: The other is required to be accessed from classes in different packages; e.g. during sentinel master switch, oss cluster reconfiguration, etc.

src/main/java/redis/clients/jedis/ClientSideCache.java
sazzad16: Before this PR, we were only invalidating keys. But we are also invalidating hashes. So I wanted have a more generic name. As we already have a method `invalidate` for the purpose of invalidate reply from Redis server, I have opted for `invalidate0` here.

src/main/java/redis/clients/jedis/util/CaffeineCSC.java
sazzad16: I guess... but these are just two classes.
There are many other packages that make sense. If we continue having new packages for one or two classes, we will end up with having way too many.

src/main/java/redis/clients/jedis/util/CaffeineCSC.java
sazzad16: Sure... we'll also need why Caffeine, why Guava, why not only one, etc.

src/main/java/redis/clients/jedis/ClientSideCache.java
chayim: Can we be more descriptive?

src/main/java/redis/clients/jedis/util/CaffeineCSC.java
chayim: Fair - but these are at least related. util is a catch-all, a dumping ground (not yours, generally). It wouldn't be bad to bring some order to this.

src/main/java/redis/clients/jedis/util/CaffeineCSC.java
chayim: 100%

src/main/java/redis/clients/jedis/ClientSideCache.java
sazzad16: Note: This is `private`.

src/main/java/redis/clients/jedis/ClientSideCache.java
sazzad16: Here we are creating a key for our key to command hashes map. Only `make` would be worse.

src/main/java/redis/clients/jedis/ClientSideCache.java
sazzad16: I felt we should use lock in following `getValue` method. IFF we use lock there, it makes sense to use the same lock here as well. Now the decision to make is if we want lock there?

src/main/java/redis/clients/jedis/ClientSideCache.java
chayim: Even so

src/main/java/redis/clients/jedis/ClientSideCache.java
chayim: I'm asking if we need the locks at all. Others don't have a lock implementation. What's your thinking?

sazzad16: > I'm asking if we need the locks at all.

I don't have a strong opinion about this.

> What's your thinking?

(stated in previous comment)

src/main/java/redis/clients/jedis/ClientSideCache.java
sazzad16: please check

src/main/java/redis/clients/jedis/ClientSideCache.java
sazzad16: some renaming are done. please re-check

src/main/java/redis/clients/jedis/ClientSideCache.java
sazzad16: removed lock

src/main/java/redis/clients/jedis/ClientSideCache.java
sazzad16: renamed

chayim (approved):

chayim (approved):

PR #3722, opened 2024-02-15
Remove commented out code

50-Course:
There is a commented code meant to initially invalidate `client` connection when the Connection is no longer listening.
However, it is currently being bypassed, and does not actually reference anything in the JVM.

This patch aims to remove it - therefore, and as such does not introduces breaking change to the actual module.


50-Course (commented):
the remove code goes thus:

```java



    ...
      // /* Invalidate instance since this thread is no longer listening */
    // this.client = null;
  }
}
```

With that, I do not think this should be left here anymore.

ggivo:
Clean up PR's. Removes 2 lines of commented-out code. Let's leave them for now as a hint till we review if they are actually needed,

PR #3726, opened 2024-02-19
Check for thread interrupt in subscribe process of PubSub

### LINKED ISSUE #3725: jedis subscription not listening to thread interrupted signal
### Issue Description

jedisCluster.ssubscribe(pubSub, channel) this line blocks forever. 

```java
 jedisCluster.ssubscribe(
         pubSub,
         "channel".getBytes()
 );
```
because, it runs in a loop keeps pulling message afterwards. see `JedisShardedPubSubBase::process()`. Here is a simplified snippet

```java
do {
      Object reply = client.getUnflushedObject();

       // process the message 
} while (isSubscribed());
``` 

As a result, this line ` jedisCluster.ssubscribe(..)` is usually invoked in a separate thread, for example a ThreadPool, to avoid blocking the main thread. 

During termination(shutdown), people will usually to shutdown the threadPool, which internally **will interrupt the threads**. 

But the process loop **does not check for interrupted flag**. Hence it is unaware that the thread that is running the process() call has been signalled to stop. 

Therefore, **the process loop keeps running while it should have exited upon interruption**. 

### Expected behavior

`JedisShardedPubSubBase::process` should exit the "do while" loop upon thread interrupt. 

### Actual behavior

`JedisShardedPubSubBase::process` keeps running despite thread interrupt

### Steps to reproduce:

1. start a `JedisCluster`
2. call `JedisCluster::ssubscribe(...)` on a different thread. can be in a thread pool or otherwise.
3. interrupt the thread that's running `JedisCluster::ssubscribe(...)`
4. You should expect to see consumption continues still, which means the thread will still be stuck at this line: `JedisCluster::ssubscribe(...)`

This is not related to configuration. 

#### Jedis version:
5.0.0

#### Redis version:
latest

#### Java version:
openjdk 17

### Suggested Fix:

```java
do {
      Object reply = client.getUnflushedObject();

       // process the message 
} while (!Thread.currentThread.isInterrupted() && isSubscribed());
```

SoilChang:
- adding thread interrupted check in `JedisShardedPubSubBase::process`
- also in `JedisPubSubBase::process`

Fixes #3725

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisPubSubBase.java
+++ b/src/main/java/redis/clients/jedis/JedisPubSubBase.java
@@ -172,7 +172,7 @@ public abstract class JedisPubSubBase<T> {
       } else {
         throw new JedisException("Unknown message type: " + reply);
       }
-    } while (isSubscribed());
+    } while (!Thread.currentThread().isInterrupted() && isSubscribed());
 
 //    /* Invalidate instance since this thread is no longer listening */
 //    this.client = null;
--- a/src/main/java/redis/clients/jedis/JedisShardedPubSubBase.java
+++ b/src/main/java/redis/clients/jedis/JedisShardedPubSubBase.java
@@ -99,7 +99,7 @@ public abstract class JedisShardedPubSubBase<T> {
       } else {
         throw new JedisException("Unknown message type: " + reply);
       }
-    } while (isSubscribed());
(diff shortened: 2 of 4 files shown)


SoilChang:
did the same set for `JedisPubSubBase`

sazzad16 (approved):

PR #3734, opened 2024-02-20
Do not rely on Redis compiled from source code

gerzse:
Issue #3710

Instead of building Redis server from source code, use Docker images. Prepare docker-compose stacks that accept the version of the Docker image as parameter, so we can run against multiple versions in CI.

The purpose of this change is to strictly move everything to Docker, without trying to change anything else. Some Java tests had to be adapted, still, because now the Redis instances are accessed via localhost, while internally in the docker-compose network they see each other with other IPs and ports.

Polish a bit the Makefile, for example to run exactly the same thing that runs in CI, so when we run the tests locally we get exactly the same coverage.


.github/workflows/integration.yml
```diff
@@ -22,20 +22,21 @@ on:
 jobs:
 
   build:
-    name: Build and Test
+    name: Build and Test - Redis Stack ${{matrix.redis-stack-version}}
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        redis-stack-version: ['6.2.6-v9', '7.2.0-RC3', 'edge']
```
chayim, line 30: Awesome. But, while here let's update the versions to the latest of each family. Current is 7.2.0-v8 (which is also the latest tag) and 6.2.6-v12.

.github/workflows/integration.yml
```diff
@@ -22,20 +22,21 @@ on:
...
+    env:
+      REDIS_STACK_VERSION: ${{matrix.redis-stack-version}}
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
```
chayim, line 34: @v4 is out. WDYT?

.github/workflows/integration.yml
```diff
@@ -22,20 +22,21 @@ on:
...
       - name: Set up publishing to maven central
         uses: actions/setup-java@v2
```
chayim, line 36: v4 is out, WDYT?

.gitignore
```diff
@@ -15,3 +15,4 @@ tags
 *.rdb
 redis-git
 appendonlydir/
+testmodule.so
```
chayim, line 18: Glob to future proof with *.so - just in case? Or, not worth it?

src/test/java/redis/clients/jedis/ClusterPipeliningTest.java
```diff
@@ -55,8 +53,8 @@ public static void setUp() throws InterruptedException {
     node3.flushAll();
 
     // add nodes to cluster
-    node1.clusterMeet(LOCAL_IP, nodeInfo2.getPort());
-    node1.clusterMeet(LOCAL_IP, nodeInfo3.getPort());
+    node1.clusterMeet(JedisClusterTestUtil.getClusterIp(2), 6379);
```
chayim, line 56: Could we possibly introduce the dotenv pattern here in general? Maybe this is overthinking it, but we did this [here](https://github.com/RedisLabs/module-tests), using an overloader [this way](https://github.com/RedisLabs/module-tests/blob/main/helpers/config.py)

src/test/java/redis/clients/jedis/JedisClusterTest.java
```diff
@@ -216,14 +216,14 @@ public void testMigrate() {
         node2.set("e", "e");
       } catch (JedisMovedDataException jme) {
         assertEquals(15363, jme.getSlot());
-        assertEquals(new HostAndPort(LOCAL_IP, nodeInfo3.getPort()), jme.getTargetNode());
+        assertEquals(new HostAndPort(JedisClusterTestUtil.getClusterIp(3), 6379), jme.getTargetNode());
```
chayim, line 219: Same general comment re ports - throughout.

test-infra/no-cluster/stunnel/pki/private.pem
```diff
@@ -0,0 +1,48 @@
+-----BEGIN RSA PRIVATE KEY-----
```
chayim, line 1: So we made this exact same mistake in redis-py, which IMHO is tech debt. When I originally committed it, I thought it smarter to commit junk certs (and a README), but it led to fallout. Since they're generated  via [this script](https://github.com/redis/redis/blob/unstable/utils/gen-test-certs.sh) from redis, maybe we should just fetch it, and run it first? We'd need a modification given the last line, but the only requirement is openssl.

chayim (commented):
Top-line comment:

1. Is there some way to remove all the hardcoded ports (or another PR?). It seems to be a constant theme throughout, and worth the debt removal.
2. That raises the question, is there a (painless) way to inject into the various configs, like stunnel and the dockers.

.github/workflows/integration.yml
gerzse: Updated.

.gitignore
gerzse: Updated.

src/test/java/redis/clients/jedis/JedisClusterTest.java
gerzse: Yes, we have to do something about this. It's so spaghetti right now, implied relations between ports and stuff. My intention was to leave that for another PR, go step by step. Here I did a minimal mapping to the new Docker IPs and ports.

test-infra/no-cluster/stunnel/pki/private.pem
gerzse: I agree, let's integrate that. In another PR?

src/test/java/redis/clients/jedis/ClusterPipeliningTest.java
gerzse: I'll have a look. We should decouple the running Redis instances from the test code, via some configuration.

ggivo:
https://github.com/redis/jedis/pull/4015 Introduces docker-based test infra and test matrix. Looks like most of  the functionality is already covered.  Since this one is marked as "will not fix" closing for now. 
In case we need to revisit feel free to reopen.

PR #3740, opened 2024-02-27
Support white-list and black-list commands and keys

sazzad16:
- Separate `csc` package
- Add config interface for only client-side caching
- Support white-list and black-list commands and keys

### DIFF
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -17,7 +17,7 @@ import redis.clients.jedis.Protocol.Command;
 import redis.clients.jedis.Protocol.Keyword;
 import redis.clients.jedis.args.ClientAttributeOption;
 import redis.clients.jedis.args.Rawable;
-import redis.clients.jedis.csc.ClientSideCache;
+import redis.clients.jedis.csc.ClientSideCacheConfig;
 import redis.clients.jedis.commands.ProtocolCommand;
 import redis.clients.jedis.exceptions.JedisConnectionException;
 import redis.clients.jedis.exceptions.JedisDataException;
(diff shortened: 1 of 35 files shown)


yangbodong22011:
Should we use allow-list and deny-list?

sazzad16:
> Should we use allow-list and deny-list?

@uglide Should we?
@chayim ?

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -66,7 +67,7 @@ public Connection(final JedisSocketFactory socketFactory, JedisClientConfig clie
     initializeConnection(clientConfig);
   }
 
-  public Connection(final JedisSocketFactory socketFactory, JedisClientConfig clientConfig, ClientSideCache csCache) {
+  public Connection(final JedisSocketFactory socketFactory, JedisClientConfig clientConfig, ClientSideCacheConfig csCache) {
```
gerzse, line 70: Also rename `csCache` to `csCacheConfig`? Same in other files.

src/main/java/redis/clients/jedis/csc/DefaultClientSideCacheConfig.java
```diff
@@ -0,0 +1,22 @@
...
+    this.csCache = clientSideCache;
+  }
+
+  @Override
+  public ClientSideCache getClientSideCache() {
+    return this.csCache;
+  }
+
+  @Override
+  public boolean isCacheable(ProtocolCommand command, Object... keys) {
```
gerzse, line 19: Given that there is a `default` in the interface, this is not needed.

src/main/java/redis/clients/jedis/util/JedisURIHelper.java
```diff
@@ -74,7 +78,7 @@ public static RedisProtocol getRedisProtocol(URI uri) {
 
   private static final Integer ZERO_INTEGER = 0;
 
-  public static ClientSideCache getClientSideCache(URI uri) {
+  public static ClientSideCacheConfig getClientSideCache(URI uri) {
```
gerzse, line 81: Maybe add a new argument, a cache config builder? Can be `Function<ClientSideCache, ClientSideCacheConfig>`. That might be a breaking change, so it can be a new method, to have an alternative.

uglide:
@sazzad16 LGTM.

PR #3137, opened 2022-09-12
Add Connection#getSocketString() method

sazzad16:
Inspiration: https://github.com/redis/jedis/discussions/3058

PR #3745, opened 2024-02-28
Get enriched Connection information

jjz921024:
Currently, the information returned by `tostring()` of connection is very poor. we can refer the info format of netty channel to represent a connection 

https://github.com/netty/netty/blob/3c65eae052b7f460e054e6ce765cfba7f433ea32/transport/src/main/java/io/netty/channel/AbstractChannel.java#L382-L419


sazzad16:
Note: Due to unique id, existing usages like `aConnection.toString(otherConnection.toString())` will fail.

jjz921024:
> Note: Due to unique id, existing usages like `aConnection.toString(otherConnection.toString())` will fail.

@sazzad16 Hi, Thanks for pointing out that this will bring breaking change. For compatibility, can we add two new method that provides `remoteAddr` and `localAddr` for user and keep `toString` method. 

I think the address info of `connection` is very useful for user at debug.

sazzad16:
@jjz921024 

> Hi, Thanks for pointing out that this will bring breaking change. For compatibility, can we add two new method that provides `remoteAddr` and `localAddr` for user and keep `toString` method.

That's possible.

You can also put all the necessary information into a single String, just without overriding toString(). This is something I started in #3137 due to #3058 but didn't move forward because of lack of traction.

jjz921024:
I have updated for this feature. Add two new `getRemoteHostAndPort()` and `getLocalHostAndPort()` method at connection.
When socket is connected, they will return remote and local address info respectively.  @sazzad16 please review, thank.

sazzad16:
@jjz921024 Apologies for the dilemma. I prefer your earlier implementation. WDYT about bringing it back; with renaming `toString()` to something like `toIdentityString()`.

jjz921024:
> @jjz921024 Apologies for the dilemma. I prefer your earlier implementation. WDYT about bringing it back; with renaming `toString()` to something else (like `toIdentityString()`).

That's fine. I had reverted the last commit. Please review.

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -67,9 +70,43 @@ public Connection(final JedisSocketFactory socketFactory, JedisClientConfig clie
     initializeFromClientConfig(clientConfig);
   }
 
-  @Override
-  public String toString() {
-    return "Connection{" + socketFactory + "}";
+  public String toIdentityString() {
```
sazzad16, line 73: Keep the existing one.
```suggestion
  @Override
  public String toString() {
    return "Connection{" + socketFactory + "}";
  }

  public String toIdentityString() {
```

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -67,9 +70,43 @@ public Connection(final JedisSocketFactory socketFactory, JedisClientConfig clie
...
+    } else {
+      StringBuilder buf = new StringBuilder(16)
+          .append("[id: 0x")
+          .append(id)
+          .append(']');
+      strVal = buf.toString();
+    }
+
+    strValActive = broken;
+    return strVal;
```
sazzad16, line 109: Include the class name using `getClass()`.

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/Connection.java
jjz921024: Ohhh.... Sorry about it. My mistake.

jjz921024:
@sazzad16 Hi, I found there are a NullPointException when invoke `toIdentityString` before connection is connected and had fixed  it. Please review.

sazzad16 (changes requested):
LGTM!
Please add unit test. At least a simple one.

jjz921024:
> LGTM! Please add unit test. At least a simple one.

@sazzad16 I had added a test case. Please review.  :)

sazzad16 (approved):

PR #3750, opened 2024-02-29
Reduce the log level of validateObject to WARN

### LINKED ISSUE #3749: Reduce the log level of validateObject to WARN
There have been confusions, misunderstandings, half-understandings regarding ERROR logs in `JedisFactory#validateObject` and `ConnectionFactor#validateObject`.

Here is a list of relevant concerns:

- #3748 
- #3723 
- #3314 
- #3106 
- #3100 
- #3067 
- #2712 

I think, reducing the log level to WARN would resolve or at least reduce these concerns to some extent.

sazzad16:
- Reduce the log level of `JedisFactory#validateObject` to `WARN`
- Reduce the log level of `ConnectionFactory#validateObject` to `WARN`

Resolves #3749

### DIFF
--- a/src/main/java/redis/clients/jedis/ConnectionFactory.java
+++ b/src/main/java/redis/clients/jedis/ConnectionFactory.java
@@ -76,7 +76,7 @@ public class ConnectionFactory implements PooledObjectFactory<Connection> {
       // check HostAndPort ??
       return jedis.isConnected() && jedis.ping();
     } catch (final Exception e) {
-      logger.error("Error while validating pooled Connection object.", e);
+      logger.warn("Error while validating pooled Connection object.", e);
       return false;
     }
   }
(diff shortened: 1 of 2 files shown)


src/main/java/redis/clients/jedis/ConnectionFactory.java
```diff
@@ -76,7 +76,7 @@ public boolean validateObject(PooledObject<Connection> pooledConnection) {
       // check HostAndPort ??
       return jedis.isConnected() && jedis.ping();
     } catch (final Exception e) {
-      logger.error("Error while validating pooled Connection object.", e);
+      logger.warn("Error while validating pooled Connection object.", e);
```
yangbodong22011, line 79: I think general business will configure the level of info. There seems to be no difference between error or warn (both will be printed). Maybe we should consider debug?

yangbodong22011 (approved):

PR #3757, opened 2024-03-06
Support Server-assisted Client-side Caching

sazzad16:
Resolves #2104

PR #3758, opened 2024-03-06
Client-side cache related naming changes

sazzad16:
Changes:
1. <strike>ClientSideCache and CommandLongHashing are good enough names on their own. But, IMO, these should have similar naming pattern; e.g. (Cache, Hash) or (Caching, Hashing). I opted for earlier option ATM.</strike>
  CommandLongHashing is renamed to CommandLongHasher.
3. Expanded the names of GuavaCSC (GuavaClientSideCache) and CaffeineCSC (CaffeineClientSideCache).

Note for reviewers: Check #3757 to see how client-side cache support currently looks and to find other changes to suggest.


src/main/java/redis/clients/jedis/csc/hash/CommandLongHash.java
```diff
@@ -5,7 +5,7 @@
 /**
  * The interface for hashing a command object for client-side caching.
  */
-public interface CommandLongHashing {
+public interface CommandLongHash {
```
gerzse, line 8: I would call this `CommandLongHasher`, and its implementations to match this.

Issue #2774, opened 2021-12-27
Revert commit 4c61eee


sazzad16:
Revert https://github.com/redis/jedis/commit/4c61eee3cb1d535cda3119c42eabc330a046d6d8

sazzad16:
Resolved by #3764

PR #3764, opened 2024-03-11
Reenable clustering tests

gerzse:
Uncomment clustering tests and adapt the code in the minimum possible way to make these tests run.

### DIFF
--- a/src/test/java/redis/clients/jedis/HostAndPorts.java
+++ b/src/test/java/redis/clients/jedis/HostAndPorts.java
@@ -36,9 +36,9 @@ public final class HostAndPorts {
     clusterHostAndPortList.add(new HostAndPort("localhost", 7383));
     clusterHostAndPortList.add(new HostAndPort("localhost", 7384));
 
-    stableClusterHostAndPortList.add(new HostAndPort("localhost", 10379));
-    stableClusterHostAndPortList.add(new HostAndPort("localhost", 10380));
-    stableClusterHostAndPortList.add(new HostAndPort("localhost", 10381));
+    stableClusterHostAndPortList.add(new HostAndPort("localhost", 7479));
+    stableClusterHostAndPortList.add(new HostAndPort("localhost", 7480));
+    stableClusterHostAndPortList.add(new HostAndPort("localhost", 7481));
   }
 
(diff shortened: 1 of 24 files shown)


src/test/java/redis/clients/jedis/HostAndPorts.java
```diff
@@ -34,6 +35,10 @@ public final class HostAndPorts {
     clusterHostAndPortList.add(new HostAndPort("localhost", 7382));
     clusterHostAndPortList.add(new HostAndPort("localhost", 7383));
     clusterHostAndPortList.add(new HostAndPort("localhost", 7384));
+
+    stableClusterHostAndPortList.add(new HostAndPort("localhost", 10379));
+    stableClusterHostAndPortList.add(new HostAndPort("localhost", 10380));
+    stableClusterHostAndPortList.add(new HostAndPort("localhost", 10381));
```
sazzad16, line 41: I'd prefer to keep these in 7000 series; i.e. `73XY` or `7X79-81`.

sazzad16 (changes requested):
Some nit-picky comments added. Also, can we not change the TestBase classes or not use hashtags `{}` in non-cluster mode?

sazzad16:
Resolves #2774

src/test/java/redis/clients/jedis/HostAndPorts.java
gerzse: Moved to 7479-7481.

gerzse:
> Some nit-picky comments added. Also, can we not change the TestBase classes or not use hashtags `{}` in non-cluster mode?

The TestBase classes already use that trick, that's how I found it. Initially I wanted to compute keys that hit the same hashslot, but this felt cleaner.

If you think, we can get rid of the hashtags, but maybe we can do it in all places, in a separate PR?

sazzad16:
A quick idea only for this PR, Keep the `...TestBase` classes untouched and override the necessary methods in `Cluster...Test` classes.

gerzse:
> A quick idea only for this PR, Keep the `...TestBase` classes untouched and override the necessary methods in `Cluster...Test` classes.

Changed as you suggested.

sazzad16 (approved):

PR #3767, opened 2024-03-12
Geo pipelined tests

gerzse:


gerzse:
> > <Pooled|Clustered|Pipelined>CommandsTest
> 
> Pipelined is not comparable to pooled/clustered. You can have pipelined with both pooled and clustered; i.e. Pooled|ClusteredPipelinedCommandsTest is possible. There could be better names but current names are better than your proposed ones.

Ok, got it. I changed back the names.

sazzad16 (approved):

PR #3770, opened 2024-03-13
Proof-of-concept for automatic key prefixing

killergerbah:
As per the suggestion from @sazzad16 in this discussion: https://github.com/redis/jedis/discussions/3765, I have created a POC that implements automatic key prefixing in the hopes of discussing possible changes that would allow applications to do this.

This POC demonstrates that key-prefixing becomes possible when 
- `UnifiedJedis` can be supplied with a custom `CommandObjects`. Some of the constructors that allow this would need to be made `public`.
- Further changes would be necessary for `CommandObjects` to not call the overridden `commandArguments` before the child class constructor finishes, which would allow the prefix to become an instance field.


sazzad16:
@killergerbah Thank you for your POC.

- This PR considers only cluster mode. Some idea showing non-cluster mode would also be helpful.
- It introduces a new client class `JedisClusterWithPrefixedKeys` which, IMO, should not be the way to go. One approach could be using config/parameter.
- Manipulating `CommandArguments` and `CommandObjects` is an interesting and good idea and can be investigated further.

killergerbah:
@sazzad16 Thanks for taking a look.

I have extended the POC to demonstrate an approach that would work for all subclasses of `UnifiedJedis`. I believe using this approach the only changes required would be to expose constructors that allow customization of `CommandObjects`. The sample `*WithPrefixedKeys` classes demonstrate that this is possible.

src/main/java/redis/clients/jedis/CommandObjects.java
```diff
@@ -63,22 +63,16 @@ protected CommandArguments commandArguments(ProtocolCommand command) {
     return new CommandArguments(command);
   }
 
-  private final CommandObject<String> PING_COMMAND_OBJECT = new CommandObject<>(commandArguments(PING), BuilderFactory.STRING);
-
   public final CommandObject<String> ping() {
-    return PING_COMMAND_OBJECT;
+    return new CommandObject<>(commandArguments(PING), BuilderFactory.STRING);
```
sazzad16, line 67: Why change this and following ones? Prefixed or not, keys won't change the command arguments.

src/main/java/redis/clients/jedis/UnifiedJedis.java
```diff
@@ -4859,7 +4859,7 @@ public PipelineBase pipelined() {
     } else if (provider instanceof MultiClusterPooledConnectionProvider) {
       return new MultiClusterPipeline((MultiClusterPooledConnectionProvider) provider, commandObjects);
     } else {
-      return new Pipeline(provider.getConnection(), true);
+      return new Pipeline(provider.getConnection(), commandObjects, true);
```
sazzad16, line 4862: :+1: 
Thank you for noticing.

src/test/java/redis/clients/jedis/PrefixedKeysTest.java
```diff
@@ -0,0 +1,36 @@
...
+public abstract class PrefixedKeysTest {
+
+    abstract UnifiedJedis prefixingJedis();
+
+    abstract UnifiedJedis nonPrefixingJedis();
+
+    @Test
+    public void hasPrefixedKeys() {
+        HostAndPort hp = new HostAndPort("127.0.0.1", 7379);
+
```
sazzad16, line 19:
```suggestion
```

src/main/java/redis/clients/jedis/JedisSentineled.java
```diff
@@ -23,6 +24,10 @@ public JedisSentineled(SentineledConnectionProvider sentineledConnectionProvider
     super(sentineledConnectionProvider);
   }
 
+  public JedisSentineled(CommandExecutor executor, SentineledConnectionProvider sentineledConnectionProvider, CommandObjects commandObjects, RedisProtocol redisProtocol) {
```
sazzad16, line 27: 1. We do not plan to change CommandExecutor from here.
```suggestion
  public JedisSentineled(SentineledConnectionProvider sentineledConnectionProvider, CommandObjects commandObjects, RedisProtocol redisProtocol) {
```
2. Add a JavaDoc stating redisProtocol will be set into commandObjects.

src/main/java/redis/clients/jedis/JedisPooled.java
```diff
@@ -394,6 +395,11 @@ public JedisPooled(PooledConnectionProvider provider) {
     super(provider);
   }
 
+  public JedisPooled(CommandExecutor executor, PooledConnectionProvider provider, CommandObjects commandObjects,
+                     RedisProtocol redisProtocol) {
```
sazzad16, line 399: 1. We do not plan to change CommandExecutor from here.
```suggestion
  public JedisPooled(PooledConnectionProvider provider, CommandObjects commandObjects, RedisProtocol redisProtocol) {
```
2. Add a JavaDoc stating redisProtocol will be set into commandObjects.

src/main/java/redis/clients/jedis/JedisCluster.java
```diff
@@ -222,6 +223,10 @@ private JedisCluster(ClusterConnectionProvider provider, int maxAttempts, Durati
     super(provider, maxAttempts, maxTotalRetriesDuration, protocol);
   }
 
+  public JedisCluster(ClusterCommandExecutor executor, ClusterConnectionProvider provider, ClusterCommandObjects commandObjects, RedisProtocol protocol) {
```
sazzad16, line 226: 1. We do not plan to change CommandExecutor from here.
```suggestion
  public JedisCluster(ClusterConnectionProvider provider, ClusterCommandObjects commandObjects, RedisProtocol protocol) {
```
2. Add a JavaDoc stating redisProtocol will be set into commandObjects.

sazzad16 (changes requested):
I like this idea.
It is now easier to think.

1. Move the new source classes in `redis.clients.jedis.util.prefix` package.
2. Move the new test classes in `redis.clients.jedis.prefix` package.

killergerbah:
Thanks @sazzad16 for reviewing.

There is one idea that just came to mind that may be worth considering.
Since it seems like you are willing for `jedis` to support key prefixing as an out-of-the-box behavior, I wonder if it would make sense to make sense to expose this behavior as an additional `prefix` field in `JedisClientConfig`. It seems that that would make it easier for callers to use the feature, rather than having to assemble the correct `CommandObjects` etc.

If that idea does not seem better, then I am happy to continue with this approach.

src/main/java/redis/clients/jedis/CommandObjects.java
killergerbah: The original problem was that the inline objects like `PING_COMMAND_OBJECT` were calling `commandArguments` which eventually called the constructor of `CommandArgumentWithPrefixedKeys` before the constructor of `CommandObjectsWithPrefixedKeys` even ran. At this point `prefixString` is null and the constructor of `CommandObjectsWithPrefixedKeys` throws an exception when it calls `SafeEncode` on a null value.

The fundamental issue seems to be the use of an overridable `protected` method inside the superclass constructor, which can cause problems because it's not possible to know if the class has actually finished constructing from the superclass.

However, I believe we can work around this adding some logic that takes into account the possibility of `prefixString` being null inside `CommandArgumentsWithPrefixedKeys`, if it is desirable to keep the cached `CommandObject` values.

killergerbah:
@sazzad16 I have made the requested changes, and additionally changes to support key-prefixing in transactions. Transaction support required adding `CommandObjects.watch`, and using `CommandObjects` for `WATCH` commands throughout the code.

However, I noticed that there does not seem to be a way to use `WATCH` with transactions using `UnifiedJedis.multi()` - is this intentional?

sazzad16 (changes requested):
Hmm :thinking:, this is getting complex.

src/test/java/redis/clients/jedis/prefix/JedisClusterPrefixedKeysTest.java
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        jedis.clusterAddSlots(slots);
+
+        try {
+            JedisClusterTestUtil.waitForClusterReady(jedis);
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+    }
```
sazzad16, line 47: Cluster reset is not done properly. For ease, you may check `HostAndPorts.getStableClusterServers()` cluster nodes and `ClusterCommandsTestHelper`.

sazzad16 (changes requested):

src/test/java/redis/clients/jedis/prefix/JedisClusterPrefixedKeysTest.java
killergerbah: Thank you. I've updated the cleanup based on `ClusterCommandsTestHelper`. The code is slightly duplicated due to `ClusterCommandsTestHelper` being package private. The tests appear to have passed on my fork.

sazzad16:
@killergerbah I have tried same feature a bit differently in #3781. Let me know what you think.

killergerbah:
> @killergerbah I have tried same feature a bit differently in #3781. Let me know what you think.

Thank you. I like your approach because it uses composition instead of inheritance to achieve the same outcome. It also does not require modification of constructors of `UnifiedJedis` and all of its subclasses. From a caller point of view it is easier to use since it requires just setting the command key processor instead of picking the right constructor to use. I would be supportive of continuing with your branch.

PR #3778, opened 2024-03-18
Extensive unit tests for PipeliningBase

gerzse:
Add extensive unit tests for the PipeliningBase abstract class. Since this class holds a lot of functionality for its subclasses, it makes sense to test it in isolation. And since the functionality is very repetitive, it makes sense to use mocking. Basically this abstract class just shuffles commands around from a CommandObjects to its subclasses, and this can be covered very well with mocked tests.

### DIFF
--- a/src/test/java/redis/clients/jedis/PipeliningBaseTest.java
+++ b/src/test/java/redis/clients/jedis/PipeliningBaseTest.java
@@ -1,0 +1,10502 @@
+package redis.clients.jedis;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.hasSize;
+import static org.hamcrest.Matchers.is;
+import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import java.util.AbstractMap;
+import java.util.ArrayList;
+import java.util.Collections;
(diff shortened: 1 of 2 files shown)


sazzad16 (commented):
10k+ lines is a bit extreme for a single class. We should consider splitting it.

E.g. we can put the variables in PipeliningBaseTestBase and have separate test classes per data type.

gerzse:
> 10k+ lines is a bit extreme for a single class. We should consider splitting it.
> 
> E.g. we can put the variables in PipeliningBaseTestBase and have separate test classes per data type.

I though about this at some point, but I was not convinced it would actually help. Some arguments why a single file is better:
- PipeliningBase itself is huge, over 4K lines, over 800 methods. Right now this single test class has the test methods written in the same order as PipeliningBase, so theoretically if you read them in parallel, you won't get lost (as long as we maintain this order). If we split up the tests into multiple sub-classes, the only way to find your way around would be search.
- Using inheritance to share just variables for testing, and not tests themselves, somehow does not feel clean. Arguably we could work around this by copying the variables (not all of them will be needed for each subclass per Redis type).

To be honest I find it more controlled to have everything in one place, rather than scattered around. But I'm open to split it, if you are really convinced it would help.

You know, after all 10K is just a number. `redis.clients.jedis.Jedis` has 9600 lines, `redis.clients.jedis.UnifiedJedis` has 4900. We have some small "monsters" in the code base, and in order to tame them we need bigger "monsters" in the test code.

sazzad16 (approved):

PR #3781, opened 2024-03-20
Support automatic namespacing

sazzad16:
Support automatic namespacing (key prefixing) by handler (key pre-processing) interface.

This is based on #3770 by @killergerbah 

Closes #3770


killergerbah:
Pasting my thoughts from https://github.com/redis/jedis/pull/3770:

> Thank you. I like your approach because it uses composition instead of inheritance to achieve the same outcome. It also does not require modification of constructors of UnifiedJedis and all of its subclasses. From a caller point of view it is easier to use since it requires just setting the command key processor instead of picking the right constructor to use. I would be supportive of continuing with your branch.

PR #3787, opened 2024-03-25
Align Pipelined commands with UnifiedJedis

gerzse:
Add some Pipelined commands that are available in UnifiedJedis but are missing the pipelined version.

### DIFF
--- a/src/main/java/redis/clients/jedis/PipeliningBase.java
+++ b/src/main/java/redis/clients/jedis/PipeliningBase.java
@@ -1607,18 +1607,8 @@ public abstract class PipeliningBase
   }
 
   @Override
-  public Response<String> scriptLoad(String script) {
-    return appendCommand(commandObjects.scriptLoad(script));
-  }
-
-  @Override
   public Response<String> scriptLoad(String script, String sampleKey) {
     return appendCommand(commandObjects.scriptLoad(script, sampleKey));
-  }
-
-  @Override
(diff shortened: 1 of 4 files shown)


src/main/java/redis/clients/jedis/commands/SampleKeyedPipelineCommands.java
sazzad16, line 1: The changes in this file ignores "sample key" term.

src/main/java/redis/clients/jedis/PipeliningBase.java
```diff
@@ -1621,6 +1631,11 @@ public Response<String> scriptFlush(String sampleKey, FlushMode flushMode) {
     return appendCommand(commandObjects.scriptFlush(sampleKey, flushMode));
   }
 
+  @Override
+  public Response<String> scriptKill() {
+    return appendCommand(commandObjects.scriptKill());
+  }
+
```
sazzad16, line 1638: We don't need SCRIPT KILL in pipeline.
```suggestion
```

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/commands/SampleKeyedPipelineCommands.java
gerzse: Reverted these changes.

sazzad16 (approved):

PR #3791, opened 2024-03-26
Add last entry id for XREADs and support XREADs reply as map

### LINKED ISSUE #3732: Support XREAD, XREADGROUP reading from last message of stream
As documented [here](https://github.com/redis/redis/pull/13117), to be tested against redis unstable.

sazzad16:
- Added a last entry id for XREAD and XREADGROUP commands (Resolves #3732)
- Support map replies for XREAD and XREADGROUP commands (currently have list of key-value pairs)

---

Task breakdown (ToDo):
- [x] new StreamEntryID for less confusion
- [x] new XREAD and XREADGROUP support as map
- [x] command tests
- [ ] pipeline tests

### DIFF
--- a/src/main/java/redis/clients/jedis/StreamEntryID.java
+++ b/src/main/java/redis/clients/jedis/StreamEntryID.java
@@ -108,7 +108,8 @@ public class StreamEntryID implements Comparable<StreamEntryID>, Serializable {
   };
 
   /**
-   * @deprecated Use {@link StreamEntryID#XGROUP_LAST_ID} or {@link StreamEntryID#XREAD_NEW_ENTRY}.
+   * @deprecated Use {@link StreamEntryID#XGROUP_LAST_ENTRY} for XREADGROUP command or
+   * {@link StreamEntryID#XREAD_NEW_ENTRY} for XREAD command.
    */
   @Deprecated
   public static final StreamEntryID LAST_ENTRY = XGROUP_LAST_ENTRY;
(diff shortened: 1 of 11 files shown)


src/main/java/redis/clients/jedis/StreamEntryID.java
```diff
@@ -151,4 +172,20 @@ public String toString() {
...
     }
   };
+
+  /**
+   * Should be used only with XREAD
+   *
+   * {@code XREAD STREAMS mystream +}
+   */
+  // TODO: unify with MAXIMUM_ID ??
+  public static final StreamEntryID XREAD_LAST_ENTRY = new StreamEntryID() {
```
gerzse, line 182: This could be `XREAD_LAST_ENTRY = MAXIMUM_ID`, to make it obvious that it is only an alias. Somehow I find it more informative to have different names, depending on the usage context.

src/main/java/redis/clients/jedis/StreamEntryID.java
```diff
@@ -129,6 +147,7 @@ public String toString() {
   /**
    * Can be used in XRANGE, XREVRANGE and XPENDING commands.
    */
+  // TODO: FIRST_ID ?
```
gerzse, line 150: I would vote for keeping `MINIMUM_ID`, i.e. preserve the naming from Redis, as much as possible. Same for `MAXIMUM_ID`.

gerzse (approved):

PR #3792, opened 2024-03-28
Fix NPE in empty XPENDING summary

gerzse:
When there are no pending messages in a stream, the summary response to XPENDING contains nulls, except the first position which is 0 (the total number of pending message). Return early from the builder, in this case, to avoid NPE.

### DIFF
--- a/src/main/java/redis/clients/jedis/BuilderFactory.java
+++ b/src/main/java/redis/clients/jedis/BuilderFactory.java
@@ -1813,6 +1813,11 @@ public final class BuilderFactory {
 
       List<Object> objectList = (List<Object>) data;
       long total = BuilderFactory.LONG.build(objectList.get(0));
+
+      if (total == 0) {
+        return new StreamPendingSummary(total, null, null, null);
+      }
+
       String minId = SafeEncoder.encode((byte[]) objectList.get(1));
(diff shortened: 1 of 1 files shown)


sazzad16 (commented):
Do you have the test where you got NPE? Could you add that?

gerzse:
> Do you have the test where you got NPE? Could you add that?

I have it in a big PR only with unit tests. It would be some effort to retrofit it. Would it be ok to merge this separately, and the test later?

src/main/java/redis/clients/jedis/BuilderFactory.java
```diff
@@ -1813,6 +1813,11 @@ public StreamPendingSummary build(Object data) {
 
       List<Object> objectList = (List<Object>) data;
       long total = BuilderFactory.LONG.build(objectList.get(0));
+
+      if (total == 0) {
+        return new StreamPendingSummary(total, null, null, null);
```
sazzad16, line 1818: We just don't return null on our own.
We return null or whatever only after reading it.

sazzad16: Check #3793

gerzse:
Fixed more properly in 7393.

PR #3795, opened 2024-04-01
Remove broken JSON.ARRAPPEND method from CommandObjects

gerzse:
One of the jsonArrAppend methods in CommandObjects has broken serialization. Instead of deserializing the response into a List<Long>, it's only expecting Long. Luckily this method is not used, so better remove it.


sazzad16 (dismissed):

sazzad16 (commented):
It's not broken. RedisJSON v1 returns a Long in this case.
RedisJSON v1 support is already deprecated. So deprecating this method is the safest approach IMO.

PR #3797, opened 2024-04-03
Implemented read-write separation based Sentinel

Vampirea:
Implemented read-write separation function based on Redis Sentinel, balancing read request load to all slave, Support Redis high availability and failover.

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisSentinelSlaveFactory.java
+++ b/src/main/java/redis/clients/jedis/JedisSentinelSlaveFactory.java
@@ -1,0 +1,137 @@
+package redis.clients.jedis;
+
+import org.apache.commons.pool2.PooledObject;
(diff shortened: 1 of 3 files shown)


sazzad16 (commented):
First of all thank you for using Jedis and for trying to improve it.

1. The term _slave_ is either removed or deprecated.
2. Currently our focus is more towards UnifiedJedis (with the help of CommandExecutor, ConnectionProvider, etc.) compared to Jedis and Pool. So, at this moment, whether you implement this with Jedis and Pool or not, we prefer an implementation with UnifiedJedis.
3. It would be great if this can be done in JedisSentineled with some configs.

uglide:
Closing as stale. Related PR https://github.com/redis/jedis/pull/4231

Issue #3566, opened 2023-10-03
StreamEntry (XREAD) does not support binary because it uses Map<String, String>


raylau1234:
StreamEntry uses a Map<String, String> for fields.

As a result, binary data which contains illegal utf-8 sequences get mangled.

Reproduce by using XADD with a key or value which contains an illegal utf-8 sequence like 0xc3 0x28.

Retrieve via XREAD.

Expected result -- API to get byte array instead of a String, which is lossy with illegal utf-8 sequences.

sazzad16:
First of all, what is XGET?

Secondly, there are `xrange(byte[] key, ...)` methods. Don't these help?

raylau1234:
Sorry, I meant XREAD.

You are correct on the alternate XRANGE signature.  But I am really using XREAD.  Let me update the title and to clarify.... XREAD only is problematic.

raylau1234:
I see there are alternate signatures for xread as well.  Let me see of those help.  I found the StreamEntry ones first.  Give me a few minutes to explore and will udpate.

raylau1234:
OK, `List<byte[]> xread(XReadParams, Entry<byte[], byte[]>... streams)` is a possible workaround, but is cumbersome bec I need to deserialize the byte into the map of key / message id / values.  i.e. I need to replicate everything that `List<Map, Entry<String, List<StreamEntry>>> xread(XReadParams, Map<String, StreamEntryID>)` is doing in terms of decoding with STREAM_READ_RESPONSE, STREAM_ENTRY_LIST, etc. (but skipping the SafeEncoder.encode() and keeping the binary there).

thachlp:
Hi @raylau1234, I would like to work on this issue.
But I see that there is `xadd` pass Map<byte[], byte[]> already
```
  public final CommandObject<byte[]> xadd(byte[] key, XAddParams params, Map<byte[], byte[]> hash) {
    return new CommandObject<>(addFlatMapArgs(commandArguments(XADD).key(key).addParams(params), hash),
        BuilderFactory.BINARY);
  }
```
Do I misunderstand?

thachlp:
@kmularise sure, please open the new pr for this issue, I will close mine 🙇

kmularise:
@thachlp Thanks for the clarification and for allowing others to pick this up.
I’ll start working on a new PR for this issue. Appreciate the prior work done on #3803!

PR #3803, opened 2024-04-06
StreamEntry support Binary

thachlp:
Issue: https://github.com/redis/jedis/issues/3566

### DIFF
--- a/src/main/java/redis/clients/jedis/BuilderFactory.java
+++ b/src/main/java/redis/clients/jedis/BuilderFactory.java
@@ -1317,6 +1317,38 @@ public final class BuilderFactory {
     }
   };
 
+    public static final Builder<List<StreamEntryBinary>> STREAM_ENTRY_BINARY = new Builder<List<StreamEntryBinary>>() {
+      @Override
+      @SuppressWarnings("unchecked")
+      public List<StreamEntryBinary> build(Object data) {
+        List<StreamEntryBinary> result = new ArrayList<>();
+        if (data == null) {
+          return result;
+        }
+        List<Object> objectList = (List<Object>) data;
+
+        for (Object obj : objectList) {
+          List<Object> entry = (List<Object>) obj;
+          String entryIdString = SafeEncoder.encode((byte[]) entry.get(0));
+          StreamEntryID entryID = new StreamEntryID(entryIdString);
+          List<byte[]> hash = (List<byte[]>) entry.get(1);
+
+          Iterator<byte[]> hashIterator = hash.iterator();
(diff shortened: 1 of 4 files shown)


sazzad16 (changes requested):
We must not break the existing implementations.

thachlp:
> We must not break the existing implementations.

Hi, @sazzad16, thanks for the review 🙇
How do we resolve this issue?

sazzad16:
> How do we resolve this issue?

By creating new methods.

---

I just realized what your title says:

> StreamEntry support Binary instead of String

Actually we want StreamEntry binary support while keeping the String ones.

thachlp:
> > How do we resolve this issue?
> 
> By creating new methods.
> 
> I just realized what your title says:
> 
> > StreamEntry support Binary instead of String
> 
> Actually we want StreamEntry binary support while keeping the String ones.

Thanks, I will update.

thachlp (commented):
Hi @sazzad16, as I understand, there is xadd support `Map<byte[], byte[]> hash` already.
```
  public final CommandObject<byte[]> xadd(byte[] key, XAddParams params, Map<byte[], byte[]> hash) {
    return new CommandObject<>(addFlatMapArgs(commandArguments(XADD).key(key).addParams(params), hash),
        BuilderFactory.BINARY);
  }
```

jfloodnet:
Now that this package supports binary value in stream entries, would someone be able to update spark redis to support binary columns also?

It is currently only Map<string, string> https://github.com/RedisLabs/spark-redis/blob/master/src/main/scala/com/redislabs/provider/redis/streaming/RedisStreamReceiver.scala#L108

Which causes the following error:

`java.lang.ClassCastException: class org.apache.spark.unsafe.types.UTF8String cannot be cast to class [B (org.apache.spark.unsafe.types.UTF8String is in unnamed module of loader 'app'; [B is in module java.base of loader 'bootstrap')
        at org.apache.spark.sql.catalyst.expressions.BaseGenericInternalRow.getBinary(rows.scala:46)`

If your stream entry schema has binary data:

```    
    public static StructType redisSchema() {
        return new StructType(new StructField[]{
                DataTypes.createStructField("msg", DataTypes.StringType, true)
        });
    }
```

sazzad16:
@jfloodnet This PR is closed by OP without being merged.

jfloodnet:
> @jfloodnet This PR is closed by OP without being merged.

oh that's true. @thachlp what was the reason you closed this PR?

thachlp:
@jfloodnet I am just ambiguous about doing it, but If it is important, I will continue on it now, I will ask @sazzad16 for help.

PR #3816, opened 2024-04-11
Renew slot cache if ClusterPipeline.sync results in MOVED response

killergerbah:
The intent of this change is to fix a couple issues `MultiNodePipelineBase`:
- Exceptions thrown from `Connection.getMany` are swallowed.
- `JedisMovedDataException` does not trigger a call to `JedisClusterInfoCache.renewClusterSlots` (and there may be no other opportunity to call `JedisClusterInfoCache.renewClusterSlots` if the application did not specify a `topologyRefreshPeriod` when constructing `JedisCluster`).

With this implementation only the first exception thrown from `Connection.getMany` is rethrown; however this matches the behavior of `Pipeline`.

### DIFF
--- a/src/main/java/redis/clients/jedis/ClusterPipeline.java
+++ b/src/main/java/redis/clients/jedis/ClusterPipeline.java
@@ -3,6 +3,7 @@ package redis.clients.jedis;
 import java.time.Duration;
 import java.util.Set;
 import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
+import redis.clients.jedis.exceptions.JedisMovedDataException;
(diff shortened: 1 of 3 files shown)


sazzad16:
Reacting to MOVED response is reasonable request.

BUT the priority is put more towards ensuring execution of pipeline in as many connections as possible.
The `finally { executorService.shutdownNow(); }` goes against that.

Moreover, there are many cases to consider and I have concerns about those. One simple to understand of those is `syncing = false;` should be within finally. Because I have concerns overall with the approach, I am not putting it as code review.

killergerbah:
Got it - I suspect there is anyway a more complete fix for redirect responses from pipelines used on clusters.

PR #3831, opened 2024-05-05
Support RediSearch DIALECT 5

sazzad16:
- [x] Avoid escaping at query time
- [ ] Alias for tag fields (EXACT)
- [x] Avoid repeating for numeral equality
- [x] New dialect (5)


uglide (commented):
I suggest changing the title to "Test with RediSearch DIALECT 5" since there are no actual changes to support it.

sazzad16:
> I see there are a bunch of "redis.clients.jedis.exceptions.JedisDataException: DIALECT requires a non negative integer >=1 and <= 4". Do you plan to address them in this PR?

@uglide These are coming from the server (RediSearch). Once the server is ready/updated, these should go away.
BTW, if you know any docker image where we can test dialect 5, let me know so we can test this PR with that.

PR #3860, opened 2024-06-12
Support IGNORE and other optional arguments for timeseries commands

sazzad16:
* Re-implement TS.ADD command with optional arguments
* Implement TS.INCRBY and TS.DECRBY commands with optional arguments
* Support IGNORE argument for TS.[ CREATE | ALTER | ADD | INCRBY | DECRBY] commands

### DIFF
--- a/src/main/java/redis/clients/jedis/CommandObjects.java
+++ b/src/main/java/redis/clients/jedis/CommandObjects.java
@@ -3946,9 +3946,15 @@ public class CommandObjects {
     return new CommandObject<>(commandArguments(TimeSeriesCommand.ADD).key(key).add(timestamp).add(value), BuilderFactory.LONG);
   }
 
+  @Deprecated
   public final CommandObject<Long> tsAdd(String key, long timestamp, double value, TSCreateParams createParams) {
-    return new CommandObject<>(commandArguments(TimeSeriesCommand.ADD).key(key)
-        .add(timestamp).add(value).addParams(createParams), BuilderFactory.LONG);
+    return new CommandObject<>(commandArguments(TimeSeriesCommand.ADD).key(key).add(timestamp).add(value)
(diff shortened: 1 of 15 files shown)


sazzad16 (commented):
To the reviewers, I am not 100% sure about the name of `TSIncrByDecrByParams`. We usually have one Params class for one command. But in this case, two different commands (TS.INCRBY and TS.DECRBY) have exactly same set of optional arguments. Could you suggest a better class to represent optional arguments of both commands.

atakavci:
you can try to extend from a generic base where generic type extends from base itself. And put almost everything into base and derive incrby and decrby .
like ...
```
public class TSArithmeticOpParams<T extends TSArithmeticOpParams<?>> implements IParams {

  private Long timestamp;
  private Long retentionPeriod;
  private EncodingFormat encoding;
  private Long chunkSize;
  private DuplicatePolicy duplicatePolicy;
  private Map<String, String> labels;

  public TSArithmeticOpParams() {
  }

  public T timestamp(long timestamp) {
    this.timestamp = timestamp;
    return (T) this;
  }

  @Override
  public void addParams(CommandArguments args) {
    // TODO Auto-generated method stub
    throw new UnsupportedOperationException("Unimplemented method 'addParams'");
  }
.
.
.
```

and derive like

```
public class TSDecrByParams extends TSArithmeticOpParams<TSDecrByParams> {

  public TSDecrByParams() {
  }

  public static TSDecrByParams params() {
    return new TSDecrByParams();
  }
.
.
```

atakavci:
for future reference if you need to reuse something in a case like this;
```
public final CommandObject<Long> tsAdd(String key, long timestamp, double value, TSCreateParams createParams) { 
```
just extend a dummy **TSAdd** from **TSCreateParams** and use it. It will pay of when the time comes.

atakavci (dismissed):
looks good, thanks

PR #3870, opened 2024-06-21
Extract messages of unsupported exception as constants

zyfx595701088:
Optimize code


src/main/java/redis/clients/jedis/util/RedisInputStream.java
sazzad16, line 1: Please don't change anything out of the scope of the PR. If you want you can do it in separate PR.

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/util/RedisInputStream.java
zyfx595701088: ok, I will do it.

PR #3900, opened 2024-07-19
Refactor Client-Side Caching implementation

atakavci:


src/test/java/redis/clients/jedis/csc/JedisPooledClientSideCacheTest.java
```diff
@@ -32,26 +34,50 @@ public class JedisPooledClientSideCacheTest {
...
+  static void setupTrustStore() {
+    setJvmTrustStore("src/test/resources/truststore.jceks", "jceks");
+  }
+
+  private static void setJvmTrustStore(String trustStoreFilePath, String trustStoreType) {
+    assertTrue(String.format("Could not find trust store at '%s'.", trustStoreFilePath),
+        new File(trustStoreFilePath).exists());
+    System.setProperty("javax.net.ssl.trustStore", trustStoreFilePath);
+    System.setProperty("javax.net.ssl.trustStoreType", trustStoreType);
+  }
```
sazzad16, line 55: Instead of having just one test method with SSL, do all tests with SSL in a new test class; i.e. SSSJedisPooledClientSideCacheTest.

sazzad16 (changes requested):

sazzad16:
In tests with pool, I have used pool with single connection in many tests because we could only have deterministic tests that way (design 1).

@atakavci Could you please write same tests with multi connection pool (or default pool) as well?

src/test/java/redis/clients/jedis/csc/JedisPooledClientSideCacheTest.java
atakavci: this is just POC of workaround(PING stuff) for SSL. Target is to replace it with solution that underlying socket utilized for availability check. But your comment is valid anyway.

src/main/java/redis/clients/jedis/csc/ClientSideCache.java
```diff
@@ -1,120 +1,210 @@
...
- * object; e.g. {@link redis.clients.jedis.csc.CaffeineClientSideCache CaffeineClientSideCache} or
- * {@link redis.clients.jedis.csc.GuavaClientSideCache GuavaClientSideCache} or a custom implementation of their own.
+ * The class to manage the client-side caching. User can provide any of implementation of this class
+ * to the client object; e.g. {@link redis.clients.jedis.csc.CaffeineClientSideCache
+ * CaffeineClientSideCache} or {@link redis.clients.jedis.csc.GuavaClientSideCache
+ * GuavaClientSideCache} or a custom implementation of their own.
  */
 @Experimental
-public abstract class ClientSideCache {
+public abstract class ClientSideCache implements Cache {
```
sazzad16, line 23: IMO, we shouldn't use both ClientSideCache and Cache here. Use `AbstractCache` or similar.

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/csc/ClientSideCache.java
atakavci: sure,, AbstractCache sounds good.

src/main/java/redis/clients/jedis/csc/AbstractCache.java
```diff
@@ -0,0 +1,223 @@
...
+
+  // protected abstract Collection<CacheKey> remove(Set<CacheKey<?>> commands);
+
+  protected abstract void clearStore();
+
+  protected abstract Boolean containsKeyInStore(CacheKey cacheKey);
+
+  // End of abstract methods to be implemented by the concrete classes
+
+  private ByteBuffer makeKeyForRedisKeysToCacheKeys(Object key) {
```
tishun, line 208: I had to read this ... several times ... and I am still not sure what it does 😄 

Could we simplify the name and put a comment that explains the purpose?

src/main/java/redis/clients/jedis/csc/CacheEntry.java
```diff
@@ -1,30 +1,58 @@
...
+
+  private static byte[] toBytes(Object object) {
+    try (ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        ObjectOutputStream oos = new ObjectOutputStream(baos)) {
+      oos.writeObject(object);
+      oos.flush();
+      oos.close();
+      return baos.toByteArray();
+    } catch (Exception e) {
+      throw new JedisCacheException("Failed to serialize object", e);
```
tishun, line 46: I am wondering - should cache serialization fail the app?

Assuming there is some value that breaks serialization for some reason (temporary or not) - we could probably continue the normal program execution without disturbing the flow, only log the error in the log, and ignore the cache. In any case falling back to reading from the server should be completely safe, correct?

tishun: Btw do we need to catch `Exception` here, could it be some more specific implementation?

src/main/java/redis/clients/jedis/csc/CacheStats.java
```diff
@@ -0,0 +1,89 @@
+package redis.clients.jedis.csc;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+public class CacheStats {
```
tishun, line 5: I like this. Will help us track some stats and see how the app works.

In the long run we should probably use a framework such as https://micrometer.io/ (Lettuce uses that)

src/main/java/redis/clients/jedis/csc/DefaultCache.java
```diff
@@ -0,0 +1,72 @@
+package redis.clients.jedis.csc;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+public class DefaultCache extends AbstractCache {
+
+    protected final Map<CacheKey, CacheEntry> cache;
+    private final LRUEviction evictionPolicy;
```
tishun, line 10: Perhaps use the interface `EvictionPolicy` to make it possible to switch policies?

Also have one (unused) constructor to enable creating the cache with a custom eviction policy?

src/main/java/redis/clients/jedis/csc/GuavaClientSideCache.java
```diff
@@ -2,34 +2,48 @@
...
+import java.util.stream.StreamSupport;
 
-public class GuavaClientSideCache extends ClientSideCache {
+@Experimental
+public class GuavaClientSideCache extends AbstractCache {
 
   private final Cache<CacheKey, CacheEntry> cache;
+  protected static final int DEFAULT_MAXIMUM_SIZE = 10_000;
+  protected static final int DEFAULT_EXPIRE_SECONDS = 100;
+  private final LRUEviction evictionPolicy;
```
tishun, line 20: Same as last comment

tishun (commented):
A very sound improvement.
Mostly small comments.

I'd like to pull the source and play around with it next week.

Two general notes:
- the [latest design doc ](https://redislabs.atlassian.net/wiki/spaces/CAE/pages/4295884832/Lightweight+client-side+caching+design#Logical-data-model) mentions we want to have `one cache per pool` and as far as I can see we are doing cache per connection in this solution, is that right?
- we are also missing the cache key prepopulation (unless I missed it) that is supposed to help avoid cache invalidation race condition

src/main/java/redis/clients/jedis/ConnectionFactory.java
```diff
@@ -7,8 +7,8 @@
 import org.slf4j.LoggerFactory;
 
 import redis.clients.jedis.annots.Experimental;
+import redis.clients.jedis.csc.Cache;
```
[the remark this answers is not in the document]
atakavci: probably we will do that, right now my focus is making progress with jedis as fast as possible.

src/main/java/redis/clients/jedis/UnifiedJedis.java
```diff
@@ -240,21 +243,23 @@ public UnifiedJedis(CommandExecutor executor, ConnectionProvider provider, Comma
       try (Connection conn = this.provider.getConnection()) {
         if (conn != null) {
           RedisProtocol proto = conn.getRedisProtocol();
-          if (proto != null) this.commandObjects.setProtocol(proto);
+          if (proto != null)
+            this.commandObjects.setProtocol(proto);
         }
-      } catch (JedisException je) { }
+      } catch (JedisException je) {
```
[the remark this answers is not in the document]
atakavci: definitely we need a closer look on why we are doing that. 
It is only formatting diff in this PR.

src/main/java/redis/clients/jedis/csc/AbstractCache.java
```diff
@@ -0,0 +1,223 @@
...
+ * CaffeineClientSideCache} or {@link redis.clients.jedis.csc.GuavaClientSideCache
+ * GuavaClientSideCache} or a custom implementation of their own.
+ */
+@Experimental
+public abstract class AbstractCache implements Cache {
+
+  private ClientSideCacheable cacheable = DefaultClientSideCacheable.INSTANCE; // TODO: volatile
+  private final Map<ByteBuffer, Set<CacheKey<?>>> redisKeysToCacheKeys = new ConcurrentHashMap<>();
+  private final int maximumSize;
+  private ReentrantLock lock = new ReentrantLock();
```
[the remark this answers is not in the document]
atakavci: i think saying that jedis is not thread safe is not correct wording/terminology to use. 
it is more like, standalone instance of type 'Jedis' is not thread-safe is a better statement. 
this is where connection pool comes into play, basically it relies on apache object pools which uses synchronization constructs.

important point here is we have a design where a pool of objects share the cache instance, which makes using safety guards mandatory to keep the cache itself stable and reliable (at cost of using locks and/or some synchronization stuff).

src/main/java/redis/clients/jedis/csc/AbstractCache.java
```diff
@@ -0,0 +1,223 @@
...
+
+  @Override
+  public Boolean delete(CacheKey cacheKey) {
+    lock.lock();
+    try {
+      boolean removed = removeFromStore(cacheKey);
+      getEvictionPolicy().reset(cacheKey);
+
+      // removing it from redisKeysToCacheKeys as well
+      // TODO: considering not doing it, what is the impact of not doing it ??
```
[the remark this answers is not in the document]
atakavci: you are right. my concern was regarding to runtime performance; removing them one by one ends up O(n) complexity.  
as u suggested, if that redis key does not invalidates or cachekey itself is not operated on after this removal, then it will remain in the set. It ends up in a dirty state.
thanks for pushing me to think over this.

src/main/java/redis/clients/jedis/csc/AbstractCache.java
atakavci: convertRedisKeyToBytes how does that sound ?
since you ask, i started to think it needs more attention; now questioning if it is same key when you have a string 'x' and a byte array 'y' ends up with same value.

src/main/java/redis/clients/jedis/csc/CacheStats.java
atakavci: i ll give it a thought,,  a bit concerned about not caching when it is the expectation. 
imho warnings are tend to be out of sight when there are no bad things under investigation.

src/main/java/redis/clients/jedis/csc/CacheEntry.java
atakavci: i ll give it a thought,,  a bit concerned about not caching when it is the expectation. 
imho warnings are tend to be out of sight when there are no bad things under investigation.

src/main/java/redis/clients/jedis/csc/CacheStats.java
atakavci: lets keep that in mind for next phase of csc

src/main/java/redis/clients/jedis/csc/DefaultCache.java
atakavci: done

src/main/java/redis/clients/jedis/csc/GuavaClientSideCache.java
atakavci: done

src/main/java/redis/clients/jedis/csc/CacheEntry.java
tishun: About the catch-everything pattern (where you catch for the `Exception` or even worse `Throwable` level) is that of handling what you caught. Some exceptions could not be handled in a meaningful way (most `RuntimeException`s e.g. `NullPointerException`, `ClassCastException`) while others could. When you wrap all of them all in a JedisCacheException the users of the driver could assume they could catch it and handle it in some way (e.g. just ignore them) and could - involuntary - handle more than they expected (and hide some bug in the driver).

IMHO a good pattern would be to let all of these exceptions propagate naturally and only wrap checked exceptions. This would make our code - more or less - in line with the "unchecked exceptions only" club (you probably know there are two conflicting philosophies on checked vs unchecked exceptions).

src/main/java/redis/clients/jedis/UnifiedJedis.java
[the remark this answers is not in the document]
tishun: yeah quite a lot of places it seems we are swallowing exceptions

src/main/java/redis/clients/jedis/csc/CacheEntry.java
atakavci: how about wrapping all into a custom cacheserialization and cachedeserialization exceptions. In case of deserialization we might choose to remove the entry from cache as an additional action. not sure about it.
for serialization fail, i am definitely in favor of throwing it regardless of logging it or not.

src/main/java/redis/clients/jedis/csc/AbstractCache.java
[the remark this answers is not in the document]
tishun: Makes sense

src/main/java/redis/clients/jedis/csc/CacheEntry.java
tishun: Would not solve the catch-everything scenario I mentioned.
Anyway, since the JedisCacheException is ultimately a RuntimeException we don't have much of a problem, as most of the users are unlikely going to catch it explicitly.
Let's leave it like this.

PR #3920, opened 2024-08-14
Avoid racing conditions as per design

tishun:


src/main/java/redis/clients/jedis/csc/CacheConnection.java
```diff
@@ -62,30 +63,39 @@ public void disconnect() {
...
     clientSideCache.getStats().miss();
+    cacheEntry = CacheEntry.inProgress(cacheKey, this);
+    clientSideCache.set(cacheKey, cacheEntry);
+
     T value = super.executeCommand(commandObject);
-    if (value != null) {
-      cacheEntry = new CacheEntry<>(cacheKey, value, this);
+    cacheEntry = clientSideCache.get(cacheKey);
+    if (shouldCacheEntry(value, cacheEntry)) {
+      cacheEntry = CacheEntry.newCacheEntry(cacheKey, this, value);
```
sazzad16, line 95: I don't this is properly doing:

> If the correct dummy entry is still there, replace it with the actual value.

The correct dummy entry is the one we used in previous `clientSideCache.set(cacheKey, cacheEntry);` but current `cacheEntry` is not guaranteed to be that one.
Furthermore, it has to be ensured to replace THAT one.

sazzad16 (changes requested):

tishun:
Hey folks, after talking to @atakavci today I believe this is not really needed.

@sazzad16 apologies for wasting your time reviewing this PR.

sazzad16:
@tishun No problem at all.

PR #3923, opened 2024-08-15
20211212 1.momomo 20240815

momomo:


sazzad16:
Thank you @momomo for using Jedis and making the suggestions.

You've reverted and even re-reverted some of your commits. I have found some commit messages are not sufficient enough to clarify the changes in the respective PRs.

It is not possible for us to clarify your intentions. We won't invest time to craft different PRs from your different commits. We won't address/review the commits until those are made into separate PRs focusing specific changes. Thank you for your understanding and considerations.

PR #3926, opened 2024-08-20
Decoding FT.SEARCH reply can be disabled at field level

sazzad16:


uglide (changes requested):
We need to merge other PR first and then update the docs in this PR. See my comment for the suggested update: https://github.com/redis/jedis/pull/3934#discussion_r1728698310

sazzad16:
@uglide We may consider merging this PR as well and create a new example set. See my reply: https://github.com/redis/jedis/pull/3934/files#r1730282809

PR #3934, opened 2024-08-22
[DOC] RediSearch usage with FTCreateParams and FTSearchParams classes

sazzad16:


docs/redisearch.md
```diff
@@ -49,18 +68,37 @@ fields.put("price", 1337);
...
-// creating a complex query
 Query q = new Query("hello world")
         .addFilter(new Query.NumericFilter("price", 0, 1000))
         .limit(0, 5);
 
-// actual search
 SearchResult sr = client.ftSearch("item-index", q);
+```
+
+Alternative searching using FTSearchParams:
```
uglide, line 89: In the field-level decoding PR we will need to change it to:

> Alternatively, you can use FTSearchParams to access more advanced capabilities, such as field-level decoding control. Here is an example:

uglide (approved):

docs/redisearch.md
sazzad16: @uglide IMO, we need a new example. I'm having difficulty making a decent example from title (text), body (text) and price (numeric) fields.

PR #3939, opened 2024-08-28
Support caching null values

atakavci:
Redis server sends invalidation messages for non-defined keys queried by client.
this pr will enable the caching of null results/responses from server side.

### DIFF
--- a/src/test/java/redis/clients/jedis/csc/ClientSideCacheFunctionalityTest.java
+++ b/src/test/java/redis/clients/jedis/csc/ClientSideCacheFunctionalityTest.java
@@ -510,4 +510,30 @@ public class ClientSideCacheFunctionalityTest extends ClientSideCacheTestBase {
     }
   }
 
+  @Test
+  public void testNullValue() throws InterruptedException {
+    int MAX_SIZE = 20;
+    String nonExisting = "non-existing-key";
+    control.del(nonExisting);
+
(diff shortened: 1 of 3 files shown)


src/test/java/redis/clients/jedis/csc/ClientSideCacheFunctionalityTest.java
```diff
@@ -510,4 +510,30 @@ public void run() {
...
+      assertEquals(1, stats.getMissCount());
+
+      val = jedis.get(nonExisting);
+      assertEquals(1, stats.getHitCount());
+      assertEquals(1, stats.getMissCount());
+
+      control.set(nonExisting, "bar");
+      val = jedis.get(nonExisting);
+      assertEquals(1, stats.getHitCount());
+      assertEquals(2, stats.getMissCount());
```
sazzad16, line 535: - assert(s) on cache size
- assert(s) on cache values; i.e. single entry and that entry contains a null value after both of first two GET.

sazzad16 (changes requested):

src/test/java/redis/clients/jedis/csc/ClientSideCacheFunctionalityTest.java
atakavci: done

uglide (approved):
LGTM

sazzad16 (approved):

PR #3944, opened 2024-08-29
DOC-4080 Sorted Set (ZADD and ZRANGE commands) examples

andy-stark-redis:
Examples for ZADD and ZRANGE, plus placeholders for future examples.

### DIFF
--- a/src/test/java/io/redis/examples/CmdsSortedSetExample.java
+++ b/src/test/java/io/redis/examples/CmdsSortedSetExample.java
@@ -4,16 +4,14 @@ package io.redis.examples;
 
 import org.junit.Assert;
 import org.junit.Test;
-import static org.junit.Assert.*;
-
-import java.util.HashMap;
-import java.util.Map;
-import java.util.List;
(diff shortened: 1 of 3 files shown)


src/test/java/io/redis/examples/CmdsSortedSetExample.java
```diff
@@ -0,0 +1,482 @@
+// EXAMPLE: cmds_sorted_set
+// REMOVE_START
+package io.redis.examples;
+
+import org.junit.Assert;
+import org.junit.Test;
+import static org.junit.Assert.*;
```
uglide, line 7: It makes sense only to remove lines like this, which are needed only for internal testing. Other imports like `import java.util.HashMap;` should be presented to the end user so they can copy-paste it and run without fixing any imports.

andy-stark-redis: Fixed.

sazzad16 (approved):

PR #3945, opened 2024-08-30
DOC-4082 code examples for DEL, EXPIRE, and TTL

andy-stark-redis:
Examples for the [DEL](https://redis.io/docs/latest/commands/del/), [EXPIRE](https://redis.io/docs/latest/commands/expire/), and [TTL](https://redis.io/docs/latest/commands/ttl/) command pages.


src/test/java/io/redis/examples/CmdsGenericExample.java
```diff
@@ -0,0 +1,435 @@
...
+
+        // STEP_START expire
+        String expireResult1 = jedis.set("mykey", "Hello");
+        System.out.println(expireResult1);  // >>> OK
+
+        long expireResult2 = jedis.expire("mykey", 10);
+        System.out.println(expireResult2);  // >>> 1
+
+        long expireResult3 = jedis.ttl("mykey");
+        System.out.println(expireResult3);  // >>> 10
```
sazzad16, line 84: time specific

src/test/java/io/redis/examples/CmdsGenericExample.java
```diff
@@ -0,0 +1,435 @@
...
+        System.out.println(expireResult6);  // >>> 0
+
+        long expireResult7 = jedis.ttl("mykey");
+        System.out.println(expireResult7);  // >>> -1
+
+        long expireResult8 = jedis.expire("mykey", 10, ExpiryOption.NX);
+        System.out.println(expireResult8);  // >>> 1
+
+        long expireResult9 = jedis.ttl("mykey");
+        System.out.println(expireResult9);  // >>> 10
```
sazzad16, line 102: time specific

src/test/java/io/redis/examples/CmdsGenericExample.java
```diff
@@ -0,0 +1,435 @@
...
+
+        // STEP_START ttl
+        String ttlResult1 = jedis.set("mykey", "Hello");
+        System.out.println(ttlResult1); // >>> OK
+
+        long ttlResult2 = jedis.expire("mykey", 10);
+        System.out.println(ttlResult2); // >>> 1
+
+        long ttlResult3 = jedis.ttl("mykey");
+        System.out.println(ttlResult3); // >>> 10
```
sazzad16, line 378: time specific

sazzad16 (commented):
Hey @andy-stark-redis, what's our current plan for time specific examples?

src/test/java/io/redis/examples/CmdsGenericExample.java
andy-stark-redis: @sazzad16 Hi. Unless it causes problems for the testing system, I'm happy just to leave them like this (I figure that it will usually take less than half a second to run the tests anyway). There's no "correct" way to document what the user will see but providing some reasonable sample output helps to describe what happens.

sazzad16: Could you use something like below? https://github.com/redis/jedis/blob/4c094726afe782d82fce000660efe78ea967a045/src/test/java/redis/clients/jedis/commands/unified/HashesCommandsTestBase.java#L641

src/test/java/io/redis/examples/CmdsGenericExample.java
```diff
@@ -0,0 +1,434 @@
...
+
+
+        // STEP_START touch
+
+        // STEP_END
+
+        // Tests for 'touch' step.
+        // REMOVE_START
+
+        // REMOVE_END
```
uglide, line 366: @andy-stark-redis Could you explain why you put those placeholders in each PR? 

If you need a way to check which steps are missing for any client, there is a better approach https://redis-stack.github.io/redis-stack-website/examples/

I built a script for the old Redis-stack website a while ago that parses and auto-generates this page as markdown and can help in setting it up for the new website.

andy-stark-redis: @uglide The only reason why the placeholders are there is because I'm using a tool that generates the boilerplate comments (`STEP_START xxx`, etc) for the whole set of examples all at once. The tool was originally designed for the tutorial pages (eg, [strings](https://redis.io/docs/latest/develop/data-types/strings/)) where it makes sense to add all the examples for a doc page together.

The approach we're using for the command pages is to batch related command examples together into one source file (eg, all the [generic group](https://redis.io/docs/latest/commands/?group=generic) commands). However, there are lots of commands, so we're adding them in small batches using a priority order based mainly on page traffic. So, the boilerplate tool adds placeholders for all the examples that will eventually live in the source file, but we only add the actual example code here and there in priority order.

If the placeholders are a problem then I'm happy to remove them and just add the boilerplate manually or modify the tool to do individual examples.

andy-stark-redis: Removed placeholders as requested.

sazzad16 (changes requested):
@andy-stark-redis Please revert back to using `Assert.assert...()`.

sazzad16 (approved):

PR #3946, opened 2024-08-30
DOC-4102 INCR command example

andy-stark-redis:
Currently just [INCR](https://redis.io/docs/latest/commands/incr/) in this file because it's the most popular string command page.


src/test/java/io/redis/examples/Cmds_string.java
```diff
@@ -0,0 +1,48 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+// REMOVE_END
+// HIDE_START
+import redis.clients.jedis.UnifiedJedis;
+// HIDE_END
+
+// HIDE_START
+public class Cmds_string {
```
sazzad16, line 14: 1. The class name must end with `Example`.
2. We avoid putting underscore (`_`) in class name.

sazzad16 (changes requested):

sazzad16 (approved):

PR #3954, opened 2024-09-11
Server version check for CSC activation 

atakavci:


src/main/java/redis/clients/jedis/csc/CacheConnection.java
```diff
@@ -11,18 +14,35 @@
...
       throw new JedisException("Client side caching is only supported with RESP3.");
     }
+    if (!cache.compatibilityMode()) {
+      if (server == null || version == null) {
+        Map<String, Object> helloResult = hello(encode(protocol.version()));
+        if (helloResult != null) {
+          server = (String) helloResult.get("server");
+          version = (String) helloResult.get("version");
+        }
+      }
```
uglide, line 39: This condition is redundant. If Redis doesn't support the HELLO command, the connection will fail earlier in the execution stack and we'll never get there.

src/main/java/redis/clients/jedis/util/Version.java
```diff
@@ -0,0 +1,40 @@
...
+        for (int i = 0; i < max; i++) {
+            int thisNumber = this.numbers.length > i ? this.numbers[i]:0;
+            int otherNumber = other.numbers.length > i ? other.numbers[i]:0;
+            if (thisNumber < otherNumber) return -1;
+            if (thisNumber > otherNumber) return 1;
+        }
+        return 0;
+    }
+
+    public final String get() {
```
sazzad16, line 28: IMO, it should simply `toString()`.

src/main/java/redis/clients/jedis/util/Version.java
```diff
@@ -0,0 +1,40 @@
...
+
+public class Version implements Comparable<Version> {
+
+    private String version;
+    private Integer[] numbers;
+
+    public Version(String version) {
+        if (version == null) throw new IllegalArgumentException("Version can not be null");
+        this.version = version;
+        this.numbers = Arrays.stream(version.split("\\.")).map(n -> Integer.parseInt(n)).toArray(Integer[]::new);
```
sazzad16, line 13: We're doing null checks but there can be many other checks; i.e. valid format, valid values, etc.

src/main/java/redis/clients/jedis/util/Version.java
```diff
@@ -0,0 +1,40 @@
+package redis.clients.jedis.util;
+
+import java.util.Arrays;
+
+public class Version implements Comparable<Version> {
```
sazzad16, line 5: The name is **_generic_** version. `5.2.0-beta5` is a valid Jedis version. This class does not handle that.

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/util/Version.java
atakavci: done

src/main/java/redis/clients/jedis/util/Version.java
atakavci: Intention here is to cover the basically what we need here instead of introducing a whole third party library and add a dependency. May be we can rename it just to emphasize it is not handling various formats and needs. Would you feel more comfortable that way ?

src/main/java/redis/clients/jedis/util/Version.java
atakavci: Right, we might choose to introduce bunch of checks besides corresponding excpetions/messages etc, this one is in the way to take it more pragmatic and put required function in place as i mentioned in the other comment. And have no further intention to grow a generic 'Version' implementation to handle all at once.

src/main/java/redis/clients/jedis/util/Version.java
sazzad16: Yes.

RedisVersion like would be safer.

I think we can even package private this class in `redis.clients.jedis` package.

atakavci: done

src/main/java/redis/clients/jedis/util/RedisVersion.java
```diff
@@ -0,0 +1,41 @@
+package redis.clients.jedis.util;
+
+import java.util.Arrays;
+
+public class RedisVersion implements Comparable<RedisVersion> {
```
sazzad16, line 5: > we can even package private this class in `redis.clients.jedis` package.

WDYT?

atakavci: it is a change of the package of class and unit tests of it (if we dont want to do some fancy code). My experience is that these decisions could easily end up with a crowded and unrelated stuff get into to the same bag, especially with the big projects. I would like to avoid moving things around if i am not %100 sure, like this one. Are u sure on moving it ? 
sorry that it is like i reversed the question somehow to you =B

sazzad16: > My experience is that these decisions could easily end up with a crowded and unrelated stuff get into to the same bag, especially with the big projects.

Agree. To handle that my approach is to avoid adding anything `public` as much as we can.
My experience is also that users find weird ways of using `public` stuff. In some rare cases, we end up unwillingly continuing to support those. That's my another reason for _not-public_.

So,

> Are u sure on moving it ?

Yes.

atakavci: done, moved it to redis.clients.jedis.csc.
still not feeling much convinced on moving things in hierarchy in general to just to make them hidden, anyway i don't take this as a big deal.

sazzad16 (approved):

stefan-zobel:
What is the rationale for this sudden CSC 7.4 min version requirement? I was quite surprised to see that CSC doesn't work by default in the 5.2.0 GA release.

sazzad16:
@stefan-zobel From @mortensi's https://github.com/redis/jedis/discussions/3979#discussioncomment-10796566 

> Redis will soon release support for client-side caching in all its offerings (Redis Software and Redis Cloud), which will be supported only for the 7.4 feature set and later. To avoid inconsistent behavior across Community Edition and the commercial offerings, we decided to streamline the user experience across all the different options. This means developers won't face any incompatibilities while migrating to Redis Software or Redis Cloud. The official announcement will come later.

PR #3970, opened 2024-09-27
Improve QueryRangeExample and QueryGeoExample

uglide:
- Do not rely on random order of unsorted search results
- Print prices for clarity


uglide:
@andy-stark-redis, I improved your example by removing the flakiness of assertions and printing prices.

andy-stark-redis:
> @andy-stark-redis, I improved your example by removing the flakiness of assertions and printing prices.

That's great, thanks! I didn't know about `Stream.of` until now but I'll use that technique for assertions like this in the future.

sazzad16:
@andy-stark-redis The trick is more in `.sorted()`, IMO :smiley:

PR #3971, opened 2024-09-27
Make JSON imports optional in OSGi environments

### LINKED ISSUE #3956: Refine OSGi metadata for JSON dependencies
### Expected behavior

When deploying in OSGi containers the Jedis bundle has two hard requirements on having the GSON and JSON bundles around. With #3223 and #3249 the dependencies were made optional at runtime (to my understanding).

Additionally, due to an unfortunate interaction between the way the org.json:json bundle does versioning and OSGi metadata requirements OSGi consumers *must* use the same version as declared in the pom.xml dependencies.

The error messages look something like this.

#### For missing OSGi bundles

```
[ERROR] [bundle-packages] redis.clients:jedis:5.1.5: Bundle is importing package org.json with start order 24 but no bundle is exporting these for that start order.
[ERROR] [bundle-packages] redis.clients:jedis:5.1.5: Bundle is importing package com.google.gson;version=[2.10,3) with start order 24 but no bundle is exporting these for that start order in the required version range.
```

#### For a version of the org.json:json bundle that is too new

```
[ERROR] [bundle-packages] redis.clients:jedis:5.1.5: Bundle is importing package org.json;version=[20231013.0,20231014) with start order 24 but no bundle is exporting these for that start order in the required version range.
```
 
### Actual behavior

The Jedis bundle should resolve in OSGi containers without the JSON and GSON bundlles for deployers who don't use JSON functionality.

The Jedis bundle should resolve in OSGi container with versions of the JSON bundle more recent than the one used in the Jedis pom.xml.

### Steps to reproduce:

1. Deploy the Jedis bundle in an OSGi container without any JSON libraries

#### Jedis version:

5.1.5

#### Redis version:

N/A, did not get so far.

#### Java version:
openjdk 21.0.4 2024-07-16

rombert:
When deploying in OSGi containers the Jedis bundle has two hard requirements on having the GSON and JSON bundles around. Since these depenendencies are optional we mark the imports as such in the OSGi manifest.

If these packages are not exported the bundle will be installed and will be unable to make use of the JSON functionality. When they are present they will be used.

Additionally, allow newer version of the org.json bundle to be used without requiring a new release of Jedis.

Fixes #3956 .


pom.xml
```diff
@@ -292,6 +292,19 @@
...
 				<version>5.1.9</version>
+				<configuration>
+					<instructions>
+						<!-- 
+							Ensure JSON dependencies are not required at run-time in OSGi environments and also
+							allow usage of newer org.json versions.
+						-->
+						<Import-Package>
+							com.google.gson;resolution:=optional,
+							org.json;resolution:=optional;version="20240303.0",
```
sazzad16, line 303: This indicates a fixed version. Does this mean whenever we update the version of org.json, we'd have to change here as well?

sazzad16 (commented):
@rombert Thank you for the PR. We're checking about this change. It may take some time. In the mean time, I have a question:

pom.xml
rombert: My understanding is that this is an open range - `["20240303.0", Infinity)`. I'll dig up some references to it and post some screenshots of it working with different versions.

rombert: According to https://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#i3189032 

> If a version range is specified as a single version, it must be interpreted as the range [version,∞).

So this means just a minimal version. Right now org.json does not have a version newer that 20240303, but I will update the bundle locally to accept older versions for org.json so I can post a test with exact match and one with a newer version.

rombert: Here's a more explicit test. I modified the Jedis bundle locally to reference a lower import range

```patch
diff --git a/pom.xml b/pom.xml
index f929411d..415273d4 100644
--- a/pom.xml
+++ b/pom.xml
@@ -300,7 +300,7 @@
 						-->
 						<Import-Package>
 							com.google.gson;resolution:=optional,
-							org.json;resolution:=optional;version="20240303.0",
+							org.json;resolution:=optional;version="20220303.0",
 							*
 						</Import-Package>
 					</instructions>

```

I then deployed it in my OSGi app and kept the same (new) 20240303 org.json bundle. In the screenshot you can see that the import is resolved at runtime with the new 20240303.0 version, even though the Import-Package declaration references 20220303.0 .

Effectively all versions newer than 20240303.0 are supported. If you want me I can lower the import range if you think that older versions can be used. Otherwise I'm happy to leave this as it is.

image

sazzad16: Thank you @rombert for your investigation. It'll be helpful for us.

tishun:
I was wondering: if we are able to use the Jedis driver without these two - shouldn't we mark them as optional in the Maven dependency tree too. If we do then the `maven-bundle-plugin` would pick this up and generate the same result in the `MANIFEST.MF`, if I am not mistaken.

rombert:
> I was wondering: if we are able to use the Jedis driver without these two - shouldn't we mark them as optional in the Maven dependency tree too. If we do then the `maven-bundle-plugin` would pick this up and generate the same result in the `MANIFEST.MF`, if I am not mistaken.

Yes, if the dependencies are marked as optional the maven-bundle-plugin should mark the imports as optional. The version pinning for the org.json import, if deemed acceptable, still needs to stay.

sazzad16:
> I was wondering: if we are able to use the Jedis driver without these two - shouldn't we mark them as optional in the Maven dependency tree too. If we do then the `maven-bundle-plugin` would pick this up and generate the same result in the `MANIFEST.MF`, if I am not mistaken.

Yes, but this is a breaking change and has been considered as unfavorable #3278

tishun:
> > I was wondering: if we are able to use the Jedis driver without these two - shouldn't we mark them as optional in the Maven dependency tree too. If we do then the `maven-bundle-plugin` would pick this up and generate the same result in the `MANIFEST.MF`, if I am not mistaken.
> 
> Yes, but this is a breaking change and has been considered as unfavorable #3278

Apologies, was not aware of #3278, makes sense.

In this case, for OSGi specifically, this seems a reasonable improvement. AFAIK most (all?) OSGi containers depend on the user to deploy the artefacts (e.g. by using a WAR file or a nested JAR file or some other way) and the MANIFEST.MF only serves to mandate what should be loaded, so we should not have the same issues as the ones described in #3278

rombert:
> OSGi containers depend on the user to deploy the artefacts (e.g. by using a WAR file or a nested JAR file or some other way) and the MANIFEST.MF only serves to mandate what should be loaded

Yes, that is correct. In the current form of Jedis the two JSON bundles must be deployed, even if they are not used by the application.

sazzad16:
@rombert Apologies. We're still wait for a resolution from the higher-ups.

FYI, making the concerned dependencies totally `optional` in next major release is being considered.

rombert:
@sazzad16 - I see the 6.0 version is now out. Has there been any discussion on the topic of optional dependencies with regards to OSGi deployments?

ggivo:
@rombert
Sorry for the lack of updates — the team got pulled into other priorities, and this one unfortunately slipped through.
I’ll make sure to include it in the next major milestone so it stays visible, and I’ll also bring it up for discussion again soon.

rombert:
Thanks @ggivo , will keep watching this.

ggivo:
Hi @rombert,

Discussed with the team and we decided to review the JSON integration first before proceeding with any changes related to removing the optional dependency.

There are also some other open questions around JSON integration, such as the deprecation of the v1 path, that we want to address.

We’ll plan to move this work to the next release.

rombert:
Thanks for the update @ggivo ,

I did not want to touch it as part of my upgrades, but if there is an option to revisit this and use a different third party library I would suggest using the Jakarta EE JSON Processing API ( https://github.com/jakartaee/jsonp-api ) and letting deployers plug in their own implementation.

PR #3972, opened 2024-09-27
Fix flaky tests with 'await'

atakavci:


sazzad16 (commented):
3 extra commands seems many. Isn't 1 enough?

atakavci:
@sazzad16  i agree with you it should have been ok with 1. Tried with 1 in the past which still get fails intermittently. Just want to get rid of this flakiness once and for all. 
TBH looking for another approach (not a Thread.sleep) but couldnt come up with a clean one.

sazzad16:
> Tried with 1 in the past which still get fails intermittently.

@atakavci In that case, I'd take the Thread.sleep(); at least those would be easier to document and understand :sweat:

atakavci:
i m strongly against to use of sleep in unit tests. Since its not sensitive and responsive to the environment like resource utilization, latencies etc. there will be always an unpredictable requirement for how long we need to wait. 
Is it good enough to add some comments and signal the significance of these dummy operation to the future maintainer ?

sazzad16:
> there will be always an unpredictable requirement for how long we need to wait.

Well, same thing can be said about the number of commands you're executing. It's possible that if we run the tests way too many times, we would find the flaky tests are failing which would've passed if 4 or 5 commands were executed instead of 3.

> i'm strongly against to use of sleep in unit tests.

I respect that. But in my humble understanding, by executing dummy commands you're just doing Thread.sleep() without writing Thread.sleep(). By executing 3 commands, you're just using a larger sleep time compared to the equivalent time of executing 1 command.

sazzad16:
@atakavci 

> there will be always an unpredictable requirement for how long we need to wait.

Well, same thing can be said about the number of commands you're executing. It's possible that if we run the tests way too many times, we would find the flaky tests are failing which would've passed if 4 or 5 commands were executed instead of 3.

> i'm strongly against to use of sleep in unit tests.

I respect that. But in my humble understanding, by executing dummy commands you're just doing Thread.sleep() without writing Thread.sleep(). By executing 3 commands, you're just using a larger sleep time compared to the equivalent time of executing 1 command.

> ... to the future maintainer

This is what I indicated in my last comment. A line of of Thread.sleep() + comments would be much easier to understand than 3 lines of dummy command + comments.

atakavci:
@sazzad16 i see your point, let me try to explain mine;

> Well, same thing can be said about the number of commands you're executing.

this is not correct. When you execute a similar task, you will get same kind of delay and they will always be proportional in average since it uses the same resources and impacted by same set of factors.
so when you do a 'read X' and then a 'read Y' you will get the same ratio of execution times in average.

> you're just doing Thread.sleep() without writing Thread.sleep(). 

right in a sense, purpose is same but,, but `sleep` is utilization agnostic. It doesn't care if the execution is faster or slower then usual/expected. If we put 1000ms, it will be 1000ms always(And and its not even guaranteed that thread will be back exactly after 1000ms, furthermore consider we start to spread these `sleep`'s here and there in tests, pipeline will start to degrade inevitably).
so, `thread.sleep` will not remain proportional to system throughput and speed as i described above,, yet command execution will.

i appricaite your stance with readability and really would like to improve it,,, without using `sleep` stuff.

killergerbah:
Hello, I am seeing some of these tests fail in my own fork of `jedis`. In my case the `simple` test is failing.

Can this be resolved by implementing an assertion method that evaluates eventual consistency? For example, repeatedly querying the key with some small `sleeps` and a very large total timeout (e.g. 30 seconds). This could look something like:

```
assertEventuallyTrue(() -> cache.getSize() == 1)
```

atakavci:
hi @killergerbah, trying to figure out best options and i d like to have a way of addressing it without having to wait a fixed long period of time in case of a failure.

tishun:
I am also not a fan of doing either Thread.sleep or sending (arbitrary) number of commands.
Ideally we should be able to react to the environment change, based on some server push, but when we are testing the server push we are left with two choices:
a) the solution that @killergerbah suggested, accepting the shortcoming that it might take us some time to fail (but this is ok as we should mostly NOT fail and would only slow down the tests in case of a regression)
b) subscribe to another alternative mechanism, e.g. [Redis keyspace notifications](https://redis.io/docs/latest/develop/use/keyspace-notifications/) where this would work

What do you think?

atakavci:
> subscribe to another alternative mechanism, e.g. [Redis keyspace notifications](https://redis.io/docs/latest/develop/use/keyspace-notifications/) where this would work
 
 this requires another connection which leaves the test case possibly flaky again. If it is the original connection to subscribe, then it is no different then making a 'write' on same connection, which will block on a response read.
 
first choice remains feasible while it is just a 'sleep' in the heart of it.

there are options like tapping the port for incoming data or pending buffer, but wouldn't it be the perfect example of over engineering.
i believe having pipeline healthy again is getting more important than this topic. let me put a sleep cycle and we ll move fwd.

src/test/java/redis/clients/jedis/csc/UnifiedJedisClientSideCacheTestBase.java
```diff
@@ -94,7 +88,7 @@ public void cacheNotEmptyTest() {
       Cache cache = jedis.getCache();
       control.set("foo", "bar");
       assertEquals(0, cache.getSize());
-      assertEquals("bar", jedis.get("foo"));
+      AssertUtil.tryAssert(() -> jedis.get("foo"), "bar", 50, 100);
```
sazzad16, line 91: Don't need here. This is the first command with the `jedis` client object. So it's always a cache miss -> a server call -> reply.

src/test/java/redis/clients/jedis/util/AssertUtil.java
```diff
@@ -149,4 +149,29 @@ private static void assertPipelineSyncAllSet(Set<?> expected, Set<?> actual) {
       assertEquals(expected, actual);
     }
   }
+
+  public static <T> void tryAssert(Callable<T> callable, T expected, long interval, int iterations) {
```
sazzad16, line 153: interval --> intervalMillis (clarifying)
iterations --> maxAttempts (aligning with error message below)

src/test/java/redis/clients/jedis/csc/UnifiedJedisClientSideCacheTestBase.java
```diff
@@ -2,16 +2,14 @@
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotSame;
-import static org.junit.Assert.assertNull;
-
```
sazzad16, line 6: nit: keep the empty line.

ggivo:
What do you think about using an already available lib for this? There is a nice one that integrates with assert, hamcrest.. like    https://github.com/awaitility/awaitility

tishun:
> What do you think about using an already available lib for this? There is a nice one that integrates with assert, hamcrest.. like https://github.com/awaitility/awaitility

Vote for that. It would only increase the dependencies in the test phase, so it does not have impact on the driver itself.

src/test/java/redis/clients/jedis/csc/UnifiedJedisClientSideCacheTestBase.java
```diff
@@ -39,7 +40,8 @@ public void simple() {
       control.set("foo", "bar");
       assertEquals("bar", jedis.get("foo"));
       control.del("foo");
-      assertNull(jedis.get("foo"));
+      await().atMost(5, TimeUnit.SECONDS).pollInterval(50, TimeUnit.MILLISECONDS)
+          .until(() -> jedis.get("foo") == null);
```
sazzad16, line 44: In Awaitility, it has `untilAsserted` method which is capable of doing assertions which is better than checking equality.

src/test/java/redis/clients/jedis/PipeliningTest.java
```diff
@@ -342,7 +344,8 @@ public void waitAof() {
 
     try (Jedis j = new Jedis(endpoint.getHostAndPort())) {
       j.auth(endpoint.getPassword());
-      assertEquals("aof", j.get("wait"));
+      Awaitility.await().atMost(5, TimeUnit.SECONDS).pollInterval(50, TimeUnit.MILLISECONDS)
+          .untilAsserted(() -> assertEquals("aof", j.get("wait")));
```
sazzad16, line 348: :heart:

sazzad16 (dismissed):

src/test/java/redis/clients/jedis/PipeliningTest.java
tishun: Love it. Got to add it to Lettuce tests too at some point!

tishun (approved):
I think this is the best solution for now.

PR #3980, opened 2024-09-30
Add SslOptions

sazzad16:
Similar to SslOptions in Lettuce library.


src/main/java/redis/clients/jedis/SslOptions.java
```diff
@@ -0,0 +1,433 @@
...
+
+            keyManagerFactory = KeyManagerFactory.getInstance(keyManagerAlgorithm);
+            keyManagerFactory.init(keyStore, keystorePassword);
+
+        }
+
+        final TrustManagerFactory trustManagerFactory;
+        if (insecureTrust) {
+
+            trustManagerFactory = InsecureTrustManagerFactory.INSTANCE;
```
atakavci, line 349: why do we need an insecure trust manager at all ?

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
```diff
@@ -122,6 +112,57 @@ public Socket createSocket() throws JedisConnectionException {
...
+
+      _sslParameters = sslParameters != null ? sslParameters : new SSLParameters(); // TODO:
+
+      if (sslHostnameVerifyMode == SslHostnameVerifyMode.HTTPS) {
+        _sslParameters.setEndpointIdentificationAlgorithm("HTTPS");
+      } else if (sslHostnameVerifyMode == SslHostnameVerifyMode.DISABLE) {
+        _sslParameters.setEndpointIdentificationAlgorithm("");
+      }
+
+      _sslSocketFactory = _sslContext.getSocketFactory();
```
atakavci, line 137: nit: would put it close the context declaration, or even more make it single line

atakavci (commented):
having a trust manager like this smells off. May be its because i am lack of context. What is the rationale ?

src/main/java/redis/clients/jedis/DefaultJedisClientConfig.java
```diff
@@ -284,6 +329,7 @@ public static DefaultJedisClientConfig create(int connectionTimeoutMillis, int s
         false);
   }
 
+  // TODO:
   public static DefaultJedisClientConfig copyConfig(JedisClientConfig copy) {
     return new DefaultJedisClientConfig(copy.getRedisProtocol(),
         copy.getConnectionTimeoutMillis(), copy.getSocketTimeoutMillis(),
```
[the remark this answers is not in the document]
sazzad16: @atakavci It depends on how this PR becomes at the end. Note that this PR is in draft mode. However I've added TODO comment.

src/main/java/redis/clients/jedis/DefaultJedisClientConfig.java
```diff
@@ -273,6 +317,7 @@ public Builder readOnlyForRedisClusterReplicas() {
     }
   }
 
+  // TOOD:
```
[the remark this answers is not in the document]
sazzad16: added comment

src/main/java/redis/clients/jedis/SslHostnameVerifyMode.java
```diff
@@ -0,0 +1,8 @@
+package redis.clients.jedis;
+
+public enum SslHostnameVerifyMode {
```
tishun, line 3: This is the equivalent of the SslVerifyMode in Lettuce, correct?

tishun:
> having a trust manager like this smells off. May be its because i am lack of context. What is the rationale ?

I have the same question. The motivation of the change is important in this case.

uglide:
@tishun @atakavci This is an attempt to create a TLS config similar to what we have in Lettuce, as a more developer-friendly option compared to the current approach with SSLSocketFactory, etc.

src/main/java/redis/clients/jedis/SslOptions.java
sazzad16: @uglide Yes.

sazzad16: @uglide Correct.

src/main/java/redis/clients/jedis/SslHostnameVerifyMode.java
sazzad16: This is exactly SslVerifyMode in Lettuce except the NONE option.
All the remaining options are specific to hostname verification.
So I named this as such.

sazzad16:
> > having a trust manager like this smells off. May be its because i am lack of context. What is the rationale ?
> 
> I have the same question. The motivation of the change is important in this case.

@tishun 
1. Lettuce has similar support.
2. Users want _insecurity_.

tishun (commented):
I may not be getting the whole idea, so please excuse me if my comments are not making sense.

My current assumption is that we _only_ want to be able to enable and disable hostname verification, as part of the SSL verification process.

src/main/java/redis/clients/jedis/SslOptions.java
sazzad16: @tishun We have same features of NONE but the control is separated from SslVerifyMode.

src/main/java/redis/clients/jedis/util/InsecureTrustManagerFactory.java
ggivo, line 1: I believe all we need from this class is
```
TrustManager[] trustAllCerts = new TrustManager[]{
new X509TrustManager() {
        @Override
        public void checkClientTrusted(X509Certificate[] chain, String s) {
            if (logger.isDebugEnabled()) {
                logger.debug("Accepting a client certificate: " + chain[0].getSubjectDN());
            }
        }

        @Override
        public void checkServerTrusted(X509Certificate[] chain, String s) {
            if (logger.isDebugEnabled()) {
                logger.debug("Accepting a server certificate: " + chain[0].getSubjectDN());
            }
        }

        @Override
        public X509Certificate[] getAcceptedIssuers() {
            return EMPTY_X509_CERTIFICATES;
        }
    };
```

src/main/java/redis/clients/jedis/SslHostnameVerifyMode.java
tishun: I understand. The `setEndpointIdentificationAlgorithm` uses [standard names](https://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html) from the Java specification, but I think they are generic and in this case could be misleading if used as part of the API of the driver. A friendly name could describe better (I do not insist on the same naming as in Lettuce, although similar names could make it easier for a app developers to switch between them).

We could always leave a note in the public JavaDoc that this specific enum value would result in using the `HTTPS` Endpoint Identification Algorithm (for verbosity).

src/main/java/redis/clients/jedis/SslOptions.java
tishun: Oh, I see, you aimed to separate them in two different configurations - one turns the verification on and off an another (when the config is on) would turn on and off the host verification.

src/main/java/redis/clients/jedis/util/InsecureTrustManagerFactory.java
sazzad16: @ggivo The `X509TrustManager` is wrapped to make it `X509ExtendedTrustManager`.

src/main/java/redis/clients/jedis/SslOptions.java
```diff
@@ -0,0 +1,502 @@
...
+                trustStore.load(truststoreStream, truststorePassword);
+            }
+
+            TrustManagerFactory trustManagerFactory = TrustManagerFactory.getInstance(trustManagerAlgorithm);
+            trustManagerFactory.init(trustStore);
+            trustManagers = trustManagerFactory.getTrustManagers();
+        }
+
+        SSLContext sslContext = SSLContext.getInstance(sslContextProtocol);
+        sslContext.init(keyManagers, trustManagers, null);
```
sazzad16, line 374: About `null` SecureRandom: https://stackoverflow.com/questions/295628/securerandom-init-once-or-every-time-it-is-needed

src/main/java/redis/clients/jedis/SslOptions.java
```diff
@@ -0,0 +1,481 @@
...
+            }
+
+            KeyManagerFactory keyManagerFactory = KeyManagerFactory.getInstance(keyManagerAlgorithm);
+            keyManagerFactory.init(keyStore, keystorePassword);
+            keyManagers = keyManagerFactory.getKeyManagers();
+        }
+
+        if (trustManagers != null) {
+            // already processed
+        } else if (truststoreResource != null) {
```
tishun, line 344: I'd simplify this, because it is a bit hard to track to:

```java
if (trustManagers == null && truststoreResource != null) {
```

src/main/java/redis/clients/jedis/SslVerifyMode.java
```diff
@@ -0,0 +1,22 @@
+package redis.clients.jedis;
+
+/**
+ * Enumeration of SSL/TLS hostname verification modes.
+ */
+public enum SslVerifyMode {
+
+    /**
+     * No verification at all.
```
tishun, line 9: I would love if we put a clear indications that this should never be used in production.
I know it seems obvious, but I want to make sure it is written down.

Eg.

```suggestion
     * No verification at all. Avoid using this in production environments, poses serious security risks.
```

tishun (approved):
Awesome, LGTM!

tishun (dismissed):

PR #2495, opened 2021-04-02
Automate code formatting using Maven (2019 rules)

walles:
Same as #2494 but with HBase rules from 2019 (the other PR uses the existing 2014 ruleset).

# Before this change

`CONTRIBUTING.md` encouraged reformatting sources using `make format`.

`make format` would reformat lots of files, not just the ones you changed.

# With this change in place

Any `mvn` operation will implicitly format sources.

[CI will validate code formatting](https://app.circleci.com/pipelines/github/redis/jedis/700/workflows/d76074e9-77d6-4b0c-9d9c-5508b7225fa4/jobs/850).

This implicitly means that people have to at least `mvn compile` their sources locally before making PRs to make sure they don't get refused by CI.

This enforces good practice, regardless of the actual formatting.

# Notes

Files to be reviewed (the rest have been auto formatted):
* `pom.xml`
* `Makefile`
* `.github/CONTRIBUTING.md`

Also, [checking the commits list](https://github.com/redis/jedis/pull/2494/commits) might be easier than looking at files changed, all reformatting is confined to the last change.

The code formatting Maven plugin being used didn't have any `validate` goal so I replaced with another one.

Still running the Eclipse formatter though, so `hbase-formatter.xml` still works.

`hbase-formatter.xml` has been updated to [a newer one from 2019](https://github.com/apache/hbase/blob/master/dev-support/hbase_eclipse_formatter.xml).

PR #2494, opened 2021-04-02
Automate code formatting using Maven

walles:
# Before this change

`CONTRIBUTING.md` encouraged reformatting sources using `make format`.

`make format` would reformat lots of files, not just the ones you changed.

# With this change in place

Any `mvn` operation will implicitly format sources.

[CI will validate code formatting](https://app.circleci.com/pipelines/github/redis/jedis/700/workflows/d76074e9-77d6-4b0c-9d9c-5508b7225fa4/jobs/850).

This implicitly means that people have to at least `mvn compile` their sources locally before making PRs to make sure they don't get refused by CI.

This enforces good practice, regardless of the actual formatting.

# Notes

Files to be reviewed (the rest have been auto formatted):
* `pom.xml`
* `Makefile`
* `.github/CONTRIBUTING.md`

Also, [checking the commits list](https://github.com/redis/jedis/pull/2494/commits) might be easier than looking at files changed, all reformatting is confined to the last change.

The code formatting Maven plugin being used didn't have any `validate` goal so I replaced with another one.

Still running the Eclipse formatter though, so `hbase-formatter.xml` still works.

`hbase-formatter.xml` is from 2014. It could potentially be replaced with [a newer one from 2019](https://github.com/apache/hbase/blob/master/dev-support/hbase_eclipse_formatter.xml), I haven't done that though.

PR #3995, opened 2024-10-15
format by hbase-formatter

atakavci:
Formatted with hbase-formatter settings, 
format issue keeps showing up with lots of files. i ll try to fix them gradually.

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisPooled.java
+++ b/src/main/java/redis/clients/jedis/JedisPooled.java
@@ -66,13 +66,14 @@ public class JedisPooled extends UnifiedJedis {
   public JedisPooled(final String host, final int port, final boolean ssl,
       final SSLSocketFactory sslSocketFactory, final SSLParameters sslParameters,
       final HostnameVerifier hostnameVerifier) {
-    this(new HostAndPort(host, port), DefaultJedisClientConfig.builder().ssl(ssl)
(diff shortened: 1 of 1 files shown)


tishun:
We may want to apply a method, similar to the one used in Lettuce, that verifies the formatting (as part of the CI pipeline) and allows to format files with a mvn command (e.g. `mvn formatter:format`)


What do you think @sazzad16 and @atakavci?

sazzad16:
We had #2495 and #2494. Could be worth checking.

IMO, we should consider increasing line-width to at least 100 (could be even 120).

atakavci:
agree with automation suggestions. disagree with formatting it all at one shot now.
we should first fix format issues in a planned manner, then add some linter/formatter to pipeline. 

> IMO, we should consider increasing line-width to at least 100 (could be even 120).

line width is already 100 in hbase-formatter and count me in for something like 120/140 (120 is common i believe) 
`<setting id="org.eclipse.jdt.core.formatter.lineSplit" value="100"/>
`

ggivo:
@atakavci 
I think if we merge  https://github.com/redis/jedis/pull/4087, it will supersede this one.
At least we need to make sure we use the same code format rules.
Do we still need this one?

atakavci:
@ggivo closing this one in favor of #4087 
lets keep in mind the suggestion from @sazzad16  around line width. 
80 might be reasonable for a lot of reasons in the past, but not anymore.

PR #3996, opened 2024-10-15
Codecove has released beta version of Test Analytics feature

ggivo:
Codecove has released beta version of Test Analytics feature
more details can be found [here](https://docs.codecov.com/docs/test-result-ingestion-beta)

Since we are already integrated with Codecov (storing code coverage data) the idea is to try out if Test Analytics feature will provide visibility over flaky tests and stability of the tests/build. Hopefully, we will have a base to compare and identify regressions.

Changes (only in build actions)
    - Add step to publish junit report files to codecov
    - Add basic codecov.yml config (filter out files that are not junit reports)


sazzad16 (approved):

.github/workflows/integration.yml
```diff
@@ -60,3 +60,10 @@ jobs:
         with:
           fail_ci_if_error: false
           token: ${{ secrets.CODECOV_TOKEN }}
+      - name: Upload test results to Codecov
+        if: ${{ github.event_name == 'schedule' || (github.event_name == 'push') || github.event_name == 'workflow_dispatch'}}
+        uses: codecov/test-results-action@v1
+        with:
+          fail_ci_if_error: false
+          files: ./target/surefire-reports/TEST*
```
tishun, line 68: Folks, sorry for the late moment but aren't we going to miss the failsafe-reports this way?

```yaml
./target/surefire-reports/TEST*
./target/failsafe-reports/TEST*
```

ggivo: @tishun Do we use the failsafe plugin in Jedis? I might have missed it.

PR #3998, opened 2024-10-17
Initialize an mkdocs-baseed Jedis doc site.

joshrotenberg:
This PR unifies the current content in `docs` and the [Wiki](https://github.com/redis/jedis/wiki) into a single, GitHub pages hosted Jedis documentation portal. So far it's a minimal effort combo of the two plus a few extra links, with the intent that it can continue to grow into a single point of reference for all things Jedis. For the most part it mirrors the [lettuce](https://redis.github.io/lettuce/) implementation in spirit.

If merged, `Pages` will also need to be enabled for the repo. Go to `Settings`, `Pages` in the left hand navigation, and select `GitHub Actions` in the `Build and deployment` dropdown. 

My fork has an up to date version running here as an example: https://joshrotenberg.com/jedis/

PR #3999, opened 2024-10-22
Mkdocs unify docs

uglide:
Continuation of https://github.com/redis/jedis/pull/3998


.github/workflows/docs.yml
```diff
@@ -0,0 +1,36 @@
+name: Publish Docs
+on:
+  push:
+    branches: ["master", "mkdocs-unify-docs"]
```
sazzad16, line 4: Can we remove the (temporary?) branch name after testing is successful?
```suggestion
    branches: ["master"]
```

sazzad16 (changes requested):

.github/workflows/docs.yml
joshrotenberg: Yeah, should be safe to remove that. I added it to build on my fork for the demo.

sazzad16 (approved):

PR #4011, opened 2024-11-06
Token based authentication integration with core extension 

atakavci:
This PR Provides the required changes to integrate the token based authentication using the `core` module from [redis-authx](https://github.com/redis/tbd-auth-entraid).

[AuthXManager](https://github.com/atakavci/jedis/blob/ali/authx2/src/main/java/redis/clients/jedis/authentication/AuthXManager.java) is the main component which handles the id provider creation and management as well as glueing the renewal process to jedis clients internal connection management.

An instance of `AuthXManager` should be provided to jedis simply via JedisClientConfiguration. 
The configuration and creation of AuthXManager can be easily achieved through identity provider-specific configuration builders.
MS EntraID identity provider is already provided within [redis-authx](https://github.com/redis/tbd-auth-entraid).

A basic usage would look like;

```java
    TokenAuthConfig tokenAuthConfig = EntraIDTokenAuthConfigBuilder.builder()
        .clientId("YOUR_CLIENT_ID").secret("YOUR_SECRET")
        .authority("YOUR_AUTHORITY").scopes("SCOPES").build();

    DefaultJedisClientConfig jedisConfig = DefaultJedisClientConfig.builder()
        .authXManager(new AuthXManager(tokenAuthConfig)).build();

    try (JedisPooled jedis = new JedisPooled(hnp, jedisConfig)) {
      ...
      ...
    }
```


pom.xml
```diff
@@ -93,6 +93,12 @@
 			<scope>test</scope>
 		</dependency>
 
+		<dependency>
```
vladvildanov, line 96: Didn't we agreed that client won't be dependent on the actual auth library? It's up to user to add a dependency if he needs, and provide an implementation that we offer to client instance?

pom.xml
```diff
@@ -9,7 +9,7 @@
 	<packaging>jar</packaging>
 	<groupId>redis.clients</groupId>
 	<artifactId>jedis</artifactId>
-	<version>5.3.0-SNAPSHOT</version>
+	<version>5.3.1-SNAPSHOT</version>
```
sazzad16, line 12: Is there any reason for this change?

sazzad16 (commented):
It would easier to review if you could avoid formatting-only changes.

src/main/java/redis/clients/jedis/authentication/TokenCredentials.java
```diff
@@ -0,0 +1,24 @@
+package redis.clients.jedis.authentication;
+
+import redis.clients.authentication.core.Token;
+import redis.clients.jedis.RedisCredentials;
+
+public class TokenCredentials implements RedisCredentials {
```
sazzad16, line 6: It seems, this can be a private class.

src/test/java/redis/clients/jedis/authentication/TokenBasedAuthenticationUnitTests.java
```diff
@@ -0,0 +1,337 @@
...
+import static org.junit.Assert.assertThrows;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.mockConstruction;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
```
sazzad16, line 14:
```suggestion
import static org.mockito.Mockito.*;
```

src/test/java/redis/clients/jedis/authentication/TokenBasedAuthenticationUnitTests.java
```diff
@@ -0,0 +1,337 @@
...
+
+import redis.clients.authentication.core.IdentityProvider;
+import redis.clients.authentication.core.IdentityProviderConfig;
+import redis.clients.authentication.core.SimpleToken;
+import redis.clients.authentication.core.Token;
+import redis.clients.authentication.core.TokenAuthConfig;
+import redis.clients.authentication.core.TokenListener;
+import redis.clients.authentication.core.TokenManager;
+import redis.clients.authentication.core.TokenManagerConfig;
+import redis.clients.authentication.core.TokenRequestException;
```
sazzad16, line 40:
```suggestion
import redis.clients.authentication.core.*;
```

src/test/java/redis/clients/jedis/authentication/RedisEntraIDIntegrationTests.java
```diff
@@ -0,0 +1,404 @@
...
+import redis.clients.jedis.DefaultJedisClientConfig;
+import redis.clients.jedis.EndpointConfig;
+import redis.clients.jedis.HostAndPort;
+import redis.clients.jedis.HostAndPorts;
+import redis.clients.jedis.JedisPooled;
+import redis.clients.jedis.exceptions.JedisAccessControlException;
+import redis.clients.jedis.exceptions.JedisConnectionException;
+import redis.clients.jedis.scenario.FaultInjectionClient;
+
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
```
sazzad16, line 60: If NAME_ASCENDING is kept, the tests should also in that for better readability/understanding.

sazzad16 (changes requested):

src/main/java/redis/clients/jedis/ConnectionFactory.java
```diff
@@ -76,13 +116,26 @@ public PooledObject<Connection> makeObject() throws Exception {
...
   public boolean validateObject(PooledObject<Connection> pooledConnection) {
     final Connection jedis = pooledConnection.getObject();
     try {
       // check HostAndPort ??
+      try {
+        jedis.reAuth();
+      } catch (Exception e) {
+        authenticationEventListener.onConnectionAuthenticationError(e);
+        throw e;
+      }
```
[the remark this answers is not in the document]
atakavci: `validateobject` works in evict operations which we use to leverage for reauth operations on idle connections

src/test/java/redis/clients/jedis/authentication/TokenBasedAuthenticationUnitTests.java
atakavci: `*` considered to be a not good practice and personally i prefer to be able to see where each class comes from without help of IDE.

src/test/java/redis/clients/jedis/authentication/RedisEntraIDIntegrationTests.java
atakavci: i dont see how it helps readability to reorder them ? 
the reason i have that is to avoid parallel run.

src/main/java/redis/clients/jedis/authentication/TokenCredentials.java
```diff
@@ -0,0 +1,24 @@
...
+
+import redis.clients.authentication.core.Token;
+import redis.clients.jedis.RedisCredentials;
+
+class TokenCredentials implements RedisCredentials {
+    private final String user;
+    private final char[] password;
+
+    public TokenCredentials(Token token) {
+        user = token.tryGet("oid");
```
ggivo, line 11: Should this be moved to the core library?
e.g. 
Token.getAuthority() to abstract in case if any future supported IdentyProvider provides the username in different attribute?

src/test/java/redis/clients/jedis/authentication/TokenBasedAuthenticationUnitTests.java
sazzad16: > `*` considered to be a not good practice

Usually yes. But I usually think the opposite when I see 9+ imports from same package.

src/main/java/redis/clients/jedis/authentication/TokenCredentials.java
atakavci: i am on it

atakavci:
using this multiple times , does it causes multiple instance of a consumer in java ? 
anotherObject.thenConsume(this::pingHandler);

src/main/java/redis/clients/jedis/JedisSafeAuthenticator.java
```diff
@@ -0,0 +1,104 @@
...
+import org.slf4j.LoggerFactory;
+
+import redis.clients.authentication.core.SimpleToken;
+import redis.clients.authentication.core.Token;
+import redis.clients.jedis.Protocol.Command;
+import redis.clients.jedis.authentication.JedisAuthenticationException;
+import redis.clients.jedis.exceptions.JedisException;
+import redis.clients.jedis.util.SafeEncoder;
+
+public class JedisSafeAuthenticator {
```
sazzad16, line 19: Do we have any `public` use case for this?

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -103,6 +103,9 @@ public JedisClusterInfoCache(final JedisClientConfig clientConfig, Cache clientS
     this.clientConfig = clientConfig;
     this.clientSideCache = clientSideCache;
     this.startNodes = startNodes;
+    if (clientConfig.getAuthXManager() != null) {
+      clientConfig.getAuthXManager().start();
+    }
```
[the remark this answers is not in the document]
atakavci: With current approach we leverage pools API and connection factory to integrate authxmanager into jedis. Cluster implementation is an exception where individual connections are created in.

src/main/java/redis/clients/jedis/JedisSafeAuthenticator.java
atakavci: i dont have one

atakavci: changed as protected

pom.xml
```diff
@@ -75,6 +75,12 @@
 			<version>2.11.0</version>
 		</dependency>
 
+		<dependency>
+			<groupId>redis.clients.authentication</groupId>
+			<artifactId>redis-authx-core</artifactId>
+			<version>0.1.0-SNAPSHOT</version>
```
tishun, line 81: Just a reminder in case we want to bump to latest

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -44,6 +46,8 @@ public class Connection implements Closeable {
   private String strVal;
   protected String server;
   protected String version;
+  private AtomicReference<RedisCredentials> currentCredentials = new AtomicReference<>(null);
```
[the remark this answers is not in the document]
atakavci: unfortunately it's not. atomic ref here helps to manage producer consumer problem, AUTH with last credentials that connection received and use them only once without need any other synchronization construct.

src/main/java/redis/clients/jedis/JedisPubSubBase.java
```diff
@@ -69,11 +76,23 @@ public final void punsubscribe(T... patterns) {
   }
 
   public final void ping() {
-    sendAndFlushCommand(Command.PING);
+    authenticator.commandSync.lock();
```
tishun, line 79: IMHO it would be good to mention why we need to lock here.

As far as I can tell this is because the AUTH command is async and we want to be sure we dont get a concurrency problem when the user and auth threads attempt to write in the same time?

tishun (dismissed):
LGTM

src/test/java/redis/clients/jedis/authentication/TokenBasedAuthenticationUnitTests.java
```diff
@@ -0,0 +1,342 @@
...
+import redis.clients.authentication.core.TokenManager;
+import redis.clients.authentication.core.TokenManagerConfig;
+import redis.clients.authentication.core.TokenManagerConfig.RetryPolicy;
+import redis.clients.jedis.ConnectionPool;
+import redis.clients.jedis.EndpointConfig;
+import redis.clients.jedis.HostAndPort;
+
+public class TokenBasedAuthenticationUnitTests {
+
+  private HostAndPort hnp = new HostAndPort("localhost", 6379);
```
uglide, line 41: You should use the `standalone0` endpoint instead of hardcoding the connection settings here.

src/test/java/redis/clients/jedis/authentication/TokenBasedAuthenticationClusterIntegrationTests.java
```diff
@@ -0,0 +1,131 @@
...
+                                System.currentTimeMillis() + 5 * 1000, System.currentTimeMillis(),
+                                null);
+                    }
+                };
+            }
+        };
+        AuthXManager manager = new AuthXManager(EntraIDTokenAuthConfigBuilder.builder()
+                .lowerRefreshBoundMillis(1000).identityProviderConfig(idpConfig).build());
+
+        HostAndPort hp = HostAndPorts.getClusterServers().get(0);
```
uglide, line 58: You should use here `cluster-entraid-acl` endpoint

src/test/java/redis/clients/jedis/authentication/TokenBasedAuthenticationClusterIntegrationTests.java
```diff
@@ -0,0 +1,131 @@
...
+        List<Connection> connections = new ArrayList<>();
+        doAnswer(invocation -> {
+            Connection connection = spy((Connection) invocation.getArgument(0));
+            invocation.getArguments()[0] = connection;
+            connections.add(connection);
+            Object result = invocation.callRealMethod();
+            return result;
+        }).when(authXManager).addConnection(any(Connection.class));
+
+        HostAndPort hp = HostAndPorts.getClusterServers().get(0);
```
uglide, line 101: Same as above.

src/test/java/redis/clients/jedis/authentication/TokenBasedAuthenticationIntegrationTests.java
```diff
@@ -0,0 +1,256 @@
...
+
+  private static EndpointConfig endpointConfig;
+
+  @BeforeClass
+  public static void before() {
+    try {
+      endpointConfig = HostAndPorts.getRedisEndpoint("standalone0");
+    } catch (IllegalArgumentException e) {
+      try {
+        endpointConfig = HostAndPorts.getRedisEndpoint("standalone");
```
uglide, line 58: I'll fix the endpoint name; there's no need for workarounds here.

src/main/java/redis/clients/jedis/JedisPubSubBase.java
atakavci: thats correct , i will provide explanation in another iteration

uglide (changes requested):
Endpoints in tests should be fixed.

src/test/java/redis/clients/jedis/authentication/TokenBasedAuthenticationClusterIntegrationTests.java
atakavci: as long as this endpoint answers to cluster commands it would be fine. Right now the tests can run without  entraid enabled endpoints and make the necessary assertions on it. It would be better to keep them running on each commit.

src/test/java/redis/clients/jedis/authentication/TokenBasedAuthenticationClusterIntegrationTests.java
atakavci: same case with other one

src/test/java/redis/clients/jedis/authentication/TokenBasedAuthenticationIntegrationTests.java
atakavci: so we will use "standalone0" instead of "standalone", right ?

src/test/java/redis/clients/jedis/authentication/TokenBasedAuthenticationClusterIntegrationTests.java
uglide: In the integration test, we do not bring up the endpoints hardcoded in the cluster config, so we need to switch to the `cluster-entraid-acl` to be able to run the tests. Let me know if I'm missing something here.

atakavci: yes , it doesnt run with entraid enabled env but runs within every integration test in jedis repo

uglide (approved):

src/test/java/redis/clients/jedis/authentication/TokenBasedAuthenticationIntegrationTests.java
uglide: yes

PR #4015, opened 2024-11-10
Introduces test matrix based on Redis versions [8.0-M1, 7.4.1, 7.2.6, 6.2.16]

### LINKED ISSUE #4016: NPE thrown on jedis.aclLog() used against Redis server 7.2.0<
New fields are added in Redis version 7.2.0 (`entry ID`, `timestamp created`, and `timestamp last updated`).
When using Jedis client against older server version where those fields are missing NPE is thrown
`java.lang.NullPointerException: Cannot invoke "java.lang.Long.longValue()" because the return value of "java.util.Map.get(Object)" is null`

### Expected behavior
Do not throw NPE.
Since `AccessControlLogEntry``timestamp created`, and `timestamp last updated`  are of   primary type return default values 
when misisng

### Actual behavior
NPE is thrown

### Steps to reproduce:
Invoke ` jedis.aclLog() `against any Redis server 6.2.16
or run test `AccessControlListCommandsTest#aclLogWithEntryID`

```
try (Jedis jedis = new Jedis( "localhost", 6379)) {
    jedis.aclLog() throws NPE
}
```

#### Jedis version:
5.2.0
#### Redis version:
6.2.16 or any other which is older than 7.2.0
#### Java version:
1.8

### LINKED ISSUE #4017: NPE thrown when invoking  Jedis#commandInfo against Redis server 6.2.16
### Expected behavior
NPE is not thrown 

### Actual behavior

Write here what happens instead ...

### Steps to reproduce:
```
try ( Jedis jedis = new Jedis ("localhost", 6379) { 
   Map<String, CommandInfo> infos = jedis.commandInfo("GET", "foo", "SET");
}
```

or  run ControlCommandsTest#commandInfo test

default
#### Jedis version:
5.2.0
#### Redis version:
6.2.16 (or Redis server 7.0.0< ) 
#### Java version:
1.8

ggivo:
We use docker composer to bring up the test env using `redislabs/client-libs-test` image.

### Example command to bring up test env for dev/test
- **Redis 8.0-M01**
```
      export REDIS_VERSION=8.0-M01
      docker compose --env-file src/test/resources/env/.env -f src/test/resources/env/docker-compose.yml up
```
- **Redis 7.4.1** 
```
      export REDIS_VERSION=7.4.1
      docker compose --env-file src/test/resources/env/.env -f src/test/resources/env/docker-compose.yml up
```
- **Redis 7.2.6** 
```
      export REDIS_VERSION=7.2.6
      docker compose --env-file src/test/resources/env/.env -f src/test/resources/env/docker-compose.yml up
```
- **Redis 6.2.16**
      - **NOTE :** 6.2.16 uses a dedicated .env.v6.12.16 file, since some of the redis configuration settings are not supported in 6.2.16
      `docker compose --env-file src/test/resources/env/.env.v6.12.16 -f src/test/resources/env/docker-compose.yml up`

### Introduces test matrix based on Redis versions [8.0-M1, 7.4.1, 7.2.6, 6.2.16]
When run against older Redis version it appears that some tests are using commands available only in newer Redis server versions. To resolve this we are introducing two new annotations/rules

 - Introduce `SinceRedisVersion` annotation/Rule - for conditionally running tests based on Redis server version contacted
 - Introduce `EnableOnCommad` annotation/Rule -  for conditionally running tests based on command availability on the server

And mark respective tests with the least Redis Version required by the test
 - SinceRedisVersion ("7.4.0") - Mark tests using commands/modifiers introduced with Redis 7.4.0
 - SinceRedisVersion ("7.2.0") - Mark tests using commands/modifiers introduced with Redis 7.2.0
 - SinceRedisVersion ("7.0.0") - Mark tests using commands/modifiers introduced with Redis 7.0.0

 The same approach used to mark CSC tests
 - Disabled client-side caching tests for versions below  7.4

### Fix in Jedis Client (Redis  6.x)
- Fix NPE on CommandInfo - some of the array elements returned are available since ( Redis server 6.0/7.0)
   (see #4017 )

- Fix NPE  when reading AccessControlLogEntry  
  (see   #4016)

### Fixed Test failures against 6.x
 - Fix JedisPooledClientSideCacheTest
 - Fix AccessControlListCommandsTest.aclLogTest:372 » NullPointer
 - Fix AccessControlListCommandsTest.aclLogWithEntryID:473 » NullPointer
 - Fix StreamsCommandsTest
 - Fix StreamsPipelineCommandsTest xadd - Starting with Redis version 7.0.0: Added support for the <ms>-* explicit ID form.

### - Test env migrated to use native Redis server TLS instead of  using stunnel
   - There are still Cluster SSL tests failing against **6.2.16** env.  see https://github.com/redis/jedis/issues/4018
      The primary reason for those failures is the Redis server 6.2.16 is returning the non-TLS port in the CLUSTER SLOTS command in contrast to Redis Server 7.2.0+. I suggest reviewing & addressing Cluster SSL related test failures with follow-up PR.

### Ignored tests during migration
 - Disabled ModuleTest in containerized test env. 
    ModuleTest uses a custom test module to test load/unload/sendCommand.  Requires pre-build test module on the same os like test container to avoid errors

 - Disable UDS tests in containerized test env 
    No easy way to make unix sockets work on MAC with docker
 
Closes #4016,    Closes #4017


tishun:
LGTM, the code looks much neater this way!

src/test/java/redis/clients/jedis/SSLJedisClusterTest.java
```diff
@@ -27,7 +29,7 @@ public class SSLJedisClusterTest extends JedisClusterTestBase {
     int port = hostAndPort.getPort();
     if (host.equals("127.0.0.1")) {
       host = "localhost";
-      port = port + 1000;
+      //port = port + 1000;
```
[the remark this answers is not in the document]
ggivo: Yes. Removed

src/test/java/redis/clients/jedis/SSLJedisClusterTest.java
```diff
@@ -37,61 +39,90 @@ public class SSLJedisClusterTest extends JedisClusterTestBase {
     if ("localhost".equals(hostAndPort.getHost())) {
       return hostAndPort;
     }
-    return new HostAndPort(hostAndPort.getHost(), hostAndPort.getPort() + 1000);
+    //return new HostAndPort(hostAndPort.getHost(), hostAndPort.getPort() + 1000);
```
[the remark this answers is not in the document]
ggivo: removed

.github/workflows/test-on-docker.yml
```diff
@@ -0,0 +1,128 @@
+---
+
+name: Build and Test using containerized environment
```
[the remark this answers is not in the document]
ggivo: Discussed offline with @sazzad16 and decided to keep also the legacy way for bootstrapping test envs(using make files) but make the containerised setup default one for tests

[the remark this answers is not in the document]
ggivo: updated the review so that also local env is not using stunnel but native Redis TLS support

atakavci (approved):
this is hard work and i appreciate how we improve it. 
i left a couple of comments and questions but none of them is a big deal TBH,
it looks pretty good to me.
one thing to note; we need to find our way through smaller PRs, at least for non-public API changes.

Makefile
```diff
@@ -209,69 +249,94 @@ daemonize yes
...
 port 7380
+tls-port 8380
 cluster-node-timeout 15000
 pidfile /tmp/redis_cluster_node2.pid
 logfile /tmp/redis_cluster_node2.log
 save ""
 appendonly no
 cluster-enabled yes
 cluster-config-file /tmp/redis_cluster_node2.conf
+tls-auth-clients no
```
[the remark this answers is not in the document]
ggivo: Yes, it's still needed. I have briefly updated the existing make file just to remove the ' stunnel` and replace it with using native `TLS` support provided by the Redis server itself. 
This way, we can preserve both ways for bootstrapping test environments (via makefile and via docker-compose) and have all tests pass till we migrate the remaining test (Unix sockets) and remove make test env completely.
Make test env is using (expired) pre-generated certificate and I don't think it makes sense to put more effort to support/upgrade it if we plan to move on with docker-compose one.

.github/CONTRIBUTING.md
```diff
@@ -25,6 +25,47 @@ Jedis unit tests use many Redis instances, so we use a ```Makefile``` to prepare
...
+This guide explains how to bootstrap and manage a test environment for Jedis using Docker Compose.
+
+## Workflow Steps
+1. **Bring up the test environment** (examples provided below).
+2. **Run tests** (via IDE, Maven, etc.).
+3. **Destroy the test environment** using `docker compose down`.
+
+### Important Note
+The default test environment uses the temporary work folder `./redis-env-work`.  
+Some tests might leave Redis nodes in an inconsistent state, so this folder should be cleaned or removed before bootstrapping the environment again.
```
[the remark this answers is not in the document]
ggivo: Updated the review. Additional targets added to make file to bootstrap test env using docker compose

.github/workflows/test-on-docker.yml
```diff
@@ -0,0 +1,128 @@
...
+          export TEST_ENV_PROVIDER=docker
+          export TEST_WORK_FOLDER=$REDIS_ENV_WORK_DIR
+          echo $TEST_WORK_FOLDER
+          if [ -z "$TESTS" ]; then
+            mvn clean compile test
+          else
+            mvn -Dtest=$SPECIFIC_TEST clean compile test
+          fi
+        env:
+          TESTS: ${{ github.event.inputs.specific_test || '' }}
```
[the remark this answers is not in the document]
ggivo: Not is provided as a workflow input. 
Used to run a single test (group of tests ) when triggering the workflow manually. 
I have used it to speed up testing during test migration work (since a complete test suit takes like 20m)

src/test/java/io/redis/test/utils/EnabledOnCommandRule.java
```diff
@@ -0,0 +1,126 @@
...
+
+    @Override
+    public Statement apply(Statement base, Description description) {
+        return new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                try (Jedis jedisClient = new Jedis(hostPort, config)) {
+                    String[] command = getCommandFromAnnotations(description);
+
+                    if (command != null && !isCommandAvailable(jedisClient, command[0],command[1])) {
```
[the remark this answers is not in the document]
ggivo: I am afraid that it will add additional complexity to test infra and a chance for erroneous behavior. I wouldn't go with such optimization unless we see it is required.,

.github/workflows/test-on-docker.yml
```diff
@@ -0,0 +1,128 @@
...
+            /var/cache/apt
+          key: jedis-${{hashFiles('**/pom.xml')}}
+      # Set up Docker Compose environment
+      - name: Set up Docker Compose environment
+        run: |
+          mkdir -m 777 $REDIS_ENV_WORK_DIR
+          export REDIS_VERSION="${{ matrix.redis_version }}"
+          export COMPOSE_ENV_FILES="src/test/resources/env/.env"
+          if [[ "${{ matrix.redis_version }}" == "6.2.16" ]]; then
+           COMPOSE_ENV_FILES+=",src/test/resources/env/.env.v${{ matrix.redis_version }}"
```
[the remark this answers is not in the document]
ggivo: Fixed

src/test/java/io/redis/test/utils/RedisVersionRule.java
```diff
@@ -0,0 +1,105 @@
...
+            private void checkRedisVersion(Jedis jedisClient, SinceRedisVersion versionAnnotation) {
+
+                // Check if the environment variable is set
+                RedisVersion currentVersion;
+
+                if (forcedVersion != null) {
+                    logger.info("Using forced Redis server version from environment variable: " + forcedVersion);
+                    currentVersion = forcedVersion;
+                } else {
+                    RedisInfo info = RedisInfo.parseInfoServer(jedisClient.info("server"));
```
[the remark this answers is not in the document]
ggivo: Same as in EnableOnCommand I would prefer to keep it simple unless we see it is needed. 
Here we also have the option to force server version and this way completely disable calls to server

src/test/resources/Arch/sentinel5/config/node-sentinel5/redis.conf
sazzad16, line 1: Why `Arch` folder?

src/test/java/redis/clients/jedis/csc/JedisPooledClientSideCacheTest.java
```diff
@@ -1,13 +1,21 @@
...
+import org.junit.ClassRule;
 import redis.clients.jedis.HostAndPorts;
 
+@SinceRedisVersion(value = "7.4.0", message = "Jedis client-side caching is only supported with Redis 7.4 or later.")
 public class JedisPooledClientSideCacheTest extends JedisPooledClientSideCacheTestBase {
 
+  @ClassRule
+  public static RedisVersionRule versionRule = new RedisVersionRule(
+          HostAndPorts.getRedisEndpoint("standalone1").getHostAndPort(),
+          HostAndPorts.getRedisEndpoint("standalone1").getClientConfigBuilder().build());
```
sazzad16, line 15: You've already added the shortcut.
```suggestion
  public static RedisVersionRule versionRule = new RedisVersionRule(
          HostAndPorts.getRedisEndpoint("standalone1"));
```

src/test/java/redis/clients/jedis/csc/ClientSideCacheTestBase.java
```diff
@@ -1,26 +1,35 @@
...
   protected static final EndpointConfig endpoint = HostAndPorts.getRedisEndpoint("standalone1");
 
   protected static final HostAndPort hnp = endpoint.getHostAndPort();
 
   protected Jedis control;
 
+  @Rule
+  public RedisVersionRule versionRule = new RedisVersionRule(
+            HostAndPorts.getRedisEndpoint("standalone1").getHostAndPort(),
+            HostAndPorts.getRedisEndpoint("standalone1").getClientConfigBuilder().build());
```
sazzad16, line 27:
```suggestion
  public RedisVersionRule versionRule = new RedisVersionRule(
            HostAndPorts.getRedisEndpoint("standalone1"));
```

src/test/java/redis/clients/jedis/csc/ClientSideCacheTestBase.java
```diff
@@ -1,26 +1,35 @@
...
+  @Rule
+  public EnabledOnCommandRule enabledOnCommandRule = new EnabledOnCommandRule(
+          HostAndPorts.getRedisEndpoint("standalone1").getHostAndPort(),
+          HostAndPorts.getRedisEndpoint("standalone1").getClientConfigBuilder().build());
```
sazzad16, line 31:
```suggestion
  public EnabledOnCommandRule enabledOnCommandRule = new EnabledOnCommandRule(
          HostAndPorts.getRedisEndpoint("standalone1"));
```

sazzad16 (changes requested):

src/test/resources/Arch/sentinel5/config/node-sentinel5/redis.conf
ggivo: It is not used and submitted by mistake.
Removed.

src/test/java/redis/clients/jedis/commands/jedis/ControlCommandsTest.java
```diff
@@ -564,6 +577,23 @@ public void commandInfoAcl() {
   }
 
   @Test
+  @Ignore( "Till https://github.com/redis/jedis/issues/4020 is resolved")
```
sazzad16, line 580: It is resolved.

src/test/java/redis/clients/jedis/commands/jedis/ControlCommandsTest.java
```diff
@@ -564,6 +576,22 @@ public void commandInfoAcl() {
...
+    assertEquals(0, aclInfo.getFlags().size());
+    assertEquals(0, aclInfo.getFirstKey());
+    assertEquals(0, aclInfo.getLastKey());
+    assertEquals(0, aclInfo.getStep());
+    assertEquals(1, aclInfo.getAclCategories().size());
+    assertEquals(0, aclInfo.getTips().size());
+    assertFalse(aclInfo.getSubcommands().isEmpty());
+  }
+
+  @Test
```
sazzad16, line 593: We don't need this test at all because of the above test `commandInfoAcl()`.
```suggestion
```

src/test/java/redis/clients/jedis/commands/jedis/AllKindOfValuesCommandsTest.java
```diff
@@ -626,7 +635,8 @@ public void restoreParams() {
     assertTrue(jedis2.pttl("foo") <= 1000);
 
     jedis2.restore("bar", System.currentTimeMillis() + 1000, serialized, RestoreParams.restoreParams().replace().absTtl());
-    assertTrue(jedis2.pttl("bar") <= 1000);
+    assertThat("ttl", jedis2.pttl("bar"), lessThanOrEqualTo(1000l + TIME_SKEW));
```
sazzad16, line 638:
```suggestion
    assertThat(jedis2.pttl("bar"), lessThanOrEqualTo(1000l + TIME_SKEW));
```

src/test/java/redis/clients/jedis/JedisTest.java
```diff
@@ -319,6 +321,7 @@ public void clientSetInfoDefault() {
   }
 
   @Test
+  @SinceRedisVersion(value = "7.2.0", message = "@see https://redis.io/docs/latest/commands/client-setinfo/")
```
sazzad16, line 324: ClientSetInfoConfig.DISABLED should work even at lower versions.
```suggestion
```

sazzad16 (changes requested):

src/test/java/redis/clients/jedis/util/RedisVersionUtil.java
```diff
@@ -1,32 +1,61 @@
...
-    return getRedisMajorVersionNumber(endpoint) >= minVersion;
+  public static RedisVersion getRedisVersion(EndpointConfig endpoint) {
+    if (forcedVersion != null) {
+      return forcedVersion;
+    }
+    
+    DefaultJedisClientConfig.Builder builder = endpoint.getClientConfigBuilder();
+    if (endpoint.isTls()) {
+      builder.sslSocketFactory(createTrustAllSslSocketFactory());
+    }
```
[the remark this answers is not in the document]
ggivo: Not sure I get this one. What do you mean by compatible inside EndpointConfig?

src/test/java/redis/clients/jedis/commands/jedis/ClusterJedisCommandsTestBase.java
```diff
@@ -71,28 +75,25 @@ public void setUp() throws InterruptedException {
...
 
   @AfterClass
   public static void cleanUp() {
     int slotTest = JedisClusterCRC16.getSlot("test");
     int slot51 = JedisClusterCRC16.getSlot("51");
-    String node3Id = getNodeId(node3.clusterNodes());
-    node2.clusterSetSlotNode(slotTest, node3Id);
-    node2.clusterSetSlotNode(slot51, node3Id);
-    node2.clusterDelSlots(slotTest, slot51);
+    if (node3 != null ) {
```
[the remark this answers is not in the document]
ggivo: It is in the case when you have all tests disabled (for example in 
 ClusterShardedPublishSubscribeCommandsTest). Then the @Before method is never run, and we end up with null for all nodes.

```
@SinceRedisVersion(value = "7.0.0", message = "Sharded Pub/Sub")
public class ClusterShardedPublishSubscribeCommandsTest extends ClusterJedisCommandsTestBase {
```

src/test/java/redis/clients/jedis/util/RedisVersionUtil.java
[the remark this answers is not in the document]
sazzad16: I mean `getClientConfigBuilder()` should check `isTls()` and do `builder.sslSocketFactory(createTrustAllSslSocketFactory())` within itself.

src/test/java/redis/clients/jedis/commands/jedis/ClusterJedisCommandsTestBase.java
sazzad16, line 85:
```suggestion
    if (node3 != null) {
```

src/test/java/redis/clients/jedis/commands/unified/cluster/ClusterBitCommandsTest.java
```diff
@@ -16,6 +17,15 @@
 @RunWith(Parameterized.class)
 public class ClusterBitCommandsTest extends BitCommandsTestBase {
 
+  @Rule
+  public RedisVersionRule versionRule = new RedisVersionRule(
+            HostAndPorts.getStableClusterServers().get(0)
+            ,DefaultJedisClientConfig.builder().password("cluster").build());
```
sazzad16, line 23:
```suggestion
            HostAndPorts.getStableClusterServers().get(0),
            DefaultJedisClientConfig.builder().password("cluster").build());
```

src/test/java/redis/clients/jedis/commands/unified/cluster/ClusterCommandsTestHelper.java
```diff
@@ -14,7 +14,7 @@ static JedisCluster getCleanCluster(RedisProtocol protocol) {
     clearClusterData();
     return new JedisCluster(
         Collections.singleton(HostAndPorts.getStableClusterServers().get(0)),
-        DefaultJedisClientConfig.builder().password("cluster").protocol(protocol).build());
+            DefaultJedisClientConfig.builder().password("cluster").protocol(protocol).build());
```
sazzad16, line 17:
```suggestion
        DefaultJedisClientConfig.builder().password("cluster").protocol(protocol).build());
```

src/test/java/redis/clients/jedis/csc/JedisClusterClientSideCacheTest.java
```diff
@@ -28,6 +32,9 @@ public class JedisClusterClientSideCacheTest extends UnifiedJedisClientSideCache
         return poolConfig;
       };
 
+  @ClassRule
+  public static RedisVersionRule   versionRule = new RedisVersionRule(hnp.iterator().next(), clientConfig.get());
```
sazzad16, line 36:
```suggestion
  public static RedisVersionRule versionRule = new RedisVersionRule(hnp.iterator().next(), clientConfig.get());
```

src/main/java/redis/clients/jedis/resps/CommandInfo.java
```diff
@@ -2,6 +2,7 @@
 
 import redis.clients.jedis.Builder;
 
+import java.util.Collections;
```
sazzad16, line 5:
```suggestion
```

src/test/java/redis/clients/jedis/JedisClusterTestBase.java
```diff
@@ -23,6 +26,10 @@ public abstract class JedisClusterTestBase {
 
   protected static final String LOCAL_IP = "127.0.0.1";
 
+  @Rule
+  public RedisVersionRule versionRule = new RedisVersionRule(nodeInfo1,DefaultJedisClientConfig.builder().password("cluster").build());
+  @Rule
+  public EnabledOnCommandRule enabledOnCommandRule = new EnabledOnCommandRule(nodeInfo1, DefaultJedisClientConfig.builder().password("cluster").build());
```
sazzad16, line 32:
```suggestion
  @Rule
  public RedisVersionRule versionRule = new RedisVersionRule(nodeInfo1,DefaultJedisClientConfig.builder().password("cluster").build());
  @Rule
  public EnabledOnCommandRule enabledOnCommandRule = new EnabledOnCommandRule(nodeInfo1, DefaultJedisClientConfig.builder().password("cluster").build());

```

src/test/java/redis/clients/jedis/commands/commandobjects/CommandObjectsTestBase.java
```diff
@@ -84,9 +86,14 @@ public void setUp() {
...
-        commandExecutor.executeCommand(commandObjects.functionFlush(FlushMode.SYNC)),
-        equalTo("OK"));
+    if ( RedisVersionUtil
+            .getRedisVersion(endpoint)
+            .isGreaterThanOrEqualTo(RedisVersion.V7_0_0)) {
+      assertThat(
+              commandExecutor.executeCommand(commandObjects.functionFlush(FlushMode.SYNC)),
+              equalTo("OK"));
+
+    }
```
sazzad16, line 96:
```suggestion
    if (RedisVersionUtil.getRedisVersion(endpoint)
            .isGreaterThanOrEqualTo(RedisVersion.V7_0_0)) {
      assertThat(
              commandExecutor.executeCommand(commandObjects.functionFlush(FlushMode.SYNC)),
              equalTo("OK"));
    }
```

src/test/java/redis/clients/jedis/EndpointConfig.java
```diff
@@ -112,6 +123,10 @@ public DefaultJedisClientConfig.Builder getClientConfigBuilder() {
       DefaultJedisClientConfig.Builder builder = DefaultJedisClientConfig.builder()
           .password(password).ssl(tls);
 
+        if (tls & environment != null) {
```
sazzad16, line 126:
```suggestion
        if (tls && environment != null) {
```

src/test/java/redis/clients/jedis/util/RedisVersionUtil.java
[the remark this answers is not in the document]
ggivo: while using trus all socket factories should be fine for utility methods like resolving server version,
`redis.clients.jedis.EndpointConfig#getClientConfigBuilder ` is used by TLS tests when configuring Jeis client, and there we should make sure we use properly configured TLS context. I guess  we can do something like

```
        if (tls && environment != null) {
          builder.sslSocketFactory(TlsUtil.sslSocketFactoryForEnv(environment));
        }
```
and have Jedis client configured to use the environment-specific trust store.

src/test/java/redis/clients/jedis/commands/jedis/BitCommandsTest.java
```diff
@@ -153,13 +155,20 @@ public void bitposModifier() {
     assertEquals(1, jedis.bitpos("mykey", true, BitPosParams.bitPosParams()));
     assertEquals(18, jedis.bitpos("mykey", true, BitPosParams.bitPosParams().start(2)));
     assertEquals(18, jedis.bitpos("mykey", true, BitPosParams.bitPosParams().start(2).end(-1)));
+  }
+
+  @Test
+  @SinceRedisVersion(value = "7.0.0", message = "7.0.0 Added the BYTE|BIT option.")
+  public void bitposModifierByte() {
+    jedis.set("mykey", "\\x00\\xff\\xf0");
```
sazzad16, line 163: The name modifier suggests the above test is specifically for BYTE|BIT. So just

```suggestion
```

and put the condition above.

src/test/java/redis/clients/jedis/commands/jedis/BitCommandsTest.java
```diff
@@ -182,6 +191,15 @@ public void bitCount() {
 
     assertEquals(3, (long) jedis.bitcount("foo", 2L, 5L));
     assertEquals(3, (long) jedis.bitcount("foo".getBytes(), 2L, 5L));
+  }
+
+  @Test
+  @SinceRedisVersion("7.0.0")
+  public void bitCountByteOptions() {
```
sazzad16, line 198: BYTE is actually the default mode and BIT is the new option. Modifier is safer name and aligns with other tests.

```suggestion
  public void bitCountModifier() {
```

src/test/java/redis/clients/jedis/commands/jedis/AccessControlListCommandsTest.java
```diff
@@ -95,20 +100,23 @@ public void addAndRemoveUser() {
...
   @Test
+  @SinceRedisVersion(value = "7.0.0", message = "Redis 6.2.x misses [~]>")
   public void aclGetUser() {
     // get default user information
     AccessControlUser userInfo = jedis.aclGetUser("default");
 
     assertFalse(userInfo.getFlags().isEmpty());
     assertEquals(1, userInfo.getPassword().size());
     assertEquals("+@all", userInfo.getCommands());
+
```
sazzad16, line 119:
```suggestion
```

src/test/java/redis/clients/jedis/commands/commandobjects/CommandObjectsTestBase.java
```diff
@@ -84,9 +86,12 @@ public void setUp() {
...
 
-    assertThat(
-        commandExecutor.executeCommand(commandObjects.functionFlush(FlushMode.SYNC)),
-        equalTo("OK"));
+    if (RedisVersionUtil.getRedisVersion(endpoint)
+            .isGreaterThanOrEqualTo(RedisVersion.V7_0_0)) {
+      assertThat(
+              commandExecutor.executeCommand(commandObjects.functionFlush(FlushMode.SYNC)),
+              equalTo("OK"));
+    }
```
sazzad16, line 94: We can move this into CommandObjectsScriptingCommandsTest.

src/test/java/redis/clients/jedis/commands/jedis/ClientCommandsTest.java
```diff
@@ -38,6 +42,11 @@ public class ClientCommandsTest extends JedisCommandsTestBase {
 
   private Jedis client;
 
+  @Rule
+  public RedisVersionRule versionRule = new RedisVersionRule(endpoint);
+  @Rule
+  public EnabledOnCommandRule enabledOnCommandRule = new EnabledOnCommandRule(endpoint);
+
```
sazzad16, line 49: Do we need these here anymore? As these conditions are already in the super-class, JedisCommandsTestBase.

src/test/java/redis/clients/jedis/commands/jedis/ClusterCommandsTest.java
```diff
@@ -35,6 +35,11 @@ public class ClusterCommandsTest {
   private static HostAndPort nodeInfo1 = HostAndPorts.getClusterServers().get(0);
   private static HostAndPort nodeInfo2 = HostAndPorts.getClusterServers().get(1);
 
+  @Rule
+  public RedisVersionRule versionRule = new RedisVersionRule(nodeInfo1, DefaultJedisClientConfig.builder().password("cluster").build());
+  @Rule
+  public EnabledOnCommandRule enabledOnCommandRule = new EnabledOnCommandRule(nodeInfo1, DefaultJedisClientConfig.builder().password("cluster").build());
```
sazzad16, line 41: Do we need this when not using EnabledOnCommand?

src/test/java/redis/clients/jedis/commands/jedis/ScriptingCommandsTest.java
```diff
@@ -39,7 +42,11 @@ public ScriptingCommandsTest(RedisProtocol redisProtocol) {
   @Override
   public void setUp() throws Exception {
     super.setUp();
-    jedis.functionFlush();
+    if ( RedisVersionUtil
+            .getRedisVersion(jedis)
```
sazzad16, line 46:
```suggestion
    if (RedisVersionUtil.getRedisVersion(jedis)
```

src/test/java/redis/clients/jedis/commands/unified/BitCommandsTestBase.java
```diff
@@ -19,6 +21,7 @@
 
 public abstract class BitCommandsTestBase extends UnifiedJedisCommandsTestBase {
 
+
```
sazzad16, line 24:
```suggestion
```

src/test/java/redis/clients/jedis/commands/unified/pooled/PooledCommandsTestHelper.java
```diff
@@ -4,7 +4,7 @@
 
 public class PooledCommandsTestHelper {
 
-  private static final EndpointConfig nodeInfo = HostAndPorts.getRedisEndpoint("standalone0");
+  public static final EndpointConfig nodeInfo = HostAndPorts.getRedisEndpoint("standalone0");
```
sazzad16, line 7:
```suggestion
  static final EndpointConfig nodeInfo = HostAndPorts.getRedisEndpoint("standalone0");
```

_package private_ could be enough. But please make sure :)

src/test/java/redis/clients/jedis/csc/ClientSideCacheTestBase.java
```diff
@@ -1,26 +1,32 @@
...
-import redis.clients.jedis.EndpointConfig;
-import redis.clients.jedis.HostAndPort;
-import redis.clients.jedis.HostAndPorts;
-import redis.clients.jedis.Jedis;
-import redis.clients.jedis.JedisClientConfig;
+import org.junit.Rule;
+import redis.clients.jedis.*;
 
-abstract class ClientSideCacheTestBase {
+@SinceRedisVersion(value = "7.4.0", message = "Jedis client-side caching is only supported with Redis 7.4 or later.")
```
sazzad16, line 15: We shouldn't put this same annotation on both super and sub classes.
It's great if putting only on super class works. Otherwise remove from this super class and put only on sub classes.

src/test/java/redis/clients/jedis/csc/ClientSideCacheTestBase.java
```diff
@@ -1,26 +1,32 @@
...
 
   protected static final HostAndPort hnp = endpoint.getHostAndPort();
 
   protected Jedis control;
 
+  @Rule
+  public RedisVersionRule versionRule = new RedisVersionRule( HostAndPorts.getRedisEndpoint("standalone1"));
+
+  @Rule
+  public EnabledOnCommandRule enabledOnCommandRule = new EnabledOnCommandRule(HostAndPorts.getRedisEndpoint("standalone1"));
```
sazzad16, line 28: EnabledOnCommand may never be used for CSC tests.
```suggestion
```

src/test/java/redis/clients/jedis/csc/ClientSideCacheTestBase.java
```diff
@@ -1,26 +1,32 @@
...
+public abstract class ClientSideCacheTestBase {
 
   protected static final EndpointConfig endpoint = HostAndPorts.getRedisEndpoint("standalone1");
 
   protected static final HostAndPort hnp = endpoint.getHostAndPort();
 
   protected Jedis control;
 
+  @Rule
+  public RedisVersionRule versionRule = new RedisVersionRule( HostAndPorts.getRedisEndpoint("standalone1"));
```
sazzad16, line 25:
```suggestion
  public RedisVersionRule versionRule = new RedisVersionRule(HostAndPorts.getRedisEndpoint("standalone1"));
```

src/test/java/redis/clients/jedis/commands/unified/pooled/PooledCommandsTestHelper.java
ggivo: used also in different package - redis.clients.jedis.commands.unified.pipeline

src/test/java/redis/clients/jedis/commands/jedis/ClusterCommandsTest.java
ggivo: removed EnabledOnCommandRule

src/test/java/redis/clients/jedis/csc/ClientSideCacheTestBase.java
ggivo: removing it  from subclasses

src/test/java/redis/clients/jedis/commands/commandobjects/CommandObjectsScriptingCommandsTest.java
```diff
@@ -35,6 +39,16 @@ public CommandObjectsScriptingCommandsTest(RedisProtocol protocol) {
     super(protocol);
   }
 
+  @Before
+  public void flushLibraries() {
```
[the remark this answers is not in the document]
ggivo: According to the junit doc (https://junit.org/junit4/javadoc/4.12/org/junit/Before.html)
The @Before methods of superclasses will be run before those of the current class, unless they are overridden in the current class

src/test/java/redis/clients/jedis/SSLOptionsJedisSentinelPoolTest.java
```diff
@@ -75,8 +80,8 @@ public void sentinelWithSslConnectsToRedisWithoutSsl() {
   public void sentinelWithSslConnectsToRedisWithSsl() {
 
     SslOptions sslOptions = SslOptions.builder()
-                .truststore(new File("src/test/resources/truststore.jceks"))
-                .trustStoreType("jceks")
+            .truststore(envTruststore("redis9-sentinel").toFile())
+            .trustStoreType("jceks")
             .sslVerifyMode(SslVerifyMode.CA).build();
```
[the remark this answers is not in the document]
ggivo: The latter one looks concise to me (e.g. sslVerifyMode on one level with trustStoreType) 
Not sure what should be the correct one. Could you advise?

src/test/java/redis/clients/jedis/SSLOptionsJedisClusterTest.java
```diff
@@ -36,49 +39,77 @@ public class SSLOptionsJedisClusterTest extends JedisClusterTestBase {
...
+
   @Test
   public void testSSLDiscoverNodesAutomatically() {
     try (JedisCluster jc2 = new JedisCluster(new HostAndPort("localhost", 8379),
         DefaultJedisClientConfig.builder().password("cluster")
-            .sslOptions(SslOptions.builder()
-                .truststore(new File("src/test/resources/truststore.jceks"))
+                .sslOptions(SslOptions.builder()
+                .truststore(envTruststore("cluster-unbound").toFile())
                 .trustStoreType("jceks").build())
```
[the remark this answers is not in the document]
ggivo: reformated

sazzad16 (changes requested):

src/test/java/redis/clients/jedis/EndpointConfig.java
```diff
@@ -51,12 +54,20 @@ public int getPort() {
...
+        return tls;
+    }
+
     public int getBdbId() { return bdbId; }
 
     public URI getURI() {
         return endpoints.get(0);
     }
 
+    public String getEnvironment() {
```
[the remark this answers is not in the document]
ggivo: @sazzad16 
Yes, it makes sense.
One caveat is that in the legacy test environment, we used one and the same pre-generated certificate to bootstrap all Redis servers. With the containerized test environment, we generate _server/client/ca_ certificates when bootstrapping each test container, which means that we end up with a different certificate per test container. 

On the other hand, we have Jedis tests using default JVM SSLContext with custom trust store (e.g `javax.net.ssl.trustStore`). Since some tests use multiple `EndpointConfig`  we can create a trust store per test & populate it with the certificate from each EndpointConfig used in the given test. What do you think?

Also, I am starting to rethink if we set "`environment`" in `EndpointConfig` and use it to construct the path to certificates or we can directly store the location of the certificates in each EndpointConfig.
```
{
  "standalone0-acl-tls": {
    "username": "acljedis",
    "password": "fizzbuzz",
    "tls": true,
    "certificates": {
      "location": "redis1-2-5-8-sentinel/work/tls"
    },
    "endpoints": [
      "rediss://localhost:6390"
    ]
  }
}
```
I mean now we specify that given EndpointConfig should use TLS  but for a successful connection, we need to configure trusted certificates & eventually client certificates.
@uglide any thoughts/ideas on how we want to manage the certificates in EndpointConfig?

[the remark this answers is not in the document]
ggivo: Also since Cluster test were not yet migrated to use EndpointConfig's we will need to keep the hardcoded values at least for them.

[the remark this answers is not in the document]
ggivo: @sazzad16 
This now should be addressed where possible (e.g. Some tests like cluster one are not using EndpointConfigs yet)
Also, I replaced the "environment" with the actual "certificatesLocation", which resolves to the root folder containing certificates for a given endpoint. Trust stores are now generated per test instead per the environment.

src/test/java/redis/clients/jedis/util/GeoRadiusResponseMatcher.java
```diff
@@ -19,15 +20,19 @@ public GeoRadiusResponseMatcher(GeoRadiusResponse expected) {
...
-        // Check if distance and rawScore match exactly
+        // Check if other attributes match exactly
         if (Double.compare(expected.getDistance(), actual.getDistance()) != 0) {
             return false;
         }
-        return expected.getRawScore() == actual.getRawScore();
+        if (Long.compare(expected.getRawScore(), actual.getRawScore()) != 0) {
+            return false;
+        }
+        return Arrays.equals(expected.getMember(), actual.getMember());
```
ggivo, line 35: Good catch!

PR #4038, opened 2024-12-05
DOC-4560 pipe/transaction examples for docs

andy-stark-redis:
[DOC-4560](https://redislabs.atlassian.net/browse/DOC-4560)

Some examples for a new page roughly equivalent to the one for [Python](https://redis.io/docs/latest/develop/clients/redis-py/transpipe/) (although Jedis handles things a bit differently, of course).

Let me know if there are any other details you'd like me to add, or if the `watch` example should show repeated attempts in a loop, etc.

[DOC-4560]: https://redislabs.atlassian.net/browse/DOC-4560?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/test/java/io/redis/examples/PipeTransExample.java
+++ b/src/test/java/io/redis/examples/PipeTransExample.java
@@ -76,8 +76,9 @@ public class PipeTransExample {
         jedis.set("shellpath", "/usr/syscmds/");
 
         // Start the transaction and watch the key we are about to update.
-        trans = jedis.multi();
-        trans.watch("shellpath");
+        trans = jedis.transaction(false); // create a Transaction object without sending MULTI command
+        trans.watch("shellpath"); // send WATCH command(s)
+        trans.multi(); // send MULTI command
 
         String currentPath = jedis.get("shellpath");
         String newPath = currentPath + ":/usr/mycmds/";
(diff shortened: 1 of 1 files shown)


src/test/java/io/redis/examples/PipeTransExample.java
```diff
@@ -0,0 +1,117 @@
...
+        Assert.assertEquals("3", jedis.get("counter:3"));
+        // REMOVE_END
+
+        // STEP_START trans_watch
+        // Set initial value of `shellpath`.
+        jedis.set("shellpath", "/usr/syscmds/");
+
+        // Start the transaction and watch the key we are about to update.
+        trans = jedis.multi();
+        trans.watch("shellpath");
```
sazzad16, line 80: IIUC, you should call all the WATCHes before the MULTI. If so, the change can be:
```suggestion
        trans = jedis.transaction(false); // create a Transaction object without sending MULTI command
        trans.watch("shellpath"); // send WATCH command(s)
        trans.multi(); // send MULTI command
```

andy-stark-redis:
@sazzad16 I wondered why both the client object and the transaction object had a `multi()` method, so thanks for clarifying that :-)

sazzad16 (approved):

PR #4042, opened 2024-12-26
Test modules ACL support

sazzad16:


src/test/java/redis/clients/jedis/modules/ConsolidatedAccessControlListCommandsTest.java
```diff
@@ -0,0 +1,186 @@
...
+      // user can now execute category commands
+      operation.accept(client);
+    }
+  }
+
+  private static String getNopermErrorMessage(boolean commandNameUpperCase, ProtocolCommand protocolCommand) {
+    String command = SafeEncoder.encode(protocolCommand.getRaw());
+    return String.format("NOPERM User %s has no permissions to run the '%s' command",
+          USER_NAME, commandNameUpperCase ? command.toUpperCase(Locale.ENGLISH) : command.toLowerCase(Locale.ENGLISH));
+  }
```
ggivo, line 185: nit: I think verifying the error code is NOPERM should be sufficient

ggivo (dismissed):
LGTM

The Integration test job is failing since it is bootstrapping test env from unstable branch  (being 7.9) but by default it is build without modules.

Should we consider migrating completely to containerised env?

src/test/java/redis/clients/jedis/modules/ConsolidatedAccessControlListCommandsTest.java
sazzad16: @ggivo It's still a consistent error message and I'm already checking it :slightly_smiling_face: 
We can relax the check later if it becomes unstable.

ggivo (approved):
LGTM

PR #4053, opened 2025-01-15
Fix sporadic test failing with OOM

ggivo:
Missing AuthXManager shutdown in tests causes the token renew thread to run after the test completion. 
When used with Mockito.mock() this leads to a growing number of mocked objects and eventually to OOM.

- Invoke AuthXManager.close()
- Updated GitHub test workflow to publish heap dump on OOM errors in the test

### DIFF
--- a/.github/workflows/test-on-docker.yml
+++ b/.github/workflows/test-on-docker.yml
@@ -83,7 +83,7 @@ jobs:
             mvn -Dtest=$TESTS clean compile test
           fi
         env:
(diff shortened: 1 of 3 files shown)


sazzad16 (changes requested):
All these `.close();` and `.stop();` will still be missed if there are errors; i.e. assertion failures before them.
These should be put in reasonable `finally` block. try-with-resources can also be considered if possible.

sazzad16 (approved):

PR #4056, opened 2025-01-15
DOC-4445 server management command examples

andy-stark-redis:
[DOC-4445](https://redislabs.atlassian.net/browse/DOC-4445) and [DOC-4460](https://redislabs.atlassian.net/browse/DOC-4460).

Jedis versions of the [`FLUSHALL`](https://redis.io/docs/latest/commands/flushall/) and [`INFO`](https://redis.io/docs/latest/commands/info/) commands.

Note: `INFO` doesn't seem to be supported by `UnifiedJedis` but it is mentioned in other places. I figured the best way to handle it was to say it was not supported but let me know if you want to show an alternative way to access it.

[DOC-4445]: https://redislabs.atlassian.net/browse/DOC-4445?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DOC-4460]: https://redislabs.atlassian.net/browse/DOC-4460?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/test/java/io/redis/examples/CmdsServerMgmtExample.java
+++ b/src/test/java/io/redis/examples/CmdsServerMgmtExample.java
@@ -1,0 +1,43 @@
+// EXAMPLE: cmds_servermgmt
+// REMOVE_START
+package io.redis.examples;
+
+import org.junit.Assert;
+import org.junit.Test;
+// REMOVE_END
+import java.util.Set;
+
+// HIDE_START
+import redis.clients.jedis.UnifiedJedis;
+
+public class CmdsServerMgmtExample {
+    @Test
+    public void run() {
+        UnifiedJedis jedis = new UnifiedJedis("redis://localhost:6379");
+
+        // STEP_START flushall
+        //REMOVE_START
+        jedis.set("testkey1", "1");
+        jedis.set("testkey2", "2");
+        jedis.set("testkey3", "3");
+        //REMOVE_END
+        String flushAllResult1 = jedis.flushAll();
+        System.out.println(flushAllResult1); // >>> OK
+
+        Set<String> flushAllResult2 = jedis.keys("*");
+        System.out.println(flushAllResult2); // >>> []
+        // STEP_END
+        // REMOVE_START
+        Assert.assertEquals("OK", flushAllResult1);
+        Assert.assertEquals("[]", flushAllResult2.toString());
+        // REMOVE_END
+
+        // STEP_START info
+
+        // Not currently supported by Jedis.
+
+        // STEP_END
+
+// HIDE_END
+    }
+}
--- a/src/test/java/io/redis/examples/CmdsServerMgmtExample.java
+++ b/src/test/java/io/redis/examples/CmdsServerMgmtExample.java
@@ -15,6 +15,7 @@ public class CmdsServerMgmtExample {
     @Test
     public void run() {
         UnifiedJedis jedis = new UnifiedJedis("redis://localhost:6379");
+// HIDE_END
 
         // STEP_START flushall
         // REMOVE_START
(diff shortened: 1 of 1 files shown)


src/test/java/io/redis/examples/CmdsServerMgmtExample.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Set;
+
+import redis.clients.jedis.Jedis;
+// HIDE_START
+import redis.clients.jedis.UnifiedJedis;
+
+public class CmdsServerMgmtExample {
+    @Test
+    public void run() {
+        UnifiedJedis jedis = new UnifiedJedis("redis://localhost:6379");
```
sazzad16, line 17: `Jedis` class supports both FLUSHALL and INFO commands.
Would you prefer to have this example with only one class (Jedis) or keep it as it is right now?

andy-stark-redis: > `Jedis` class supports both FLUSHALL and INFO commands. Would you prefer to have this example with only one class (Jedis) or keep it as it is right now?

@sazzad16 Am I right in thinking we recommend using `UnifiedJedis` in preference to `Jedis` where possible? Almost all of the examples assume `UnifiedJedis` throughout, so I think we should stick to that except for cases like this where there are unsupported features. Remember that the user usually sees the examples as separate short snippets rather than seeing the whole file, so we should probably mention explicitly where a command is only available in `Jedis`.

src/test/java/io/redis/examples/CmdsServerMgmtExample.java
```diff
@@ -0,0 +1,55 @@
...
+
+        // STEP_START flushall
+        //REMOVE_START
```
sazzad16, line 20:
```suggestion
        // REMOVE_START
```

src/test/java/io/redis/examples/CmdsServerMgmtExample.java
```diff
@@ -0,0 +1,55 @@
...
+        jedis.set("testkey1", "1");
+        jedis.set("testkey2", "2");
+        jedis.set("testkey3", "3");
+        //REMOVE_END
```
sazzad16, line 24:
```suggestion
        // REMOVE_END
```

src/test/java/io/redis/examples/CmdsServerMgmtExample.java
```diff
@@ -0,0 +1,55 @@
...
+        // is much longer than this).
+        System.out.println(infoResult.substring(0, 8)); // >>> # Server
+
+        jedis2.close();
+        // STEP_END
+        // REMOVE_START
+        Assert.assertEquals("# Server", infoResult.substring(0, 8));
+        // REMOVE_END
+        
+// HIDE_END
```
sazzad16, line 53: This will HIDE everything from line 11. This should be fixed by adding more HIDE clause.

src/test/java/io/redis/examples/CmdsServerMgmtExample.java
sazzad16, line 17: Please call `jedis.close();` at a proper place.

sazzad16 (changes requested):

src/test/java/io/redis/examples/CmdsServerMgmtExample.java
andy-stark-redis: > Please call `jedis.close();` at a proper place.

@sazzad16 I've put it at the end of the file on the basis that we will probably have a lot more `UnifiedJedis` examples in this file in the future. But if you think it looks out-of-place with the current content of the file then I'll move it to the end of the `flushall` example.

sazzad16: It's fine.

sazzad16 (approved):

PR #4070, opened 2025-01-22
Multi node pipeline executor

dolavb:
This allow clients to provide an ExecutorService at pipeline creation instead of having a new ExecutorService created at every pipeline Sync call. An Executor service creation will create new threads, which are expensive resource to creates. In a high throughput application developed internally, we are writing at a rate of ~100k set per seconds, on a six node cluster, on an instance equivalent to an EC2 m5.12xlarge. The creation of threads uses 40% of CPU, and adds substantial latency. This new approach will allow clients to send a pooled Executor that is tuned to there load patterns.

This change is non breaking, but will come with a slight optimization for the clients currently using the created thread pool. In the current approach even if a pipeline has a single connection to close the Executor service will create MULTI_NODE_PIPELINE_SYNC_WORKERS threads. In the default mode would mean wasting 2 thread creation.

---
From: https://stackoverflow.com/questions/5483047/why-is-creating-a-thread-said-to-be-expensive

> Thread lifecycle overhead. Thread creation and teardown are not free. The actual overhead varies across platforms, but thread creation takes time, introducing latency into request processing, and requires some processing activity by the JVM and OS. If requests are frequent and lightweight, as in most server applications, creating a new thread for each request can consume significant computing resources.
> 
> From Java Concurrency in Practice
> By Brian Goetz, Tim Peierls, Joshua Bloch, Joseph Bowbeer, David Holmes, Doug Lea
> Print ISBN-10: 0-321-34960-1

> 
> Java thread creation is expensive because there is a fair bit of work involved:
> 
> * A large block of memory has to be allocated and initialized for the thread stack.
> * System calls need to be made to create / register the native thread with the host OS.
> * Descriptors need to be created, initialized and added to JVM-internal data structures.


sazzad16:
@dolavb Thank you for your effort to improve Jedis.
Your concern is justified. But we have our hands full ATM. We'll try to get to this PR ASAP.

dolavb:
Executed some more load testing, and we have host spending 60% our total CPU usage on starting threads.

sazzad16:
cc @uglide @ggivo @atakavci

atakavci:
lets start considering on a pool of executorService instances. And discuss tradeoffs of each solution we have. For sure we are doing this for performance but we need to regard other aspects from API standpoint, like principles of information hiding and encapsulation .

src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
```diff
@@ -84,44 +103,47 @@ public final void sync() {
...
-    executorService.shutdownNow();
+  private ExecutorService getExecutorService() {
+    if (executorService == null) {
+      return Executors.newFixedThreadPool(Math.min(this.pipelinedResponses.size(), MULTI_NODE_PIPELINE_SYNC_WORKERS));
+    }
+    return executorService;
+  }
 
-    syncing = false;
+  private void closeConnection(Map.Entry<HostAndPort, Queue<Response<?>>> entry) {
```
ggivo, line 133: The method name is misleading. It is actually reading the command responses rather than closing the connections

src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
```diff
@@ -84,44 +103,47 @@ public final void sync() {
...
+    ExecutorService executorService = getExecutorService();
+    CompletableFuture[] futures
+            = pipelinedResponses.entrySet().stream()
+            .map(e -> CompletableFuture.runAsync(() -> closeConnection(e), executorService))
+            .toArray(CompletableFuture[]::new);
+    CompletableFuture awaitAllCompleted = CompletableFuture.allOf(futures);
     try {
-      countDownLatch.await();
+        awaitAllCompleted.get();
+        if (executorService != this.executorService) {
```
ggivo, line 114: I suggest a more explicit approach.
e.g, when creating the MultiNodePipelineBase, if an external executor is provided, have a flag `useSharedExecutor = true`.
This way it will be clearer that lifecycle is managed externaly

ggivo (commented):
I think the described use-case make sense and also the direction of the PR will help resolve performance consideration.
As @atakavci pointed it is good to discuss from API point of view if it make sense to expose directly  ExecutorService 
@atakavci Do you have anything in mind?

src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
dolavb: Thanks, will address.

src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
dolavb: Thanks, will address.

dolavb:
@atakavci  Yes, I think the API considerations are important. It is unfortunate that this now have an API that is essentially a global variable in the static field. Ohterwise, I would have chosen one of the typical approaches:
* Have an optional ExecutorService in the client config.
  * Handle the option at the implementation level, if present parallel calls, if absent, serial calls.
* Have a thread pool configuration in the client config (threadcount and the such).
  * And expose the resulting threadpool in some way for clients who are adding JMX sensor or other things to their threadpools, like in my usecase.
  
These approaches would keep changes to the API small by avoiding injecting the ExecutorService in the client's pipeline method. The executorService would be a field in the client config that would be injected from there when creating a pipeline. 

I did not go for any of these approach because of the unfortunate public static variable. There might be a way to do it, but I am not sure it would be simpler if we want to maintain backward compatibility. So do we want to break people who have started depending on that static field?

As for information hiding, I am not sure I understand, a ThreadPool is not information it is resources, if it is the responsibility of the client to create it, it needs to be exposed in some way. I am running a system where resources are important, we monitor all of our threadpools for tuning, and we do this using JMX sensors. I would not use a lib that hides a threadpool.

dolavb:
@atakavci I agree there are a couple of ways to do it without adding a breaking change, but I believe the complexity is not worth it. But your call.

atakavci:
@dolavb , @ggivo please take a look at [this](https://github.com/redis/jedis/compare/master...atakavci:jedis:ali/MultiNodePipelineExecutor) and let me know what you think of.

ggivo:
@atakavci  
> downside of this, as is, the lack of mechanism to collect the `servicePool`
The downside mentioned is a critical one since there is no way to free the resources at will
Also, with this approach, we still have the configuration of the pool/executors hidden to the user, and they lack the option to optimise it for their use case. 

@dolavb @atakavci 
Providing the option for using external ExecutorService will allow for customisable `ExecutorService` based on the user's needs, and users will have control over the lifecycle of the `ExecutorService`. 

Why do we consider this to be a breaking change if we preserve the default behaviour to be the existing one?
- If an external `ExecutorService` is not explicitly provided, we still use the existing approach (create new `ExecutorService` using the statically provided configuration )
- To use shared `ExecutorService`, users should explicitly set it (via `JedisClientConfig`, or provide it when constructing the pipeline)

atakavci:
to get it clear, no breaking solutions proposed in this PR as i see.
yet i am on behalf of confirming if we are comfortable at adding Executors to public API? 
Its nice that it gives great control to the app! 
May be we should emphasize/declare more on two modes of execution with `MultiNodePipelineExecutor`, not sure.
Question: Does it make sense to introduce a simple interface that provides only `submit(...)` on it ? 
Having **shared** one and **owned** one interchangeable like the way here, could it lead to potential confusion or misuse in the future ?

src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
```diff
@@ -37,6 +38,24 @@ public MultiNodePipelineBase(CommandObjects commandObjects) {
...
+
+  /**
+   * Sub-classes must call this method, if graph commands are going to be used.
+   * @param connectionProvider connection provider
+   */
+  protected final void prepareGraphCommands(ConnectionProvider connectionProvider) {
+    GraphCommandObjects graphCommandObjects = new GraphCommandObjects(connectionProvider);
+    graphCommandObjects.setBaseCommandArgumentsCreator((comm) -> this.commandObjects.commandArguments(comm));
+    super.setGraphCommands(graphCommandObjects);
+  }
```
ggivo, line 57: Graph module support got removed with https://github.com/redis/jedis/pull/4073

dolavb:
@ggivo I agree this is a better approach.

@atakavci upon review of your gist I have the following question, why the countdownlatch? Since the introduction of CompletableFuture (JDK8), future composition is the idiomatic way to wait/merge threads. I don't see any backward compatibility concern here since JDK7 is EOL.

atakavci:
i dont mind if its countdownlatch or Futures, i just did it in min lines of change with existing code. rest is the diff editor's trick.

dolavb:
While working on this I saw [this](https://github.com/redis/jedis/pull/2731/files#r1966418712).  Do I understand correctly that we shutdown the topologyRefreshExecutor in the JedisClusterInfoCache whenever we close a pipeline?

_I apologize for the lack of clean reference. I am not the most versed with GitHub, and could not find a way to better reference this here._

ggivo:
> While working on this I saw [this](https://github.com/redis/jedis/pull/2731/files#r1966418712). Do I understand correctly that we shut down the topologyRefreshExecutor in the JedisClusterInfoCache whenever we close a pipeline?

I  don't think that is the case, but let's keep the discussion in referred PR, not to clutter this one

dolavb:
As part of this PR, should we mark `public static volatile int MULTI_NODE_PIPELINE_SYNC_WORKERS = 3;` as deprecated? With a deprecation comment like this:
`* @deprecated Client using this approach are paying the thread creation cost for every pipeline sync. Clients
   * should use refer to {@link JedisClientConfig#getPipelineExecutorProvider()} to provide a single Executor for
   * gain in performance.
 `
 I would advise for having a default Executor to be provided in the future and this would be in preparation of that change. Otherwise I will keep the comment, but remove the deprecated annotation.

dolavb:
@ggivo This makes for a bigger change but offer a path away from the current static implementation. Let me know what you think.

ggivo:
> @ggivo This makes for a bigger change but offer a path away from the current static implementation. Let me know what you think.

Sorry for the delay. Got distracted with other tasks, but I will take a look till the end of the week

src/main/java/redis/clients/jedis/JedisClientConfig.java
```diff
@@ -115,4 +115,15 @@ default boolean isReadOnlyForRedisClusterReplicas() {
...
+
+  /**
+   * If different then DEFAULT this will provide an Executor implementation that will sync/close multi node pipelines
+   * in parallel. This replaces the deprecated internal usage of new Executor Services for every pipeline, resulting in
+   * high thread creation rates and impact on latency.
+   *
+   * @return PipelineExecutorProvider
+   */
+  default PipelineExecutorProvider getPipelineExecutorProvider() {
+    return PipelineExecutorProvider.DEFAULT;
```
atakavci, line 127: i think there is an issue here with the multithread/multi-client/cluster cases compared to "as-is" implementation; they will need to wait each other to complete/release the executors which was not the case before.
i like it better the previous approach in this PR where we provide `ExecutorService` to each pipeline execution explicitly.

ggivo:
> As part of this PR, should we mark `public static volatile int MULTI_NODE_PIPELINE_SYNC_WORKERS = 3;` as deprecated? With a deprecation comment like this:

> I would advise for having a default Executor to be provided in the future and this would be in preparation of that change. Otherwise I will keep the comment, but remove the deprecated annotation.

@dolavb @atakavci 
We have 3 different issues, and we should tackle them separately. 

1)  **Performance issues caused by creating an executor per  pipeline**
     We can resolve this by allowing usage of an external executor in Pipelines, similar to the initial solution suggested with this PR. This way the lifecycle/configuration of the executor can be managed by the user of the client library in optimal way.
     
2) **Shared executor across all pipelines** from the same Jedis Client
    This should be further discussed since it will introduce a new set of issues. For example, there is no common default configuration that will work out of the box for all users having different usage patterns. If one Pipeline misbehaves, it will interfere with others using the same executor. I  suggest opening a dedicated **Issue** for this
and discuss it further.

3) **Static configuration** of default pipeline executor created on each pipeline execution,
This makes sense to me to be configurable per Jedis client. I don't think that the solution is to replace it with 2) but rather introducing a ClientOption, which will override the static one if provided. This way, keeping also the strategy "executor per pipeline execution" and preserving backward compatibility. Again I will suggest opening a dedicated issue to discuss it.

I think in this PR we should concentrate on **1)** since it will resolve the performance issue and bring immediate value and open a separate Issues/PR for **2)** and **3)**

dolavb:
Ultimately I don't mind. We have had that forked and rolled out and is working well with performance issues elsewhere in Redis.

On 1, I agree, if the previous approach is preferred, then I do not understand some of the comments in this PR.

On 2, I think this approach gives you just that, you can provide the executor you want to the client. I believe the cross pipeline concern is a non issue, executors offer enough implementations or offers you ways to implement what you need enough to cover all usage pattern. If you think of it in relation of your connection count, then it becomes simple to reason about. This is what we did by monitoring both thread and connection pool under load test.

On 3, I disagree, the static configuration as a global variable is quite a code smell and should be taken out to avoid anybody else assuming this is something accepted in this code base.

But ultimately I want the solution to the creation of executor services to be solved as it is not just bad for us, but for everyone using this, resulting in a wasteful CPU usage. I can revert back to what I had done before, even though this does I think this is not the ideal implementation. You tell me what to do, and will update the PR.

xrayw:
I think it's better to provide a Executor/ExecutorService by user, instead of create a new one for each pipeline. It's can't used in high-concurrency/cost scenarios.

If the cluster has only 3 nodes, we can even run it directly in the worker thread by
```
Executor executor = run -> run.run()
```

src/main/java/redis/clients/jedis/JedisCluster.java
```diff
@@ -376,9 +395,14 @@ public void ssubscribe(BinaryJedisShardedPubSub jedisPubSub, final byte[]... cha
 
   @Override
   public ClusterPipeline pipelined() {
-    return new ClusterPipeline((ClusterConnectionProvider) provider, (ClusterCommandObjects) commandObjects);
+    if (clusterPipelineExecutor == null) {
```
xrayw, line 398: Can we provide an overload method with executor parameter?
like 
```java
public ClusterPipeline pipelined(Executor pipelineExecutor) {
   xxx
}
```

src/main/java/redis/clients/jedis/JedisCluster.java
```diff
@@ -334,6 +347,12 @@ private JedisCluster(ClusterConnectionProvider provider, int maxAttempts, Durati
     super(provider, maxAttempts, maxTotalRetriesDuration, protocol, clientSideCache);
   }
 
+  @Override
+  public void close() {
+    super.close();
+    IOUtils.closeQuietly(this.clusterPipelineExecutor);
+  }
+
```
ggivo, line 355: Considering `JedisClientConfig` could be reused among multiple clients. Closing the `clusterPipelineExecutor` when one of the clients completes, could cause other clients initialized from the same `JedisClientConfig` to fail. 
I have concerns that the contract for who is managing the pipeline executor lifecycle is not clear. We can end up with the implementation of `PipelineExecutorProvider` which creates a new executor each time `getPipelineExecutorProvider()` is invoked, or another implementation which reuses the same instance.

ggivo (commented):
HI @dolavb 
I took a look at the second approach you are suggesting, and though it makes sense I have concerns around managing the lifecycle of the executor. 

I am rethinking if the `JedisClientConfig` is the proper place for storing the executor since it can be easily reused between multiple `JedisCllients`, and then it is not clear who should close/free the resources.

For example, the Lettuce library has separated `client resources` from `client configuration` and has provided a clear option when resources are managed by the client vs externally.

ggivo:
> Ultimately I don't mind. We have had that forked and rolled out and is working well with performance issues elsewhere in Redis.
> 
> On 1, I agree, if the previous approach is preferred, then I do not understand some of the comments in this PR.
> 
> On 2, I think this approach gives you just that, you can provide the executor you want to the client. I believe the cross pipeline concern is a non issue, executors offer enough implementations or offers you ways to implement what you need enough to cover all usage pattern. If you think of it in relation of your connection count, then it becomes simple to reason about. This is what we did by monitoring both thread and connection pool under load test.
> 
> On 3, I disagree, the static configuration as a global variable is quite a code smell and should be taken out to avoid anybody else assuming this is something accepted in this code base.
> 
> But ultimately I want the solution to the creation of executor services to be solved as it is not just bad for us, but for everyone using this, resulting in a wasteful CPU usage. I can revert back to what I had done before, even though this does I think this is not the ideal implementation. You tell me what to do, and will update the PR.

The initial solution  proposed has a small API surface change and solves the main issue, my thinking  was if wit make sense to  do the change in stages since the second solution  requires a bigger change and might require more  discussion till we get to  final solution

qingdaoheze:
Can you provide the timeout parameters? In my scenario, when the pipeline takes a long time, it is desirable to timeout and release the connection and thread early.

qingdaoheze:
Another suggestion, grouping the request params by node first, and then execute them separately, finally merge them. So that connection can be used efficiently, not wait for each other.

binbin0325 (commented):
Is it necessary to close the link again in the close method?

ggivo:
@dolavb 
We have implemented the simple approach, allowing for an external executor with https://github.com/redis/jedis/pull/4479

I believe it should allow enough flexibility to avoid creating new executors and prevent high CPU usage,
Feedback appreciated!

Let me know if it covers your use case, and if we can resolve this PR in favour of 4479

PR #4074, opened 2025-01-30
Fix the Java example code for Lists using RPUSH

sanaulla123:
In the example under Capped List on the documentation page: https://redis.io/docs/latest/develop/data-types/lists/
The Java-Sync library code example is not matching the one given in Redis CLI and also not matching the problem statement: "For example, if you're adding bikes on the end of a list of repairs, but only want to worry about the 3 that have been on the list the longest:" 

The right solution is to use `rpush` and `ltrim` the first X elements. 

This PR fixes the code to match the Redis CLI example.

### DIFF
--- a/src/test/java/io/redis/examples/ListExample.java
+++ b/src/test/java/io/redis/examples/ListExample.java
@@ -176,7 +176,7 @@ public class ListExample {
         // REMOVE_START
         assertEquals(5, res27);
         assertEquals("OK", res28);
-        assertEquals("[bike:5, bike:4, bike:3]", res29.toString());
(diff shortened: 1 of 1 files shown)


sazzad16:
@sanaulla123 Thank you for noticing.

@andy-stark-redis No other examples match with redis-cli example.

If you confirm, we can merge this PR. If you want to fix differently, let us know.

andy-stark-redis (approved):
Approved. I don't know why these were changed in the client examples, but I guess we should get them all in line with the CLI example.

sazzad16 (approved):

sazzad16:
Merged. Thanks @sanaulla123 for contribution!

PR #4077, opened 2025-01-31
[code cleanup] Jedis client to implement CommandCommands interface

ggivo:
When support for COMMAND was added we have missed adding the interface to Jedis client itself


tishun (approved):
CommandCommands interface needs to implement ObjectObjects but we can leave it for another time 🤣

PR #4078, opened 2025-02-02
Support INFO command in UnifiedJedis

sazzad16:
This PR target's  to resolve https://github.com/redis/jedis/issues/4094

In PR [4079](https://github.com/redis/jedis/pull/4079 ) is suggested simplified approach were in case of cluster, INFO` `command is send to arbitrary node.

This PR suggests broadcasting the command to all node's from the cluster and return's aggregated response in complex object

### DIFF
--- a/src/main/java/redis/clients/jedis/UnifiedJedis.java
+++ b/src/main/java/redis/clients/jedis/UnifiedJedis.java
@@ -42,6 +42,7 @@ import redis.clients.jedis.search.aggr.FtAggregateIteration;
 import redis.clients.jedis.search.schemafields.SchemaField;
 import redis.clients.jedis.timeseries.*;
 import redis.clients.jedis.util.IOUtils;
+import redis.clients.jedis.util.JedisBroadcastReplies;
 import redis.clients.jedis.util.JedisURIHelper;
 import redis.clients.jedis.util.KeyValue;
 
@@ -333,6 +334,10 @@ public class UnifiedJedis implements JedisCommands, JedisBinaryCommands,
     this.commandObjects.setBroadcastAndRoundRobinConfig(this.broadcastAndRoundRobinConfig);
   }
 
+  public final JedisBroadcastReplies broadcastCommandDifferingReplies(CommandObject commandObject) {
+    return executor.broadcastCommandDifferingReplies(commandObject);
+  }
+
   public Cache getCache() {
     return cache;
   }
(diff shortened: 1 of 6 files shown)


src/main/java/redis/clients/jedis/util/JedisBroadcastReplies.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * The collection of replies where a command is broadcasted to multiple nodes, but the replies are expected to differ
+ * even in ideal situation.
+ */
+public class JedisBroadcastReplies {
+
+  private static final Object DUMMY_OBJECT = new Object();
+
+  private final Map<Object, Object> replies;
```
uglide, line 15: Why not `Map<HostAndPort, Object> replies;` or NodeID?

sazzad16: Because we don't a specified solution for this, e.g. in ConnectionProvider and CommandExecutor interfaces, yet. For example, in ConnectionProvider interface the respective method definition is `Map<?, ?> getConnectionMap()`.

src/main/java/redis/clients/jedis/UnifiedJedis.java
```diff
@@ -353,6 +358,16 @@ public String configSet(String parameter, String value) {
     return checkAndBroadcastCommand(commandObjects.configSet(parameter, value));
   }
 
+  public final JedisBroadcastReplies info() {
```
ggivo, line 361: It will be a bit strange API when connecting to standalone Redis  using let's say UnifiedJedis.
Then you will get a `JedisBroadcastReplies` which makes sense for the cluster. 
But in case of standalone, you will get `JedisBroadcastReplies`, and to get to actual value need to invoke 'JedisBroadcastReplies::getReplies(with host/port)"

ggivo:
@uglide @sazzad16 
Spent some time looking at the suggestion for adding INFO command to UnifiedJedis, and here are my suggestions

1. In UnifedJedis we provide ` info()`  command implementation with proper interface for standalone invocation

```
  public  String info()
...
  public String info(String section) 
```
2. In `JedisCluster`, we provide an additional `infoAllNodes() `method that broadcasts to all nodes and returns replies from each node.

```
public final JedisBroadcastReplies info() {
  ...
public final JedisBroadcastReplies info(String section) {
  
```
3. `broadcastCommandDifferingReplies` is moved toward `ClusterCommandExecutor`  and removed from `'CommandExecutor'`
No need to broadcast in UnifiedJedis itself, and this will allow us to get rid of the DUMMY key and use typed reply map  `Map<HostPort, T>` for the reply list.

Here is a draft with the proposal. 
https://github.com/ggivo/jedis/commit/c4cbd1af2cea8f058603ed0bf411f6e7e7360d13

ggivo:
After a discussion with the team was decided to proceed with approach selected in PR [4079](https://github.com/redis/jedis/pull/4079),

PR #4087, opened 2025-02-07
Enforce code style format

atakavci:
This is an initiative to make the code formatting enforced on pipeline. 
We will start with small steps and try to cover whole code base in multiple iterations.
The plan is step by step something like this (i guess this makes some sense as is, but lets improve the plan itself as well starting with this PR);
- improve the style file `hbase-formatter.xml` as we see the opportunity
- introduce mandatory validation step into pipeline starting with a [small perimeter](https://github.com/atakavci/jedis/blob/39a8466e662c079e0bd4df11fefc6abbc97e4571/pom.xml#L302) (only for a single folder with a few files as of this PR)
- make sure the files covered by mandatory validation are well formatted and cause no errors on pipeline 
- add another workflow(`format_check.yml`) to the pipeline to follow if the set of files in any PR is well formatted
- add more directories in iterations with fixing formatting errors and fine tuning style format itself as suggested above

i think this approach will help us to cover more and more directories adding it into [pom.xml](https://github.com/atakavci/jedis/blob/39a8466e662c079e0bd4df11fefc6abbc97e4571/pom.xml#L301). Later at some point we ll have full coverage and get rid of directories section in `pom.xml`

### DIFF
--- a/.github/workflows/format_check.yml
+++ b/.github/workflows/format_check.yml
@@ -1,0 +1,44 @@
+name: Java Format Check
+
+on:
+  pull_request:
+    paths:
+      - '**/*.java' # Only trigger for Java file changes
+
+jobs:
+  check-format:
+    runs-on: ubuntu-latest
+
+    steps:
+      # Step 1: Checkout the PR code
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      # Step 2: Set up Java (if needed for format check tools)
+      - name: Set up JDK
+        uses: actions/setup-java@v3
+        with:
+          java-version: '11' # Use the version you need for your project
+
+      # Step 3: Get a list of changed Java files in the PR
+      - name: Get changed Java files
+        id: changed_files
+        run: |
+          echo "::group::Changed Java Files"
+          git diff --name-only origin/${{ github.event.pull_request.base.ref }} | grep '\.java$' || true
+          echo "::endgroup::"
+          echo "::set-output name=files::$(git diff --name-only origin/${{ github.event.pull_request.base.ref }} | grep '\.java$' || true)"
+
+      # Step 4: Check formatting of the changed files
+      - name: Check Java file format
+        run: |
+          CHANGED_FILES="${{ steps.changed_files.outputs.files }}"
+          if [ -z "$CHANGED_FILES" ]; then
+            echo "No Java files changed."
+          else
+            for FILE in $CHANGED_FILES; do
+              echo "Checking $FILE"
+              FILE_NAME=$(basename "$FILE")
+              mvn formatter:validate -f formatter-pom.xml "-Dformatter.includes=**/Pipeline.java" "$FILE"
+            done
+          fi
--- a/.github/workflows/format_check.yml
+++ b/.github/workflows/format_check.yml
@@ -30,7 +30,8 @@ jobs:
         id: changed_files
         run: |
           echo "::group::Changed Java Files"
-          CHANGED_FILES=$(git diff --name-only origin/master | grep '\.java$' || true)
+          echo Base Branch: ${{ github.event.pull_request.base.ref }}
+          CHANGED_FILES=$(git diff --name-only origin/${{ github.event.pull_request.base.ref }} | grep '\.java$' || true)
           echo "$CHANGED_FILES"
           echo "::endgroup::"
           # Write the multiline content to a file
--- a/.github/workflows/format_check.yml
+++ b/.github/workflows/format_check.yml
@@ -31,20 +31,21 @@ jobs:
         id: changed_files
         run: |
           echo "::group::Changed Java Files"
-          git diff --name-only origin/${{ github.event.pull_request.base.ref }} | grep '\.java$' || true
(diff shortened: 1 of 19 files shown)


.github/workflows/format_check.yml
```diff
@@ -0,0 +1,68 @@
+name: Java Format Check
```
tishun, line 1: Do we need an additional action just for that? Could we do it as part of the integration?

If we keep it as a separate action it would do a lot of unnecessary new things  - power up a container, fetch the contents of the PR, set Java up ...

.github/workflows/format_check.yml
```diff
@@ -0,0 +1,68 @@
...
+
+      # Step 3: Fetch latest changes
+      - name: Fetch latest changes
+        run: git fetch origin
+
+      - name: Get changed Java files
+        id: changed_files
+        run: |
+          echo "::group::Changed Java Files"
+          CHANGED_FILES=$(git diff --name-only origin/master | grep '\.java$' || true)
```
tishun, line 33: This is a bit brittle when we work with branches different than the main branch, we might want to add a parameter to indicate the branch we are merging to

formatter-pom.xml
```diff
@@ -0,0 +1,37 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
```
tishun, line 1: How does having an additional pom.xml help in this case?

tishun (commented):
Much needed PR I think, thanks for working on that!

.github/workflows/format_check.yml
atakavci: The goal here is to confirm that everything developer touches within the PR is well formatted. And, if not, signal this to the developer. If we add this to integration and it fails, the PR won't meet the merge criteria. If we add it and it doesn't fail, then we won't have added a sufficiently noticeable warning.
My intention here is to avoid forcing developer to have everything well formatted with just a single line change in file.
Not to force him, but to give a striking warning.

formatter-pom.xml
atakavci: Check my previous response around having that a more noticeable warning but not forcing it.

tishun: Having a separate pom.xml means that any PR contributor needs to know that they have to call it (as a separate step to, say `make test`) and this makes it harder for contributors to know what they need to do before they can merge a PR.

In an ideal world `mvn install` or `make test` should be the only thing a developer needs to run (successfully) to ensure the CI will pass.

.github/workflows/format_check.yml
tishun: So the goal is to only add an action that shows formatting issues without enforcing them?

Do we want even more code being merged without following the formatting rules?

.github/workflows/format_check.yml
tishun: Thanks!

.github/workflows/format_check.yml
atakavci: True. On the other hand, in case we force it, having a single line of change, contributor needs to fix all formatting issues in entire file to pass. this is a readability issue in PRs, especially with large files, both for contributor and reviewers.
Having a perimeter that we **enforce** to keep it formatted and gradually expanding it with `<directories>` in [pom.xml](https://github.com/redis/jedis/pull/4087/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8) is a good approach.

tishun: Okay, I see your point.

IMHO we need to reformat the entire code base with one PR and pay the price of code difference when merging back to older branches. But if you and @ggivo are on the same page I am okay with this intermediate step.

formatter-pom.xml
tishun: As with my other comment I think we should just make the change today. Just an opinion.

We could also merge this as it is and see if it is helping use get where we want to get or not.

tishun (approved):
I leave it to you folks to decide if we need this intermediate step.
Overall the change looks good.

.github/workflows/format_check.yml
ggivo: Both approaches come with cons.
     If we do it gradually, PR's reviews will be harder, containing lots of formatting changes only
     If we go with formatting everything at once, most likely, we will end up with a number of pending PR's in conflict state

I am rather inclined to, as @tishun mentioned, "pay the price at once" and reformat all the code at once and enforce from then on code style. 

I also think that if we go with the direction to have it as a warning only and not mandatory, it will take a long time before ending up in good state

ggivo (approved):
LGTM

I think it is a step in the right direction. Will get at least a warning.
Let's see how it goes and we can revisit it  decide if we want to reformat all at once

atakavci:
> Let's see how it goes and we can revisit it decide if we want to reformat all at once

 right !! this is not a one way road, we can go for a single shot whenever we decide.

PR #4090, opened 2025-02-11
Extend EXECABORT with "previous errors" #4084

### LINKED ISSUE #4084: Extend EXECABORT with "previous errors"
In a MULTI/EXEC environment, Jedis might receive the following response when an issue is detected:

```
JedisDataException: EXECABORT Transaction discarded because of previous errors.
```

In such cases, no information is included in the exception and because of the exception, the `redis.clients.jedis.Response` value is never set and throws an exception when the user tries to read it.

```
IllegalStateException: Please close pipeline or multi block before calling this method.
```

Which is a bit misleading because the multi block was closed.

InpuStream example:
```
+OK
+QUEUED
-CROSSSLOT Keys in request don't hash to the same slot (context='within MULTI', command='get', original-slot='9842', wrong-slot='5649', first-key='1', violating-key='2')
-CROSSSLOT Keys in request don't hash to the same slot (context='within MULTI', command='get', original-slot='9842', wrong-slot='1584', first-key='1', violating-key='3')
```

Is it possible to include these in the exception or to populate the `Response` objects and let the user find the relevant errors?

ggivo:
Addresses #4084

[see : Transactions](https://redis.io/docs/latest/develop/interact/transactions/)
A command may fail to be queued, so there may be an error before EXEC is called. For instance the command may be syntactically wrong (wrong number of arguments, wrong command name, ...), 

Fix:  Errors were already collected but not reported. With this change, they are added to the final exception as suppressed.
 
closes #4084

### DIFF
--- a/src/main/java/redis/clients/jedis/Transaction.java
+++ b/src/main/java/redis/clients/jedis/Transaction.java
@@ -169,11 +169,25 @@ public class Transaction extends TransactionBase {
     try {
       // ignore QUEUED (or ERROR)
       // processPipelinedResponses(pipelinedResponses.size());
-      connection.getMany(1 + pipelinedResponses.size());
+      List<Object> queuedCmdResponses = connection.getMany(1 + pipelinedResponses.size());
+
 
       connection.sendCommand(EXEC);
 
-      List<Object> unformatted = connection.getObjectMultiBulkReply();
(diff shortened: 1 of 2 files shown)


tishun (approved):
LGTM

PR #4091, opened 2025-02-18
Add tests for vector search INT8/UINT8 types

ggivo:
Adding test to verify support for new vector types INT8/UINT8 added to RediSearch


atakavci (approved):
LGTM ! 
is this complies with formatting already? how about making line length 120

ggivo:
> LGTM ! is this complies with formatting already? how about making line length 120

Initially, I formatted the complete files, but it became unreadable (actual change mixed with formatting). In the end, I formatted only the new code. I think we should review the formatting approach. 
I am fine with making the line 120

PR #4093, opened 2025-02-18
Migrate ‘Build and Test’ Job to Containerized Test Environment with Redis Version Matrix

ggivo:
The `“Build and Test”` job is currently required for PR merges but relies on a legacy approach that sets up the test infrastructure from Redis source. With this change, we are switching `"Build and Test using containerized environment"` to be used in place `"Build and Test”`  for PR verification. 

✅ “Build and Test” to use a containerized test environment instead of the legacy setup.
✅ Test matrix to run tests against multiple Redis server versions is now mandatory
✅ Aggregating test results so that if any matrix job fails, the “Build and Test” job is marked as failed.


ggivo:
Ended up marking `Build and Test (7.2, 7.4, 8.0)` as required for main branch and 
existing `Build and Test` against Redis `unstable` Redis branch as an optional one.

PR #4101, opened 2025-02-25
Update stale issue workflow

ggivo:
- Mark the Issue as stale only if marked with waiting-for-feedback
  - Do not mark Issue/PR as stale if marked with feedback-provided
  - Do not mark the Issue/PR as stale if assigned to a milestone

### DIFF
--- a/.github/workflows/stale-issues.yml
+++ b/.github/workflows/stale-issues.yml
@@ -18,7 +18,7 @@ jobs:
         stale-issue-message: 'This issue is marked stale. It will be closed in 30 days if it is not updated.'
         stale-pr-message: 'This pull request is marked stale. It will be closed in 30 days if it is not updated.'
         days-before-stale: 30
-        days-before-close: 14
+        days-before-close: 30
         stale-issue-label: "stale"
         stale-pr-label: "stale"
         operations-per-run: 10
--- a/.github/workflows/stale-issues.yml
+++ b/.github/workflows/stale-issues.yml
@@ -17,12 +17,12 @@ jobs:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         stale-issue-message: 'This issue is marked stale. It will be closed in 30 days if it is not updated.'
         stale-pr-message: 'This pull request is marked stale. It will be closed in 30 days if it is not updated.'
-        days-before-stale: 365
-        days-before-close: 30
+        days-before-stale: 30
+        days-before-close: 14
         stale-issue-label: "stale"
         stale-pr-label: "stale"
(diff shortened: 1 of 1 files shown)


atakavci (commented):
i think waiting a year on an issue without labels `waiting-for-feedback` or `waiting-for-triage` is good enough reason to close. Are we confident with the exempt rules ?

ggivo:
> i think waiting a year on an issue without labels `waiting-for-feedback` or `waiting-for-triage` is good enough reason to close. Are we confident with the exempt rules ?
@atakavci 
The issue is that 'stale' rule was introduced an year ago, and there are number of issue which wait for team action but are still marked as stale and get removed. If we remove the stale tag in 30 days it is again , marked as stale and will need to review it again

tishun:
Probably because it is Friday afternoon but I do not get anything here 🤯 

The typical flow is:
- any issue - if untriaged - is skipped by the flow
- if an issue is triaged with "waiting-for-feedback" and more than X days have passed we mark it as stale
- if an issue has been stale for Y days we close it 

Why would this flow not work right now?

Btw 365 days is insane IMHO. My suggestions for X and Y are 30 days and 14 days respectively (month an a half total)

ggivo:
> Why would this flow not work right now?

@tishun 
Current flow does not consider `only-labels: 'waiting-for-feedback'` and will mark any issue as stale after the inactivity period is reached, even issues which are not waiting on any feedback but rather to someone to start working on them, or need triage.

The suggested workflow & periods make sense to me let me update the days-before-xx settigns
```
        days-before-stale: 30
        days-before-close: 14
```

atakavci:
> even issues which are not waiting on any feedback but rather to someone to start working 
> on them, or need triage.

so it sounds like it would be ok if we introduce another label than `waiting-for-triage`,  and tag the ones "waiting for someone" with it. isnt it `waiting-for-triage` enough?

ggivo:
> so it sounds like it would be ok if we introduce another label than `waiting-for-triage`, and tag the ones "waiting for someone" with it. isnt it `waiting-for-triage` enough?

@atakavci  
That is pretty much the idea. 
`waiting-for-triage` - means the team needs to triage the issue and process it, at that step we decide if we need additional info, is there a reproducible example provided ... , and we can decide to request this info and mark it `'waiting-for-feedback'` 

`waiting-for-feedback`  - means we have triaged the issue and requested additional inputs, if it is not provided in a reasonable time frame, mark the issue as stale and schedule it for closure.

atakavci (approved):
LGTM

PR #4113, opened 2025-03-13
Test support for DefaultAzureCredential

atakavci:


ggivo (approved):
LGTM

src/test/java/redis/clients/jedis/authentication/RedisEntraIDIntegrationTests.java
```diff
@@ -378,4 +382,22 @@ private void triggerNetworkFailure() {
...
+        .build();
+
+    DefaultJedisClientConfig jedisConfig = DefaultJedisClientConfig.builder()
+        .authXManager(new AuthXManager(tokenAuthConfig)).build();
+
+    try (JedisPooled jedis = new JedisPooled(hnp, jedisConfig)) {
+      String key = UUID.randomUUID().toString();
+      jedis.set(key, "value");
+      assertEquals("value", jedis.get(key));
+      jedis.del(key);
```
ggivo, line 400: How about verifying the actual username logged in e.g. (acl whoami)
Clearer intent that we are testing auth, and prevent to some degree green test if server does not require auth

ggivo (commented):
On as second thought  if possible we can use `acl whoami ` verify actually logged use is excpected one

src/test/java/redis/clients/jedis/authentication/RedisEntraIDIntegrationTests.java
atakavci: do you mean the username in the jedisConfig instance or hardcoding some user on specific to azure env ?
First would be the same since it already uses the one in config,, latter is kind of future maintenance issue in case of changing config or working in another setup.

ggivo: I guess we can extract the expeced user from the AuthXManager used and validate that the connection is actually using it 

``` 
AuthXManager manager = new AuthXManager(tokenAuthConfig);
manager.get().getUser()
```

atakavci: @ggivo , sorry i just saw this one. i think both solutions does good enough for the test. What you suggest also confirms the endpoint configuration is the right one in the test setup. Would you like me to drop another PR ?

PR #4114, opened 2025-03-18
Update redis server test versions 

ggivo:
Test with 8.0-M05-pre


ggivo:
Bump test infra to use 8.0-M05-pre

PR #4115, opened 2025-03-18
Add TTL support for CSC

wp973:
Add TTL support for CSC.

### DIFF
--- a/src/main/java/redis/clients/jedis/csc/AbstractCache.java
+++ b/src/main/java/redis/clients/jedis/csc/AbstractCache.java
@@ -23,15 +23,17 @@ public abstract class AbstractCache implements Cache {
   private Cacheable cacheable;
   private final Map<ByteBuffer, Set<CacheKey<?>>> redisKeysToCacheKeys = new ConcurrentHashMap<>();
   private final int maximumSize;
-  private ReentrantLock lock = new ReentrantLock();
+  private final long ttl;
+  ReentrantLock lock = new ReentrantLock();
   private volatile CacheStats stats = new CacheStats();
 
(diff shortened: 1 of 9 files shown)


wp973:
Can someone help me take a look at this PR? :D

ggivo:
Hi @wp973 

Thank you for your contribution! The team is currently busy, so it may take some time before we can take a closer look at the suggested changes.

A brief review suggests that this PR could introduce significant performance considerations, especially for larger caches. Given that, I don’t think we should go down the path of implementing a full-blown cache within the project.

This feature request requires further discussion. Let's open an Issue where we can better understand the use cases and explore alternative approaches. One potential direction could be providing an example of an adapter using a third-party caching library instead of maintaining a custom implementation.

ggivo:
> Using popular third-party caching library(e.g, Caffeine) is a good choice, but it will also introduce third-party dependencies. Is this acceptable to Jedis?

I don't think we should add a third-party dependency but just document/provide an example how to create an adapter of the `Cache` interface that can be used with Jedis using an arbitrary third-party library.

ggivo:
Related to  https://github.com/redis/jedis/issues/4117

ggivo:
@wp973 
Is this PR still relevant or it is superseded by https://github.com/redis/jedis/pull/4120?

wp973:
> > Using popular third-party caching library(e.g, Caffeine) is a good choice, but it will also introduce third-party dependencies. Is this acceptable to Jedis?
> 
> I don't think we should add a third-party dependency but just document/provide an example how to create an adapter of the `Cache` interface that can be used with Jedis using an arbitrary third-party library.

Got it. Maybe I can try to provide an example. :D

PR #4118, opened 2025-03-21
Fix failing network partition test

atakavci:
in the scenario test environment (the only place where this test works as of today), we have identified a critical timing-related issue with the impact of failure injection. Failure injection is initiated by the test itself, but there are no mechanism to:
- confirm the exact moment when the failure injection takes effect
- validate the completion of propagation of the injected failure

due to a bunch of reasons, failure injection might take longer than seconds, while our tests proceed relatively fast. As a result, tests fail prematurely—before the failure injection even takes affect.

To address this issue,  we will try to utilize the injection API's built-in check mechanism. This will help us keep asserting and allow a sufficient time to receive the impact. 
New implementation (using the injection API's `isCompleted` ) will keep trying **during** the given time window **after** the propogation is done and status confirmed from the calling side.


src/test/java/redis/clients/jedis/authentication/RedisEntraIDIntegrationTests.java
```diff
@@ -341,34 +346,44 @@ public void networkPartitionEvictionTest() {
...
+      TriggerActionResponse actionResponse = triggerNetworkFailure();
 
       JedisConnectionException aclException = assertThrows(JedisConnectionException.class, () -> {
-        for (int i = 0; i < 50; i++) {
-          String key = UUID.randomUUID().toString();
-          jedis.set(key, "value");
-          assertEquals("value", jedis.get(key));
-          jedis.del(key);
+        while (!actionResponse.isCompleted(ONE_HUNDRED_MILLISECONDS, TWO_SECONDS, FIVE_SECONDS)) {
+          for (int i = 0; i < 50; i++) {
```
ggivo, line 353: How about using a blocking command like  `bloop`.  Should be more predictable than relaying  on multiple get/set

- issue blpop (guarantee that we have active connection)
- trigger network failure (this should drop the connection )
- wait for action to complete and validate there was a connection exception

tishun: TBH these are really two different test scenarios:
- what happens when there is a network failure during a blocking operation execution
- what happens when there is a network failure during executing multiple short-lived operations

In a perfect world we should be testing both, but in this case I would go for either.

tishun (approved):

src/test/java/redis/clients/jedis/authentication/RedisEntraIDIntegrationTests.java
ggivo: don't think we should test both, 
Looking at the test, we are trying to simulate network failure while the connection is being actively used and check for exceptions. Blocking operation should keep the connection opened till network issue is simulated

ggivo: @atakavci 
What about adding a timeout to the test itself?

atakavci: > Should be more predictable than relaying on multiple get/set ...

i am not strongly against it but its not clear to me what would improve between;
- multiple calls where we expect any one to fail in n seconds
- single call we expect to fail in n seconds

ggivo: I thought some flakiness might come from not throwing an exception between the calls if the server is reconnected too fast,  but it seems not very likely.
Agree, let's keep it as it is.

Issue #4117, opened 2025-03-21
Add TTL support for CSC


wp973:
#### Feature Request

The default local cache implementation of CSC does not include the TTL. This may cause the following problems:

- The data in the local cache is **inconsistent** with the data in Redis.

#### Some suggestions:
- Add support for TTL in the default cache implementation.
- Provide adapter examples for using third-party cache libraries(e.g, caffeine).

ggivo:
The idea of CSC is a bit different.
In the case of client-side caching, the protocol propagates invalidation signals and triggers cache evictions.

If more customizations are required, there are a number of libraries that support a rich set of caching options like caffeine, guava... I think we should not add additional complexity to Jedis itself by implementing/maintaining additional generic cache features.

Still we are missing a good documentation how to use external cache implentaiton

ggivo:
By missing a good documentation I mean for example document/provide an example of how to create an adapter of the Cache interface that can be used with Jedis using an arbitrary third-party library.

PR #4120, opened 2025-03-23
Feature/csc ttl support

PengJingzhao:
Hello, I add a ttl support for the jedis csc (extend the DefaultCache) to solve a issue of the jedis.You can read my commit record for the details

### DIFF
--- a/src/main/java/redis/clients/jedis/TTLCache.java
+++ b/src/main/java/redis/clients/jedis/TTLCache.java
@@ -50,10 +50,9 @@ public class TTLCache extends DefaultCache {
 
     private void initCleanupTask() {
 
-        // decide the time interval of the cleaning up
-        long timeInterval = Math.max(this.defaultMilis / 10, 1000);
-
         if (defaultMilis > 0) {
+            // decide the time interval of the cleaning up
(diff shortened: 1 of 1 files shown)


PengJingzhao:
If somebody help me review it? Thankyou very much!

ggivo:
Related to https://github.com/redis/jedis/issues/4117

ggivo:
There is an ongoing discussion regarding adding TTL to CSC in Jedis  tracked with https://github.com/redis/jedis/issues/4117
Can we take a step back and try to identify what is the issue we are trying to solve before getting to the implementation?

PengJingzhao:
Of course! I can explain in detail what my PR does. In my PR, I extend the DefaultCache class under the csc package. In the extended class, I use a Map called expirationTimes to maintain an expiration time for each cache key. Additionally, I start a thread to periodically clean up keys that have already expired.

ggivo:
My question was about the problem we want to solve before going into concrete implementations. 
CSC is not intended to be a generic cache and is driven by notification from Redis server. Instead of providing an TTL implementation inside Jedis, another approach is to use integrate a third-party cache library instead of the DefaultCache implementstion provided.

Issue #2320, opened 2020-12-21
Use user other than default in ACL tests


sazzad16:
We are using the `default` user for ACL tests. Because of that some bugs are getting uncaught even when the user param gets missed.

sazzad16:
#2318

sazzad16:
Partially resolved by #2325 in 1a9de794847c6f7bfda51249225ead066f9f4387 . These are mostly Jedis, Jedis with SSL and JedisPool. We still have more tests to cover.

YoHanKi:
Hi, my friend and I are interested in taking on this good first issue. We’re still junior developers, so it might be a bit challenging for us, but we’re really eager to learn more about Jedis and contribute where we can. Before we dive in, could you please confirm if it’s alright for us to work on this issue? Additionally, any guidance on what the test code should target and where we should start would be greatly appreciated. Thanks for your help!

ggivo:
Hi @YoHanKi 

Anyone is welcome to contribute. Feel free to work on this issue.

I would suggest reviewing the fix for the initial [issue](https://github.com/redis/jedis/issues/2318) reported and also https://github.com/redis/jedis/pull/2325 where tests are added. 

The original issue itself was missing `username` parameter in the constructor,  which led to using **default** instead of the provided `username`.  I would start with reviewing  submitted tests and check if there are other places were we configure username/password missing tests.

I would also try to figure out if this can be tested with unit tests (not integration ones) 
Can we have a dedicated Unit test for constructors of JedisPool/UnifeidJedis..  which verifies that the provided username/password reaches ConnectionFactory. Without the need of explicit test env

YoHanKi:
Hi @ggivo,

I’ve implemented some tests based on your guidance, but there are a few points that I need clarification on. Could you please provide some clarification on the questions mentioned in the PR description?

Thank you very much for your time and support! 😊

MathJAY10:
Hey @ggivo, I am a beginner looking to make my first open-source contribution and this issue aligns perfectly with my Java background. I see that @YoHanKi has already submitted a PR for JedisPool and UnifiedJedis constructors.

Could you please let me know if there are other areas or connection components where ACL tests are still missing? I would love to pick up the remaining parts of this issue. Thanks!

ggivo:
Hi @MathJAY10 , thanks for the enthusiasm — great to see you looking to contribute!

[...]

PR #4121, opened 2025-03-23
#2320: Unit Tests for ACL Credential Propagation in JedisPool/UnifiedJedis

YoHanKi:
- #2320 Unit test for constructors of JedisPool/UnifeidJedis which verifies that the provided username/password reaches ConnectionFactory. Without the need of explicit test env

### Questions
- In your previous feedback, you mentioned, "I would start with reviewing submitted tests and check if there are other places where we configure username/password with missing tests." Does this mean that username/password should be passed in every test, or are you referring to specific tests (e.g., JedisPoolTest or JedisPooledTest)?

- Regarding the approach of using a test-specific subclass (TestJedisFactory) to verify, without using Reflection, that the ACL credentials provided to the constructor are correctly reflected in the internal JedisClientConfig—is this approach acceptable? Are there any additional aspects we should check?

- Additionally, we would like to confirm if the placement of these tests is appropriate.

### DIFF
--- a/src/test/java/redis/clients/jedis/JedisPoolTest.java
+++ b/src/test/java/redis/clients/jedis/JedisPoolTest.java
@@ -19,6 +19,7 @@ import org.junit.Test;
 import redis.clients.jedis.exceptions.InvalidURIException;
 import redis.clients.jedis.exceptions.JedisConnectionException;
 import redis.clients.jedis.exceptions.JedisException;
+import redis.clients.jedis.providers.PooledConnectionProvider;
 
 public class JedisPoolTest {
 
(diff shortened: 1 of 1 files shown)


ggivo:
Hi @YoHanKi, 
Thanks for spending time on this issue an sorry I was not more specific.

With  [commit](https://github.com/redis/jedis/commit/1a9de794847c6f7bfda51249225ead066f9f4387#diff-d626b1fec1b34ead15e76b90317c5dcb4bb9112ce057808fda60d7f3d6736a3c)   we have introduced **dedicated** test user **acljedis** different than default one

`user acljedis on all commands allkeys >fizzbuzz `

and added integration tests `JedisWithCompleteCredentialsTest.java,` `JedisPoolWithCompleteCredentialsTest.java,` `SSLJedisWithCompleteCredentialsTest.java`  to connect with non-**default** user.

In the comment https://github.com/redis/jedis/issues/2320#issuecomment-751587925 it is mentioned that there are still missing tests, what I meant is to try to identify them and add any missing ones. 

Optionally we can create a unit test for JedisPoolUnitTest, ..., which exercises JedisPool object constructions.
A unit test should be placed in a separate test class.

> `Regarding the approach of using a test-specific subclass (TestJedisFactory) to verify, without using Reflection, that the ACL credentials provided to the constructor are correctly reflected in the internal JedisClientConfig—is this approach acceptable? Are there any additional aspects we should check?`

Another option instead of using Reflection  is to use [Mocking object construction](https://javadoc.io/static/org.mockito/mockito-core/3.5.9/org/mockito/Mockito.html#mocked_construction) 

Hope it helps...

YoHanKi:
Hello @ggivo ,

Thank you for your guidance.

I have written tests based on your feedback and did my best, though I feel my skills are still quite limited.
If you could kindly point out any additional improvements or errors, I would be more than happy to revise them. 

I really appreciate your support for a junior developer like me.

ggivo:
@YoHanKi 
Sorry for the delayed response. Got dragged on other assignments.
ACLJedisPoolTest & ACLJedisTest looks good to me.
I have my doubts about mixing Mock's with actual env in JedisPoolUnitTest.java and need to take a closer look. Will try to spend some time on it in the upcoming days.

YoHanKi:
Hello, @ggivo

Thank you very much for your kind attention and review. Before you looked into JedisPoolUnitTest, I was concerned that the previous usage of magic numbers (such as context.arguments().get(1)) might lead to potential issues, so I modified it accordingly. If you feel that the previous approach is more suitable, I would be more than happy to revert the change.

Once again, I sincerely appreciate your thoughtful review and support.

PR #4125, opened 2025-03-27
Document pgp keys

### LINKED ISSUE #3125: Please document your gpg keys
We use [Dependency Verification](https://docs.gradle.org/current/userguide/dependency_verification.html) and currently there is no documentation stating which keys are safe

Could you please document this?

Here are some examples of other projects documenting what key they use to sign their artifacts.

https://github.com/qos-ch/slf4j/blob/master/SECURITY.md#verifying-contents
https://square.github.io/okhttp/security/security/#verifying-artifacts
https://downloads.apache.org/logging/KEYS

ggivo:
Closes #3125


uglide (changes requested):
I think we should move this page to the `docs` folder and add it to the navbar as "Verifying artifacts"

PR #4131, opened 2025-04-03
Fix failing ModuleTest & test against 8.0-RC1-pre

ggivo:
**Fix failing ModuleTest.java**

The vector-sets module is a part of Redis Core and is available by default with  commit https://github.com/redis/redis/commit/78e0d87177f21b3d6823bbf78ac7a45b34a9bb25
  
Module test was expecting there are no other modules loaded on the bare Redis server which is no longer the case.
Update the test to check explicitly for loading/unloading of the test module

**Fix SearchWithParamsTest**
After bump to 8.0-RC1 error message got changed,


atakavci (approved):
LGTM

PR #4134, opened 2025-04-05
Update "Pipelines and transactions" doc example to include pipeline.close()

### LINKED ISSUE #4132: Update "Pipelines and transactions" doc example to include pipeline.close()
https://redis.io/docs/latest/develop/clients/jedis/transpipe/

The existing example does not perform close on the pipeline at the end and in case we try to open more than  max available pool connections is it will block.

Erroneus pipeline usage 
```java
   //  If pipeline is not closed, and max pool size is reached, it will block
    try (UnifiedJedis jedis = new UnifiedJedis("redis://localhost:6379")) {
      for (int i = 0; i < 10; i++) {
        AbstractPipeline pipeline = jedis.pipelined();
        pipeline.set("var", "do some work");
      }
```
 Correct: 
```java
   try (UnifiedJedis jedis = new UnifiedJedis("redis://localhost:6379")) {
      for (int i = 0; i < 10; i++) {
        try (AbstractPipeline pipeline = jedis.pipelined()) {
          pipeline.set("var", "do some work");
        }
      }
    }
```

ggivo:
The example PipeTransExample.java  is updated to use a try-with-resources block to ensure the pipeline is properly closed after use.

Failing to close pipelines can lead to resource leaks. Specifically, if pipelines are repeatedly created without being closed, and the number of open connections exceeds the maximum allowed by the connection pool, the application will block.

closes #4132

### DIFF
--- a/src/test/java/io/redis/examples/PipeTransExample.java
+++ b/src/test/java/io/redis/examples/PipeTransExample.java
@@ -27,42 +27,49 @@ public class PipeTransExample {
         // REMOVE_END
 
         // STEP_START basic_pipe
-        AbstractPipeline pipe = jedis.pipelined();
+        // Pipeline is a way to send multiple commands to Redis in a single request.
+        // It can be used to improve performance by reducing the number of round trips to the server.
+        // Make sure to close the pipeline after use to release resources and return the connection to the pool.
+        try (AbstractPipeline pipe = jedis.pipelined()) {
 
-        for (int i = 0; i < 5; i++) {
-            pipe.set(String.format("seat:%d", i), String.format("#%d", i));
+            for (int i = 0; i < 5; i++) {
+                pipe.set(String.format("seat:%d", i), String.format("#%d", i));
+            }
+
+            pipe.sync();
         }
 
-        pipe.sync();
+        try (AbstractPipeline pipe = jedis.pipelined()) {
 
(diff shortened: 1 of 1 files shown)


src/test/java/io/redis/examples/PipeTransExample.java
```diff
@@ -27,42 +27,49 @@ public void run() {
         // REMOVE_END
 
         // STEP_START basic_pipe
-        AbstractPipeline pipe = jedis.pipelined();
+        // Pipeline is a way to send multiple commands to Redis in a single request.
+        // It can be used to improve performance by reducing the number of round trips to the server.
+        // Make sure to close the pipeline after use to release resources and return the connection to the pool.
```
andy-stark-redis, line 32: We probably don't need the explanation part here (the code is used in [this page](https://redis.io/docs/latest/develop/clients/jedis/transpipe/), which already has an explanation of pipelines). Also, the width of the code example tab is quite small and we haven't implemented scrolling yet, so it looks a bit neater if we avoid the long comment line.
```suggestion
        // Make sure you close the pipeline after use to release resources
        // and return the connection to the pool.
```

andy-stark-redis (approved):
I've tested this with a doc build and the example looks fine. Just a minor suggestion but otherwise LGTM.

PR #4143, opened 2025-04-14
Update README.md

jujn:
Original Text:
"The table below highlights version compatibility of the most-recent library versions and Redis versions."
Proposed Change:
"The table below highlights the version compatibility of the most-recent library versions with Redis versions and JDK versions."
Explanation:
This change provides a more comprehensive description of the table's content by explicitly mentioning JDK version compatibility.


ggivo (approved):
Thanks for the contribution! The clarification makes sense. Made slide rewording

Issue #4148, opened 2025-04-23
[MultiNodePipelineBase] perf cluster pipeline if all the keys on same node


xrayw:
I think there are some issues at here.

https://github.com/redis/jedis/blob/1d1b268f9be5b093f6684bc814d2fdab376a1ee3/src/main/java/redis/clients/jedis/MultiNodePipelineBase.java#L98

```java

ExecutorService executorService = Executors.newFixedThreadPool(MULTI_NODE_PIPELINE_SYNC_WORKERS);

CountDownLatch countDownLatch = new CountDownLatch(pipelinedResponses.size());
Iterator<Map.Entry<HostAndPort, Queue<Response<?>>>> pipelinedResponsesIterator
    = pipelinedResponses.entrySet().iterator();
while (pipelinedResponsesIterator.hasNext()) {
  Map.Entry<HostAndPort, Queue<Response<?>>> entry = pipelinedResponsesIterator.next();
  HostAndPort nodeKey = entry.getKey();
  Queue<Response<?>> queue = entry.getValue();
  Connection connection = connections.get(nodeKey);
  executorService.submit(() -> {                        // we can run the last entry on current thread to save resources
    try {
      List<Object> unformatted = connection.getMany(queue.size());
      for (Object o : unformatted) {
        queue.poll().set(o);
      }
    } catch (JedisConnectionException jce) {
      log.error("Error with connection to " + nodeKey, jce);
      // cleanup the connection
      pipelinedResponsesIterator.remove();            // should not be used cross thread, it's not thread-safe, and when run to here, the iter may already reached to the end(hasNext=false)
      connections.remove(nodeKey);
      IOUtils.closeQuietly(connection);               // the connection should be return to pool in finally block
    } finally {
      countDownLatch.countDown();
    }
  });
}
```

and do we need to clear the `pipelinedResponses` and `connections` at the end? instead of remove the exceptioned one at catch block

xrayw:
```java
} catch (JedisConnectionException jce) {
          log.error("Error with connection to " + nodeKey, jce);
          // cleanup the connection
          // TODO these operations not thread-safe and
          pipelinedResponsesIterator.remove();
          connections.remove(nodeKey);
}

```
the code also has bug.

ggivo:
Hi, just to confirm we’re aligned — based on the discussion [here](https://github.com/redis/jedis/pull/4149#discussion_r2056517413), it seems there’s no actual connection leak.

[...]

PR #4149, opened 2025-04-23
Run pipeline in current thread if all the keys on same node

xrayw:
fix https://github.com/redis/jedis/issues/4148


src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
```diff
@@ -80,22 +81,30 @@ public void close() {
 
   @Override
   public final void sync() {
+    if (pipelinedResponses.isEmpty()) {
+      return;
+    }
+
```
ggivo, line 87: Don't think this will bring a lot of value.

src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
```diff
@@ -104,22 +113,27 @@ public final void sync() {
         } catch (JedisConnectionException jce) {
           log.error("Error with connection to " + nodeKey, jce);
           // cleanup the connection
+          // TODO these operations not thread-safe
           pipelinedResponsesIterator.remove();
           connections.remove(nodeKey);
-          IOUtils.closeQuietly(connection);
         } finally {
-          countDownLatch.countDown();
+          IOUtils.closeQuietly(connection);
```
ggivo, line 120: Looking at the current connection handling model,  only broken connections are closed.
The current connection handling model suggests closing the connection when the pipeline itself is closed, and it make sense to me. We can have multiple `sync()` reusing the same connection. I think it should be preserved.

ggivo, line 120: Looking at the current connection handling model,  only broken connections are closed.
```suggestion
          IOUtils.closeQuietly(connection);
```
The current connection handling model suggests closing the connection when the pipeline itself is closed, and it make sense to me. We can have multiple `sync()` reusing the same connection. I think it should be preserved.

src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
```diff
@@ -80,22 +81,30 @@ public void close() {
...
         = pipelinedResponses.entrySet().iterator();
     while (pipelinedResponsesIterator.hasNext()) {
       Map.Entry<HostAndPort, Queue<Response<?>>> entry = pipelinedResponsesIterator.next();
       HostAndPort nodeKey = entry.getKey();
       Queue<Response<?>> queue = entry.getValue();
       Connection connection = connections.get(nodeKey);
-      executorService.submit(() -> {
+
+      // last node run in current thread directly
+      Executor executor = pipelinedResponsesIterator.hasNext() ? executorService : Runnable::run;
```
ggivo, line 106: In case of multiple pipeline responses, I think it is better to preserve the same behaviour for all responses.
It becomes hard to reason/troubleshoot if one of the responses is managed in the calling thread and all the other in  executor thread. Brings complexity for not much value.

ggivo (changes requested):

src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
ggivo: Also need to add tests

xrayw: > In case of multiple pipeline responses, I think it is better to preserve the same behaviour for all responses. It becomes hard to reason/troubleshoot if one of the responses is managed in the calling thread and all the other in executor thread. Brings complexity for not much value.

```java
@Nonnull
Executor executorService = onlyOneNode ? Runnable::run : Executors.newFixedThreadPool(MULTI_NODE_PIPELINE_SYNC_WORKERS);

```
~~Executor executor = pipelinedResponsesIterator.hasNext() ? executorService : Runnable::run;~~

executor.execute(() -> {xxx});


Do you said that we can use threadpool at the multi node scenario to reduce the complex?

src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
xrayw: And even if connections can be reused, should we clear pipelinedResponses after this sync?

xrayw: And if we don't clear the pipelinedResponses after `sync()`, 

prev: {node1: [xxx,xx], node2:[xx, xx], node3: [xx,xx]}
after sync(): {{node1: [], node2:[], node3: []}}

new pipeline: {{node1: [], node2:[], node3: [,yy, y]}}

we'll submit the empty task for node1 & node2 to executorService. It's unnecessary.
I think the current logic is too complex. and there are some unexpected behavior

ggivo: Got your point but this will not work. We need to use same connection to send the command and read the response from Redis server. Connection is acquired and cached on `appendCommand`. It is then used to send the command to Redis server. During sync, we need to use the same connection to read the responses for commands sent.

src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
ggivo: I suggest, in case of multiple pipelineResources to keep the current behaviour. I don't think the performance advantages from running the last node in current thread worth the complexity added.

src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
xrayw: I reverted some code, Now it's focus on run in current thread when only single node. 
and the bugs at // TODO need additional fixes




> Got your point but this will not work. We need to use same connection to send the command and read the response from Redis server. Connection is acquired and cached on `appendCommand`. It is then used to send the command to Redis server. During sync, we need to use the same connection to read the responses for commands sent via this connection.

But when executed failed, It will remove the connection from `connections: Map`, This break the cache model. 
Does this mean that re-get the connection is also ok when each sync().

ggivo: > Does this mean that re-get the connection is also ok when each sync().
No. 

The current model is that at any time we have a single connection per node, on each connection we can have multiple commands sent (respectively, their responses we are waiting to receive ). If an error is detected connection & related Responses are dropped.

ggivo: What I mean is the connection should be the one used to send the command, not the one from the previous sync.
Connections are created and cached during appendCommand().

Command's responses are available only on the connection used to send the command to the Redis server. Hereby, it is necessary to use one and the same connection to send the command and later to read the response from the Redis server. That's why all pending Responses are cleared on connection failure.

xrayw: It's ok if we sure It should be like this, but we need fix these bug in it.

ggivo: > It's ok if we sure It should be like this, but we need fix these bug in it. we need to use concurrent map

@xrayw
Thanks again for taking the time and helping make Jedis better. 

You are right, it is not thread safe.
I’m not entirely sure what the original intention behind #3330 was, and I’m not aware of any use case where MultiNodePipelineBase would be used across multiple threads — especially since it’s clearly not thread-safe.

Even if we use a concurrent map internally, that alone wouldn’t make MultiNodePipelineBase thread-safe. Given that, I’m not sure we should move in the direction of making it thread-safe at all. 

Let me know about your opinion

xrayw: > > It's ok if we sure It should be like this, but we need fix these bug in it. we need to use concurrent map
> 
> @xrayw Thanks again for taking the time and helping make Jedis better.
> 
> You are right, it is not thread safe. I’m not entirely sure what the original intention behind #3330 was, and I’m not aware of any use case where `MultiNodePipelineBase` would be used across multiple threads — especially since it’s clearly **not** thread-safe.
> 
> Even if we use a concurrent map internally, that alone wouldn’t make MultiNodePipelineBase thread-safe. Given that, I’m not sure we should move in the direction of making it thread-safe at all.
> 
> Let me know about your opinion

The purpose of using concurrent map is not to make `MultiNodePipelineBase` thread-safe, but to safely remove the corresponding connection in the catch block.

```java

executorService.submit(() -> {
  try {

  }
  catch (xx) {
    iter.remove();                                 // here, it's not thread-safe
    connections.remove(nodeKey);   // and here
  }
});
```

But anyway, using concurrentmap in a single thread is a bit crazy. Maybe we can do it in other ways

ggivo (commented):
LGTM 
I think we should also add dedicated tests to verify `MultiNodePipeline` with single command to a node and `MultiNodePipeline` multiple commands to the same node are processed successfully

xrayw:
@ggivo May i check should i add the test to `ClusterPipeliningTest` class?

ggivo:
> @ggivo May i check should i add the test to `ClusterPipeliningTest` class?

Yes, I would suggest adding them there.

xrayw:
@ggivo I add a test, can you help me to check it. cause i don't have the cluster environment. 🙏

ggivo:
> @ggivo I add a test, can you help me to check it. cause i don't have the cluster environment. thank you very much

@xrayw 
Hey, thanks for working on it! If you have Docker, you should be able to bootstrap a test environment locally  using
`make start-test-env`
This will also include the Redis Cluster used by other tests.

details can be found in [contribution guideline](https://github.com/redis/jedis/blob/master/.github/CONTRIBUTING.md)

xrayw:
@ggivo test passed now, if the test code is correct.
i can't run `make start-test-env` on Colima ( MacOS ), So i test it by `make test`.

ggivo:
@xrayw 
Thanks again, will take a final look and provide an update.

> @ggivo test passed now, if the test code is correct. i can't run `make start-test-env` on Colima ( MacOS ), So i test it by `make test`.

`make test` will do in this case but we are planning to replace it with the Dockerised environment at some point and rely only on containers.  I am on Sonoma ( MacOS ), and it works for me. Curious what is the error you are getting so we can try to address it.

xrayw:
image


Error response from daemon: invalid reference format
Started test environment with Redis version 8.0.

Although it says it is started, but,  actually there is no redis-server started.

ggivo (approved):
LGTM

PR #4172, opened 2025-06-02
fix(transaction): connections not returned to pool after exec() or discard() in JedisPooled

raphaeldelio:
When using JedisPooled.multi(), the returned Transaction object borrows a connection from the pool. However, the connection would not be released after exec() or discard() was called. Over time, this led to connection exhaustion and blocked threads.

This change ensures that the connection is returned to the pool automatically by calling close() at the end of exec() and discard().

### DIFF
--- a/src/main/java/redis/clients/jedis/Transaction.java
+++ b/src/main/java/redis/clients/jedis/Transaction.java
@@ -214,6 +214,7 @@ public class Transaction extends TransactionBase {
       if (jedis != null) {
         jedis.resetState();
       }
+      close();
     }
   }
 
@@ -241,6 +242,7 @@ public class Transaction extends TransactionBase {
       if (jedis != null) {
         jedis.resetState();
       }
+      close();
     }
   }
 }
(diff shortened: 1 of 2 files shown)


ggivo:
@raphaeldelio 
Thanks for taking the time to report it.

**Transaction**/**Pipeline** API implementation in Jedis is a bit ambiguous, it allows reusing the same `Transaction` object for multiple transaction executions. Closing the connection on exec/discard is introducing a breaking change, and if we decide to go in that direction will need to introduce it in a major version. We will consider that for future improvement.

To resolve your issue, ‘After eight transactions, we run out of connections, ’  you need to either call `Transaction#close` manually, or use **try-with-resources** block. You can refer to the official example [Example with try-with-resources](https://redis.io/docs/latest/develop/clients/jedis/transpipe/) 
```
try ( AbstractTransaction trans = jedis.multi()) {
    trans.incrBy("counter:1", 1);
    trans.incrBy("counter:2", 2);
    trans.incrBy("counter:3", 3);
    trans.exec();
}
```

raphaeldelio:
@ggivo Got it! I thought I could be missing something. My intuition was that JedisPooled should take care of managing all resources. Thank you for the explanation.

PR #4173, opened 2025-06-06
Align Search APIs and behaviour with Redis 8.0+

uglide:
- Introduce AggregateIterator for proper Cluster support
- Deprecate APIs that rely on legacy RediSeach behaviour

Based on https://github.com/redis/jedis/pull/4411

---

> [!NOTE]
> **Medium Risk**
> Touches client-side cursor iteration and connection-provider behavior; correctness depends on proper node selection and cursor lifecycle handling in both standalone and cluster deployments.
> 
> **Overview**
> Introduces `AggregateIterator`, a new cursor-based iterator for `FT.AGGREGATE` that keeps cursor reads/deletes on the same Redis node and borrows/returns pooled connections per batch to avoid pool exhaustion/leaks; `UnifiedJedis` and `RediSearchCommands` gain `ftAggregateIterator(...)` to expose it.
> 
> Deprecates legacy iteration/cursor APIs (`FtAggregateIteration`, `FtSearchIteration` and `UnifiedJedis.ftSearchIteration(...)`, plus `ftCursorRead`/`ftCursorDel` in `RediSearchCommands`) and extends `AggregationBuilder` with cursor batch size tracking plus `AggregationResult.isEmpty()` to support iterator semantics; adds standalone+cluster integration tests and unit tests, and fixes `PooledConnectionProvider.getPrimaryNodesConnectionMap()` to return the pool rather than borrowing a connection.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 464d231df53d4a37682a63046f393b7d35dfa8dc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/main/java/redis/clients/jedis/UnifiedJedis.java
+++ b/src/main/java/redis/clients/jedis/UnifiedJedis.java
@@ -40,6 +40,7 @@ import redis.clients.jedis.params.*;
 import redis.clients.jedis.providers.*;
 import redis.clients.jedis.resps.*;
 import redis.clients.jedis.search.*;
+import redis.clients.jedis.search.aggr.AggregateIterator;
 import redis.clients.jedis.search.aggr.AggregationBuilder;
 import redis.clients.jedis.search.aggr.AggregationResult;
 import redis.clients.jedis.search.aggr.FtAggregateIteration;
@@ -4483,6 +4484,7 @@ public class UnifiedJedis implements JedisCommands, JedisBinaryCommands,
    * @param params limit will be ignored
    * @return search iteration
    */
+  @Deprecated
   public FtSearchIteration ftSearchIteration(int batchSize, String indexName, String query, FTSearchParams params) {
     return new FtSearchIteration(provider, commandObjects.getProtocol(), batchSize, indexName, query, params);
   }
@@ -4499,6 +4501,7 @@ public class UnifiedJedis implements JedisCommands, JedisBinaryCommands,
    * @param query limit will be ignored
    * @return search iteration
    */
+  @Deprecated
   public FtSearchIteration ftSearchIteration(int batchSize, String indexName, Query query) {
     return new FtSearchIteration(provider, commandObjects.getProtocol(), batchSize, indexName, query);
   }
@@ -4540,8 +4543,38 @@ public class UnifiedJedis implements JedisCommands, JedisBinaryCommands,
    * @param aggr cursor must be set
    * @return aggregate iteration
    */
+  @Deprecated
   public FtAggregateIteration ftAggregateIteration(String indexName, AggregationBuilder aggr) {
     return new FtAggregateIteration(provider, indexName, aggr);
+  }
+
+  /**
+   * Creates an iterator for aggregation results with cursor support.
+   * This method provides a clean, connection-aware iterator that ensures cursor operations
+   * are executed on the same Redis node.
+   *
+   * <p>Usage example:
+   * <pre>{@code
+   * AggregationBuilder aggr = new AggregationBuilder()
+   *     .groupBy("@category", Reducers.sum("@price").as("total"))
+   *     .cursor(50, 30000);
+   *
+   * try (AggregateIterator iterator = jedis.ftAggregateIterator("products", aggr)) {
+   *     while (iterator.hasNext()) {
+   *         AggregationResult batch = iterator.next();
+   *         // Process batch - access rows via batch.getRows()
+   *     }
+   * }
+   * }</pre>
+   *
+   * @param indexName the search index name
+   * @param aggr aggregation builder with cursor configuration
+   * @return aggregate iterator for cursor-based pagination
+   * @throws IllegalArgumentException if aggregation doesn't have cursor configured
+   * @since 6.1.0
+   */
+  public AggregateIterator ftAggregateIterator(String indexName, AggregationBuilder aggr) {
+    return new AggregateIterator(provider, indexName, aggr);
   }
 
   @Override
--- a/src/main/java/redis/clients/jedis/UnifiedJedis.java
+++ b/src/main/java/redis/clients/jedis/UnifiedJedis.java
@@ -4571,7 +4571,7 @@ public class UnifiedJedis implements JedisCommands, JedisBinaryCommands,
    * @param aggr aggregation builder with cursor configuration
    * @return aggregate iterator for cursor-based pagination
    * @throws IllegalArgumentException if aggregation doesn't have cursor configured
-   * @since 6.1.0
+   * @since 8.0
    */
   public AggregateIterator ftAggregateIterator(String indexName, AggregationBuilder aggr) {
     return new AggregateIterator(provider, indexName, aggr);
--- a/src/main/java/redis/clients/jedis/search/FtSearchIteration.java
+++ b/src/main/java/redis/clients/jedis/search/FtSearchIteration.java
@@ -9,6 +9,14 @@ import redis.clients.jedis.providers.ConnectionProvider;
 import redis.clients.jedis.search.SearchResult.SearchResultBuilder;
 import redis.clients.jedis.util.JedisCommandIterationBase;
 
+/**
+ * Iterator for FT.SEARCH results across cluster nodes.
+ *
+ * @deprecated Since Redis 8.0, FT.SEARCH automatically retrieves results from all cluster nodes,
+ *             eliminating the need for manual iteration across nodes. Use the standard
+ *             {@code ftSearch} methods directly instead.
+ */
+@Deprecated
 public class FtSearchIteration extends JedisCommandIterationBase<SearchResult, Document> {
 
   private int batchStart;
--- a/src/main/java/redis/clients/jedis/search/FtSearchIteration.java
+++ b/src/main/java/redis/clients/jedis/search/FtSearchIteration.java
@@ -10,11 +10,15 @@ import redis.clients.jedis.search.SearchResult.SearchResultBuilder;
 import redis.clients.jedis.util.JedisCommandIterationBase;
 
 /**
- * Iterator for FT.SEARCH results across cluster nodes.
+ * Iterator for paginating through FT.SEARCH results in batches.
+ * <p>
+ * This class provides an iteration mechanism over {@link Document} results from a RediSearch
+ * FT.SEARCH command, automatically handling pagination through the result set.
  *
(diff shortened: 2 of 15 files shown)


src/main/java/redis/clients/jedis/UnifiedJedis.java
```diff
@@ -4022,24 +4009,25 @@ public SearchResult ftSearch(String indexName, Query query) {
    * @param query limit will be ignored
    * @return search iteration
    */
+  @Deprecated
   public FtSearchIteration ftSearchIteration(int batchSize, String indexName, Query query) {
     return new FtSearchIteration(provider, commandObjects.getProtocol(), batchSize, indexName, query);
   }
 
   @Override
   @Deprecated
```
ggivo, line 4510: Same here, we should add a note why deprecated, what need to be used instead

src/main/java/redis/clients/jedis/CommandObjects.java
```diff
@@ -63,1968 +63,2806 @@ protected CommandArguments commandArguments(ProtocolCommand command) {
...
-    return new CommandObject<>(commandArguments(ECHO).add(msg), BuilderFactory.STRING);
+    return new CommandObject<>(commandArguments(ECHO).add(msg), BuilderFactory.STRING,
+        EnumSet.of(CommandObject.CommandFlag.LOADING, CommandObject.CommandFlag.STALE,
+          CommandObject.CommandFlag.FAST));
   }
 
-  private final CommandObject<String> FLUSHALL_COMMAND_OBJECT = new CommandObject<>(commandArguments(FLUSHALL), BuilderFactory.STRING);
+  private final CommandObject<String> FLUSHALL_COMMAND_OBJECT = new CommandObject<>(
+      commandArguments(FLUSHALL), BuilderFactory.STRING,
+      EnumSet.of(CommandObject.CommandFlag.WRITE));
```
ggivo: Performance-wise and to save some mem/GC load, we could reuse the  `CommandFlag` enum set instead of creating a new one on each command.

We can have a  CommandMetadataRegistry with predefined command flags per command: subcommand, and reuse them.  

This way, we can also have the option to customize/override the metadata per command  by the users

src/main/java/redis/clients/jedis/executors/ClusterCommandExecutor.java
```diff
@@ -90,6 +96,63 @@ public final <T> T executeCommand(CommandObject<T> commandObject) {
...
   }
 
+  @Override
+  public final <T> T executeKeylessCommand(CommandObject<T> commandObject) {
+    Instant deadline = Instant.now().plus(maxTotalRetriesDuration);
+    int consecutiveConnectionFailures = 0;
+    Exception lastException = null;
+
+    RequiredConnectionType connectionType;
+    if (flags.getFlags(commandObject.getArguments())
```
ggivo: What about if there  are no Replicas?

Should the default be even for readonly commands to run on PRIMARY? 

We can have client side setting similar to lettuce allowing to configure 
`ReadFromReplica` - if there is no replica connection fail the commans
`ReadFromReplicaPrefered` - if there is no replica connection fail-back to PRIMARY connection
`ReadFromPrimary` - default?

src/main/java/redis/clients/jedis/executors/ClusterCommandExecutor.java
```diff
@@ -90,6 +96,63 @@ public final <T> T executeCommand(CommandObject<T> commandObject) {
     return doExecuteCommand(commandObject, false);
   }
 
+  @Override
+  public final <T> T executeKeylessCommand(CommandObject<T> commandObject) {
```
ggivo: code duplication with doExecuteCommand

src/main/java/redis/clients/jedis/executors/ClusterCommandExecutor.java
```diff
@@ -157,7 +220,49 @@ private <T> T doExecuteCommand(CommandObject<T> commandObject, boolean toReplica
...
+    ConnectionPool pool = selectedEntry.getValue();
+
+    return pool.getResource();
+  }
+
+  private List<Map.Entry<String, ConnectionPool>> selectNextConnectionPool(
+      RequiredConnectionType connectionType) {
+    Map<String, ConnectionPool> connectionMap;
+
+    // NOTE(imalinovskyi): If we need to connect to replica, we use all nodes, otherwise we use only
```
ggivo: Ok. Now I see how it will work if there are no replica nodes.
In this case we can consider introducing client side configuration as separate PR. 

`RequiredConnectionType` is a bit misleading

src/main/java/redis/clients/jedis/search/aggr/AggregateIterator.java
```diff
@@ -0,0 +1,207 @@
...
+    AggregationResult result = AggregationResult.SEARCH_AGGREGATION_RESULT_WITH_CURSOR
+        .build(rawReply);
+
+    cursorId = result.getCursorId();
+    return result;
+  }
+
+  private Connection acquireConnection(AggregationBuilder aggregationBuilder) {
+    // Create the initial FT.AGGREGATE command
+    redis.clients.jedis.CommandArguments args = new redis.clients.jedis.ClusterCommandArguments(
```
ggivo: class import

src/main/java/redis/clients/jedis/search/aggr/AggregateIterator.java
```diff
@@ -0,0 +1,207 @@
...
+    if (!aggregationBuilder.isWithCursor()) {
+      throw new IllegalArgumentException("AggregationBuilder must have cursor configured");
+    }
+
+    this.connectionProvider = connectionProvider;
+    this.indexName = indexName;
+    this.batchSize = aggregationBuilder.getCursorCount();
+
+    // Get a dedicated connection for this cursor session
+    this.connection = acquireConnection(aggregationBuilder);
```
ggivo: Do we need to keep the connection till cursor is closed?

multiple AggregateIterators, could exaust the pool, any reason not to acuire it whenever we need to perform an operation?


Can we stick to the connection pool instead of the connection?
then it. should be fine to acuire connection/perform operation/return the connection to the pool

ggivo (commented):
Looks good.
Added some comments that can be addressed later.
And also sharing a concern around sticky connection management, which will lock the connection from the pool for the time of iteration.

src/main/java/redis/clients/jedis/executors/ClusterCommandExecutor.java
uglide: This can be added later, and is out of scope for now.

src/main/java/redis/clients/jedis/executors/ClusterCommandExecutor.java
uglide: I'm open for a better naming suggestions :)

src/main/java/redis/clients/jedis/search/aggr/AggregateIterator.java
```diff
@@ -0,0 +1,227 @@
...
+  public AggregateIterator(ConnectionProvider connectionProvider, String indexName,
+      AggregationBuilder aggregationBuilder) {
+    if (!aggregationBuilder.isWithCursor()) {
+      throw new IllegalArgumentException("AggregationBuilder must have cursor configured");
+    }
+
+    this.indexName = indexName;
+    this.batchSize = aggregationBuilder.getCursorCount();
+
+    // Get connection pool entry - use getPrimaryNodesConnectionMap() to get pool-based connections
```
ggivo, line 89: shuffle nodes?

src/main/java/redis/clients/jedis/search/aggr/AggregateIterator.java
```diff
@@ -0,0 +1,227 @@
...
+    Map<?, ?> connectionMap = connectionProvider.getPrimaryNodesConnectionMap();
+    if (connectionMap.isEmpty()) {
+      throw new JedisException("No connections available from connection provider");
+    }
+    // Get the first (or only) entry from the map
+    this.connectionEntry = connectionMap.entrySet().iterator().next();
```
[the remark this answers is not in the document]
ggivo: This seems legitimate issue to me.

src/main/java/redis/clients/jedis/UnifiedJedis.java
```diff
@@ -4499,6 +4501,7 @@ public SearchResult ftSearch(String indexName, Query query) {
    * @param query limit will be ignored
    * @return search iteration
    */
+  @Deprecated
```
ggivo, line 4504: FtSearchIteration serves also the purpose  to provide  search results in batches when used in STANDALONE mode, e.g read large results in chunks.

Current logic used does not enforce any sorting and rely on limit which might produce duplicates/incorrect event now.

Discussed offline, and aggreed we are deprecating this behaviour in favour of using directly ftSearch with params (sort/limit)

src/main/java/redis/clients/jedis/UnifiedJedis.java
```diff
@@ -4540,10 +4543,40 @@ public String ftCursorDel(String indexName, long cursorId) {
...
+   *         // Process batch - access rows via batch.getRows()
+   *     }
+   * }
+   * }</pre>
+   *
+   * @param indexName the search index name
+   * @param aggr aggregation builder with cursor configuration
+   * @return aggregate iterator for cursor-based pagination
+   * @throws IllegalArgumentException if aggregation doesn't have cursor configured
+   * @since 6.1.0
```
ggivo: @since 8.0.0

src/main/java/redis/clients/jedis/search/FtSearchIteration.java
```diff
@@ -9,6 +9,14 @@
 import redis.clients.jedis.search.SearchResult.SearchResultBuilder;
 import redis.clients.jedis.util.JedisCommandIterationBase;
 
+/**
```
ggivo, line 12: javadoc is misleading


In STANDALONE  mode with  FT.SEARCH it will retrieve results in batches but does not enforce any sort and might end up with duplicated results

When used in CUSTER mode will perform the FT.SEARCH  query on each node and exaust the results on each node, which in some setups migth lead to duplicate results. For exampke since Redis 8.0, FT.SEARCH automatically retrieves results from all cluster nodes

ggivo (commented):
LGTM, 

There are comments by cursor that seems ligitime

ggivo (approved):
LGTM

PR #4175, opened 2025-06-10
Implement failover retry for in-flight commands

ggivo:
This change adds support for retrying in-flight commands when a Redis connection fails and the client automatically fails over to another cluster. The feature is configurable through the FailoverOptions builder.

Key changes:
- Added retryFailedInflightCommands option to FailoverOptions
- Implemented retry logic in CircuitBreakerCommandExecutor
- Added integration tests to verify both retry and no-retry behavior
- Created utility methods for test setup and configuration

This enhancement improves resilience for long-running commands like blocking operations, allowing them to transparently continue on the failover cluster without client-side errors.


ggivo:
Current implementation performs actual failover, on first incoming command after CB threshold is reached. 
Discussed with @atakavci that we need to perform failover imediately when CB switches to OPEN state.
Will work on this with separate PR.

ggivo:
@atakavci updated, the review could you take a second look?

src/main/java/redis/clients/jedis/UnifiedJedis.java
```diff
@@ -237,7 +238,19 @@ public UnifiedJedis(ConnectionProvider provider, int maxAttempts, Duration maxTo
...
+
+  /**
+   * Constructor which supports multiple cluster/database endpoints each with their own isolated connection pool.
+   * <p>
+   * With this Constructor users can seamlessly failover to Disaster Recovery (DR), Backup, and Active-Active cluster(s)
+   * by using simple configuration which is passed through from Resilience4j - https://resilience4j.readme.io/docs
+   * <p>
+   */
+  @Experimental
+  public UnifiedJedis(MultiClusterPooledConnectionProvider provider, FailoverOptions failoverOptions) {
```
atakavci, line 252: if this is going to be the only place that we introduce `FailoverOptions` into `UnifiedJedis`, how about putting it into `MultiClusterClientConfig`  and may be holding a reference to it in provider itself. 
My reasoning here is that `retryOnFailover` is not going to be the only thing that we need from config and i guess we will need it in provider as well.
Not perfectly sure about it though.

atakavci: @ggivo on a second thought, its better to do it later when we actually need it. lets keep it as it is.

ggivo: This was my initial thinking, but it is rather setting of the executor, and decided to put it in the Executor, with the idea to add additional configuration settings in future if needed without the need to introduce new constructors..

src/main/java/redis/clients/jedis/mcf/CircuitBreakerCommandExecutor.java
```diff
@@ -49,9 +53,31 @@ public <T> T executeCommand(CommandObject<T> commandObject) {
   private <T> T handleExecuteCommand(CommandObject<T> commandObject, Cluster cluster) {
     try (Connection connection = cluster.getConnection()) {
       return connection.executeCommand(commandObject);
+    } catch (Exception e) {
+
+      if (retryOnFailover() && !isActiveCluster(cluster)
```
atakavci, line 58: trying to explore if we get a case where a failover already executed and a delayed exception landed on old cluster,, does this trigger a second _unnecessary_ failover?

atakavci (approved):
i just put a question in my comment where i am not sure about the answer.
Other than that LGTM.

src/main/java/redis/clients/jedis/mcf/CircuitBreakerCommandExecutor.java
ggivo: @atakavci 
I think this should not be different compared to how we handle currently multiple parallel threads trying to execute a command just after the **CB** threshold is reached, and it transitions to the **OPEN** state. 
Actual synchronization should happen in `CircuitBreakerFailoverBase#clusterFailover`
 https://github.com/ggivo/jedis/blob/2ae7186d2e17788327cab0f0d32caf28c6880a64/src/main/java/redis/clients/jedis/mcf/CircuitBreakerFailoverBase.java#L40-L40

PR #4178, opened 2025-06-11
[automatic failover] Improve failover resilience by evaluating circuit breaker before retries

ggivo:
## Problem
Currently, the circuit breaker is evaluated only after all retry attempts are exhausted. This delays failover to healthy endpoints in unstable network conditions, as the system must wait for all retries to complete before marking an endpoint as unhealthy.

## Solution
This PR changes the evaluation order so that the circuit breaker counts each connection error separately, even during retry attempts. This allows the circuit breaker to open more quickly when an endpoint is experiencing issues, enabling faster failover to healthy endpoints.

## Changes
- Modified the order of evaluation between circuit breaker and retry mechanisms
- Added integration test to verify the new behavior

### DIFF
--- a/formatter-pom.xml
+++ b/formatter-pom.xml
@@ -20,7 +20,7 @@
 			<plugin>
 				<groupId>net.revelc.code.formatter</groupId>
 				<artifactId>formatter-maven-plugin</artifactId>
-				<version>2.23.0</version>
+				<version>2.16.0</version>
 				<configuration>
 					<configFile>${project.basedir}/hbase-formatter.xml</configFile>
 				</configuration>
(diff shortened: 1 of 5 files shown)


formatter-pom.xml
```diff
@@ -20,7 +20,7 @@
 			<plugin>
 				<groupId>net.revelc.code.formatter</groupId>
 				<artifactId>formatter-maven-plugin</artifactId>
-				<version>2.23.0</version>
+				<version>2.16.0</version>
```
atakavci, line 23: so what is the issue with 2.23 ?

atakavci (approved):
LGTM

formatter-pom.xml
ggivo: @atakavci 
2.23 requires Java 11.
There is a slight difference in the formatted code if we format the code using 2.23.0 vs 2.16.0 (used in the pom.xml with regular build). 
To avoid the difference decided to align the versions.

PR #4186, opened 2025-06-26
Fix:JedisCluster throws NullPointerException when maxAttempts is set to 0

### LINKED ISSUE #4185: JedisCluster throws NullPointerException when maxAttempts is set to 0
When maxAttempts is set to 0 in JedisCluster, it leads to an unhandled NullPointerException during command execution. The cause is that lastException remains null, and Throwable.addSuppressed(null) is called, which is not allowed.

This could be avoided with a guard clause to validate the value of maxAttempts or by checking for null before calling addSuppressed.

**Reproduction**
```
JedisCluster jedis = new JedisCluster(
    new HostAndPort("localhost", 6379),
    2000,       // connection timeout
    2000,       // soTimeout
    0,          // maxAttempts
    null,       // password
    new JedisPoolConfig()
);
jedis.set("key", "value");  // NPE thrown
```

**Stacktrace**
`
java.lang.NullPointerException: Cannot suppress a null exception.
    at java.util.Objects.requireNonNull(Objects.java:235)
    at java.lang.Throwable.addSuppressed(Throwable.java:1074)
    at redis.clients.jedis.JedisClusterCommand.runWithRetries(JedisClusterCommand.java:158)
`

**Cause**
In RetryableCommandExecutor.java, the code attempts to call lastException.addSuppressed(ex) without checking whether lastException is null.

`e.addSuppressed(lastException); // lastException is null when maxAttempts = 0`

This was tried with Jedis 3.9.0, but the same issue is also present in the latest master branch.
Relevant line in the current code:
https://github.com/redis/jedis/blob/master/src/main/java/redis/clients/jedis/executors/RetryableCommandExecutor.java#L71

ggivo:
This change adds input validation to the **RetryableCommandExecutor** constructor to prevent usage with invalid configuration values, specifically when **maxAttempts** is set to **0**.

Previously, setting maxAttempts to 0 in JedisCluster would result in a `NullPointerException` during command execution, due to `lastException` being uninitialized when no attempt is made. With this fix, an `IllegalArgumentException` is thrown early if `maxAttempts < 1`, providing feedback to the user.

Added unit tests to verify the validation behavior.

Co-authored-by: shivincible <shivangi.gupta.jiit@gmail.com>

Fixes #4185

### DIFF
--- a/src/test/java/redis/clients/jedis/misc/AutomaticFailoverTest.java
+++ b/src/test/java/redis/clients/jedis/misc/AutomaticFailoverTest.java
@@ -1,5 +1,6 @@
 package redis.clients.jedis.misc;
 
+import java.net.UnknownHostException;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


atakavci (approved):
LGTM!
nit: copilot warns around this 

> * @throws IllegalArgumentException if the object array contains a {@code null} element

PR #4189, opened 2025-06-27
[automatic failover] Implement HealtStatusManager + weighted endpoints

atakavci:
- weighted cluster seleciton
- Healtstatus manager with initial listener and registration logic
- pluggable health checker strategy  introduced,  these are draft NoOpStrategy, EchoStrategy, LagAwareStrategy,
- fix failing integration tests impacted by weighted clusters


src/main/java/redis/clients/jedis/HostAndPort.java
```diff
@@ -2,7 +2,9 @@
 
 import java.io.Serializable;
 
-public class HostAndPort implements Serializable {
+import redis.clients.jedis.mcf.Endpoint;
+
+public class HostAndPort implements Serializable, Endpoint {
```
a-TODO-rov, line 7: getHost() and getPort() methods are not marked as being overridden

src/main/java/redis/clients/jedis/UnifiedJedis.java
```diff
@@ -238,19 +238,7 @@ public UnifiedJedis(ConnectionProvider provider, int maxAttempts, Duration maxTo
    */
   @Experimental
   public UnifiedJedis(MultiClusterPooledConnectionProvider provider) {
-    this(new CircuitBreakerCommandExecutor(provider, FailoverOptions.builder().build()), provider);
```
a-TODO-rov, line 241: Unused import for `redis.clients.jedis.mcf.FailoverOptions`.

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
```diff
@@ -240,8 +270,54 @@ public void validateTargetConnection(int multiClusterIndex) {
      * provided new index. Special care should be taken to confirm cluster/database availability AND potentially
      * cross-cluster replication BEFORE using this capability.
      */
-    public void setActiveMultiClusterIndex(int multiClusterIndex) {
+    // public void setActiveMultiClusterIndex(int multiClusterIndex) {
```
a-TODO-rov: Do we still need this commented code ?
Also, there is a link [https://github.com/redis/jedis/pull/4189/files#diff-51a7d130f973e51d707bbad52d03c1aaf39fcd7923a04c2f3593d545decbb02aR40](url) in the documentation, that prevents tests to run successfully.

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
```diff
@@ -147,71 +148,100 @@ public MultiClusterPooledConnectionProvider(MultiClusterClientConfig multiCluste
...
-                lastClusterCircuitBreakerForcedOpen = true;
+        clusterWithHealthChange.setHealthStatus(newStatus);
 
-                throw new JedisConnectionException(
-                    "Cluster/database endpoint could not failover since the MultiClusterClientConfig was not "
-                        + "provided with an additional cluster/database endpoint according to its prioritized sequence. "
-                        + "If applicable, consider failing back OR restarting with an available cluster/database endpoint.");
-            } else activeMultiClusterIndex++;
+        if (newStatus.isHealthy()) {
+            if (clusterWithHealthChange.isFailbackEnabled() && activeCluster != clusterWithHealthChange) {
```
a-TODO-rov, line 196: You might try to skip the nested if.

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
```diff
@@ -147,71 +148,100 @@ public MultiClusterPooledConnectionProvider(MultiClusterClientConfig multiCluste
...
+                // lets check if weighted switching is possible
+                Map.Entry<Endpoint, Cluster> failbackCluster = findWeightedFailbackCluster();
+                if (failbackCluster == clusterWithHealthChange
+                    && clusterWithHealthChange.getWeight() > activeCluster.getWeight()) {
+                    setActiveCluster(clusterWithHealthChange, false);
+                }
+            }
+        } else if (clusterWithHealthChange == activeCluster) {
```
a-TODO-rov, line 204: May be try to move as first condition with early return.

a-TODO-rov, line 204: `clusterWithHealthChange == activeCluster` or `clusterWithHealthChange.equals(activeCluster)`

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
a-TODO-rov, line 196: `activeCluster != clusterWithHealthChange` or `!activeCluster.equals(clusterWithHealthChange)` ?

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
```diff
@@ -147,71 +148,100 @@ public MultiClusterPooledConnectionProvider(MultiClusterClientConfig multiCluste
...
-            String originalClusterName = getClusterCircuitBreaker().getName();
+        Cluster clusterWithHealthChange = multiClusterMap.get(endpoint);
 
-            // Only increment if it can pass this validation otherwise we will need to check for NULL in the data path
-            if (activeMultiClusterIndex + 1 > multiClusterMap.size()) {
+        if (clusterWithHealthChange == null) return;
+        if (clusterWithHealthChange.isForcedUnhealthy()) return;
 
-                lastClusterCircuitBreakerForcedOpen = true;
+        clusterWithHealthChange.setHealthStatus(newStatus);
```
a-TODO-rov, line 193: What if `clusterWithHealthChange.equals(activeCluster)`?
Will another thread using the activeCluster at this moment be notified about the status switch ?

src/main/java/redis/clients/jedis/mcf/EchoStrategy.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    @Override
+    public int getTimeout() {
+        return timeout;
+    }
+
+    @Override
+    public HealthStatus doHealthCheck(Endpoint endpoint) {
+        return "OK".equals(echo.apply(endpoint)) ? HealthStatus.HEALTHY : HealthStatus.UNHEALTHY;
```
vladvildanov: As I understand real echo execution will be added later, right?

src/main/java/redis/clients/jedis/mcf/HealthCheck.java
```diff
@@ -0,0 +1,107 @@
...
+    public Endpoint getEndpoint() {
+        return endpoint;
+    }
+
+    public HealthStatus getStatus() {
+        return statusRef.get().getValue();
+    }
+
+    public void start() {
+        scheduler.scheduleAtFixedRate(this::healthCheck, 0, strategy.getInterval(), TimeUnit.MILLISECONDS);
```
vladvildanov, line 42: Sometimes healthcheck isn't a recurring action (f.e initial healthcheck to define initial state of databases), It's nice to have an interface to just call it once

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
atakavci: could you elaborate this?

atakavci: i think we should proceed only if its the same cluster instance in memory(since all the operations going on/through that particular one), not by using a some custom comparison, if we ever provide a custom comparison.

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
atakavci: same as the one above

atakavci: it will require to check same condition multiple times, is that correct?

src/main/java/redis/clients/jedis/mcf/HealthCheck.java
atakavci: @vladvildanov is this referring to this statement?
"_The health of a database is either verified in the background (at a specific check interval) or when the circuit enters the half-open state._"

src/main/java/redis/clients/jedis/UnifiedJedis.java
atakavci: done

src/main/java/redis/clients/jedis/HostAndPort.java
atakavci: done

src/main/java/redis/clients/jedis/mcf/HealthCheck.java
vladvildanov: @atakavci it refers to this statement:


```The MultiDBClient establishes a connection to each of the databases. It performs an initial health-check to determine if the DB is considered healthy. ```

File: src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java (hunk 3 above)
Comment (atakavci) [reply src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java]: clean up done

File: src/main/java/redis/clients/jedis/mcf/EchoStrategy.java (hunk 7 above)
Comment (atakavci) [reply src/main/java/redis/clients/jedis/mcf/EchoStrategy.java]: correct! 
its almost ready to commit a change containing it. working on tests.

File: src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java (hunk 6 above)
Comment (atakavci) [reply src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java:193]: its just to flag the cluster with the new health status, setting status itself does not trigger or interrupt other threads. 
if the cluster receiving status is the active one and its not healthy anymore, then it needs to do a failover. 
To summarize the cases and related behavior for other threads/commands in case of failover;
- if command request shows up after switching to new cluster, it will be send to new cluster
- if command request sent but response not received, it will wait for a response/timeout (or we might go for fast termination of command, we are not there yet. TBH not even sure if there is a clean approach to do it, may be closing the related socket ?!)
- if command response received, then it basically means command completed
- if command request failed during request/response, then command would be retried on new cluster in respect to `retryOnFailover` setting

File: src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java (hunk 4 above)
Comment (a-TODO-rov) [reply src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java:196]: Yes. It's related to the comment bellow (on the else-if statement). If you move `else if (clusterWithHealthChange == activeCluster)` as first condition with early return than you will not need to check the second condition in `if (clusterWithHealthChange.isFailbackEnabled() && activeCluster != clusterWithHealthChange)`. And this way you will be able to remove the nested if statement by stacking the conditions like `if (newStatus.isHealthy() && clusterWithHealthChange.isFailbackEnabled())`.
Basically this suggestion is very optional and not a deal breaker, just a cleaner solution in my opinion.

File: src/main/java/redis/clients/jedis/mcf/EchoStrategy.java [hunk 9]
```diff
@@ -0,0 +1,50 @@
...
+        return interval;
+    }
+
+    @Override
+    public int getTimeout() {
+        return timeout;
+    }
+
+    @Override
+    public HealthStatus doHealthCheck(Endpoint endpoint) {
```
ggivo, line 42: Seems like the injected Endpoint is not useful, since in most cases you need additional configuration to perform a connection/call.

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
a-TODO-rov: Yes. Also related to the comment above.

src/main/java/redis/clients/jedis/mcf/FailoverOptions.java
```diff
@@ -8,9 +10,42 @@
 @Experimental
 public class FailoverOptions {
     private final boolean retryOnFailover;
+    private final StrategySupplier healthCheckStrategySupplier;
+    private final float weight;
```
ggivo: Shouldn't  weight  be a property of the ClusterConfig's itself, and not part of the FailoverOptions

src/main/java/redis/clients/jedis/mcf/FailoverOptions.java
```diff
@@ -8,9 +10,42 @@
...
+    private boolean failback;
```
ggivo: Can we document the semantics of configuration options?
If I am getting it right, **failback**  means that this **Cluster** is **DISABLED** and should not be used even if `HEALTHY`?

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
a-TODO-rov: Thanks for the summary.
I think i wasn't clear enough in this comment ...
I was referring to this change:
`private volatile Integer activeMultiClusterIndex = 1;` -> `private volatile Cluster activeCluster;`
So lets say you are switching the status here to the activeCluster -> when effectively this change will be visible for a concurrent thread ?
Basically the writes are synchronized with lock, but the reads are synchronized by the variable being declared volatile. But initially this variable was of a primitive kind and now is not. Having this in mind, should we expose a public setter for it at all ?
So to summarize all of the above my concern is - do you think we might be introducing some concurrent problem with the change, mentioned above ?

src/main/java/redis/clients/jedis/mcf/HealthStatusManager.java
```diff
@@ -0,0 +1,79 @@
...
+
+    public void unregisterListener(Endpoint endpoint, HealthStatusListener listener) {
+        endpointListeners.computeIfPresent(endpoint, (k, v) -> {
+            v.remove(listener);
+            return v;
+        });
+        listeners.remove(listener);
+    }
+
+    public void notifyListeners(HealthStatusChangeEvent eventArgs) {
```
ggivo, line 33: Since endpoint listeners are automatically also registered as global listeners then we will notify same listener twice.

Is this intentional?

src/main/java/redis/clients/jedis/mcf/FailoverOptions.java
```diff
@@ -8,9 +10,42 @@
 @Experimental
 public class FailoverOptions {
     private final boolean retryOnFailover;
```
ggivo: My initial tought was that to be a Client wide configuration not Cluster wide.
If I remember correctly it's  meaning is if we want to retry failed inflight command during failover.

Do we want it configurable per cluster?

src/main/java/redis/clients/jedis/mcf/CircuitBreakerFailoverBase.java
```diff
@@ -48,25 +48,29 @@ protected void clusterFailover(CircuitBreaker circuitBreaker) {
...
+            // if its the same one and there are no more clusters to failover to, 
+            // then throw an exception
+            else if (circuitBreaker == provider.getCluster().getCircuitBreaker() && !provider.canIterateOnceMore()) {
                 throw new JedisConnectionException(
                     "Cluster/database endpoint could not failover since the MultiClusterClientConfig was not "
                         + "provided with an additional cluster/database endpoint according to its prioritized sequence. "
                         + "If applicable, consider failing back OR restarting with an available cluster/database endpoint");
             }
+            // Ignore exceptions since we are already in a failure state
+        } catch (RuntimeException e) {
```
ggivo: Whats' the idea to here?

src/main/java/redis/clients/jedis/mcf/EchoStrategy.java
```diff
@@ -0,0 +1,50 @@
...
+
+    private int interval;
+    private int timeout;
+    private UnifiedJedis jedis;
+
+    public EchoStrategy(HostAndPort hostAndPort, JedisClientConfig jedisClientConfig) {
+       this(hostAndPort, jedisClientConfig, 1000, 1000);
+    }
+
+    public EchoStrategy(HostAndPort hostAndPort, JedisClientConfig jedisClientConfig, int interval, int timeout) {
```
ggivo, line 20: can we use some sort of Configration/Option  to wrap interval,timeout object so that in future we can extend with additinal oproperties without having to update the constructor

src/main/java/redis/clients/jedis/mcf/LagAwareStrategy.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    @Override
+    public int getTimeout() {
+        return timeout;
+    }
+
+    @Override
+    public HealthStatus doHealthCheck(Endpoint endpoint) {
+        RedisRestAPIHelper helper = new RedisRestAPIHelper(endpoint.getHost(), String.valueOf(endpoint.getPort()),
```
ggivo, line 28: make sure we don't forget hardcoded values

src/main/java/redis/clients/jedis/mcf/RedisRestAPIHelper.java
```diff
@@ -0,0 +1,95 @@
...
+ */
+class RedisRestAPIHelper {
+    // Required connection information - replace with your host, port, username, and
+    // password
+    private final String host;
+    private final String port;
+    private final String userName;
+    private final String password;
+
+    RedisRestAPIHelper(String host, String port, String userName, String password) {
```
ggivo, line 26: I would go with generic options class here, so that we can easily extend it, e.g tls, proxy ..

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
```diff
@@ -130,7 +135,7 @@ public MultiClusterPooledConnectionProvider(MultiClusterClientConfig multiCluste
         for (ClusterConfig config : clusterConfigs) {
             GenericObjectPoolConfig<Connection> poolConfig = config.getConnectionPoolConfig();
 
-            String clusterId = "cluster:" + config.getPriority() + ":" + config.getHostAndPort();
+            String clusterId = "cluster:" + config.getFailoverOptions().getWeight() + ":" + config.getHostAndPort();
```
ggivo: Do we want to include the weight in cluster id?
In future they might become calculated dynamically at some point.

I was also under the impression that weight should be modifiable within the current design?

src/main/java/redis/clients/jedis/mcf/HealthStatus.java
```diff
@@ -0,0 +1,19 @@
+package redis.clients.jedis.mcf;
+
+public enum HealthStatus {
+    HEALTHY(0x01), UNHEALTHY(0x02), FORCED_UNHEALTHY(0x04);
```
ggivo: What is the intended use of FORCED_UNHEALTHY?

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
atakavci: i dont think there is conccurrency issue since each time an event received, its only possible that related healtcheck thread is executing the `handleStatusChange`. So that thread is bound to send events on that individual cluster, at least the implementation does that with its current stage.

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
```diff
@@ -147,71 +152,103 @@ public MultiClusterPooledConnectionProvider(MultiClusterClientConfig multiCluste
...
+                // lets check if weighted switching is possible
+                Map.Entry<Endpoint, Cluster> failbackCluster = findWeightedFailbackCluster();
+                if (failbackCluster == clusterWithHealthChange
+                    && clusterWithHealthChange.getWeight() > activeCluster.getWeight()) {
+                    setActiveCluster(clusterWithHealthChange, false);
+                }
+            }
+        } else if (clusterWithHealthChange == activeCluster) {
+            iterateActiveCluster();
+        }
```
atakavci: @a-TODO-rov, is this what you suggest? if not, could you provide code snippet on your suggestion? 

```suggestion
        if (clusterWithHealthChange == activeCluster && !newStatus.isHealthy()) {
            iterateActiveCluster();
            return;
        }
        if (newStatus.isHealthy() && clusterWithHealthChange.isFailbackEnabled() && activeCluster != clusterWithHealthChange) {
                // lets check if weighted switching is possible
                Map.Entry<Endpoint, Cluster> failbackCluster = findWeightedFailbackCluster();
                if (failbackCluster == clusterWithHealthChange
                    && clusterWithHealthChange.getWeight() > activeCluster.getWeight()) {
                    setActiveCluster(clusterWithHealthChange, false);
                }
            }
        } 
```

dengliming:
不好Sincerely,  产品测试专用（勿动）
@atakavci commented on this pull request.



In src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java:
> +        if (newStatus.isHealthy()) {
+            if (clusterWithHealthChange.isFailbackEnabled() && activeCluster != clusterWithHealthChange) {
+                // lets check if weighted switching is possible
+                Map.Entry<Endpoint, Cluster> failbackCluster = findWeightedFailbackCluster();
+                if (failbackCluster == clusterWithHealthChange
+                    && clusterWithHealthChange.getWeight() > activeCluster.getWeight()) {
+                    setActiveCluster(clusterWithHealthChange, false);
+                }
+            }
+        } else if (clusterWithHealthChange == activeCluster) {
+            iterateActiveCluster();
+        }

@a-TODO-rov, is this what you suggest? if not, could you provide code snippet on your suggestion?
⬇️ Suggested change
-        if (newStatus.isHealthy()) {
-            if (clusterWithHealthChange.isFailbackEnabled() && activeCluster != clusterWithHealthChange) {
-                // lets check if weighted switching is possible
-                Map.Entry<Endpoint, Cluster> failbackCluster = findWeightedFailbackCluster();
-                if (failbackCluster == clusterWithHealthChange
-                    && clusterWithHealthChange.getWeight() > activeCluster.getWeight()) {
-                    setActiveCluster(clusterWithHealthChange, false);
-                }
-            }
-        } else if (clusterWithHealthChange == activeCluster) {
-            iterateActiveCluster();
-        }
+        if (clusterWithHealthChange == activeCluster && !newStatus.isHealthy()) {
+            iterateActiveCluster();
+            return;
+        }
+        if (newStatus.isHealthy() && clusterWithHealthChange.isFailbackEnabled() && activeCluster != clusterWithHealthChange) {
+                // lets check if weighted switching is possible
+                Map.Entry<Endpoint, Cluster> failbackCluster = findWeightedFailbackCluster();
+                if (failbackCluster == clusterWithHealthChange
+                    && clusterWithHealthChange.getWeight() > activeCluster.getWeight()) {
+                    setActiveCluster(clusterWithHealthChange, false);
+                }
+            }
+        } 



—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you are subscribed to this thread.Message ID: ***@***.***>
[
{
***@***.***": "http://schema.org",
***@***.***": "EmailMessage",
"potentialAction": {
***@***.***": "ViewAction",
"target": "https://github.com/redis/jedis/pull/4189#pullrequestreview-3005372340",
"url": "https://github.com/redis/jedis/pull/4189#pullrequestreview-3005372340",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
***@***.***": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
```diff
@@ -250,79 +293,80 @@ public void setActiveMultiClusterIndex(int multiClusterIndex) {
...
-            if (activeMultiClusterIndex == multiClusterIndex)
+            if (activeCluster == cluster)
                 log.warn("Cluster/database endpoint '{}' successfully closed its circuit breaker", originalClusterName);
             else log.warn("Cluster/database endpoint successfully updated from '{}' to '{}'", originalClusterName,
-                getClusterCircuitBreaker(multiClusterIndex).getName());
-
-            activeMultiClusterIndex = multiClusterIndex;
-            lastClusterCircuitBreakerForcedOpen = false;
+                cluster.circuitBreaker.getName());
+            if (!activeCluster.isFailbackEnabled()) {
```
ggivo, line 307: I think we should keep the HealtStatus accurate based on actual health checks and keep the `failbaclEnabled` separate. Where needed, we can do  (cluster.isHealthy && cluster.isEnabled()

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
```diff
@@ -147,71 +152,103 @@ public MultiClusterPooledConnectionProvider(MultiClusterClientConfig multiCluste
...
+        if (clusterWithHealthChange.isForcedUnhealthy()) return;
 
-            // Only increment if it can pass this validation otherwise we will need to check for NULL in the data path
-            if (activeMultiClusterIndex + 1 > multiClusterMap.size()) {
+        clusterWithHealthChange.setHealthStatus(newStatus);
 
-                lastClusterCircuitBreakerForcedOpen = true;
+        if (newStatus.isHealthy()) {
+            if (clusterWithHealthChange.isFailbackEnabled() && activeCluster != clusterWithHealthChange) {
+                // lets check if weighted switching is possible
```
ggivo, line 197: Looks like there is a concurency issue here.

If multiple Clusters change state to healthy, we can end up with `setActiveCluster()`  one that is  with the least weight,

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
```diff
@@ -147,71 +152,103 @@ public MultiClusterPooledConnectionProvider(MultiClusterClientConfig multiCluste
...
+                Map.Entry<Endpoint, Cluster> failbackCluster = findWeightedFailbackCluster();
+                if (failbackCluster == clusterWithHealthChange
+                    && clusterWithHealthChange.getWeight() > activeCluster.getWeight()) {
+                    setActiveCluster(clusterWithHealthChange, false);
+                }
+            }
+        } else if (clusterWithHealthChange == activeCluster) {
+            iterateActiveCluster();
```
ggivo: Should this be syncronised with CB failover?
If CB try to failover, at the same time HealtCheck triggers handleStatusChange
we can end up invoking iterateActiveCluster twice.

In general I think is good idea if we can manage the failover in single place.
Currently if failover is initiated by CB transition to OPEN state, failover is triggered  in the CircuitBreakerCommandExecutor (with proper locking), if it is triggred by handleStatusChange,  iterateActiveCluster() is invoked there and this I think can lead to syncronisation problems.

atakavci:
hey @dengliming , 
what about these comments you had via email. Could you explain and if this is something not planned, please remove or disable the automation or whatever process this is. 
This is kind of spamming our PR's , so thank you for your cooperation.

dengliming:
Not goodSincerely,  产品测试专用（勿动）atakavci left a comment (redis/jedis#4189)
hey @dengliming ,
what about these comments you had via email. Could you explain and if this is something not planned, please remove or disable the automation or whatever process this is.
This is kind of spamming our PR's , so thank you for your cooperation.

—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>

[
{
***@***.***": "http://schema.org",
***@***.***": "EmailMessage",
"potentialAction": {
***@***.***": "ViewAction",
"target": "https://github.com/redis/jedis/pull/4189#issuecomment-3061286205",
"url": "https://github.com/redis/jedis/pull/4189#issuecomment-3061286205",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
***@***.***": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

src/main/java/redis/clients/jedis/mcf/EchoStrategy.java
atakavci: this method signature is from [HealthCheckStrategy](https://github.com/redis/jedis/pull/4189/files#diff-410dd555814c16bbb7ee407e03065e8764a89de2ca5cee065c510ebe9360fae9) interface, where it is responsible for identifiying related Endpoint. Any other information is not considered to be relevant, at least at this level. 
The configuration can be passed through the `StrategySupplier`  in 'FailoverOptions` where both `endpoint` and `clientConfig` is provided to supplier.
Design is flexible and covers almost all possible use cases, which might be over-flexible. Also brings some complexity as we discussed.

src/main/java/redis/clients/jedis/mcf/FailoverOptions.java
atakavci: i ll check if i can move ;
- `weight` and `healthCheckStrategySupplier` to `ClusterConfig`
- `retryOnFailover` and `failback` to `MultiClusterClientConfig`

src/main/java/redis/clients/jedis/mcf/FailoverOptions.java
atakavci: thanks for pointing to this;
i think this needs to be revised anyway so i will get help from AI to add javadoc especially for these unclear definitions later stages. Let's agree on rules and expectations on failback behaviour first.

atakavci: to answer your second Q;
this is intended for enabling/disabling "failing back to this cluster" once its failed over from.

src/main/java/redis/clients/jedis/mcf/HealthStatusManager.java
atakavci: good catch, it is a mistake for sure.

src/main/java/redis/clients/jedis/mcf/FailoverOptions.java
atakavci: i ll check if i can move ;
- `weight` and `healthCheckStrategySupplier` to `ClusterConfig`
- `retryOnFailover` and `failback` to `MultiClusterClientConfig`

src/main/java/redis/clients/jedis/mcf/CircuitBreakerFailoverBase.java
atakavci: its left over from my initial experiments on circuit breaker behaviour..
clean up done.

src/main/java/redis/clients/jedis/mcf/EchoStrategy.java
atakavci: not sure if this deserves a new type for option, let me follow and add the type if we need another field/parameter.

src/main/java/redis/clients/jedis/mcf/LagAwareStrategy.java
atakavci: FWIW, `LagAwareStrategy` is nothing more than an idea as public API at this stage;  sorry that i couldn't properly filter my changes/commits to this PR. And its bigger than i initially anticipated.
I take this this in to its own small PR.

src/main/java/redis/clients/jedis/mcf/RedisRestAPIHelper.java
atakavci: same comment as in the one for `LagAwareStrategy`

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
atakavci: Absolutly right, TBH i dont even know what this value is doing there.

src/main/java/redis/clients/jedis/mcf/HealthStatus.java
atakavci: i ll do the change to introduce an aggregated status function(to gather the values from CB, HealtChecks and cluster enabled/disables) as we agreed in our dedicated session.

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
atakavci: i ll go for an aggregated status function but still need to get my head around for this to be optimal and final(may be not final but mature). 
Want to avoid tracking multiple states here and there, as much as possible. 
To me, this kind of setup is a prime candidate for turning into status hell.

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
atakavci: there is no race condition here since both `findWeightedFailbackCluster` and `setActiveCluster` is thread safe, where former is inherently safe while latter is safe by use of a `ReentrantLock`.
to explain better, if we assume two healthy cluster shows up, what is important here is sequence; 
if the less weighted shows up early, even with involving concurrency constructs, there is no escape from switching to it since we never know higher weighted would show up.

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
atakavci: @a-TODO-rov, provided a code block [here](https://github.com/redis/jedis/pull/4189#discussion_r2197528925) to verify we are on same track, please check and let me know.

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
atakavci: @a-TODO-rov, provided a code block [here](https://github.com/redis/jedis/pull/4189#discussion_r2197528925) to verify we are on same track, please check and let me know.

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
atakavci: if they both trigger `iterateActiveCluster` they will  end up with same highest weighted healthy cluster in the iterate process anyway, 
which would do no-change in theory.
But i admit i am not %100 comfortable with how transitions are managed and agree to build the logic and safety into single place. 
i will save time to investigate and build corner case test for this.

dengliming:
nubi 要吃牛肉，真是伤脑筋qianming

ggivo (approved):
LGTM

Let's start merging them in the feature branch so we have a consistent state there

PR #4194, opened 2025-07-01
Push notification listener

ggivo:
This PR enhances Jedis and UnifiedJedis to provide better support for RESP3 push messages and introduces a configurable option for relaxed timeout behavior.

**Background**

Starting with RESP3, Redis servers can send push messages to clients out-of-band (i.e., outside the standard request/response flow). See the [RESP3 Push Specification](https://redis.io/docs/latest/develop/reference/protocol-spec/#pushes) for more details.

In the current state of the client:
- 	Jedis supports only known **Push** notifications like **invalidate**.  Push notifications are supported using dedicated client-side caching (CSC) connection  (CacheConnection)
- 	This leads to runtime errors if `CLIENT TRACKING ON `is enabled in a context other than CSC, or if other **RESP3** push events are received.

✅ Changes Introduced
	•	Added support for **processing and discarding all RESP3 push messages**, preventing errors in non-CSC scenarios.
	•	Introduced support for **registering a custom PushListener**, enabling applications to handle push notifications as needed.
	•	**Pub/Sub push messages** are still handled as before and **propagated to the client** for backward compatibility.
	•	Added a **configurable option to enable relaxed timeout behavior**, intended for usage during maintenance events or when expecting slow responses from the server.


src/main/java/redis/clients/jedis/csc/CacheConnection.java
```diff
@@ -8,6 +9,9 @@
 import redis.clients.jedis.JedisClientConfig;
 import redis.clients.jedis.JedisSocketFactory;
 import redis.clients.jedis.Protocol;
+import redis.clients.jedis.PushConsumer;
+import redis.clients.jedis.PushConsumerContext;
+import redis.clients.jedis.PushHandler;
```
a-TODO-rov: Unused

src/main/java/redis/clients/jedis/Jedis.java
a-TODO-rov, line 1: Not much relevant changes here

src/main/java/redis/clients/jedis/JedisShardedPubSubBase.java
a-TODO-rov, line 1: Not much relevant changes in this file

src/main/java/redis/clients/jedis/JedisClientConfig.java
```diff
@@ -7,6 +7,8 @@
 
 import redis.clients.jedis.authentication.AuthXManager;
 
+import static redis.clients.jedis.PushHandler.NOOP;
```
a-TODO-rov: Unused

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -127,9 +129,9 @@ private static String[] parseTargetHostAndSlot(String clusterRedirectResponse) {
     return response;
   }
 
-  private static Object process(final RedisInputStream is) {
+  private static Object process(final RedisInputStream is, PushConsumer pushConsumer) {
     final byte b = is.readByte();
-    // System.out.println("BYTE: " + (char) b);
+    //System.out.println("BYTE: " + (char) b);
```
a-TODO-rov: ?

src/main/java/redis/clients/jedis/util/JedisAsserts.java
a-TODO-rov, line 1: Not relevant changes here

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -68,15 +79,61 @@ public Connection(final HostAndPort hostAndPort, final JedisClientConfig clientC
...
+       */
+      MaintenanceEventHandler maintenanceEventHandler = config.getMaintenanceEventHandler();
+      if (maintenanceEventHandler != null) {
+        this.pushConsumer.add(new MaintenanceEventConsumer(maintenanceEventHandler));
+
+        if (config.isProactiveRebindEnabled()) {
+          maintenanceEventHandler.addListener(new ConnectionRebindHandler());
+        }
+
+        if (TimeoutOptions.isRelaxedTimeoutEnabled(config.getTimeoutOptions().getRelaxedTimeout())) {
```
atakavci, line 127: are we sure relaxedTimeout s are something that makes sense as standalone item? All this should be considered part of hitless upgrade setup and shouldnt be some standalone item in the configuration.

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -614,4 +694,219 @@ protected boolean isTokenBasedAuthenticationEnabled() {
...
+          }
+        });
+      } catch (Exception e) {
+        // Log notification failures
+      }
+    }
+  }
+
+
+  private static class MaintenanceEventConsumer implements PushConsumer {
```
atakavci, line 803: nit: these types(`ListenerNotificationConsumer`, `MaintenanceEventConsumer`, ..) better be located in their own files for the sake of readability, maintainability etc..
May be as package private.

src/main/java/redis/clients/jedis/PushConsumerChain.java
```diff
@@ -0,0 +1,210 @@
...
+
+    @Override
+    public void accept(PushConsumerContext context) {
+      if (isPubSubType(context.getMessage().getType())) {
+        // Ensure pub/sub events are propagated to application
+        context.setProcessed(false);
+      }
+    }
+
+    private boolean isPubSubType(String type) {
```
atakavci: nit: not sure it adds any value to have this as a method.

src/main/java/redis/clients/jedis/util/ReflectionTestUtils.java
```diff
@@ -0,0 +1,44 @@
+package redis.clients.jedis.util;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
+
+public class ReflectionTestUtils {
```
atakavci: lets move this to test folder

src/main/java/redis/clients/jedis/TimeoutOptions.java
```diff
@@ -0,0 +1,85 @@
...
+    return  relaxedTimeout != null && !relaxedTimeout.equals(DISABLED_TIMEOUT);
+  }
+
+  public static boolean isRelaxedTimeoutDisabled(Duration relaxedTimeout) {
+    return  relaxedTimeout == null || relaxedTimeout.equals(DISABLED_TIMEOUT);
+  }
+
+  public static boolean isRelaxedTimeoutDisabled(int relaxedTimeout) {
+    return  relaxedTimeout == DISABLED_TIMEOUT_MS;
+  }
```
atakavci: do we really need all of this? first one should be enough, likely to be the optimal one.

personal POV; if there is nothing planned to built/add more on `TimeoutOptions` at the moment, i'd say this options class itself is an overkill.

src/main/java/redis/clients/jedis/MaintenanceEventHandler.java
```diff
@@ -0,0 +1,18 @@
...
+  void addListener(MaintenanceEventListener listener);
+
+
+  void removeListener(MaintenanceEventListener listener);
+
+
+  void removeAllListeners();
+
+
+  Collection<MaintenanceEventListener> getListeners();
```
atakavci: nit: this looks pretty generic, how about making a generic interface and use it for `PushHandler` too?
And even a generic base impl as well.
```suggestion

public interface ListenerHandler<T> {
  void addListener(T listener);
  void removeListener(T listener);
  void removeAllListeners();
  Collection<T> getListeners();
}

public interface MaintenanceEventHandler extends ListenerHandler<MaintenanceEventListener> {

}
```

src/main/java/redis/clients/jedis/csc/CacheConnection.java
```diff
@@ -43,10 +63,11 @@ protected void initializeFromClientConfig(JedisClientConfig config) {
   }
 
   @Override
-  protected Object protocolRead(RedisInputStream inputStream) {
+  protected Object protocolRead(RedisInputStream inputStream, PushConsumer listener) {
```
atakavci: here a bit confused with listener.. why it is not receiving what is read from stream?
it should be same object from base `Connection`, but using what is provided as argument is kind of best practice. May be i am missing something?

src/main/java/redis/clients/jedis/PushConsumerChain.java
```diff
@@ -0,0 +1,210 @@
...
+    return new PushConsumerChain(handlers);
+  }
+
+  /**
+   * Add a handler to be executed after this chain.
+   *
+   * @param handler The handler to add
+   * @return A new chain with the handler added
+   */
+  public PushConsumerChain then(PushConsumer handler) {
```
atakavci: what is the use case of this ? lets remove anything (at least from public API) if we are not using it right now, including other instance methods.

src/main/java/redis/clients/jedis/PushConsumerChain.java
```diff
@@ -0,0 +1,210 @@
...
+   */
+  public static final PushConsumer PROPAGATE_ALL_HANDLER = (context) -> {
+    // mark as not-processed, always propagate
+    context.setProcessed(false);
+  };
+
+
+  /**
+   * Handler that allows only pub/sub related events to be propagated to the client.
+   * Marks non-pub/sub events as processed, preventing their propagation.
```
atakavci, line 31: comment is not matching with how it functions. i am not sure around using consumer interface just for filtering overall.

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -211,22 +214,52 @@ private static List<KeyValue> processMapKeyValueReply(final RedisInputStream is)
...
+    } while (isPush(reply) && isProcessed((PushConsumerContext) reply) );
+
+    if ( isPush(reply)) {
+      return ((PushConsumerContext) reply).getMessage().getContent();
+    }
+
+    return reply;
   }
 
+  private static boolean isProcessed(PushConsumerContext reply) {
```
atakavci: nit: doesnt seem necessary or adding value

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -211,22 +214,52 @@ private static List<KeyValue> processMapKeyValueReply(final RedisInputStream is)
...
         final List<KeyValue> ret = new ArrayList<>(num);
         for (int i = 0; i < num; i++) {
-          ret.add(new KeyValue(process(is), process(is)));
+          ret.add(new KeyValue(process(is, null), process(is,null)));
         }
         return ret;
     }
   }
 
   public static Object read(final RedisInputStream is) {
```
atakavci, line 223: is this method remains just for test purposes?

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -211,22 +214,52 @@ private static List<KeyValue> processMapKeyValueReply(final RedisInputStream is)
...
   @Experimental
-  public static Object read(final RedisInputStream is, final Cache cache) {
-    Object unhandledPush = readPushes(is, cache, false);
-    return unhandledPush == null ? process(is) : unhandledPush;
+  public static Object read(final RedisInputStream is, PushConsumer pushConsumer) {
+    // read until we have a non-push event,
+    // or push-event is not handled and need to be propagated to application
+    Object reply;
+    do {
+      reply = process(is, pushConsumer);
```
atakavci, line 244: its only here that we expect a push message (instance of `PushConsumerContext`) can be returned. so lets return a wrapper type that contains if its a push message or command response. Then, we wont need to use class type checking like within `isPush`.

src/main/java/redis/clients/jedis/PushHandler.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.Collection;
+import java.util.Collections;
+
+/**
+ * A handler object that provides access to {@link PushListener}s.
+ *
+ * @author Ivo Gaydajiev
+ * @since 6.1
+ */
+public interface PushHandler {
```
atakavci: same for `MaintenanceEventHandler`

src/main/java/redis/clients/jedis/ConnectionPool.java
```diff
@@ -78,4 +86,31 @@ private void attachAuthenticationListener(AuthXManager authXManager) {
...
+
+    public RebindHandler(ConnectionPool pool, ConnectionFactory factory) {
+      this.pool = pool;
+      this.factory = factory;
+    }
+
+    @Override
+    public void onRebind(HostAndPort target) {
+      HostAndPort previous = rebindTarget.getAndSet(target);
+      if (previous != target) {
```
atakavci: is it possible to trigger same handler with multiple rebind on different targets?
not sure how it should be but if we rely on host+port equality, lets use `equals`.

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -614,4 +694,219 @@ protected boolean isTokenBasedAuthenticationEnabled() {
...
+          socket.setSoTimeout(soTimeout);
+        }
+      } catch (SocketException ex) {
+        setBroken();
+        throw new JedisConnectionException(ex);
+      }
+    }
+  }
+
+  private static int safeToInt(long millis) {
```
atakavci: nit: move it to some util class.. may be a new one like `NumberUtils` and which also exposes `DoublePrecision` methods and retire `DoublePrecision` as well.

src/main/java/redis/clients/jedis/Connection.java
ggivo: I look at them as a functionality not bound to maintenance events (hitless upgrade.  Maintenance events are just a way to trigger a relaxed timeout for a period of time, but this can also be done from different external triggers.
Moreover, similar to Lettuce, at some point, we can add a configurable command timeout per command type.

src/main/java/redis/clients/jedis/TimeoutOptions.java
ggivo: removed "`isRelaxedTimeoutDisabled`"alternatives. 
but decided to add `isRelaxedTimeoutEnabled(int relaxedTimeout)`

Since `Socket.getSoTimeout()`/`Socket.setSoTimeout()` works with `int` I added `isRelaxedTimeoutEnabled(int relaxedTimeout)`. 
I use it as a sort of optimisation, not to convert between `Duration` -> `int` each time we check it.

src/main/java/redis/clients/jedis/Connection.java
ggivo: I prefer to keep them private so we have more freedom in modifying the actual implementation in the future. 
Those consumers are not intended to be used directly but via exposed `MaintenanceEventListener` for example

src/main/java/redis/clients/jedis/ConnectionPool.java
ggivo: Fixed. Now using equality

src/main/java/redis/clients/jedis/Protocol.java
ggivo: I thought of this. Currently, we are creating wrapper objects only for push notifications. 
For the command output (which I expect to be the regular code path), we are directly returning the result (no wrapper). That's the reason to prefer wrapping only push notifications and use `instanceof` to avoid unnecessary object allocations.

src/main/java/redis/clients/jedis/PushConsumerChain.java
ggivo: removed unused method.

src/main/java/redis/clients/jedis/Protocol.java
ggivo: removed

src/main/java/redis/clients/jedis/Protocol.java
ggivo: @atakavci 
Which one? 
If you refer to `process(is)`, it is now removed.

src/main/java/redis/clients/jedis/MaintenanceEventHandler.java
ggivo: Introduced shared interface & abstract implementation

src/main/java/redis/clients/jedis/PushHandler.java
ggivo: Introduced shared interface and abstract implementation with  MaintenanceEventHandler

src/main/java/redis/clients/jedis/csc/CacheConnection.java
ggivo: @atakavci 
Not sure I get the question. Let's discuss it if available.
The difference in `CacheConnection` compared to base `Conection`  is that it register's invalidation push handler to react on **INVALIDATE** push notification. And also it uses some internal locking in `protocolRead` compared to base implementation in `Connection`, which is why I preserve `protocolRead` in `CacheConnection` as it was.

src/main/java/redis/clients/jedis/PushConsumerChain.java
ggivo: Agree, not the best naming, I can update the comments, but opened for suggestions here.
Let's also discuss this.

My initial idea is that PushConsumers* are used only internally to process incoming PushNotifications so we can handle different types of existing Connections (CSC, Pub/Sub) in a backward compatible way. 
Then, to the end user of the library, we have `PushHandler` where we register `PushListeners` which propagate each `PushMessage` to the end user if interested.

ggivo:
We can simplify the logic by switching the default and removing the need for **CONSUME_ALL_HANDLER**.
**(Done)**

However, all applicable consumers must still be invoked if present. For example, if a user registers a listener for `maintenance events`, the `MaintenanceEventConsumer` must be called; likewise, we need the `ListenerNotificationConsumer` to notify any registered `PushListener`.

Until we refactor the Pub/Sub system and verify that no external code relies on propagated push events, we still need a mechanism to determine whether a given Push event should be propagated. Any consumer may update this decision, so the order of consumers remains important. 

Implementing an option for skipping a consumer is possible, but if propagation is disabled too early, again we end up with the same uncertainty. So I think the final resolution is to make sure no push messages are propagated at all, but don't think this can be handled easily and need to be tackled separately.

My idea is that **PushConsumers** should be kept internal and should not be exposed to users. Instead, users should interact with the system via PushListener.

ggivo:
> looks good!! there are couple of things not entirely clear to me. Just to clear them on my mind i will summarize as i understand; all push messages goes through all the consumers in the chain, in current impl, they go through;
> 
> * `CONSUME_ALL_HANDLER`
> * `PUBSUB_ONLY_HANDLER`
> * `MaintenanceEventConsumer`
> * `ListenerNotificationConsumer` - helps to plug customer app push listeners
> * `PushInvalidateConsumer` (in case of CSC -`CacheConnection`-)
> 
> what i don't like is each push messages visit all consumers no matter what. And the last one in the list, it kind of decides to set if data handed back to command execution or not. so we expect each consumer to do whatever it wants to do with message and set the context appropriately. this setup brings both flexibility and chaotic behaviour as well, because if someone shows up and marks it as they wish, they challenge to other consumer implementations plugged previously.
> 
> i think no push messages, other than pub/sub ones should be handed back to command executer anyway. (_and at some point, even they are not going to following the command response path in the case we introduce a way to customer app handle pub/sub messages via push listeners_). IMHO there should be something separate from letting client received/notified of push messages as well. May be with `PUBSUB_ONLY_HANDLER` functions the same way with `ListenerNotificationConsumer` but only for pub/sub messages.
> 
> let me try to explain the way i see it;
> 
> * a consumer chooses if each message propagates further to next consumer or not, and returns it as a value of consume result.
>   Each consumer along the chain should receive only what is let by its preceding consumers. this secures if messages should be seen or not, by any other consumer. Also this is more efficient in case there is a load of invalidation and/or pubsub messages.
> 
> ```java
> interface PushConsumer {
>    bool consume(PushMessageContext);
> }
> ```
> 
> * no messages other than pub/sub ones handed back to customer app (or received by command executor). isProcessed flag in context still remains the same to identify if message is chosen to be returned to the app
> * each push message initializes with default `isProcessed`=true and there is no need for a `CONSUME_ALL_HANDLER`. Then current naming of flag "isProcessed" might be suboptimal in this case. Alternatives would be `forwardToClient` or `isRequesterBound`

We can simplify the logic by switching the default and removing the need for **CONSUME_ALL_HANDLER**.
**(Done)**

However, all applicable consumers must still be invoked if present. For example, if a user registers a listener for `maintenance events`, the `MaintenanceEventConsumer` must be called; likewise, we need the `ListenerNotificationConsumer` to notify any registered `PushListener`.

Until we refactor the Pub/Sub system and verify that no external code relies on propagated push events, we still need a mechanism to determine whether a given Push event should be propagated. Any consumer may update this decision, so the order of consumers remains important. 

Implementing an option for skipping a consumer is possible, but if propagation is disabled too early, again we end up with the same uncertainty. So I think the final resolution is to make sure no push messages are propagated at all, but don't think this can be handled easily and need to be tackled separately.

My idea is that **PushConsumers** should be kept internal and should not be exposed to users. Instead, users should interact with the system via PushListener.

src/main/java/redis/clients/jedis/Protocol.java
ggivo: `read(is)` method is left for backward compatibility. In case anyone is using it outside the client code itself.
Will mark it as deprecated and remove in next major release.

ggivo:
Discussed with @atakavci  and we came up with the idea that it will be more optimal if instead of invoking the consumers, for each push message and each consumer decides if it is interested to process it, it will be better for consumers to register themself for a specific message type e.g. ("INVALIDATE") and get invoked only for it. 

Considering this an optimisation, we will try to improve it with a separate PR.

ggivo:
closed in favour of #4473

#4473 
- rebased on the latest code
- stripped down code not required by maintenance events 
   - removed support for user-provided listener for push notifications/maintenance events

PR #4205, opened 2025-07-19
Fix memory leak in JedisClusterInfoCache - replica nodes not cleared

### LINKED ISSUE #4191: Fix memory leak in JedisClusterInfoCache - replica nodes not cleared
If the client is configured with isReadOnlyForRedisClusterReplicas == true, calls to reset() within JedisClusterInfoCache will clear out the slots and slotNodes, but will not wipe out replicaSlots. This means that any calls to assignSlotsToReplicaNode, will cause the lists to grow indefinitely. 

This can cause a memory leak when a node experiences a failure due to repeated calls to refresh the slots.

thachlp:
- Updates the reset() method in JedisClusterInfoCache to clear the replicaSlots list when resetting the cluster cache.
- No external behavior or API is affected

Closes #4191

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
@@ -491,7 +491,8 @@ public class JedisClusterInfoCache {
 
   @SuppressWarnings("unchecked")
   private List<Object> executeClusterSlots(Connection jedis) {
-    CommandArguments clusterSlotsCmd = new ClusterCommandArguments(Protocol.Command.CLUSTER).add("SLOTS");
+    CommandArguments clusterSlotsCmd = new ClusterCommandArguments(Protocol.Command.CLUSTER).add(
+        "SLOTS");
     return (List<Object>) jedis.executeCommand(clusterSlotsCmd);
   }
 
--- a/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
@@ -10,6 +10,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Set;
 
 import java.util.concurrent.Executors;
@@ -68,8 +69,7 @@ public class JedisClusterInfoCache {
     }
   }
 
-  public JedisClusterInfoCache(final JedisClientConfig clientConfig,
-      final Set<HostAndPort> startNodes) {
+  public JedisClusterInfoCache(final JedisClientConfig clientConfig, final Set<HostAndPort> startNodes) {
     this(clientConfig, null, null, startNodes);
   }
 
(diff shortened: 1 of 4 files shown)


src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -454,6 +458,16 @@ public void reset() {
       nodes.clear();
       Arrays.fill(slots, null);
       Arrays.fill(slotNodes, null);
+      if (clientSideCache != null) {
+        clientSideCache.flush();
+      }
```
ggivo: Not sure if we need to flush also csc. Let me take another look.

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -454,6 +458,16 @@ public void reset() {
...
+      if (clientConfig.isReadOnlyForRedisClusterReplicas() && replicaSlots != null) {
```
ggivo: `replicaSlots != null` should be sufficient

ggivo (commented):
Hi, @thachlp 
Thank you for the PR! The overall approach looks good and addresses the issue.
I have a couple of small items I'd like to clean up before merging:

1. Formatting changes: I'll revert the formatting-only changes since we're planning a mass reformatting soon, and having formatting mixed with functional changes makes reviews more difficult.
2. Unit test: I'll also add a unit test to verify the proper cleanup behavior to ensure this fix works as expected and prevents regressions.

I'll make these changes myself and get this merged. Thanks for the great contribution!

src/test/java/redis/clients/jedis/commands/jedis/ControlCommandsTest.java
```diff
@@ -429,23 +432,23 @@ public void memoryUsageBinary() {
...
+    byte[] bbar = { 0x05, 0x06, 0x07, 0x08 };
+    byte[] bfoobar = { 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08 };
 
     jedis.set(bfoo, bbar);
     assertThat(jedis.memoryUsage(bfoo), greaterThan(20l));
 
-    jedis.lpush(bfoobar, new byte[]{0x01, 0x02}, new byte[]{0x05, 0x06}, new byte[]{0x00});
-    assertThat(jedis.memoryUsage(bfoobar, 2), greaterThan(40l));
+    jedis.lpush(bfoobar, new byte[] { 0x01, 0x02 }, new byte[] { 0x05, 0x06 }, new byte[] { 0x00 });
+    assertThat(jedis.memoryUsage(bfoobar, 2), greaterThanOrEqualTo(40l));
```
ggivo: This should not be related to the current issue.

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
ggivo: Don't see a reason to flush it. will leave it aside for now.

PR #4204, opened 2025-07-17
[automatic failover] Implement failback with given grace period

atakavci:
This PR is based on changes in previous #4200.

Commits essential to this one are;
- [- periodic failback checks](https://github.com/redis/jedis/commit/ff163302c9390484d0b4e6c0bf4792bd418e1915)

PR #4207, opened 2025-07-19
 [automatic failover] Implement wait on healthCheck results during client init

atakavci:
This PR is based on changes in previous #4204.

Commits essential to this one are;
- [- introduce StatusTracker with purpose of waiting initial healthcheck…](https://github.com/redis/jedis/commit/c39fda1472e80fd85a51bdfbe02dbc8d7e18df95) 
- [- introduce forceActiveCluster by duration](https://github.com/redis/jedis/commit/975ab784d8f92d119c3ed9337e29a36c46f4d082)


src/main/java/redis/clients/jedis/mcf/CircuitBreakerFailoverBase.java
```diff
@@ -39,34 +40,43 @@ protected void clusterFailover(CircuitBreaker circuitBreaker) {
...
-            // Check state to handle race conditions since incrementActiveMultiClusterIndex() is
+            // Check state to handle race conditions since iterateActiveCluster() is
             // non-idempotent
             if (!CircuitBreaker.State.FORCED_OPEN.equals(circuitBreaker.getState())) {
 
                 // Transitions state machine to a FORCED_OPEN state, stopping state transition, metrics and
                 // event publishing.
                 // To recover/transition from this forced state the user will need to manually failback
+
+                Cluster activeCluster = provider.getCluster();
```
ggivo: Instead why , not provide the actual cluster to the clusterFailover.
We already have reference to it, when building the invocation

protected void clusterFailover(Cluster cluster)

This way will be sure we are performing the operation on the correct cluster.

src/main/java/redis/clients/jedis/mcf/CircuitBreakerFailoverBase.java
```diff
@@ -39,34 +40,43 @@ protected void clusterFailover(CircuitBreaker circuitBreaker) {
...
             }
-
-            // Once the priority list is exhausted only a manual failback can open the circuit breaker so
-            // all subsequent operations will fail
-            else if (provider.isLastClusterCircuitBreakerForcedOpen()) {
+            // this check relies on the fact that many failover attempts can hit with the same CB,
+            // only the first one will trigger a failover, and make the CB FORCED_OPEN.
+            // when the rest reaches here, the active cluster is already the next one, and should be different than
+            // active CB. If its the same one and there are no more clusters to failover to, then throw an exception
+            else if (circuitBreaker == provider.getCluster().getCircuitBreaker() && !provider.canIterateOnceMore()) {
```
ggivo: If we provide the cluster, this check could also be performed on the actual cluster

src/main/java/redis/clients/jedis/mcf/LagAwareStrategy.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    @Override
+    public int getTimeout() {
+        return timeout;
+    }
+
+    @Override
+    public HealthStatus doHealthCheck(Endpoint endpoint) {
+        RedisRestAPIHelper helper = new RedisRestAPIHelper(endpoint.getHost(), String.valueOf(endpoint.getPort()),
```
ggivo: Note not to forget to address hardcoded values

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
```diff
@@ -50,33 +68,46 @@ public class MultiClusterPooledConnectionProvider implements ConnectionProvider
...
     private List<Class<? extends Throwable>> fallbackExceptionList;
 
+    private HealthStatusManager healthStatusManager = new HealthStatusManager();
+    private StatusTracker statusTracker;
+
+    // Flag to control when handleHealthStatusChange should process events (only after initialization)
+    private volatile boolean initializationComplete = false;
+
+    // Queue to hold health status events during initialization
+    private final ConcurrentLinkedQueue<HealthStatusChangeEvent> pendingHealthStatusChanges = new ConcurrentLinkedQueue<>();
```
a-TODO-rov: Currently we store all events that happened. Don't we just need the latest per Endpoint ?

atakavci: well spotted!! 
i guess order of events also important, so let me think of something that regards the order of events while using latest per endpoint.

atakavci: @a-TODO-rov , i got this simplistic approach;
```java
    private void processPendingHealthStatusChanges() {
        HealthStatusChangeEvent event;
        // Synchronize to ensure the order of events when consuming the queue
        synchronized (pendingHealthStatusChanges) {
            // Process all queued events
            while ((event = pendingHealthStatusChanges.poll()) != null) {
                Endpoint endpoint = event.getEndpoint();
                boolean latestInTheQueue = !pendingHealthStatusChanges.stream().anyMatch(e -> e.getEndpoint().equals(endpoint));
                if (latestInTheQueue) {
                    processStatusChangeEvent(event);
                }
            }
        }
    }
```

i understand and accept "_the possibility of processing multiple events per endpoint_" to avoid complexity. 
it is still safe in terms of not skipping the latest changes on an endpoint.

to provide further guarantees relative to my suggestion above, guarantees of ordering and using latest events, it gets real complex. At least i am not able to come up with a nice one quickly. 
i am open to suggestions..

a-TODO-rov: I was thinking more like changing the Collection to `Map<Endpoint, HealthStatusChangeEvent>` and then processing the values. I am not 100% sure that in that case the order of events is crucial (having in mind you have only one event per Endpoint), but if you think it is, you can wrap the HealthStatusChangeEvent into some new Object with timestamp:
{
    private HealthStatusChangeEvent event;
    private long timestamp;
}

atakavci: map is better in performance and clear in implementation.
my only concern with "map solution" is that not preserving order would end up different behaviour than the version its preserved.
@ggivo also share his idea around not queuing them but at least setting the values directly to each cluster while initialization in progress. And start consuming after initialization completes. 
i think both would do better than current one. So i will try the approach without collecting them, and fallback to what you suggest if it doesnt go smooth.

atakavci: the commit related to topic is [this one ](https://github.com/redis/jedis/pull/4207/commits/b8d4e87427cf90e25c307371c9dc35d5823ccaff).

src/test/java/redis/clients/jedis/mcf/FailbackMechanismIntegrationTest.java
```diff
@@ -271,7 +271,7 @@ void testMultipleClusterFailbackPriority() throws InterruptedException {
                 MultiClusterPooledConnectionProviderHelper.onHealthStatusChange(provider, endpoint3, HealthStatus.UNHEALTHY, HealthStatus.HEALTHY);
 
                 // Wait for failback
-                Thread.sleep(200);
+                Thread.sleep(250);
```
a-TODO-rov: Can't we use:
`Awaitility.await().atMost(ONE_SECOND).until(() -> failback())`

src/main/java/redis/clients/jedis/mcf/HealthCheck.java
```diff
@@ -27,7 +27,7 @@ public class HealthCheck {
         this.endpoint = endpoint;
         this.strategy = strategy;
         this.statusChangeCallback = statusChangeCallback;
-        statusRef.set(new SimpleEntry<>(0L, HealthStatus.HEALTHY));
+        statusRef.set(new SimpleEntry<>(0L, HealthStatus.UNKNOWN));
```
ggivo, line 30: I suggest introducing a dedicated entry class, e.g `HealthCheckResult (initiatedAt, healthStatus) ` instead using generic `SimpleEntry`

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
```diff
@@ -282,25 +299,30 @@ private void addClusterInternal(MultiClusterClientConfig multiClusterClientConfi
         StrategySupplier strategySupplier = config.getHealthCheckStrategySupplier();
         if (strategySupplier != null) {
             HealthCheckStrategy hcs = strategySupplier.get(config.getHostAndPort(), config.getJedisClientConfig());
+            // Register listeners BEFORE adding clusters to avoid missing events
+            healthStatusManager.registerListener(config.getHostAndPort(), this::onHealthStatusChange);
             healthStatusManager.add(config.getHostAndPort(), hcs);
+        } else {
+            cluster.setHealthStatus(HealthStatus.HEALTHY);
```
ggivo, line 306: It will be better to have cluster status in a single place vs trying to update it on a change in `HealthStatusManager`

src/main/java/redis/clients/jedis/mcf/StatusTracker.java
```diff
@@ -0,0 +1,77 @@
...
+        try {
+            // Double-check status after registering listener (race condition protection)
+            currentStatus = healthStatusManager.getHealthStatus(endpoint);
+            if (currentStatus != HealthStatus.UNKNOWN) {
+                return currentStatus;
+            }
+
+            // Wait for the health status change event
+            // just for safety to not block indefinitely
+            boolean completed = latch.await(60, TimeUnit.SECONDS);
```
ggivo, line 63: this is made configurable with follow up PR

ggivo (approved):
Agreed that comments in this PR will be adrressed with separate PR against feature branch to allow merging other PR's  pending on this one and reduce conflict's resolution

src/main/java/redis/clients/jedis/mcf/StatusTracker.java
atakavci: see https://github.com/redis/jedis/issues/4233

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
atakavci: see https://github.com/redis/jedis/issues/4233

src/main/java/redis/clients/jedis/mcf/HealthCheck.java
atakavci: see https://github.com/redis/jedis/issues/4233

src/main/java/redis/clients/jedis/mcf/LagAwareStrategy.java
atakavci: see https://github.com/redis/jedis/issues/4233

src/main/java/redis/clients/jedis/mcf/CircuitBreakerFailoverBase.java
atakavci: see https://github.com/redis/jedis/issues/4233

src/main/java/redis/clients/jedis/mcf/CircuitBreakerFailoverBase.java
atakavci: see https://github.com/redis/jedis/issues/4233

PR #4218, opened 2025-07-29
Clarify why new stream entries aren't deleted with XDELEX

uglide:


ggivo (commented):
What about something like?

RETAINED
IN_USE
REFERENCED

and a comment to clarify when such state is returned

Issue #4144, opened 2025-04-15
JedisCluster functionLoadReplace throws JedisBroadcastException


oscarboher:
**Expected behavior**

functionLoadReplace(script) using a JedisCluster should load the function without errors.

The script is a dummy script:
```
#!lua name=mylib
redis.register_function({
    function_name = 'mylib',
    callback = function() return 0 end,
    description = "hi how are you"
})
```

Using Jedis 5.2.0

**Actual behavior**

functionLoadReplace throws a JedisBroadcastException error:
Caused by: redis.clients.jedis.exceptions.JedisBroadcastException: A failure occurred while broadcasting the command.
	at redis.clients.jedis.executors.ClusterCommandExecutor.broadcastCommand(ClusterCommandExecutor.java:47) ~[jedis-5.2.0.jar:?]
	at redis.clients.jedis.UnifiedJedis.broadcastCommand(UnifiedJedis.java:312) ~[jedis-5.2.0.jar:?]
	at redis.clients.jedis.UnifiedJedis.checkAndBroadcastCommand(UnifiedJedis.java:325) ~[jedis-5.2.0.jar:?]
	at redis.clients.jedis.UnifiedJedis.functionLoadReplace(UnifiedJedis.java:3552) ~[jedis-5.2.0.jar:?]

Checking in the cluster the function does exist, so I understand that in some node the load was successful, but failed in the broadcast.

This happens for both functions that write to the ddbb and for functions with the 'no-write' flag, so It would seem it is not because of trying to upload the function to a replica node.

Doing functionLoadReplace on each of the nodes of the cluster using a Jedis resource does work, both for replicas and masters, with or without the `no-write` flag.

ggivo:
@oscarboher 
Thanks for the report. 
It will speed up the investigation if you could provide a  reproducible example which will  help to be on the same page how is JedisCluster created/configured/

oscarboher:
Sure, this is the simplest way to test it, using Jedis 5.2.0.
We are using user/pwd authentication, and ssl, but can be removed to the same effect.

I checked against a redis-cluster installed locally (as well as our environment) - The error only happens when the cluster has replicas, with only 3 masters the JedisBroadcastException does not appear.

```
package org.example;

import redis.clients.jedis.ConnectionPoolConfig;
import redis.clients.jedis.DefaultJedisClientConfig;
import redis.clients.jedis.HostAndPort;
import redis.clients.jedis.JedisCluster;

import java.time.Duration;
[...]

Issue #3303, opened 2023-02-27
Broadcast and Round-Robin commands in JedisCluster


uglide:
Cluster connection classes should provide overrides for specific Redis Core and module commands to make them easy to use with OS Cluster API.

- [x] Commands that should be executed on primaries with Round-robin load balancing/scheduling
    - [x] [FT.SEARCH](https://redis.io/commands/ft.search/)
    - [x] [FT.AGGREGATE](https://redis.io/commands/ft.aggregate/)
    - [x] [TS.MGET](https://redis.io/commands/ts.mget/)
    - [x] [TS.MRANGE](https://redis.io/commands/ts.mrange/)
- [ ] Provide an abstraction that handles `FT.AGGREGATE` and executes `FT.CURSOR READ` or `FT.CURSOR DEL` on the correct node (on the node where FT.AGGREGATE was executed) automatically 
- [x] Commands that should be broadcasted to all primaries
  - [x] [CONFIG SET](https://redis.io/commands/config-set/)
  - [x] [FLUSHALL](https://redis.io/commands/flushall/)
  - [x] [FLUSHDB](https://redis.io/commands/flushdb/)
  - [x] [FUNCTION DELETE](https://redis.io/commands/function-delete/)
  - [x] [FUNCTION FLUSH](https://redis.io/commands/function-flush/)
  - [x] [FUNCTION KILL](https://redis.io/commands/function-kill/)
  - [x] [FUNCTION RESTORE](https://redis.io/commands/function-restore/)
  - [x] [SCRIPT FLUSH](https://redis.io/commands/script-flush/)
  - [x] [SCRIPT KILL](https://redis.io/commands/script-kill/)
  - [x] [SLOWLOG RESET](https://redis.io/commands/slowlog-reset/)
  - [x] [FT.ALIASADD](https://redis.io/commands/ft.aliasadd/)
  - [x] [FT.ALIASDEL](https://redis.io/commands/ft.aliasdel/)
  - [x] [FT.ALIASUPDATE](https://redis.io/commands/ft.aliasupdate/)
  - [x] [FT.ALTER](https://redis.io/commands/ft.alter/)
  - [x] [FT.CREATE](https://redis.io/commands/ft.create/)
  - [x] [FT.DROPINDEX](https://redis.io/commands/ft.dropindex/)
 - [ ] JSON.MGET should return all documents from the cluster by broadcasting and merging results

PR #4219, opened 2025-07-31
Fix JedisBroadcastException in functionLoadReplace for Redis Cluster

Kguswo:
### Change Summary

Fix JedisBroadcastException in functionLoadReplace by restricting FUNCTION commands to primary nodes only.

As FUNCTION operations are write commands, they should not be broadcast to read-only replica nodes. This change updates the broadcast logic to filter primary nodes for these commands and adds tests to verify the behavior.

Open to suggestions or corrections!

### Related Issue
#4144

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
@@ -425,6 +425,27 @@ public class JedisClusterInfoCache {
     }
   }
 
+  public Map<String, ConnectionPool> getPrimaryNodes() {
+    r.lock();
+    try {
+      Map<String, ConnectionPool> primaryNodes = new HashMap<>();
+      Set<ConnectionPool> addedPools = new HashSet<>();
+
+      for (int slot = 0; slot < slots.length; slot++) {
+        ConnectionPool pool = slots[slot];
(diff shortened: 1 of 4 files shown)


ggivo (commented):
Hi @Kguswo,

Thank you for your contribution and for taking the time to work on this issue! 🙏

After reviewing your PR, I think the root cause of the problem goes back to the initial implementation when broadcasting was introduced ([#3303](https://github.com/redis/jedis/issues/3303)). The original goal was to broadcast commands only to primary nodes, but it seems that currently we’re broadcasting them to all nodes. This affects all broadcasted commands, not just the FUNCTION subcommands.

To address this for all commands, we can introduce a cache of primary nodes (similar to your suggestion) and use that for all broadcasts—no need to filter by command for now. Since topology refreshes are relatively infrequent, it’s better to keep this primary-node cache updated after each topology change.

I’ve applied these changes directly in this PR so we can build on your work, and I’ll also ping the original requester of #3303 for review.

Thanks again for your effort—it’s much appreciated!

ggivo:
This change is somewhat related to broadcasting changes in https://github.com/redis/jedis/pull/4173
In my opinion, they are not contradicting. @uglide, could you also take a look?

PR #4222, opened 2025-08-01
Add support for SVS-VAMANA vector indexing

uglide:
Adds support for [SVS-VAMANA](https://redis.io/docs/latest/develop/ai/search-and-query/vectors/#svs-vamana-index) vector index types.

### DIFF
--- a/src/main/java/redis/clients/jedis/search/Schema.java
+++ b/src/main/java/redis/clients/jedis/search/Schema.java
@@ -138,6 +138,21 @@ public class Schema {
     return this;
   }
 
+  /**
+   * Add a Vamana vector field to the schema using the SVS-VAMANA algorithm.
+   * This method provides a convenient way to add SVS-VAMANA vector fields.
+   *
+   * @param name the field's name
+   * @param attributes the SVS-Vamana algorithm configuration attributes
+   * @return the schema object
+   */
+  public Schema addSvsVamanaVectorField(String name, Map<String, Object> attributes) {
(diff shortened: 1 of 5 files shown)


src/main/java/redis/clients/jedis/search/Schema.java
```diff
@@ -350,7 +365,8 @@ public static class VectorField extends Field {
 
     public enum VectorAlgo {
       FLAT,
-      HNSW
+      HNSW,
+      SVS_VAMANA
```
ggivo: Seems this will not work without making it Rawable similar  to 

`public enum VectorAlgorithm implements Rawable`

We need also a  test similar to `redis.clients.jedis.modules.search.SearchTest#testHNSWVectorSimilarity`

ggivo (approved):
LGTM

src/test/java/redis/clients/jedis/modules/search/SearchTest.java
```diff
@@ -1179,6 +1179,30 @@ public void testHNSWVVectorSimilarity() {
     assertEquals("0", doc1.get("__v_score"));
   }
 
+  @Test
+  public void testSvsVamanaVectorSimilarity() {
```
ggivo, line 1185:
```suggestion
  @SinceRedisVersion("8.1.240")
  public void testSvsVamanaVectorSimilarity() {
```

ggivo (approved):

ggivo (approved):

PR #4225, opened 2025-08-05
Restore binary compatibility of SetParams

### LINKED ISSUE #4224: Binary backward incompatibility in 6.0.0
Some methods, like `SetParams.px()` were moved to `BaseSetExParams` which is source compatible change but is not binary compatible (code compiled with an old version does not work with 6.0.0 and vice versa). This make it hard for library author to support both 6.0.0 and older version. Can you please reintroduce the methods back to `SetParams`?

See https://github.com/lukas-krecan/ShedLock/issues/2762

ggivo:
### Background

In Jedis 6.0.0, methods like `SetParams.px()` and `ex()` were moved to the superclass `BaseSetExParams`. While this change is source compatible, it introduces a binary incompatibility: code compiled against Jedis 5.x expecting these methods directly in `SetParams` will fail at runtime with `NoSuchMethodError`.

### Changes

- Reintroduce delegated methods (`px()`, `ex()`, etc.) in `SetParams`.
- Delegated methods call the corresponding superclass implementations.

Closes #4224


atakavci (approved):
LGTM!

PR #4230, opened 2025-08-12
Fix: EchoStrategy does not recover on connection error

ggivo:


ggivo:
Alternative solution is implemented with #4226 ([see here](  https://github.com/redis/jedis/pull/4226/commits/397f4374dc2b32ffdfb564a0862262c163dc45e7#diff-81e4182c8db8d77cf0d0d53384388fd34e81dab27b74a307b8cd3325a95d4592)  using UnifiedJedis client and implementing close on helathchecks

We are still planning to use the test's from this PR once #4226 is  merged, let's get back and update it

atakavci (approved):
LGTM!

PR #4231, opened 2025-08-13
Implemented read-write separation based on JedisSentineled

koleter:
Implemented read-write separation based on JedisSentineled

### DIFF
--- a/src/main/java/redis/clients/jedis/Protocol.java
+++ b/src/main/java/redis/clients/jedis/Protocol.java
@@ -23,7 +23,7 @@ public final class Protocol {
 
   public static final String DEFAULT_HOST = "127.0.0.1";
   public static final int DEFAULT_PORT = 6379;
-  public static final int DEFAULT_SENTINEL_PORT = 26378;
+  public static final int DEFAULT_SENTINEL_PORT = 26379;
   public static final int DEFAULT_TIMEOUT = 2000;
   public static final int DEFAULT_DATABASE = 0;
   public static final int CLUSTER_HASHSLOTS = 16384;
--- a/src/main/java/redis/clients/jedis/Protocol.java
+++ b/src/main/java/redis/clients/jedis/Protocol.java
@@ -23,7 +23,7 @@ public final class Protocol {
 
   public static final String DEFAULT_HOST = "127.0.0.1";
   public static final int DEFAULT_PORT = 6379;
-  public static final int DEFAULT_SENTINEL_PORT = 26379;
+  public static final int DEFAULT_SENTINEL_PORT = 26378;
   public static final int DEFAULT_TIMEOUT = 2000;
   public static final int DEFAULT_DATABASE = 0;
   public static final int CLUSTER_HASHSLOTS = 16384;
--- a/src/main/java/redis/clients/jedis/Protocol.java
+++ b/src/main/java/redis/clients/jedis/Protocol.java
@@ -283,45 +283,64 @@ public final class Protocol {
 
   public static enum Command implements ProtocolCommand {
 
-    PING, AUTH, HELLO, SET, GET, GETDEL, GETEX, EXISTS, DEL, UNLINK, TYPE, FLUSHDB, FLUSHALL, MOVE,
-    KEYS, RANDOMKEY, RENAME, RENAMENX, DUMP, RESTORE, DBSIZE, SELECT, SWAPDB, MIGRATE, ECHO, //
-    EXPIRE, EXPIREAT, EXPIRETIME, PEXPIRE, PEXPIREAT, PEXPIRETIME, TTL, PTTL, // <-- key expiration
-    MULTI, DISCARD, EXEC, WATCH, UNWATCH, SORT, SORT_RO, INFO, SHUTDOWN, MONITOR, CONFIG, LCS, //
-    GETSET, MGET, SETNX, SETEX, PSETEX, MSET, MSETNX, DECR, DECRBY, INCR, INCRBY, INCRBYFLOAT,
-    STRLEN, APPEND, SUBSTR, // <-- string
-    SETBIT, GETBIT, BITPOS, SETRANGE, GETRANGE, BITCOUNT, BITOP, BITFIELD, BITFIELD_RO, // <-- bit (string)
-    HSET, HGET, HSETNX, HMSET, HMGET, HINCRBY, HEXISTS, HDEL, HLEN, HKEYS, HVALS, HGETALL, HSTRLEN,
-    HEXPIRE, HPEXPIRE, HEXPIREAT, HPEXPIREAT, HTTL, HPTTL, HEXPIRETIME, HPEXPIRETIME, HPERSIST,
-    HRANDFIELD, HINCRBYFLOAT, HSETEX, HGETEX, HGETDEL, // <-- hash
-    RPUSH, LPUSH, LLEN, LRANGE, LTRIM, LINDEX, LSET, LREM, LPOP, RPOP, BLPOP, BRPOP, LINSERT, LPOS,
-    RPOPLPUSH, BRPOPLPUSH, BLMOVE, LMOVE, LMPOP, BLMPOP, LPUSHX, RPUSHX, // <-- list
-    SADD, SMEMBERS, SREM, SPOP, SMOVE, SCARD, SRANDMEMBER, SINTER, SINTERSTORE, SUNION, SUNIONSTORE,
-    SDIFF, SDIFFSTORE, SISMEMBER, SMISMEMBER, SINTERCARD, // <-- set
-    ZADD, ZDIFF, ZDIFFSTORE, ZRANGE, ZREM, ZINCRBY, ZRANK, ZREVRANK, ZREVRANGE, ZRANDMEMBER, ZCARD,
-    ZSCORE, ZPOPMAX, ZPOPMIN, ZCOUNT, ZUNION, ZUNIONSTORE, ZINTER, ZINTERSTORE, ZRANGEBYSCORE,
-    ZREVRANGEBYSCORE, ZREMRANGEBYRANK, ZREMRANGEBYSCORE, ZLEXCOUNT, ZRANGEBYLEX, ZREVRANGEBYLEX,
-    ZREMRANGEBYLEX, ZMSCORE, ZRANGESTORE, ZINTERCARD, ZMPOP, BZMPOP, BZPOPMIN, BZPOPMAX, // <-- zset
-    GEOADD, GEODIST, GEOHASH, GEOPOS, GEORADIUS, GEORADIUS_RO, GEOSEARCH, GEOSEARCHSTORE,
-    GEORADIUSBYMEMBER, GEORADIUSBYMEMBER_RO, // <-- geo
-    PFADD, PFCOUNT, PFMERGE, // <-- hyper log log
-    XADD, XLEN, XDEL, XTRIM, XRANGE, XREVRANGE, XREAD, XACK, XGROUP, XREADGROUP, XPENDING, XCLAIM,
-    XAUTOCLAIM, XINFO, // <-- stream
-    EVAL, EVALSHA, SCRIPT, EVAL_RO, EVALSHA_RO, FUNCTION, FCALL, FCALL_RO, // <-- program
-    SUBSCRIBE, UNSUBSCRIBE, PSUBSCRIBE, PUNSUBSCRIBE, PUBLISH, PUBSUB,
-    SSUBSCRIBE, SUNSUBSCRIBE, SPUBLISH, // <-- pub sub
-    SAVE, BGSAVE, BGREWRITEAOF, LASTSAVE, PERSIST, ROLE, FAILOVER, SLOWLOG, OBJECT, CLIENT, TIME,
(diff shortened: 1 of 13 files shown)


src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -283,47 +283,66 @@ public static final byte[] toByteArray(final double value) {
...
 
     @Override
     public byte[] getRaw() {
       return raw;
     }
+
+    @Override
+    public boolean isWriteCommand() {
+      return isWriteCommand;
+    }
```
ggivo: In some cases, we might have `COMMAND`, having both read-only & write sub-commands.
For example, `FUNCTION LIST` vs `FUNCTION LOAD`. I think we should not hardcode this based on the command type.  As pointed in the review comment, I think we can have an approach similar to Lettuce, where we can use a configurable predicate to identify read-only commands.

koleter:
@ggivo  OK, I modified the submitted code, please review it again

src/main/java/redis/clients/jedis/providers/SentineledConnectionProvider.java
```diff
@@ -167,6 +224,79 @@ private ConnectionPool createNodePool(HostAndPort master) {
     }
   }
 
+  private void initSlaves(List<HostAndPort> slaves) {
+    List<PoolInfo> removedSlavePools = new ArrayList<>();
+    try {
+      synchronized (slavePools) {
```
ggivo: `synchronized` blocks cause carrier thread pinning, which undermines Loom’s scalability model. To avoid this, we replaced `synchronized` blocks with `ReentrantLock` in other parts of the driver. (#3480)

We should avoid introducing new synchronized sections here, and instead rely on other type syncronisation.

src/main/java/redis/clients/jedis/providers/SentineledConnectionProvider.java
```diff
@@ -102,13 +116,52 @@ public SentineledConnectionProvider(String masterName, final JedisClientConfig m
...
+      Connection slaveConn = getSlaveResource();
+      if (slaveConn != null) {
+        return slaveConn;
+      }
+      if (!masterClientConfig.isFallbackToMaster()) {
+        throw new JedisException("can not get Connection, all slave is invalid");
+      }
+    }
     return pool.getResource();
   }
```
ggivo, line 307: See my comment about preserving backword compatibility

src/main/java/redis/clients/jedis/JedisClientConfig.java
```diff
@@ -115,4 +116,21 @@ default boolean isReadOnlyForRedisClusterReplicas() {
...
+  }
+
+  /**
+   * check a Command is READONLY
+   * @param args
+   * @return
+   */
+  default boolean isReadCommand(CommandArguments args) {
+    return Commands.ReadOnlyCommands.contains(args.getCommand());
+  }
```
ggivo: We are introducing support for **READ commands** only for the `JedisSentineled` client.
Because of this, the read-from settings should be configured specifically in `JedisSentineled` (or its `SentineledConnectionProvider`) rather than being placed in the generic client configuration.

In addition, I think it is better to determine whether a command is a READ command by exposing a  configurable Predicate. This gives users flexibility and avoids the need to override methods. I like how this is achieved  in Lettuce:
	•	Predicate interface: [ReadOnlyPredicate](https://github.com/redis/lettuce/blob/21bbbb8971166ecd9409c01c9444e467fe1f0ee1/src/main/java/io/lettuce/core/protocol/ReadOnlyCommands.java#L91-L91)
	•	Example implementation: [ReadOnlyCommands](https://github.com/redis/lettuce/blob/21bbbb8971166ecd9409c01c9444e467fe1f0ee1/src/main/java/io/lettuce/core/protocol/ReadOnlyCommands.java#L38-L38)
	•	Configuration: [ClientOptions](https://github.com/redis/lettuce/blob/21bbbb8971166ecd9409c01c9444e467fe1f0ee1/src/main/java/io/lettuce/core/ClientOptions.java#L369-L369)

src/main/java/redis/clients/jedis/providers/SentineledConnectionProvider.java
```diff
@@ -275,24 +423,58 @@ public void run() {
...
+                  addSlave(new HostAndPort(slaveIp, slavePort));
+
+                  String masterIp = switchMsg[6];
+                  int masterPort = Integer.parseInt(switchMsg[7]);
+                  removeSlave(new HostAndPort(masterIp, masterPort));
+                  break;
               }
             }
-          }, "+switch-master");
+          }, "+switch-master", "+sdown", "-sdown", "+slave");
```
ggivo, line 660: Based on  [High availability with Redis Sentinel ](https://redis.io/docs/latest/operate/oss_and_stack/management/sentinel/)
```
+sentinel <instance details> -- A new sentinel for this master was detected and attached.
+sdown <instance details> -- The specified instance is now in Subjectively Down state.
-sdown <instance details> -- The specified instance is no longer in Subjectively Down state.
+odown <instance details> -- The specified instance is now in Objectively Down state.
-odown <instance details> -- The specified instance is no longer in Objectively Down state.
```
Not clear the difference  between `sdown`/`odown` events. If I am reading it right `odown` events are send when majority of sentinels consider the node down, and `sdown` is send when one sentinel considers the node down. 

Probably we should react on `odown` events.

src/main/java/redis/clients/jedis/providers/SentineledConnectionProvider.java
```diff
@@ -167,6 +224,79 @@ private ConnectionPool createNodePool(HostAndPort master) {
     }
   }
 
+  private void initSlaves(List<HostAndPort> slaves) {
+    List<PoolInfo> removedSlavePools = new ArrayList<>();
+    try {
+      synchronized (slavePools) {
+        Loop:
```
ggivo: nit: The nested loop with continue Loop reduces readability .

src/main/java/redis/clients/jedis/providers/SentineledConnectionProvider.java
koleter: There is a problem here. When I kill a redis node, +odown is triggered, but when the node comes back online, there is no -odown event.

koleter: And +odown will only be triggered when the master goes down, my redis version is 6.2.6

koleter:
@ggivo The majority opinion has been revised, but there are some situations regarding sdown/odown events that may need to remain as is.

ggivo:
> @ggivo The majority opinion has been revised, but there are some situations regarding sdown/odown events that may need to remain as is.

Sorry for the late reply. I will take a look at it until the end of the week.

uglide:
Hi @koleter 
I know you put a lot of effort in this PR, but we doing some internal improvements that affect your work. See https://github.com/redis/jedis/pull/4332/commits/acb6cfde74bb5f9fcfeb763e4eb605e8e78f6272 for more details.

koleter:
@uglide @ggivo Should I keep the existing logic or use CommandFlag?

ggivo:
> Should I keep the existing logic or use CommandFlag?

IMO, using CommandFlags for distinguishing the read-only commands makes sense, but probably we need to come up with a solution for how to handle also sub-command flags.
@koleter @uglide  thoughts/suggestions?

koleter:
@ggivo Perhaps I could start with CommandFlags and, like Lettuce, provide a user-configurable function hook for distinguishing read-only commands. The way subcommand flags are handled can be extended later on top of CommandFlags.

ggivo:
@koleter 
We have just updated the CommandFlags to support also sub-commands, you can take a look and let me know if this works for your use case also.

> like Lettuce, provide a user-configurable function hook for distinguishing read-only commands.

I was also thinking it might be beneficial to have a user defined hook. 
Similar to Lettuce we can use CommandFlags as source of info for readonly commands for default isReadOnly check, but users can provide their own. If you mean the same.

koleter:
@ggivo Running the test cases is taking too long. Could you please check if the current modifications are feasible?

ggivo:
@koleter 
Hey, thanks for moving this forward.
We are a bit stretched currently, wrapping up some work  8.4 release, will do my best to provide feedback early next week

koleter:
Hi maintainers,
It looks like the GitHub Actions workflow is awaiting approval.
Could you please approve and trigger the CI when you have time? Thanks!

koleter:
@ggivo 

the failure in Build and Test (8.2) seems unrelated to the PR changes and is caused by a Maven Central connection reset during dependency:go-offline (OSGi artifacts).
Could you please help re-run the workflow?

Thanks a lot!

koleter:
Hi maintainers,
when you have time, could you please help take a look at this PR?
Thanks a lot!

ggivo:
@koleter
First of all, sorry for the slow response here.
The team has been quite stretched recently, which has delayed our review, but we’ll do our best to provide feedback as soon as we have some time.

Really appreciate the effort and persistence you’ve put into this PR — thank you for sticking with it!

koleter:
@ggivo Does the logic regarding sdown/odown events in this PR need to be changed to odown? I've noticed this PR hasn't been merged into master mode for a long time. Is there anything else I need to do?

src/main/java/redis/clients/jedis/providers/SentineledConnectionProvider.java
```diff
@@ -339,6 +558,24 @@ public void run() {
 
           sentinelJedis = sentinelConnectionFactory.createConnection(node, sentinelClientConfig);
 
+          List<Map<String, String>> slaveInfos = sentinelJedis.sentinelSlaves(masterName);
```
ggivo, line 561: connections to slave should not be initialised unless we plan to use them. Which by default is disabled

src/main/java/redis/clients/jedis/providers/SentineledConnectionProvider.java
```diff
@@ -65,6 +77,16 @@ public class SentineledConnectionProvider implements ConnectionProvider {
...
 
+  private final ReadFrom readFrom;
+
+  private ReadOnlyPredicate readOnlyPredicate;
+
+  private final List<PoolInfo> slavePools = new ArrayList<>();
+
+  private final Lock slavePoolsLock = new ReentrantLock(true);
+
+  private int poolIndex;
```
ggivo, line 88: Round robin lgic should be encapsulated in dedicated class.
E.g ConnectionResolver.resolve(List<Nodes>), then we can have RoundRobinConnectionResolver or RandomConnectionResolver

src/main/java/redis/clients/jedis/providers/SentineledConnectionProvider.java
```diff
@@ -164,18 +216,94 @@ public SentineledConnectionProvider(String masterName, final JedisClientConfig m
...
 
+    this.readFrom = readFrom;
+    this.readOnlyPredicate = readOnlyPredicate;
+
     HostAndPort master = initSentinels(sentinels);
     initMaster(master);
   }
 
+
+  private Connection getSlaveResource() {
```
ggivo, line 227: see my comment above, e.g connection of selection logic should be encapsulated

src/main/java/redis/clients/jedis/providers/SentineledConnectionProvider.java
ggivo: I still have concern here, that relyin gon events to update the topology easily can get us in a satte which is not correct. especially relaying on +-/sdown events which are not agrreed by majority of the servers.
Starting to thing that we should use events only as triigers to reload the topology (e.g use the sentinel master/slaves) to read the actual state

src/test/java/redis/clients/jedis/SentineledConnectionProviderTest.java
```diff
@@ -239,12 +249,102 @@ public void testResetValidPassword() {
...
+    DefaultRedisCredentialsProvider credentialsProvider
+            = new DefaultRedisCredentialsProvider(new DefaultRedisCredentials(null, password));
+
+    try (JedisSentineled jedis = new JedisSentineled(MASTER_NAME, DefaultJedisClientConfig.builder()
+            .timeoutMillis(2000).credentialsProvider(credentialsProvider).database(2)
+            .clientName("my_shiny_client_name").build(), new ConnectionPoolConfig(),
+            sentinels, DefaultJedisClientConfig.builder().build(), ReadFrom.REPLICA)) {
+
+      Thread.sleep(1000);
+      Whitebox.setInternalState(jedis.provider, "slavePools", new ArrayList<>());
```
ggivo, line 288: Probably we need to refactore  the provider, and make it testable instead

ggivo:
> Does the logic regarding sdown/odown events in this PR need to be changed to odown? I've noticed this PR hasn't been merged into master mode for a long time. Is there anything else I need to do?

@koleter 

Just going through the PR, and reviewing it.
I plan to spend some more time on it tomorrow also, but let me post a few things I spotted and think we need to reconsider. The main one is the comment above.  Still consider that relying on events to update the internal state might easily get us out-of-sync and a probably safer approach would be on an event to reread the topology.

I am also trying to figure out if we can refactor code in SentineledConnectionProvider so we don't need to use PowerMockRunner, and better isolate responsibilities for loading the master/replica topology, round-robin between connections.  

Posting my preliminray notes, and will spend some more time later on

koleter:
@ggivo  I think I should wait until you finalize a detailed plan or until you've completed the refacting.

PR #4235, opened 2025-08-15
 [automatic failover] Implement lag-aware api

### LINKED ISSUE #4239: LagAwareStrategy implementation is kind of draft and should be implemented properly
`LagAwareStrategy` implementation is kind of draft and should be implemented properly, suggeste [here](https://github.com/redis/jedis/pull/4207#discussion_r2224658189).

atakavci:
- Closes/Fixes #4239


src/main/java/redis/clients/jedis/mcf/LagAwareStrategy.java
```diff
@@ -32,19 +50,32 @@ public int minConsecutiveSuccessCount() {
...
-                if ("available".equals(helper.checkBdbAvailability(bdbs.get(0)))) {
-                    return HealthStatus.HEALTHY;
+            String bdb = bdbId;
+            if (bdb == null) {
+                List<String> bdbs = redisRestAPI.getBdbs();
+                if (bdbs.size() > 0) {
+                    bdb = bdbs.get(0);
+                    bdbId = bdb;
                 }
             }
```
ggivo: This way might be checking the Health status of an arbitrary BDB available.
Should we enforce the bdbId as required configuration or resolve it somehow?

src/main/java/redis/clients/jedis/mcf/RedisRestAPI.java
```diff
@@ -0,0 +1,163 @@
...
+        java.util.Arrays.fill(passBytes, (byte) 0);
+        java.util.Arrays.fill(combined, (byte) 0);
+        return "Basic " + encodedAuth;
+    }
+
+    /**
+     * Parses the response body and extracts the list of UIDs.
+     */
+    static List<String> parseUidsFromResponse(String responseBody) {
+        return JsonParser.parseString(responseBody).getAsJsonArray().asList().stream().map(e -> {
```
ggivo: Extract bdbinfo from the response, avoiding hard dependency on Gson JsonParser. 
Going forward, Gson should be an optional dependency

pom.xml
```diff
@@ -341,8 +341,11 @@
 						<include>src/main/java/redis/clients/jedis/mcf/*.java</include>
 						<include>src/test/java/redis/clients/jedis/failover/*.java</include>
 						<include>src/test/java/redis/clients/jedis/mcf/EchoStrategyIntegrationTest.java</include>
+						<include>src/test/java/redis/clients/jedis/mcf/LagAwareStrategyUnitTest.java</include>
```
atakavci, line 344: nit: adding these using wildcard is more convenient instead of adding each file.

ggivo: Agreee, just there are set of files not formated in the folder, didn't want to cluter this review. Will do it separately

src/main/java/redis/clients/jedis/mcf/RedisRestAPI.java
ggivo: Seems not trivial and will use the exisgin gson dependency

src/main/java/redis/clients/jedis/mcf/LagAwareStrategy.java
```diff
@@ -1,50 +1,226 @@
...
         }
         return HealthStatus.UNHEALTHY;
     }
+
+    public static class Config extends HealthCheckStrategy.Config {
+
+        public static final boolean EXTENDED_CHECK_DEFAULT = true;
+        public static final Duration AVAILABILITY_LAG_TOLERANCE_DEFAULT = Duration.ofMillis(100);
+
+        private final Endpoint endpoint;
```
atakavci: its useful to set a name that mentions its different than db endpoint
```suggestion
        private final Endpoint restEndpoint;
```

src/main/java/redis/clients/jedis/mcf/LagAwareStrategy.java
```diff
@@ -1,50 +1,226 @@
...
+        } catch (Exception e) {
+            log.error("Error while checking database availability", e);
+            bdbId = null;
         }
         return HealthStatus.UNHEALTHY;
     }
+
+    public static class Config extends HealthCheckStrategy.Config {
+
+        public static final boolean EXTENDED_CHECK_DEFAULT = true;
```
atakavci, line 80: is it for sure extended check is the default one,, or it is just for failback?

src/main/java/redis/clients/jedis/mcf/LagAwareStrategy.java
```diff
@@ -1,50 +1,226 @@
...
+        }
+
+        /**
+         * Perform standard datapath validation only.
+         * <p>
+         * Extended checks like lag validation is disabled by default. To enable extended checks, use
+         * {@link #lagAware(Endpoint, Supplier)} or {@link #lagAwareWithTolerance(Endpoint, Supplier, Duration)}
+         * </p>
+         */
+        public static Config dataPathAvailability(Endpoint endpoint, Supplier<RedisCredentials> credentialsSupplier) {
```
atakavci: nits: it might be confusing for users to see the name "dataPathAvailability", since its never referred as "data path" within the official docs and help pages for availability. i think its just "database availability".

src/main/java/redis/clients/jedis/mcf/LagAwareStrategy.java
```diff
@@ -1,50 +1,226 @@
...
+         * Builder for LagAwareStrategy.Config.
+         */
+        public static class ConfigBuilder extends HealthCheckStrategy.Config.Builder<ConfigBuilder, Config> {
+            private final Endpoint endpoint;
+            private final Supplier<RedisCredentials> credentialsSupplier;
+
+            // Maximum acceptable lag in milliseconds (default: 100);
+            private Duration availabilityLagTolerance = AVAILABILITY_LAG_TOLERANCE_DEFAULT;
+
+            // Enable extended lag checking (default: false - performs only standard datapath validation)
```
atakavci: here comment suggests default is false for `extendedCheckEnabled` while value of EXTENDED_CHECK_DEFAULT is true

ggivo:
> all minor things in my comments... thank you for working on and completing this PR @ggivo i decided to approve, but looks like I am not the approver for this one.

Thanks for the review!
I see, since you are the author.
Let me address the comments and then will merge it!

src/main/java/redis/clients/jedis/mcf/LagAwareStrategy.java
ggivo: make sense. will do.

src/main/java/redis/clients/jedis/mcf/LagAwareStrategy.java
ggivo: You mean to switch the default to be only a standard `databaseAvailability` check?
I was thinking since the primary goal of this health check is to be based on the `lag`, I enabled the extended check by default.

atakavci: i see the documentation on extended availability check stresses this API would not be a best fit on "failover" but on "failback".

ggivo: I saw the same, but then it would be AvailabilityStrategy vs LagAwareStrategy, and  I think it makes sense to use the lag-aware strategy for failover, also in some setups.

atakavci: lets check this with @uglide as well.

PR #4244, opened 2025-08-15
 [automatic failover] Refactor cluster to replace `HealthStatus` with `HealthCheck` itself

### LINKED ISSUE #4240: Refactor `Cluster` to keep a reference to `HealthCheck` itselft instead of tracking/updating `HealthStatus` with events
Refactor Cluster to keep a reference to HealthCheck itselft instead of tracking/updating HealthStatus with events, suggested [here](https://github.com/redis/jedis/pull/4207#discussion_r2272461879).

### LINKED ISSUE #4242: Move `Endpoint` into proper package

### LINKED ISSUE #4243: Set sleep duration 1 ms for handling all the waiters in `TrackingConnectionPool.forceDisconnect`
Set sleep duration 1 ms for handling all the waiters in `TrackingConnectionPool.forceDisconnect` [here](https://github.com/redis/jedis/pull/4226/files#diff-c2c16a7d4ab22053b10551fc8612174f552820376ab8d16b0858f21eaf2094edR183).

atakavci:
- Closes/Fixes #4240 
- Closes/Fixes #4242
- Closes/Fixes #4243


src/main/java/redis/clients/jedis/mcf/HealthCheckImpl.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    private static final Logger log = LoggerFactory.getLogger(HealthCheck.class);
+
+    private Endpoint endpoint;
+    private HealthCheckStrategy strategy;
+    private AtomicReference<HealthCheckResult> resultRef = new AtomicReference<HealthCheckResult>();
+    private Consumer<HealthStatusChangeEvent> statusChangeCallback;
+
+    private ScheduledExecutorService scheduler;
```
ggivo, line 46: **Note for future improvement:**

Currently, each HealthCheck instance uses its own `ScheduledExecutorService` for scheduling (via `scheduleAtFixedRate`) and a separate executor (`Executors.newCachedThreadPool()`) for running the actual health check tasks.

That works but we might save some resources by sharing the scheduler among all HealthChecks

**Future enhancement idea:**
Consider introducing a shared scheduler and/or executor service across all HealthCheck instances:
- 	**Single shared ScheduledExecutorService**: schedule health checks for all endpoints, reducing thread footprint.
- 	**Shared worker executor pool**: power the health check logic from a common pool, avoiding too many cached pools.

src/main/java/redis/clients/jedis/mcf/HealthCheckImpl.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+    }
+
+    private void notifyListeners(HealthStatus oldStatus, HealthStatus newStatus) {
+        if (statusChangeCallback != null) {
+            statusChangeCallback.accept(new HealthStatusChangeEvent(endpoint, oldStatus, newStatus));
+        }
+    }
+
+    public long getTimeout() {
```
ggivo: A bit confusing — I would expect this to be the `timeout` for a single `HealthCheck`, but instead it represents the **maximum time to wait for an Endpoint to reach a status during initialization**.

Currently it’s exposed via `HealthStatusManager.getMaxWaitFor(Endpoint).`
How about calculating this directly there instead?
(We’d likely need to extend the interface to provide a reference to the strategy used for creating the **HealthCheck**.)

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
```diff
@@ -671,11 +683,7 @@ public CircuitBreaker getCircuitBreaker() {
         }
 
         public HealthStatus getHealthStatus() {
-            return healthStatus;
-        }
-
-        public void setHealthStatus(HealthStatus healthStatus) {
-            this.healthStatus = healthStatus;
+            return noHealthChecks ? HealthStatus.HEALTHY : healthCheck.getStatus();
```
ggivo: Do we need `noHelathChecks`? Can we perform the check based on `healthCheck == null?`

ggivo (approved):
Approving to unblock follow-up PRs. LGTM overall, left some comments to address later.

src/main/java/redis/clients/jedis/mcf/HealthCheckImpl.java
atakavci: this was my initial thinking.  
i  believe right approach is to utilize a dynamic pool,, then i got lazy and fallback to individual executors since  health checks should be isolated from one another, ensuring that any possible issue on any health check don't impact others or exhaust shared resources.

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
atakavci: well spotted, this is from a previous attempt where healthCheck is not set during constructor and set externally via some 'setHealthCheck" method.

src/main/java/redis/clients/jedis/mcf/HealthCheckImpl.java
atakavci: Conceptually it is suboptimal for status manager to contain that logic. 
We can just rename it back to something like `getMaxWaitFor` or `getMaxResultDuration`, etc..
i agree with "confusing" part.

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
ggivo: will update the review

src/main/java/redis/clients/jedis/mcf/HealthCheckImpl.java
ggivo: tracked with : [#4254](https://github.com/redis/jedis/issues/4254)

src/main/java/redis/clients/jedis/mcf/HealthCheckImpl.java
ggivo: Still think this method should not be part of HealthCheck. It uses only info from strategy it was build from
Let's rename it for now

atakavci: well,, the source of info is one thing,, while the logic/implementation that relies on that info is another.
So having all info provided from strategy is not the only argument to come to decision. 
i think this is something to discuss with less priority at another time.

PR #4249, opened 2025-08-19
[automatic failover] Improve AA failover scenario test

uglide:


ggivo:
Note: Accidentally  ActiveActiveFailoverTest was formatted in the feature branch. Reverting also the formating wiht this PR

Issue #4255, opened 2025-08-22
[automatic failover] Tracking Issue: HealthCheckStatus Incorrectly Overwritten Leading to False State Transitions


ggivo:
## Issue Summary
The `safeUpdate()` method in `HealthCheckImpl` has a bug where it notifies listeners of status changes even when the actual stored state is not updated, leading to false failover triggers and HEALTHY/UNHEALTHY state toggling.

### Bug
- Scenario: An older health check result arrives after a newer one has already been stored
   - Expected Behavior: State should NOT be updated (timestamp check prevents this )
   - Listeners should NOT be notified (no actual state change occurred)
- Actual Behavior:
   - State is correctly NOT updated
   - But listeners are incorrectly notified because oldResult.getStatus() != status compares the stored status against the rejected input status

###
Example Scenario 
```
Time: T1 - Health check starts (timestamp: 1000)
Time: T2 - Health check starts (timestamp: 2000)  
Time: T3 - T2 completes first: HEALTHY → stored (timestamp: 2000)
Time: T4 - T1 completes later: UNHEALTHY → rejected (older timestamp)
         - But listeners notified: HEALTHY → UNHEALTHY (FALSE!)
         - Actual state remains: HEALTHY
 ```

**Problematic Code**:
```java
private void safeUpdate(long owner, HealthStatus status) {
    HealthCheckResult newResult = new HealthCheckResult(owner, status);
    HealthCheckResult oldResult = resultRef.getAndUpdate(current -> {
        if (current.getTimestamp() < owner) {
            return newResult;  // State updated
        }
        return current;        // State NOT updated
    });
    if (oldResult.getStatus() != status) {  // : Always compares against input status
        // notify listeners
        notifyListeners(oldResult.getStatus(), status);  // False notification!
    }
}

PR #4256, opened 2025-08-22
[automatic failover ]fix: only notify health status listeners on actual state changes #4255

ggivo:
- Replace overlapping health checks with sequential execution
  - use wasUpdated flag to prevent false notifications from timestamp conflicts. Eliminates race conditions causing spurious failovers.

Closes: #4255

### DIFF
--- a/src/main/java/redis/clients/jedis/mcf/HealthCheckImpl.java
+++ b/src/main/java/redis/clients/jedis/mcf/HealthCheckImpl.java
@@ -8,6 +8,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 
@@ -15,6 +16,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import redis.clients.jedis.Endpoint;
+import redis.clients.jedis.annots.VisibleForTesting;
 
 public class HealthCheckImpl implements HealthCheck {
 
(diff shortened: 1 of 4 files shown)


src/main/java/redis/clients/jedis/mcf/HealthCheckImpl.java
```diff
@@ -115,19 +141,33 @@ private void healthCheck() {
...
         HealthCheckResult newResult = new HealthCheckResult(owner, status);
+        AtomicBoolean wasUpdated = new AtomicBoolean(false);
+
         HealthCheckResult oldResult = resultRef.getAndUpdate(current -> {
             if (current.getTimestamp() < owner) {
+                wasUpdated.set(true);
                 return newResult;
             }
             return current;
         });
```
atakavci: this block in `getAndUpdate` should be "side effect free" since it might be called multiple times and leave no trace of previous executions.


```suggestion
        HealthCheckResult oldResult = resultRef.getAndUpdate(current -> {
            if (current.getTimestamp() < owner) {
                wasUpdated.set(true);
                return newResult;
            }
            wasUpdated.set(false);
            return current;
        });
```

src/main/java/redis/clients/jedis/mcf/HealthCheckImpl.java
```diff
@@ -115,19 +141,33 @@ private void healthCheck() {
             safeUpdate(me, HealthStatus.UNHEALTHY);
             Thread.currentThread().interrupt(); // Restore interrupted status
             log.warn("Health check interrupted for {}", endpoint, e);
+        } finally {
+            // Schedule the next health check after this one completes (success, timeout, or failure)
+            scheduleNextHealthCheck(strategy.getInterval());
```
atakavci: are we **%100** sure about the -schedule next- approach??  each time it is possible to get an execution/timeout delay which means actual value  is "strategy interval+strategy timeout" or "strategy interval+execution duration". This doesn't sound right to me.
i know there are consequences of executing it in a fixed rate as well. 
thats why we have a single threaded scheduler and thread pooled worker(executor).

ggivo: :) not 100%, 
I prefer this one as a safer approach with fewer overlapping checks.

atakavci:
i think overlapping is either a configuration issue or something unexpected going on... isnt it suboptimal  to see a delayed result when something unexpected is going on?

Yahoo Mail: Search, organise, conquer 
 
  On Fri, 22 Aug 2025 at 14:30, Ivo ***@***.***> wrote:   



@ggivo commented on this pull request.

In src/main/java/redis/clients/jedis/mcf/HealthCheckImpl.java:
> @@ -115,19 +141,33 @@ private void healthCheck() {
             safeUpdate(me, HealthStatus.UNHEALTHY);
             Thread.currentThread().interrupt(); // Restore interrupted status
             log.warn("Health check interrupted for {}", endpoint, e);
+        } finally {
+            // Schedule the next health check after this one completes (success, timeout, or failure)
+            scheduleNextHealthCheck(strategy.getInterval());

:) not 100%,
I prefer this one as a safer approach with fewer overlapping checks.

src/main/java/redis/clients/jedis/mcf/HealthCheckImpl.java
atakavci: i think overlapping is either a configuration issue or something unexpected going on... isnt it suboptimal  to see a delayed result when something unexpected is going on?

ggivo: Agree about the configuration issue, let me revert to the previous approach

Issue #4233, opened 2025-08-13
[automatic failover] Follow up items regarding to feedbacks and potential issues


atakavci:
The list of items to be completed on [feature branch](https://github.com/redis/jedis/tree/feature/automatic-failover) before merging to master.

- [X] Tests on EchoStrategy are to be included in the feature branch, suggestion [here](https://github.com/redis/jedis/pull/4230). Fixed [here](https://github.com/redis/jedis/pull/4230)
- [x] Connection.forceDisconnect should also set the connection as broken ( closeQuietly ??) , [reference to line](https://github.com/redis/jedis/pull/4226/files#diff-a5cc442f670c55cac4ad9b268105292c6ecda839a3c10e3d0cddacf9fa359437R328). Fixed [here](https://github.com/redis/jedis/pull/4258)

atakavci:
- #4236

PR #4258, opened 2025-08-22
Connection.forceDisconnect should also set the connection as broken #4233

ggivo:
Ensure connection is disposed when returned to the pool

### DIFF
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -326,7 +326,8 @@ public class Connection implements Closeable {
   }
 
   public void forceDisconnect() throws IOException {
-    socket.close();
+    setBroken();
+    IOUtils.closeQuietly(socket);
   }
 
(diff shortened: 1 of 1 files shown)


ggivo:
As of now, `forcedDisconnect` depends/waits till all borrowed connections are closed, which means the close of the pool might be blocked.
I think we should consider if :
-  forcedDisconnect closes connections borrowed by the pool, and not only disconnects the socket. But ensure they are returned & disposed of in the pool

```
  public void forceDisconnect() throws IOException {
    setBroken();
    IOUtils.closeQuietly(socket);
    if (this.memberOf != null) {
      this.memberOf = null;
      pool.returnBrokenResource(this);
    }
  }
```
or [ObjectPool.html#invalidateObject](https://commons.apache.org/proper/commons-pool/apidocs/org/apache/commons/pool2/ObjectPool.html#invalidateObject(T))
- wait for borrowed connections with timeout (forceDisconnect(timeout)) 
- both

PR #4263, opened 2025-08-28
Add builders for classes based on UnifiedJedis

uglide:
## Motivation
This change creates a foundation for providing a new client class(es) for AA failover, allowing developers to easily instantiate JedisPooled, JedisSentineled, JedisCluster, etc.

# Review

## Summary
This commit introduces a formal Builder pattern for UnifiedJedis-based clients (JedisPooled, JedisCluster, JedisSentineled) and adds a shared builder base (AbstractClientBuilder) plus specialized builders for standalone, cluster, and sentinel deployments. It also exposes a couple of configuration hooks, aligns RediSearch default dialect handling, and adds reflection-based tests to ensure builder coverage over the existing constructor surfaces. Build tooling is adjusted to preserve parameter names for those tests.

---

## High-level Impact
- Additive public API: new static builder() entry points and builder classes under redis.clients.jedis.builders.
- Internal visibility tweak: UnifiedJedis constructor made package-visible to allow subclass super(...) calls.
- Slightly broader access: CommandObjects.setKeyArgumentPreProcessor() is now public.
- Tests and tooling: reflection-based tests to validate builder coverage; Maven profile to compile with parameter names.
- Behavioral defaults remain intact (e.g., RediSearch DEFAULT_DIALECT still 2, now referenced via constant).

---

### 1) Builders
- New classes:
  - redis.clients.jedis.builders.AbstractClientBuilder<T,C>
  - redis.clients.jedis.builders.StandaloneClientBuilder<C>
  - redis.clients.jedis.builders.ClusterClientBuilder<C extends JedisCluster>
  - redis.clients.jedis.builders.SentinelClientBuilder<C>

Key capabilities (AbstractClientBuilder):
- Common options: poolConfig, cache/cacheConfig, connectionProvider, commandExecutor, commandObjects, redisProtocol, key preprocessor, JSON mapper, search dialect.
- Build flow: validate -> optional cache from cacheConfig -> default provider/executor if unset -> apply common config to CommandObjects -> instantiate concrete client.
- Cache implies RESP3: calling cache(...) or cacheConfig(...) forces redisProtocol=RESP3 (UnifiedJedis guards against cache+RESP2).
- applyCommonConfiguration wires:
  - CommandObjects.setKeyArgumentPreProcessor(...)
  - CommandObjects.setJsonObjectMapper(...)
  - CommandObjects.setDefaultSearchDialect(...) when non-default

Specialized builders:
- StandaloneClientBuilder
  - hostAndPort(String,int|HostAndPort), clientConfig(JedisClientConfig)
  - fromURI(String|URI) builds DefaultJedisClientConfig based on URI and sets host/port
  - Default provider: PooledConnectionProvider(hostAndPort, config, cache, poolConfig)
  - Validation intends to ensure either URI or host/port, not both
- ClusterClientBuilder
  - nodes(Set<HostAndPort>), clientConfig, maxAttempts(int), maxTotalRetriesDuration(Duration), topologyRefreshPeriod(Duration)
  - Default provider: ClusterConnectionProvider(nodes, config, cache, poolConfig, topologyRefreshPeriod)
  - Validates nodes non-empty and non-negative durations/attempts
- SentinelClientBuilder
  - masterName(String), sentinels(Set<HostAndPort>), clientConfig, sentinelClientConfig
  - Default provider: SentineledConnectionProvider(masterName, clientConfig, cache, poolConfig, sentinels, sentinelClientConfig)
  - Validates required fields

### 2) Jedis* classes integrate builders
- JedisPooled, JedisCluster, JedisSentineled:
  - Add private constructors accepting (CommandExecutor, ConnectionProvider, CommandObjects, RedisProtocol, Cache) and delegating to super(...)
  - Add public static inner Builder extends respective specialized builder
  - Add public static Builder builder() factory method

### 3) UnifiedJedis visibility
- UnifiedJedis(CommandExecutor, ConnectionProvider, CommandObjects, RedisProtocol, Cache) constructor changed from private to package-visible to enable subclass super(...) calls from Jedis*.

### 4) CommandObjects tuning
- Replace hard-coded default search dialect (2) with SearchProtocol.DEFAULT_DIALECT.
- setKeyArgumentPreProcessor made public (was @Experimental with package access). This enables configuration via builders.

### 5) RediSearch protocol constant usage
- SearchDefaultDialectTest and CommandObjects now reference SearchProtocol.DEFAULT_DIALECT rather than literals. SearchProtocol.DEFAULT_DIALECT remains 2.

### 6) Build tooling to support reflection tests
- Makefile: mvn-test-local and mvn-test now pass -Dwith-param-names=true.
- pom.xml: new profile tests-with-params activated via -Dwith-param-names=true, enabling:
  - maven-compiler-plugin <parameters>true</parameters> (preserve method/parameter names)
  - maven-surefire-plugin argLine includes --add-opens java.base/java.lang=ALL-UNNAMED ${JVM_OPTS}

### 7) Tests
- New reflection-based coverage tests:
  - builders/JedisPooledConstructorReflectionTest
  - builders/JedisClusterConstructorReflectionTest
  - builders/JedisSentineledConstructorReflectionTest
  - builders/UnifiedJedisConstructorReflectionTest
- Purpose: verify that for every constructor of the target classes, the builder surface offers an equivalent configuration path (maps each parameter type/name to a builder option). Tests assert full coverage and log any gaps.

---

## API/Compatibility Assessment
- Additive: New builder APIs are additive; existing constructors remain.
- Internal constructor visibility change is not exposed as public API but is a package-level relaxation; safe and contained.
- Making setKeyArgumentPreProcessor public increases surface area; consider de-@Experimental or documenting stability expectations.
- No default behavioral changes beyond wiring options when using the builder.


src/main/java/redis/clients/jedis/builders/AbstractClientBuilder.java
```diff
@@ -0,0 +1,312 @@
...
+   * Sets the client-side cache for caching Redis responses.
+   * <p>
+   * Client-side caching can improve performance by storing frequently accessed data locally,
+   * reducing the number of round trips to the Redis server.
+   * @param cache the cache instance
+   * @return this builder
+   */
+  public T cache(Cache cache) {
+    this.cache = cache;
+    this.redisProtocol = RedisProtocol.RESP3;
```
ggivo: Not sure about silently overriding the protocol. 
I will suggest validating it during build if cache is configured and configured protocol does not support it..

src/main/java/redis/clients/jedis/builders/AbstractClientBuilder.java
```diff
@@ -0,0 +1,312 @@
...
+   * Sets the cache configuration for client-side caching.
+   * <p>
+   * Client-side caching can improve performance by storing frequently accessed data locally. The
+   * cache will be created from this configuration during the build process.
+   * @param cacheConfig the cache configuration
+   * @return this builder
+   */
+  public T cacheConfig(CacheConfig cacheConfig) {
+    this.cacheConfig = cacheConfig;
+    this.redisProtocol = RedisProtocol.RESP3;
```
ggivo: Same as above, don't think we should silently set the protocol;

src/main/java/redis/clients/jedis/builders/AbstractClientBuilder.java
```diff
@@ -0,0 +1,312 @@
...
+
+  /**
+   * Validates common configuration parameters.
+   * <p>
+   * This method can be called by concrete builders to validate the common configuration before
+   * building the client.
+   * @throws IllegalArgumentException if any common configuration is invalid
+   */
+  protected void validateCommonConfiguration() {
+    if (poolConfig == null) {
```
ggivo: What about using custom connection provider that does not require poolConfig?

I think pool config should be optional

src/main/java/redis/clients/jedis/builders/AbstractClientBuilder.java
```diff
@@ -0,0 +1,312 @@
...
+   * Each builder implementation should create their specific client type (JedisPooled,
+   * JedisCluster, etc.) using the provided parameters.
+   * @param commandExecutor the command executor
+   * @param connectionProvider the connection provider
+   * @param commandObjects the command objects
+   * @param redisProtocol the Redis protocol
+   * @param cache the client-side cache (may be null)
+   * @return the configured Redis client instance
+   */
+  protected abstract C createClient(CommandExecutor commandExecutor,
```
ggivo: Adding new components (metrics, tracing, etc.) would require breaking changes. No need to pass explicitly `commandExecutor`,... They are already available to extending classes

src/test/java/redis/clients/jedis/builders/ClientBuilderTest.java
```diff
@@ -0,0 +1,107 @@
...
+        .searchDialect(3).build();
+
+    client.ftSearch("idx", "q", new FTSearchParams());
+    verify(exec, atLeastOnce()).executeCommand(cap.capture());
+    List<String> args = argsToStrings(cap.getValue());
+    assertThat(args, contains("FT.SEARCH", "idx", "q", "DIALECT", "3"));
+  }
+
+  @Test
+  void cacheEnablesRESP3() {
```
ggivo: related to previous comment. Don't think we should silently set the protocol when enabling client side cache

src/test/java/redis/clients/jedis/builders/ClientBuilderTest.java
```diff
@@ -0,0 +1,107 @@
...
+    JedisPooled client = JedisPooled.builder().commandExecutor(exec).connectionProvider(provider)
+        .cache(cache).build();
+
+    client.ping();
+    verify(exec).broadcastCommand(cap.capture());
+    assertThat(argsToStrings(cap.getValue()).get(0), containsString("PING"));
+  }
+
+  @Test
+  void standaloneValidateHostPortRequired() {
```
ggivo, line 100: Relates to previous comment. We should not enforce poolConfig as required. For example when using custom ConnectionProvide we might not use appache commons pool

src/test/java/redis/clients/jedis/builders/ClientBuilderTest.java
```diff
@@ -0,0 +1,107 @@
...
+    assertThrows(IllegalArgumentException.class, () -> new JedisPooled.Builder() {
+    }.poolConfig(null).build(), "Pool configuration cannot be null");
+  }
+
+  @Test
+  void sentinelValidateMasterAndSentinels() {
+    assertThrows(IllegalArgumentException.class, () -> new JedisSentineled.Builder() {
```
ggivo: What is the reason for Illegal argument exception? Should we have sensible defaults and build() pass with them?

src/test/java/redis/clients/jedis/builders/ClusterClientBuilderTest.java
```diff
@@ -0,0 +1,145 @@
...
+  @Test
+  void keyPreprocessorAppliedInCluster() {
+    JedisCluster client = JedisCluster.builder().nodes(someNodes()).connectionProvider(provider)
+        .commandExecutor(exec).keyPreProcessor(k -> "prefix:" + k).build();
+
+    client.set("k", "v");
+    verify(exec).executeCommand(cap.capture());
+    List<String> args = argsToStrings(cap.getValue());
+    // SET prefix:k v
+    org.hamcrest.MatcherAssert.assertThat(args.get(0), containsString("SET"));
```
ggivo: static imports

src/test/java/redis/clients/jedis/builders/JedisClusterConstructorReflectionTest.java
```diff
@@ -0,0 +1,267 @@
...
+import redis.clients.jedis.providers.ClusterConnectionProvider;
+import redis.clients.jedis.providers.ConnectionProvider;
+
+/**
+ * Reflection-based coverage test for JedisCluster constructors against builder API. The test
+ * verifies each constructor parameter can be provided via: - ClusterClientBuilder methods, or -
+ * DefaultJedisClientConfig.Builder methods, or - Custom ConnectionProvider (manual coverage)
+ */
+@EnabledIfSystemProperty(named = "with-param-names", matches = "true")
+public class JedisClusterConstructorReflectionTest {
```
ggivo, line 33: I have some concerns about the current reflection-based testing approach in `JedisPooledConstructorReflectionTest`.  
 The tests rely on validating hardcoded internal state, which can become outdated quickly and may fail to catch real functional differences.  
  
Some issues with this approach:  
 - It can lead to false positives.  
 - It only validates that parameters can be provided, not that they work correctly.  
  
 Instead of reflection-based coverage validation, have you considered testing **functional equivalence** between constructor-created and builder-created clients?  
 This way, the tests also serve a documentation purpose while validating actual functionality parity.  
  
 For example:  
  
 ```java  
 @Test  
 void constructorBuilderEquivalence() {  
     // Ensure constructor and builder produce functionally equivalent clients  
     
     // Constructor approach  
     JedisPooled constructorClient = new JedisPooled("localhost", 6379, "user", "pass");  
     
     // Builder approach    
     JedisPooled builderClient = JedisPooled.builder()  
         .hostAndPort("localhost", 6379)  
         .clientConfig(DefaultJedisClientConfig.builder()  
             .user("user")  
             .password("pass")  
             .build())  
         .build();  
     
     // Validate functional equivalence  
     assertClientsEquivalent(constructorClient, builderClient);  
 }  
 ```  
  
 Where `assertClientsEquivalent(...)` could validate the equivalence of initialized client components such as:  
 - `ConnectionProvider provider`  
 - `Pool<Connection> pool`  
 - `CommandExecutor executor`  
 - `CommandObjects commandObjects`  
 - `Cache cache`

ggivo (commented):
Approach with the builders looks good. 
I have put some comments, and have concerns around tests validating the constructor coverage.

src/test/java/redis/clients/jedis/builders/ClientBuilderTest.java
uglide: Primary name and sentinels config are required and cannot have a default value. See https://github.com/redis/jedis/pull/4263/files/183b00d9046022a45ebe162c1342663c4f53d1b2#diff-af5d7cdfaeacc02ac468d5007e9860e001b9f90f80b80288ce3dccd22ae390b7R93-R104 for more details

src/test/java/redis/clients/jedis/builders/JedisClusterConstructorReflectionTest.java
uglide: The approach I used serves as a safeguard to prevent the addition of new constructors and provides an efficient way to check the builder's coverage. Once we reach Jedis v9, we should drop it altogether with all deprecated constructors.

Another problem with functional tests is that they only work for simple cases like the one you have mentioned. Also, creating reasonable `assertClientsEquivalent` is quite tricky. I gave up on that approach after seeing that it became a reflection monster to expose all the internal states of UnifiedJedis and co. At best, with a functional approach, we end up testing that both client instances can connect to the same redis-server instance, which is not sufficient and can be achieved by just instantiating UnifiedJedis and co with the new builder in plenty of integration tests that we already have.

src/main/java/redis/clients/jedis/builders/AbstractClientBuilder.java
uglide: Good catch, thanks!

src/test/java/redis/clients/jedis/builders/ClientBuilderTest.java
uglide: This test is just off; it should test the host and port instead. I'm going to fix it.

src/main/java/redis/clients/jedis/builders/AbstractClientBuilder.java
uglide: Fixed

src/main/java/redis/clients/jedis/builders/AbstractClientBuilder.java
uglide: Fixed

src/main/java/redis/clients/jedis/builders/AbstractClientBuilder.java
uglide: Fixed

src/main/java/redis/clients/jedis/builders/AbstractClientBuilder.java
uglide: Fixed

src/test/java/redis/clients/jedis/builders/ClientBuilderTest.java
uglide: Fixed

src/test/java/redis/clients/jedis/builders/ClientBuilderTest.java
uglide: Done

src/test/java/redis/clients/jedis/builders/ClusterClientBuilderTest.java
uglide: Fixed

src/test/java/redis/clients/jedis/builders/JedisClusterConstructorReflectionTest.java
ggivo: I see your point. 
 
Yes, I was thinking that we need a mix of functional tests and unit tests. We can still use the reflection, but based on comparing the actual objects built.  e.g, ConnectionProvider built via the constructor is equivalent to ConnectionProvider built with the corresponding builder.

IMO, the current approach is also hard to verify via review. 

It might take some time till v9 is out. Existing tests will give some confidence, but I  still think it makes sense to open a follow-up issue and add additional tests focused on object build using a concrete constructor vs the corresponding one using builder.

src/main/java/redis/clients/jedis/builders/AbstractClientBuilder.java
```diff
@@ -0,0 +1,304 @@
...
+  }
+
+  /**
+   * Applies common configuration to the CommandObjects instance.
+   * <p>
+   * This method is called by concrete builders to configure the CommandObjects with the common
+   * settings like key preprocessor, JSON mapper, and search dialect.
+   * @param commandObjects the CommandObjects instance to configure
+   */
+  public void applyCommandObjectsConfiguration(CommandObjects commandObjects) {
```
atakavci, line 305: missing important part is that this instance of commandObjects is not managed in other abstract or concrete builders. 
In addition to that commandobjects is a kind of factory itself and should be left to be instantiated/configured by typed builders(like the other fields in builder) instead of being set as default by abstract one.

uglide: Good point! Fixed.

src/test/java/redis/clients/jedis/builders/JedisClusterConstructorReflectionTest.java
atakavci: what we can improve here is to annotate each ctor we want to check and identify where these args value/instance should land on with the client object vs builders. Here is a sample how to annotate;
```java
  @ParamSpec({"provider.cache.nodes", "provider.cache.clientConfig", "executormaxAttempts"})
  public UnifiedJedis(Set<HostAndPort> jedisClusterNodes, JedisClientConfig clientConfig, int maxAttempts) {
    this(jedisClusterNodes, clientConfig, maxAttempts,
        Duration.ofMillis(maxAttempts * clientConfig.getSocketTimeoutMillis()));
  }
```

now we know where the value should land on , just provide a value for each parameter in a dynamic fashion and run ctor and set it on builder as well with reflection. Then compare it same approach @ggivo  suggested but just for the fields in `ParamSpec` annotation;

```java
    // Validate functional equivalence  
    assertClientsEquivalent(constructorClient, builderClient);  
```

we can get rid of all when we get rid of ctors.

ggivo: @atakavci
There are settings that can not be mapped directly one-to-one. Some of them might lead to creating a specific kind of object, or being propagated to multiple places. 
I still think it will be simpler to generate test for constructors and corresponding builders directly. The tricky part being asserting that built objects are equivalent.

ggivo: @uglide @atakavci 
Discussed & decided to open a follow-up issue to create tests confirming parity between the constructor & builders

ggivo: Follow up issue to add additional test
https://github.com/redis/jedis/issues/4301

ggivo (approved):

PR #4273, opened 2025-09-04
[automatic failover] Implement health check retries

### LINKED ISSUE #4272: [automatic failover]  Implement retry logic with health checks, introduce `delayInBetweenRetries`

atakavci:
Closes #4272 
- replace minConsecutiveSuccessCount with numProbes
- add probing mechanism and relevant classes interfaces for healtCheckImpl
- apply changes to strategy implementations config classes
- add delayer executor pool to healthcheckımpl
- adjustments on  worker pool of healthCheckImpl for shared use of workers

TODO
- [X] needs some further polishing
- [X] fix failing tests

### DIFF
--- a/src/test/java/redis/clients/jedis/mcf/HealthCheckTest.java
+++ b/src/test/java/redis/clients/jedis/mcf/HealthCheckTest.java
@@ -761,7 +761,7 @@ public class HealthCheckTest {
     CountDownLatch healthyLatch = new CountDownLatch(1);
 
     TestHealthCheckStrategy strategy = new TestHealthCheckStrategy(
-        HealthCheckStrategy.Config.builder().interval(5).timeout(200).numProbes(5)
+        HealthCheckStrategy.Config.builder().interval(5000).timeout(200).numProbes(5)
             .policy(BuiltIn.MAJORITY_SUCCESS).delayInBetweenProbes(5).build(),
         e -> {
           int c = callCount.incrementAndGet();


src/main/java/redis/clients/jedis/mcf/ProbePolicy.java
```diff
@@ -0,0 +1,100 @@
...
+        return Decision.CONTINUE;
+      }
+    }
+
+    /*
+     * A database is healthy if the majority of probes returned ‘healthy’. This means you can stop
+     * probing as soon as the majority can’t be guaranteed any more (e.g., you have 4 probes and 2
+     * of them failed), or as soon as the majority is reached (e.g., 3 out of 4 were healthy)
+     */
+    private static class MajoritySuccessPolicy implements ProbePolicy {
```
ggivo: What will be the use case for using MajoritySuccessPolicy policy?

src/main/java/redis/clients/jedis/mcf/HealthCheckImpl.java
```diff
@@ -77,61 +145,107 @@ public HealthStatus getStatus() {
...
+  private void delay(int delayInBetweenProbes) {
+    try {
+      delayer.schedule(() -> {
+        /* No Op */
+      }, (long) delayInBetweenProbes, TimeUnit.MILLISECONDS).get();
+    } catch (InterruptedException | ExecutionException e) {
+      Thread.currentThread().interrupt();
+      return;
+    }
+  }
```
ggivo: Why do we need a delayer executor? How is this different than block using Thread.sleep?

src/main/java/redis/clients/jedis/mcf/HealthCheckImpl.java
```diff
@@ -77,61 +145,107 @@ public HealthStatus getStatus() {
...
 
-  // just to avoid to replace status with an outdated result from another healthCheck
+  /**
+   * just to avoid to replace status with an outdated result from another healthCheck
+   *
+   * <pre>
+   * Health Check Race Condition Prevention
+   *
+   * Problem: Async health checks can complete out of order, causing stale results
+   * to overwrite newer ones.
```
ggivo, line 213: nit:  can extend the description with examnple how can end up with parallel health-checks

For example, a parallel health check can happen in case a scheduled health check is canceled because of  a timeout, but the actual health check operation is still running in the background.

ggivo (commented):
LGTM in general.

Wondering if it can be simplified a bit by removing the dedicated executor used to apply  delays between health check probes

src/main/java/redis/clients/jedis/mcf/ProbePolicy.java
atakavci: the two other built in implementations provides the edge approaches of all and none. 
This one suggest a middle way to say, like you dont want to wait for all attempts to fail if the network is gone but you would like to give it a couple of chances before a failover on glitches. 

- "all" is a bit too quick i think, because it causes a failover on any network glitch , including a timeout or network device patch
- "any" is not durable against swinging health check results and does not cover enough duration between check interval windows, if we want to be proactive with health checks we should cover some to avoid false positives due to swings

but these are all reasonable choices depending on the case and how users interpret availability/stability ratio.

src/main/java/redis/clients/jedis/mcf/HealthCheckImpl.java
atakavci: i have no strong opinion here; there might be no additional benefit other than **properly responding** graceful terminations attempts with thread interruptions.

atakavci: > Wondering if it can be simplified a bit by removing the dedicated executor used to apply delays between health check probes

TBH, initial implementation was with Thread.Sleep. And can go back if you have strong opinion with trading of "simplicity vs interruption-cooperative". 
i think i start making up words :)

ggivo: :) I would vote for simplicity. It becomes a bit complex already, and till we need "interruption-cooperative" I would go with simpler implmentation

atakavci: i will change it, but i wonder what you mean with "needing it" ?
its not a functional requirement, it s a best practice thing.

src/main/java/redis/clients/jedis/mcf/ProbePolicy.java
ggivo: Not directly related to this PR, but my personal concerns are that during failover and failback, we might have different requirements, and the current approach lacks flexibility in that direction. Also, I think health-check should use something similar to a moving window. e.g, after x consecutive successful health checks, we consider Endpoint as healthy, and in the other direction, after y consecutive failures, Endpoint is considered **UNHEALTHY**

I see other redis clients are following the approach implemented with this review

src/main/java/redis/clients/jedis/mcf/HealthCheckImpl.java
atakavci: done

src/test/java/redis/clients/jedis/failover/FailoverIntegrationTest.java
```diff
@@ -137,6 +138,9 @@ public void cleanup() throws IOException {
   public void testAutomaticFailoverWhenServerBecomesUnavailable() throws Exception {
     assertThat(getNodeId(failoverClient.info("server")), equalTo(JEDIS1_ID));
 
+    await().atMost(1, TimeUnit.SECONDS).pollInterval(50, TimeUnit.MILLISECONDS)
+        .until(() -> provider.getCluster(endpoint2.getHostAndPort()).isHealthy());
+
```
ggivo, line 143: nit: seems like repeats in multiple test. Can this be part of the test setup? E.g await till we are sure all cluster/databses are healthy

atakavci: i ll see what i can come up with, simple and reusable..  but most probably will do with another pr.

PR #4277, opened 2025-09-08
[automatic failover] Minor optimizations on fast failover

atakavci:
The change with looping condition from

```java
while (numWaiters.get() > 0 || getNumActive() > 0 || getNumIdle() > 0) {
```
to 
```java
while (numWaiters.get() > 0 || numOfConnected > 0) {
```
this lets '`failover`' proceeed without waiting all the connections back to pool. 
`numWaiters` makes sure there is no blocked user thread waiting on `pool.getResource` whereas checking `numOfConnected` ensures all the connections out there is properly disconnected/dropped. 
The rest around handling a disconnected `Connection` is in user thread responsibility as before this change.


src/main/java/redis/clients/jedis/mcf/TrackingConnectionPool.java
```diff
@@ -53,8 +53,12 @@ public PooledObject<Connection> makeObject() throws Exception {
         } finally {
           factoryTrackedObjects.remove(object.getObject());
         }
+        // this can make a marginal improvement on fast failover duration!
+        if (failFast) {
```
uglide, line 57: Probably makes sense to make `failFast` a `volatile` variable.

uglide (commented):
See my comment about `failFast` variable

Issue #4279, opened 2025-09-11
CacheEntry Failed to serialize object


pinyht:
## Description
When config and use client-side caching, then execute a json command like jsonGet, it throw a JedisCacheException, after debug, I found this problem is jsonGet return object type is JSONObject, JSONObject has not implements Serializable, it can't convert to bytes, exception throw in this file

```java
// src/main/java/redis/clients/jedis/csc/CacheEntry.java

 private static byte[] toBytes(Object object) {
    try (ByteArrayOutputStream baos = new ByteArrayOutputStream();
        ObjectOutputStream oos = new ObjectOutputStream(baos)) {
      oos.writeObject(object);
      oos.flush();
      oos.close();
      return baos.toByteArray();
    } catch (IOException e) {
      throw new JedisCacheException("Failed to serialize object", e);
    }
  }
``` 
when execute oos.writeObject(object) will throw IOException

## Error log

```
redis.clients.jedis.exceptions.JedisCacheException: Failed to serialize object
	at redis.clients.jedis.csc.CacheEntry.toBytes(CacheEntry.java:44)
	at redis.clients.jedis.csc.CacheEntry.<init>(CacheEntry.java:21)
	at redis.clients.jedis.csc.CacheConnection.executeCommand(CacheConnection.java:93)
	at redis.clients.jedis.executors.DefaultCommandExecutor.executeCommand(DefaultCommandExecutor.java:24)
	at redis.clients.jedis.UnifiedJedis.executeCommand(UnifiedJedis.java:314)
	at redis.clients.jedis.UnifiedJedis.jsonGet(UnifiedJedis.java:4325)
	at com.pinyht.youhuizhi.infrastructure.cache.utils.CacheUtils.jsonGet(CacheUtils.java:187)
	at com.pinyht.youhuizhi.application.admin.controller.TestController.testHttp(TestController.java:522)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:360)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.invokeJoinpoint(ReflectiveMethodInvocation.java:196)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:163)
[...]

PR #4286, opened 2025-09-19
Fix client-side caching for non-Serializable objects

kavindu-udara:
Fix client-side caching for non-Serializable objects (e.g., JSONObject)

## Problem:
When using client-side caching (CacheEntry) with JSONObject returned from JSON commands (jsonGet), the cache threw a JedisCacheException. This happened because JSONObject does not implement Serializable, so the original CacheEntry tried to serialize it to bytes using ObjectOutputStream, which failed.

Issue #4279

## Solution:

Added logic in CacheEntry to fallback to storing a String when the object is not Serializable.

## Updated unit tests to cover:

- JSONObject (non-Serializable) → stored as String.
- Serializable objects (e.g., String) → stored as-is.
- `getCacheKey()` and `getConnection()` behaviors.

## Benefits:

Fixes caching for JSON commands without breaking existing Serializable objects.
Ensures client-side cache works smoothly for all object types.

### Unit tests added in CacheEntryTest.java:

- `testJsonObjectCaching() ` → verifies fallback for JSONObject.
- `testSerializableObjectCaching()` → verifies normal Serializable objects.

### DIFF
--- a/src/main/java/redis/clients/jedis/csc/CacheEntry.java
+++ b/src/main/java/redis/clients/jedis/csc/CacheEntry.java
@@ -36,6 +36,12 @@ public class CacheEntry<T> {
   private static byte[] toBytes(Object object) {
     try (ByteArrayOutputStream baos = new ByteArrayOutputStream();
         ObjectOutputStream oos = new ObjectOutputStream(baos)) {
+
+      if (!(object instanceof java.io.Serializable)) {
+        // Fallback: store as string
+        object = object.toString();
+      }
+      
       oos.writeObject(object);
       oos.flush();
       oos.close();
--- a/src/test/java/redis/clients/jedis/csc/CacheEntryTest.java
+++ b/src/test/java/redis/clients/jedis/csc/CacheEntryTest.java
@@ -1,0 +1,50 @@
+package redis.clients.jedis.csc;
+
+import org.json.JSONObject;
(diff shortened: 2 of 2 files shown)


src/main/java/redis/clients/jedis/csc/CacheEntry.java
```diff
@@ -36,6 +36,12 @@ public CacheConnection getConnection() {
   private static byte[] toBytes(Object object) {
     try (ByteArrayOutputStream baos = new ByteArrayOutputStream();
         ObjectOutputStream oos = new ObjectOutputStream(baos)) {
+
+      if (!(object instanceof java.io.Serializable)) {
+        // Fallback: store as string
+        object = object.toString();
+      }
```
ggivo, line 43: The suggested approach will not work because it changes the return type inconsistently:

**API Contract Violation:** Commands like `jsonGet()` expect to return `JSONObject`, but with this change they would return `String` when the result comes from cache.

- Direct server responses return `JSONObject`
- Cached responses return `String`
- This breaks the fundamental expectation that caching should be transparent to the caller

This actually can be seen by the tests submitted which expect's the type to be String

kavindu-udara: Oh, I understand. I'll send you another PR

PR #4293, opened 2025-09-23
[automatic failover] Implement max number of failover attempts

atakavci:
With this PR, provider will try failing over and if there is no healthy clusters, then it throws a `JedisTemporarilyNotAvailableException` with provided delay (`delayInBetweenFailoverAttempts`) until it reaches to max number of attempts (`maxNumFailoverAttempts`). 
When it reaches to max value, exception type changes and it starts to throw `JedisPermanentlyNotAvailableException` on any failover attempt from that point.
This is useful when user can choose to fallback to another resource or proceed with a brand new client/provider instance.

### DIFF
--- a/src/test/java/redis/clients/jedis/mcf/MultiClusterFailoverAttemptsConfigTest.java
+++ b/src/test/java/redis/clients/jedis/mcf/MultiClusterFailoverAttemptsConfigTest.java
@@ -112,8 +112,8 @@ public class MultiClusterFailoverAttemptsConfigTest {
 
     await().atMost(Durations.TWO_HUNDRED_MILLISECONDS).pollInterval(Duration.ofMillis(10))
         .until(() -> {
-          Exception e = assertThrows(JedisFailoverException.class,
-            () -> provider.switchToHealthyCluster(SwitchReason.HEALTH_CHECK, provider.getCluster()));
+          Exception e = assertThrows(JedisFailoverException.class, () -> provider
+              .switchToHealthyCluster(SwitchReason.HEALTH_CHECK, provider.getCluster()));
           return e instanceof JedisPermanentlyNotAvailableException;
         });
   }
@@ -130,17 +130,17 @@ public class MultiClusterFailoverAttemptsConfigTest {
     // Expect exactly 'maxAttempts' temporary exceptions, then a permanent one
     assertThrows(JedisTemporarilyNotAvailableException.class,
       () -> provider.switchToHealthyCluster(SwitchReason.HEALTH_CHECK, provider.getCluster())); // attempt
-                                                                                              // 1
+    // 1
     assertThrows(JedisTemporarilyNotAvailableException.class,
       () -> provider.switchToHealthyCluster(SwitchReason.HEALTH_CHECK, provider.getCluster())); // attempt
-                                                                                              // 2
+    // 2
 
     // Next should exceed max and become permanent
     assertThrows(JedisPermanentlyNotAvailableException.class,
       () -> provider.switchToHealthyCluster(SwitchReason.HEALTH_CHECK, provider.getCluster())); // attempt
-                                                                                              // 3
-                                                                                              // ->
-                                                                                              // permanent
+    // 3
+    // ->
+    // permanent
   }
 
   // ======== Test helper methods (reflection) ========


src/main/java/redis/clients/jedis/mcf/MultiClusterPooledConnectionProvider.java
```diff
@@ -281,6 +290,9 @@ public void remove(Endpoint endpoint) {
     } finally {
       activeClusterChangeLock.unlock();
     }
+    if (notificationData != null) {
```
ggivo, line 284: What is the reason for delaying the notification, and not performing it inside setActiveCluster for example?

src/main/java/redis/clients/jedis/mcf/MultiClusterPooledConnectionProvider.java
```diff
@@ -350,7 +362,7 @@ void onHealthStatusChange(HealthStatusChangeEvent eventArgs) {
     if (initializationComplete) {
       if (!newStatus.isHealthy() && clusterWithHealthChange == activeCluster) {
         clusterWithHealthChange.setGracePeriod();
-        iterateActiveCluster(SwitchReason.HEALTH_CHECK);
+        iterateActiveCluster(SwitchReason.HEALTH_CHECK, clusterWithHealthChange);
```
ggivo: nit: `iterateActivceCluster` is a bit confusing. Probably, we need a better name. 
It is actually  performing a `failover`  ( to the most weighted available one)

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
```diff
@@ -450,30 +462,59 @@ void periodicFailbackCheck() {
     }
   }
 
-  public Endpoint iterateActiveCluster(SwitchReason reason) {
-    Map.Entry<Endpoint, Cluster> clusterToIterate = findWeightedHealthyClusterToIterate();
+  public Endpoint iterateActiveCluster(SwitchReason reason, Cluster iterateFrom) {
```
ggivo: Not part of this review I guess. Since it is public it is easy to be invoked with custom parameters? Do we want to expose it this way. E.g someone can invoke it and set the reason to anything (like CIRCUIT_BREAKER).
Should it be hidden?

src/main/java/redis/clients/jedis/mcf/JedisFailoverException.java
```diff
@@ -0,0 +1,38 @@
+package redis.clients.jedis.mcf;
+
+import redis.clients.jedis.exceptions.JedisConnectionException;
+
+public class JedisFailoverException extends JedisConnectionException {
```
ggivo, line 13: Add java docs to clarify when different  exception is expected to be thrown

src/main/java/redis/clients/jedis/mcf/CircuitBreakerFailoverBase.java
```diff
@@ -63,19 +63,16 @@ protected void clusterFailover(Cluster cluster) {
...
       // different than
       // active CB. If its the same one and there are no more clusters to failover to, then throw an
       // exception
-      else if (cluster == provider.getCluster() && !provider.canIterateOnceMore()) {
-        throw new JedisConnectionException(
-            "Cluster/database endpoint could not failover since the MultiClusterClientConfig was not "
-                + "provided with an additional cluster/database endpoint according to its prioritized sequence. "
-                + "If applicable, consider failing back OR restarting with an available cluster/database endpoint");
+      else if (cluster == provider.getCluster()) {
+        provider.iterateActiveCluster(SwitchReason.CIRCUIT_BREAKER, cluster);
```
ggivo: This puzzles me?

`iterateActiveCluster()` should have been performed by the thread switching the CIRCUIT_BREAKER state to FORCED_OPEN.

In case other failover attempt from another thread is perfomed at the same time, but the cluster is still the same shouldn't we perform only `assertOperability` and propagate the exception to the calling thread?

Why do we invoke `iterateActiveCluster()` for failover attempts from threads that did not perform actual CB switch?

src/test/java/redis/clients/jedis/mcf/MultiClusterPooledConnectionProviderTest.java
```diff
@@ -252,4 +252,34 @@ void testHealthChecksStopAfterProviderClose() throws InterruptedException {
       testProvider.close();
     }
   }
+
+  @Test
+  public void userCommand_firstTemporary_thenPermanent_inOrder() {
```
ggivo, line 242: We can probably remove `testIterateActiveClusterOutOfRange`
This test covers it.


userCommand_firstTemporary_thenPermanent_inOrder()  -> testMaxNumFailoverAttemptsExceedsThreshold()

src/main/java/redis/clients/jedis/mcf/MultiClusterPooledConnectionProvider.java
ggivo: I assume it is to reduce the lock time

src/main/java/redis/clients/jedis/providers/MultiClusterPooledConnectionProvider.java
atakavci: this is now package private since #4294

src/main/java/redis/clients/jedis/mcf/MultiClusterPooledConnectionProvider.java
atakavci: right, we can not assume user would know it might blocking the whole failover mechanism just by listening an event.

src/main/java/redis/clients/jedis/mcf/MultiClusterPooledConnectionProvider.java
atakavci: open to naming suggestions. `failover` itself is a too broad/general name to express such operation.

src/main/java/redis/clients/jedis/mcf/CircuitBreakerFailoverBase.java
atakavci: it is a bit tricky i guess, but it is no big deal to choose `assertOperability()` as well.
if failover itself failed with initial attempt (with `JedisTemporarilyNotAvailableException`/`JedisPermanentlyNotAvailableException`), then other threads blocked on `clusterFailover(Cluster cluster)` will receive nothing but an exception in case we `assertOperability()` here.
On the other hand, with `iterateActiveCluster`, they will have another chance for a heal-fast if selected cluster is still FORCED_OPEN but a healthy cluster emerged in between `clusterFailover(Cluster cluster)` executions.

atakavci: otherwise they will need to request another command and wait for retries.

ggivo: Got it. Don't see an issue from a performance perspective between the two approaches. I think it's fine to keep it as it is.

src/test/java/redis/clients/jedis/mcf/MultiClusterPooledConnectionProviderTest.java
atakavci: added one other test `userCommand_connectionExceptions_thenMultipleTemporary_thenPermanent_inOrder`

src/main/java/redis/clients/jedis/mcf/MultiClusterPooledConnectionProvider.java
ggivo: :) not sure, throwing some ideas

**`switchToHealthyCluster(SwitchReason reason, Cluster active)`**

`switchActiveCluster(SwitchReason reason, Cluster from)`

`activateNextHealthyCluster(SwitchReason reason, Cluster active)`

src/test/java/redis/clients/jedis/mcf/MultiClusterPooledConnectionProviderTest.java
```diff
@@ -252,4 +237,81 @@ void testHealthChecksStopAfterProviderClose() throws InterruptedException {
...
+      endpointStandalone0.getClientConfigBuilder().build()).weight(0.5f).build();
+    clusterConfigs[1] = ClusterConfig.builder(endpointStandalone1.getHostAndPort(),
+      endpointStandalone1.getClientConfigBuilder().build()).weight(0.3f).build();
+
+    // ATTENTION: these configuration settings are not random and
+    // adjusted to get exact numbers of failures with exact exception types
+    // and open to impact from other defaulted values withing the components in use.
+    MultiClusterPooledConnectionProvider testProvider = new MultiClusterPooledConnectionProvider(
+        new MultiClusterClientConfig.Builder(clusterConfigs).delayInBetweenFailoverAttempts(50)
+            .maxNumFailoverAttempts(2).retryMaxAttempts(1).circuitBreakerSlidingWindowMinCalls(2)
```
ggivo: I suggest to consider also disabling health checks, to prevent them interfering with CB logic

src/main/java/redis/clients/jedis/mcf/MultiClusterPooledConnectionProvider.java
atakavci: will replace with this one;
 **`switchToHealthyCluster(SwitchReason reason, Cluster active)`**

PR #4295, opened 2025-09-26
 [automatic failover] Add dual thresholds (min num of failures + failure rate) capabililty to circuit breaker

atakavci:
## Summary
Introduce dual-threshold circuit-breaker failover for Jedis multi‑cluster, simplify configuration, and align docs/tests. Failover now occurs only when BOTH of the following are exceeded: minimum number of failures and failure‑rate percentage.

## Key changes
- Dual-threshold enforcement
  - New evaluation in CircuitBreakerFailoverBase: evaluateThresholds(cluster) checks metrics and opens the CB when thresholds are exceeded; clusterFailover(...) then forces open and switches clusters.
  - CircuitBreakerCommandExecutor and CircuitBreakerFailoverConnectionProvider now call evaluateThresholds before execution and on tracked exceptions to keep decisions consistent and deterministic.
- New CircuitBreakerThresholdsAdapter
  - Maps MultiClusterClientConfig → resilience4j CircuitBreakerConfig.
  - Special handling:
    - failureRateThreshold = 0.0f maps to 100.0f (resilience4j limitation).
    - minimumNumberOfCalls is derived from configured minFailures and rate; when rate == 0.0f, sets Integer.MAX_VALUE to prevent the CB from auto-opening on its own.
    - Uses TIME_BASED window; size comes from config.
- MultiClusterClientConfig simplification and new defaults
  - Added circuitBreakerMinNumOfFailures (default 1000) and validation that rejects both minFailures==0 and rate==0 simultaneously.
  - Defaults updated: failureRateThreshold=10.0f, slidingWindowSize=2.
  - Removed slow-call knobs (duration and rate) and sliding-window type/min-calls from the public API.
- MultiClusterPooledConnectionProvider
  - Builds the CircuitBreaker using CircuitBreakerThresholdsAdapter.
  - Exposes cluster getters for getCircuitBreakerMinNumOfFailures() and getCircuitBreakerFailureRateThreshold().
- Documentation
  - Updated failover.md to the new defaults and simplified table; removed slow-call and sliding window type/min-calls; clarified the dual-threshold model.
- Tests
  - New CircuitBreakerFailoverBaseTests: matrix coverage of minFailures × failure-rate using real CB metrics with mocked wiring.
  - New/updated CircuitBreakerThresholdsTest: exercises real provider + real executor + real CB/Retry with only the connection pool mocked (no network), validating threshold behavior and failover switching.
  - Integration tests (FailoverIntegrationTest, ActiveActiveLocalFailoverTest) adjusted to the new configuration style and defaults.

## Behavior
Failover is triggered when BOTH conditions hold:
- failures >= circuitBreakerMinNumOfFailures
- (failed / (failed + successful)) × 100 >= circuitBreakerFailureRateThreshold

Notes
- With failureRateThreshold=0.0f, adapter maps to 100.0f and sets minimumNumberOfCalls=Integer.MAX_VALUE so resilience4j does not auto-open; Jedis still enforces the dual-threshold policy in the executor/base.
- Automatic transition from OPEN to HALF_OPEN remains disabled; failback is handled by existing health‑check/failback mechanisms.

## Migration
Use these settings:
- circuitBreakerMinNumOfFailures(int)
- circuitBreakerFailureRateThreshold(float)
- circuitBreakerSlidingWindowSize(int)  // time-based seconds

Removed settings (no longer applicable):
- slidingWindowType, slidingWindowMinCalls
- slowCallDurationThreshold, slowCallRateThreshold

If you previously relied on slow-call based opening, migrate to failure-based thresholds or external performance monitors.

## Validation
- Unit/parameterized tests cover edge thresholds (0%, 100%), small/large sample sizes.
- Integration tests updated to the new API; docs reflect defaults and examples.
- CircuitBreakerThresholdsTest now validates end‑to‑end interaction among provider + executor + circuit breaker without using network calls.


src/test/java/redis/clients/jedis/mcf/HealthCheckTest.java
```diff
@@ -785,7 +785,7 @@ void testPolicy_Majority_EarlyFailStopsAtTwo() throws Exception {
     CountDownLatch unhealthyLatch = new CountDownLatch(1);
 
     TestHealthCheckStrategy strategy = new TestHealthCheckStrategy(
-        HealthCheckStrategy.Config.builder().interval(5).timeout(200).numProbes(4)
+        HealthCheckStrategy.Config.builder().interval(5000).timeout(200).numProbes(4)
```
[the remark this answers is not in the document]
atakavci: intended

src/main/java/redis/clients/jedis/mcf/CircuitBreakerThresholdsAdapter.java
```diff
@@ -0,0 +1,58 @@
...
+        // IMPORTANT: minimumNumberOfCalls is calculated based on the configured thresholds.
+        minimumNumberOfCalls = calculateMinTotalCalls(
+            multiClusterClientConfig.getThresholdMinNumOfFailures(),
+            multiClusterClientConfig.getCircuitBreakerFailureRateThreshold());
+
+        slidingWindowSize = multiClusterClientConfig.getCircuitBreakerSlidingWindowSize();
+    }
+
+    private static int calculateMinTotalCalls(int failures, float rate) {
+        int minCalls = (int) (failures * rate * 100);
```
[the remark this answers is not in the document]
atakavci: calculations revisited.

src/test/java/redis/clients/jedis/mcf/HealthCheckTest.java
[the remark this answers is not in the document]
atakavci: this is a controlled verification on result of an individual healthCheck which consists of given number of probes. Dont want+need next healthcheck iteration to corrupt the controlled states.

src/main/java/redis/clients/jedis/mcf/MultiClusterPooledConnectionProvider.java
```diff
@@ -334,6 +329,12 @@ private void addClusterInternal(MultiClusterClientConfig multiClusterClientConfi
     }
 
     multiClusterMap.put(config.getHostAndPort(), cluster);
+
+    // this is the place where we listen tracked errors and check if
+    // thresholds are exceeded for the cluster
```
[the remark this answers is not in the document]
atakavci: this is complete wrong observation since each circuitBreaker, eventPublisher is created for a unique cluster instance and there is 1-1 relation between each of them

src/main/java/redis/clients/jedis/mcf/CircuitBreakerThresholdsAdapter.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  public int getSlidingWindowSize() {
+    return slidingWindowSize;
+  }
+
+  public CircuitBreakerThresholdsAdapter(MultiClusterClientConfig multiClusterClientConfig) {
+    // IMPORTANT: this is due to we can not set failureRateThreshold to 0.0f in Resilience4j
+    failureRateThreshold = multiClusterClientConfig.getCircuitBreakerFailureRateThreshold();
+    failureRateThreshold = failureRateThreshold == 0.0f ? 100.0f : failureRateThreshold;
```
ggivo: Adding a  FAILURE_RATE_TRESHOLD_DISABLED = 0.0f might help  clarifying what  0.0 stands for

src/main/java/redis/clients/jedis/mcf/CircuitBreakerThresholdsAdapter.java
```diff
@@ -0,0 +1,58 @@
+package redis.clients.jedis.mcf;
+
+import io.github.resilience4j.circuitbreaker.CircuitBreakerConfig.SlidingWindowType;
+import redis.clients.jedis.MultiClusterClientConfig;
+
+public class CircuitBreakerThresholdsAdapter {
```
ggivo: Can we add API docs.   
It has some special behaviour such as returning MAX_INT for multiClusterClientConfig.getCircuitBreakerFailureRateThreshold() to efectively disable the standard CB treshold failure rate detection.

And should we consider marking the class as internal?

src/main/java/redis/clients/jedis/mcf/CircuitBreakerThresholdsAdapter.java
```diff
@@ -0,0 +1,58 @@
...
+    minimumNumberOfCalls = calculateMinTotalCalls(
+      multiClusterClientConfig.getCircuitBreakerMinNumOfFailures(),
+      multiClusterClientConfig.getCircuitBreakerFailureRateThreshold());
+
+    slidingWindowSize = multiClusterClientConfig.getCircuitBreakerSlidingWindowSize();
+  }
+
+  private static int calculateMinTotalCalls(int failures, float rate) {
+    int minCalls = (int) (failures * rate);
+    if (failures == 0) {
```
ggivo: IF  zero means MinNumOfFailures is disabled. Can we add some constant to hint it.
 
Also what is the reasoning arround returning 1? Shouldn't in this case return the default (CircuitBreakerConfig.DEFAULT_MINIMUM_NUMBER_OF_CALLSs)

src/main/java/redis/clients/jedis/mcf/CircuitBreakerThresholdsAdapter.java
```diff
@@ -0,0 +1,58 @@
...
+    // IMPORTANT: minimumNumberOfCalls is calculated based on the configured thresholds.
+    minimumNumberOfCalls = calculateMinTotalCalls(
+      multiClusterClientConfig.getCircuitBreakerMinNumOfFailures(),
+      multiClusterClientConfig.getCircuitBreakerFailureRateThreshold());
+
+    slidingWindowSize = multiClusterClientConfig.getCircuitBreakerSlidingWindowSize();
+  }
+
+  private static int calculateMinTotalCalls(int failures, float rate) {
+    int minCalls = (int) (failures * rate);
```
ggivo: Not sure I get this one.

If we configure ( CircuitBreakerMinNumOfFailures = 10, CircuitBreakerFailureRateThreshold=20 in percents)

we get minimumNumberOfCalls = 200?
Is that the intention here? Shoudl't it be 50. 
e.g 20% of 50 =10?

src/main/java/redis/clients/jedis/mcf/CircuitBreakerThresholdsAdapter.java
```diff
@@ -0,0 +1,58 @@
...
+      minCalls = 1;
+    }
+    // This will prevent the CB from OPENing by itself in all non-0.0 rate configurations by
+    // setting minimumNumberOfCalls to a value that practically guarantees it won't be reached
+    // (Though it might get to MAX_VALUE, CB is going to receive exceptions and be %100 failure
+    // rate from this adapter).
+    //
+    // This keeps the CB CLOSED and lets our executor’s dual-threshold condition decide
+    // failover.
+    if (rate == 0) {
```
ggivo: similar here.  0 -> FAILURE_RATE_TRESHOLD_DISABLED?

src/main/java/redis/clients/jedis/mcf/MultiClusterPooledConnectionProvider.java
```diff
@@ -847,6 +856,30 @@ public void close() {
...
+    }
+
+    private static boolean isThresholdsExceeded(Cluster cluster, boolean lastFailRecorded) {
+      Metrics metrics = cluster.getCircuitBreaker().getMetrics();
+      // ATTENTION: this is to increment fails in regard to the current call that is failing,
+      // DO NOT remove the increment, it will change the behaviour in case of initial requests to
+      // cluster fail
+      int fails = metrics.getNumberOfFailedCalls() + (lastFailRecorded ? 0 : 1);
+      int succ = metrics.getNumberOfSuccessfulCalls();
+      if (fails >= cluster.getCircuitBreakerMinNumOfFailures()) {
```
ggivo, line 873: What about
CircuitBreakerFailureRateThreshold = 0

how do we ensure

`The value of 0.0 for the thresholdRate says that only the thresholdMinNumOfFailures is taken into account.`

src/main/java/redis/clients/jedis/mcf/CircuitBreakerThresholdsAdapter.java
```diff
@@ -0,0 +1,58 @@
...
+    return SlidingWindowType.TIME_BASED;
+  }
+
+  public int getSlidingWindowSize() {
+    return slidingWindowSize;
+  }
+
+  public CircuitBreakerThresholdsAdapter(MultiClusterClientConfig multiClusterClientConfig) {
+    // IMPORTANT: this is due to we can not set failureRateThreshold to 0.0f in Resilience4j
+    failureRateThreshold = multiClusterClientConfig.getCircuitBreakerFailureRateThreshold();
```
ggivo: Should update the docs for getCircuitBreakerFailureRateThreshold()
adding a note for special meaning of 0.

src/main/java/redis/clients/jedis/mcf/CircuitBreakerThresholdsAdapter.java
```diff
@@ -0,0 +1,58 @@
...
+    failureRateThreshold = failureRateThreshold == 0.0f ? 100.0f : failureRateThreshold;
+
+    // IMPORTANT: minimumNumberOfCalls is calculated based on the configured thresholds.
+    minimumNumberOfCalls = calculateMinTotalCalls(
+      multiClusterClientConfig.getCircuitBreakerMinNumOfFailures(),
```
ggivo: similar here

Should update the docs for getCircuitBreakerMinNumOfFailures()
adding a note for special meaning of 0.

src/main/java/redis/clients/jedis/mcf/CircuitBreakerCommandExecutor.java
```diff
@@ -39,8 +39,17 @@ public <T> T executeCommand(CommandObject<T> commandObject) {
     supplier.withRetry(cluster.getRetry());
     supplier.withFallback(provider.getFallbackExceptionList(),
       e -> this.handleClusterFailover(commandObject, cluster));
-
-    return supplier.decorate().get();
+    try {
+      return supplier.decorate().get();
+    } catch (Exception e) {
+      if (isCircuitBreakerTrackedException(e, cluster)) {
+        cluster.evaluateThresholds(true);
```
ggivo: Hey,


looking at `evaluateThresholds` it will propagate call to `isThresholdsExceeded` and in case `lastFailRecord=true` will consider failed calls with one more. 


Didn't CB already counted the error? In which case we need to add  one to failed calls before evaluating the thresholds?

ggivo (commented):
Still going trough the complete review, have some initial questions

src/main/java/redis/clients/jedis/mcf/CircuitBreakerCommandExecutor.java
atakavci: those lines 45-46 are dropped after our conversation which rises my awareness around the tracked exceptions check and threshold evaluation; both is already done before executions commences to this block of code.
So those two lines bring no value and and trimmed in this commit [e49f2e7](https://github.com/redis/jedis/pull/4295/commits/e49f2e7c79e55517f00ab4c264676a5df5fd5599)

src/main/java/redis/clients/jedis/mcf/CircuitBreakerThresholdsAdapter.java
atakavci: after careful consideration this calculation is not necessary at all since we inject our own way evaluating thresholds. This adapter will only function to avoid resillience4j evaluation kick in and OPEN the circuit.
This is done by just setting the max rate and max min number of total calls in the adapter object.

src/main/java/redis/clients/jedis/mcf/CircuitBreakerThresholdsAdapter.java
atakavci: done with updating java docs.
class is now package private.

src/main/java/redis/clients/jedis/mcf/CircuitBreakerThresholdsAdapter.java
atakavci: after careful consideration this calculation is not necessary at all since we inject our own way evaluating thresholds. This adapter will only function to avoid resillience4j evaluation kick in and OPEN the circuit.

src/main/java/redis/clients/jedis/mcf/CircuitBreakerThresholdsAdapter.java
atakavci: same comment as above.

src/main/java/redis/clients/jedis/mcf/CircuitBreakerThresholdsAdapter.java
atakavci: same comment as above.

src/main/java/redis/clients/jedis/mcf/MultiClusterPooledConnectionProvider.java
atakavci: i dont understand question exactly,
if the `getCircuitBreakerMinNumOfFailures` threshold is exceeded, then calculated `getCircuitBreakerFailureRateThreshold` is always equal or higher than the rate threshold.
```java
    private static boolean isThresholdsExceeded(Cluster cluster, boolean lastFailRecorded) {
      Metrics metrics = cluster.getCircuitBreaker().getMetrics();
      // ATTENTION: this is to increment fails in regard to the current call that is failing,
      // DO NOT remove the increment, it will change the behaviour in case of initial requests to
      // cluster fail
      int fails = metrics.getNumberOfFailedCalls() + (lastFailRecorded ? 0 : 1);
      int succ = metrics.getNumberOfSuccessfulCalls();
      if (fails >= cluster.getCircuitBreakerMinNumOfFailures()) {
        float ratePercentThreshold = cluster.getCircuitBreakerFailureRateThreshold();// 0..100
        int total = fails + succ;
        if (total == 0) return false;
        float failureRatePercent = (fails * 100.0f) / total;
        return failureRatePercent >= ratePercentThreshold;
      }
      return false;
    }  
```

src/main/java/redis/clients/jedis/mcf/CircuitBreakerThresholdsAdapter.java
atakavci: added info on what 0 means for both `getCircuitBreakerFailureRateThreshold` and `getCircuitBreakerMinNumOfFailures`

src/main/java/redis/clients/jedis/mcf/CircuitBreakerThresholdsAdapter.java
atakavci: added some explanation around what 0 means and how it works with dual threshold approach.

src/main/java/redis/clients/jedis/mcf/MultiClusterPooledConnectionProvider.java
ggivo: I ment in case `getCircuitBreakerMinNumOfFailures = 10` & `getCircuitBreakerFailureRateThreshold=0`
We need to consider only `getCircuitBreakerMinNumOfFailures`. But did not realize that in this case if we already have a failure, than `failureRatePercent >= ratePercentThreshold` will always hold true

src/main/java/redis/clients/jedis/mcf/MultiClusterPooledConnectionProvider.java
```diff
@@ -847,6 +856,30 @@ public void close() {
...
+        getCircuitBreaker().transitionToOpenState();
+      }
+    }
+
+    private static boolean isThresholdsExceeded(Cluster cluster, boolean lastFailRecorded) {
+      Metrics metrics = cluster.getCircuitBreaker().getMetrics();
+      // ATTENTION: this is to increment fails in regard to the current call that is failing,
+      // DO NOT remove the increment, it will change the behaviour in case of initial requests to
+      // cluster fail
+      int fails = metrics.getNumberOfFailedCalls() + (lastFailRecorded ? 0 : 1);
```
ggivo, line 871: Not sure I get this one.
In which case, metrics are not increased yet?

src/main/java/redis/clients/jedis/MultiClusterClientConfig.java
```diff
@@ -1295,74 +1179,16 @@ public Builder retryIgnoreExceptionList(List<Class> retryIgnoreExceptionList) {
...
-     * @param circuitBreakerSlidingWindowMinCalls minimum number of calls for failure rate
-     *          calculation
-     * @return this builder instance for method chaining
-     */
-    public Builder circuitBreakerSlidingWindowMinCalls(int circuitBreakerSlidingWindowMinCalls) {
-      this.circuitBreakerSlidingWindowMinCalls = circuitBreakerSlidingWindowMinCalls;
-      return this;
-    }
-
-    /**
```
ggivo: Ah, I remember we discussed this earlier, but it slipped my mind.
Could you remind me whether the limitation is technical (i.e., not supporting a COUNT_BASED moving window), and what the reasoning was for not supporting it?

src/test/java/redis/clients/jedis/mcf/CircuitBreakerThresholdsTest.java
```diff
@@ -0,0 +1,257 @@
...
+    executor = new CircuitBreakerCommandExecutor(spyProvider);
+
+    dummyCommand = new CommandObject<>(new CommandArguments(Protocol.Command.PING),
+        BuilderFactory.STRING);
+
+    // Replace the cluster's pool with a mock to avoid real network I/O
+    poolMock = mock(TrackingConnectionPool.class);
+    Field f = MultiClusterPooledConnectionProvider.Cluster.class.getDeclaredField("connectionPool");
+    f.setAccessible(true);
+    f.set(cluster, poolMock);
```
ggivo: Code duplication. How about introducing a test util[
Had a similar need and introducing one here [ReflectionTestUtil.java](https://github.com/redis/jedis/pull/4300/files#diff-f17c3b7bba08d89d87c0f42c3df16fd9644fcc2330cb7c265743720709ec15f0)

We should also use a common utility method for getting the `connectionPool` in case the field name changes in the future

src/test/java/redis/clients/jedis/mcf/CircuitBreakerThresholdsTest.java
```diff
@@ -0,0 +1,257 @@
...
+    Cluster c = sp.getCluster();
+    try (CircuitBreakerCommandExecutor ex = new CircuitBreakerCommandExecutor(sp)) {
+      CommandObject<String> cmd = new CommandObject<>(new CommandArguments(Protocol.Command.PING),
+          BuilderFactory.STRING);
+
+      TrackingConnectionPool pool = mock(TrackingConnectionPool.class);
+      Field f = MultiClusterPooledConnectionProvider.Cluster.class
+          .getDeclaredField("connectionPool");
+      f.setAccessible(true);
+      f.set(c, pool);
```
ggivo: Same as above

src/test/java/redis/clients/jedis/mcf/CircuitBreakerThresholdsTest.java
```diff
@@ -0,0 +1,257 @@
...
+    try (CircuitBreakerCommandExecutor ex = new CircuitBreakerCommandExecutor(spy)) {
+
+      CommandObject<String> cmd = new CommandObject<>(new CommandArguments(Protocol.Command.PING),
+          BuilderFactory.STRING);
+
+      TrackingConnectionPool pool = mock(TrackingConnectionPool.class);
+      Field f = MultiClusterPooledConnectionProvider.Cluster.class
+          .getDeclaredField("connectionPool");
+      f.setAccessible(true);
+      f.set(c, pool);
```
ggivo: same as above

ggivo (commented):
Some of the new properties need to be renamed to align the naming with the design:

circuitBreakerMinNumOfFailures(int)       -> minNumOfFailuresThreshold
circuitBreakerFailureRateThreshold(float) -> failureRateThreshold
circuitBreakerSlidingWindowSize(int)        -> failureDetectionWindowSize

We need to drop also circuitBreaker prefix from others.
Probably with follow-up PR?

src/test/java/redis/clients/jedis/mcf/CircuitBreakerThresholdsTest.java
atakavci: done

src/test/java/redis/clients/jedis/mcf/CircuitBreakerThresholdsTest.java
atakavci: done

src/test/java/redis/clients/jedis/mcf/CircuitBreakerThresholdsTest.java
atakavci: done

src/main/java/redis/clients/jedis/MultiClusterClientConfig.java
atakavci: there is nothing technical to not use COUNT_BASED, we need to propose the change and agree with design covers this and alignment across libs.

ggivo: Thanks!

ggivo (approved):
LGTM

PR #4322, opened 2025-10-10
docs: Correct builder to use hostAndPort method

ChrisMcD1:
I just noticed that `hostAndPort` did not actually exist when upgrading my project to 7.0. I presume this is the recommended alternative.

I see no notes in the CONTRIBUTING.md about a CLA or anything similar, but sorry if I'm missing something standard for this project!


ggivo:
@ChrisMcD1 
Thanks for getting back! You’re not missing anything — all looks good. I’ll merge it once the tests pass.

PR #4328, opened 2025-10-13
Fix connection leak in scanIteration with JedisSentineled #4323

### LINKED ISSUE #4323: ScanIteration (JedisCommandIterationBase) connection leak with JedisSentineled
Not sure I fully understand what's going on, so looking for some idea on how to even approach fixing this.

```java
            var jedis = new JedisSentineled(....);
            var it = j.scanIteration(100, cacheObjKey()+"*");
            // doesn't really matter what happens past the line above
            while (!it.isIterationCompleted()) {
                var data = it.nextBatchList().toArray(new String[0]);
                if (data.length > 0) {
                    j.del(data);
                }
            }
```

This leaks a connection for me.
`JedisCommandIterationBase`'s constructor calls connectionProvider.getConnectionMap(), which is default `ConnectionProvider`'s `getConnectionMap()`, which just allocates a connection, puts in a single-element map, and returns that. `JedisCommandIterationBase` is not a `Closeable`, and never clears that connection out (because it doesn't think it needs to).

It looks like pool-based connection providers must return a pool from a `getConnectionMap()`, like `PooledConnectionProvider` does. `SentinelConnectionProvider` doesn't do that. But it's also not easy for it to do that because its pool is volatile.

Any ideas how to work around this in the meantime?

ggivo:
## Problem
 
When using `scanIteration()` with `JedisSentineled`, connections are leaked from the pool and never returned.
 
`SentineledConnectionProvider` uses the default `ConnectionProvider.getConnectionMap()` implementation, which allocates a raw `Connection` from the pool. `JedisCommandIterationBase` uses this connection directly (line 68-69) without returning it to the pool.
 
## Solution

Override `getConnectionMap()` and `getPrimaryNodesConnectionMap()` in `SentineledConnectionProvider` to return the pool itself, consistent with `PooledConnectionProvider`:
 
```java
@Override
public Map<?, Pool<Connection>> getConnectionMap() {
  return Collections.singletonMap(currentMaster, pool); 
}
 
@Override
public Map<?, Pool<Connection>> getPrimaryNodesConnectionMap() {
   return Collections.singletonMap(currentMaster, pool);
}
```
 
This allows `JedisCommandIterationBase` to properly manage connections using try-with-resources (lines 71-73).
 
**Note on volatile pool**: The iteration captures the pool reference at construction time. Connections from the old pool remain valid even if failover occurs during iteration.

**Note on  returning a `Pool<Connection>` by `getConnectionMap` vs  `Connection`** to:
 - `JedisCommandIterationBase` already handles both `Connection` and `Pool` types (lines 68-76)
 - `getConnectionMap()` is only used internally by iteration classes
 - This change aligns `SentineledConnectionProvider` with `PooledConnectionProvider`'s existing behavior
 - Previous behavior would cause connections to be exhausted using multilpe times `scan` operaions 

Fixes #4323

### DIFF
--- a/src/test/java/redis/clients/jedis/JedisShardedPubSubBaseTest.java
+++ b/src/test/java/redis/clients/jedis/JedisShardedPubSubBaseTest.java
@@ -53,7 +53,7 @@ public class JedisShardedPubSubBaseTest {
         });
         thread.start();
 
-        assertTrue(countDownLatch.await(20, TimeUnit.MILLISECONDS));
+        assertTrue(countDownLatch.await(100, TimeUnit.MILLISECONDS));
 
     }
 }


atakavci (approved):
LGTM

PR #4332, opened 2025-10-20
Adding a registry that maps command names to their flags

uglide:
Introduces command flags to simplify command routing.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -188,7 +188,7 @@
 		<dependency>
 			<groupId>org.apache.httpcomponents.client5</groupId>
 			<artifactId>httpclient5-fluent</artifactId>
-			<version>5.5</version>
+			<version>5.5.1</version>
 			<scope>test</scope>
 		</dependency>
 
--- a/src/test/java/redis/clients/jedis/JedisPubSubBaseTest.java
+++ b/src/test/java/redis/clients/jedis/JedisPubSubBaseTest.java
@@ -55,6 +55,6 @@ public class JedisPubSubBaseTest  {
         });
         thread.start();
 
-        assertTrue(countDownLatch.await(10, TimeUnit.MILLISECONDS));
+        assertTrue(countDownLatch.await(30, TimeUnit.MILLISECONDS));
     }
 }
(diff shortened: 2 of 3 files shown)


src/main/java/redis/clients/jedis/CommandObject.java
```diff
@@ -1,10 +1,694 @@
...
+import java.util.HashMap;
 import java.util.Iterator;
+import java.util.Map;
 import redis.clients.jedis.args.Rawable;
+import redis.clients.jedis.commands.ProtocolCommand;
+import redis.clients.jedis.util.SafeEncoder;
 
 public class CommandObject<T> {
 
+    /**
```
atakavci: lets not do this and put everything into a seperate completely auto-generate class like CommandFlagRegistry or some provider class,, and use generated content at once without need of regex or any other searching/parsing. Then just consume it from CommandObject as needed.

uglide: done

src/main/java/redis/clients/jedis/StaticCommandFlagsRegistry.java
```diff
@@ -0,0 +1,925 @@
...
+   * all CommandObject instances.
+   * @param cmd the protocol command
+   * @return EnumSet of CommandFlag for this command, or empty set if command has no flags
+   */
+  @Override
+  public EnumSet<CommandFlag> getFlags(ProtocolCommand cmd) {
+    byte[] raw = cmd.getRaw();
+    String commandName = SafeEncoder.encode(raw).toUpperCase();
+    return COMMAND_FLAGS_REGISTRY.getOrDefault(commandName, EMPTY_FLAGS);
+  }
```
ggivo: This does not seem to work for subcommands like "CLUSTER INFO", "CONFIG GET" ...
In this case, ProtocolCommand is actually only "CLUSTER"

uglide: I've addressed this problem. See my latest commit https://github.com/redis/jedis/pull/4332/commits/ff09e22a7e965fd177b0f4736dd31420617a7419

ggivo:
@uglide 
I suggest also separating generated code (populating the StaticRegistry) from actual implmenattion so it is easier for maintaining in future
You can check PR with proposal against your branch here :   : https://github.com/redis/jedis/pull/4345

PR #4340, opened 2025-10-31
Moves MutltiNodePipeline Executor to JedisClientConfig

dolavb:
This allow the configuration of ClusterPipelineExecutor to sync pipeline in parallel. The default implementation remain problematic. This new approach will allow clients to address the performance issue fo the default approach.


dolavb:
Wrong branch.

PR #4353, opened 2025-11-06
Add support for CAS/CAD, DELEX, DIGEST commands

### LINKED ISSUE #4341: Add support for CAS/CAD, DELEX, DIGEST commands
#### Set the key only when a specified condition is met

Options:
- IFEQ match-value - Set the key’s value and expiration only if its current value is equal to match-value. If the key doesn’t exist, it won’t be created
- IFNE match-value - Set the key’s value and expiration only if its current value is not equal to match-value. If the key doesn’t exist, it will be created
- IFDEQ match-digest - Set the key’s value and expiration only if the digest of its current value is equal to match-digest. If the key doesn’t exist, it won’t be created
- IFDNE match-digest - Set the key’s value and expiration only if the digest of its current value is not equal to match-digest. If the key doesn’t exist, it will be created

Reply:
- If GET was not specified, any of the following:
  - Nil reply if either
    - the key does not exist and XX/IFEQ/IFDEQ was specified. The key was not created.
    - the key exists and NX was specified or a specified IFEQ/IFNE/IFDEQ/IFDNE condition is false. The key was not set.
  - Simple string reply: OK: The key was set.
- If GET was specified, any of the following:
  - Nil reply: The key didn't exist before the SET.
  - Bulk string reply: The previous value of the key (Regardless of the conditional IFEQ/IFNE/IFDEQ/IFDNE results).

#### Removes the specified key and optionally applies only when a specified condition is met

Options:
- IFEQ match-value - Only delete the key if match-value equals the current value
- IFNE match-value - Only delete the key if match-value does not equal the current value
- IFDEQ match-digest - Only delete the key if match-digest equals the current digest value
- IFDNE match-digest - Only delete the key if match-digest does not equal the current digest value

Reply:
Integer reply: the number of keys that were removed

#### Return a hex signature of the value stored in the specified key.

Reply:
- Bulk string reply: the hex signature of the key or null if the key does not exist
- Simple error reply: the command returns an error if the key is not a string

a-TODO-rov:
resolves #4341

### DIFF
--- a/src/main/java/redis/clients/jedis/CommandObjects.java
+++ b/src/main/java/redis/clients/jedis/CommandObjects.java
@@ -498,27 +498,31 @@ public class CommandObjects {
   }
 
   public final CommandObject<String> set(String key, String value, SetParams params, ValueCondition cond) {
-    CommandArguments ca = commandArguments(Command.SET).key(key).add(value).addParams(params);
+    CommandArguments ca = commandArguments(Command.SET).key(key).add(value);
     cond.addTo(ca);
+    ca.addParams(params);
     return new CommandObject<>(ca, BuilderFactory.STRING);
   }
 
   public final CommandObject<String> setGet(String key, String value, SetParams params, ValueCondition cond) {
-    CommandArguments ca = commandArguments(Command.SET).key(key).add(value).addParams(params);
+    CommandArguments ca = commandArguments(Command.SET).key(key).add(value);
     cond.addTo(ca);
+    ca.addParams(params);
     ca.add(Keyword.GET);
     return new CommandObject<>(ca, BuilderFactory.STRING);
   }
 
   public final CommandObject<String> set(byte[] key, byte[] value, SetParams params, ValueCondition cond) {
-    CommandArguments ca = commandArguments(Command.SET).key(key).add(value).addParams(params);
+    CommandArguments ca = commandArguments(Command.SET).key(key).add(value);
     cond.addTo(ca);
+    ca.addParams(params);
     return new CommandObject<>(ca, BuilderFactory.STRING);
   }
 
   public final CommandObject<byte[]> setGet(byte[] key, byte[] value, SetParams params, ValueCondition cond) {
-    CommandArguments ca = commandArguments(Command.SET).key(key).add(value).addParams(params);
+    CommandArguments ca = commandArguments(Command.SET).key(key).add(value);
     cond.addTo(ca);
+    ca.addParams(params);
     ca.add(Keyword.GET);
     return new CommandObject<>(ca, BuilderFactory.BINARY);
   }


pom.xml
a-TODO-rov, line 1: Should revert

src/main/java/redis/clients/jedis/commands/KeyCommands.java
```diff
@@ -418,6 +420,17 @@ public interface KeyCommands {
    */
   long del(String... keys);
 
+  /**
+   * Compare-and-delete: delete key if optional value/digest condition matches.
+   * @return 1 if the key was deleted, 0 otherwise
+   */
```
a-TODO-rov, line 426: Code for removal

src/test/java/redis/clients/jedis/commands/jedis/StringValuesCommandsTest.java
```diff
@@ -40,6 +41,44 @@ public void setAndGet() {
...
+
+    // IFEQ on missing key: false -> no create
+    jedis.del("k");
+    assertNull(jedis.set("k", "vy", ValueCondition.valueEq("anything")));
+    assertNull(jedis.get("k"));
+  }
+
+
+
+
```
a-TODO-rov, line 81: some extra lines

src/main/java/redis/clients/jedis/commands/StringCommands.java
```diff
@@ -50,6 +59,24 @@ public interface StringCommands extends BitCommands {
...
+   * Experimental: SET with SetParams and compare-and-* condition.
+   */
+  @Experimental
+  String set(String key, String value, SetParams params, ValueCondition condition);
+
+  /**
+   * Experimental: SET+GET with SetParams and compare-and-* condition.
+   */
+  @Experimental
+  String setGet(String key, String value, SetParams params, ValueCondition condition);
```
ggivo, line 78: We could add new params directly to SetParams and avoid having more overloaded methods.

considering that actually CAS is extension to existing GET command

src/main/java/redis/clients/jedis/commands/KeyBinaryCommands.java
```diff
@@ -60,6 +62,13 @@ public interface KeyBinaryCommands {
 
   long del(byte[] key);
 
+
+  /**
+   * Experimental: Compare-and-delete guarded by value/digest condition.
+   */
+  @Experimental
+  long delex(byte[] key, ValueCondition condition);
```
ggivo, line 70: As mentioned in the spec
DELEX is a new command

I think we should go with :
`DelExParams with the same ifeq/ifne/ifdeq/ifdne methods from SetParams.`

a-TODO-rov:
closed in favor of #4363

PR #4361, opened 2025-11-12
Add MSETEX command support

### LINKED ISSUE #4343: Add support for MSETEX command
Sets the given keys to their respective values.
This command is an extension of the MSETNX that adds expiration and XX options.

Options:
- EX seconds - Set the specified expiration time, in seconds.
- PX milliseconds - Set the specified expiration time, in milliseconds.
- EXAT timestamp-seconds - Set the specified Unix time at which the keys will expire, in seconds.
- PXAT timestamp-milliseconds - Set the specified Unix time at which the keys will expire, in milliseconds.
- KEEPTTL - Retain the time to live associated with the keys.
- XX - Only set the keys and their expiration if all already exist.
- NX - Only set the keys and their expiration if none exist

Reply:
Integer reply: 0 if no keys were set
Integer reply: 1 if all the keys were set

a-TODO-rov:
resolves #4343

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -544,6 +544,7 @@
 						<include>**/ClientTestUtil.java</include>
 						<include>**/ReflectionTestUtil.java</include>
 						<include>**/*CommandFlags*.java</include>
+						<include>**/*MSetExParams*.java</include>
 					</includes>
 				</configuration>
 				<executions>


src/main/java/redis/clients/jedis/UnifiedJedis.java
```diff
@@ -978,6 +983,11 @@ public String mset(byte[]... keysvalues) {
     return executeCommand(commandObjects.mset(keysvalues));
   }
 
+  @Override
+  public boolean msetex(redis.clients.jedis.params.SetParams params, byte[]... keysvalues) {
```
ggivo: package qualifier not needed

ggivo (commented):
LGTM in general
There are some formatting to be addressed
New functions miss an extra new line and has additional one after them.

And also we should add a test for "XX"

PR #4363, opened 2025-11-17
Add support for CAS/CAD, DELEX, DIGEST commands

### LINKED ISSUE #4341: Add support for CAS/CAD, DELEX, DIGEST commands

a-TODO-rov:
resolves #4341


src/main/java/redis/clients/jedis/util/CompareCondition.java
```diff
@@ -0,0 +1,163 @@
...
+  public static CompareCondition valueNe(String value) {
+    if (value == null) {
+      throw new IllegalArgumentException("value must not be null");
+    }
+    return new CompareCondition(Condition.VALUE_NOT_EQUAL, value);
+  }
+
+  public static CompareCondition valueEq(byte[] value) {
+    if (value == null) {
+      throw new IllegalArgumentException("value must not be null");
```
ggivo: There is a utility method that can be used for param validation
```
 JedisAsserts.notNull(provider, "IFEQ value must not be null");
```

ggivo (approved):
LGTM

ggivo (approved):
LGTM

PR #4366, opened 2025-11-20
Add helpers to Jedis pool

oscar-besga-panel:
This mininal chage of code will allow to use Jedis connections without making a try/catch

Before
```java
    try (Jedis jedis = jedisPool.getResource()) {
      //some code
       jedis.setnx("hello", "there");
      //some code
    }
```

After
```java
jedisPool.withResource( (jedis) -> {
      //some code
       jedis.setnx("hello", "there");
      //some code
 });
```

Cleaner in my opinion

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -546,6 +546,7 @@
 						<include>**/*CommandFlags*.java</include>
 						<include>**/*CompareCondition*.java</include>
 						<include>**/*MSetExParams*.java</include>
+						<include>**/*UnitTest.java</include>
 					</includes>
 				</configuration>
 				<executions>
--- a/src/test/java/redis/clients/jedis/JedisPoolTest.java
+++ b/src/test/java/redis/clients/jedis/JedisPoolTest.java
@@ -9,8 +9,10 @@ import org.apache.commons.pool2.PooledObjectFactory;
 import org.apache.commons.pool2.impl.DefaultPooledObject;
 import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
 
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInfo;
 import redis.clients.jedis.exceptions.InvalidURIException;
 import redis.clients.jedis.exceptions.JedisAccessControlException;
 import redis.clients.jedis.exceptions.JedisConnectionException;
@@ -18,9 +20,10 @@ import redis.clients.jedis.exceptions.JedisException;
 
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static org.awaitility.Awaitility.await;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertInstanceOf;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertSame;
 import static org.junit.jupiter.api.Assertions.assertThrows;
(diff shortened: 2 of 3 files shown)


oscar-besga-panel:
retest

ggivo:
@oscar-besga-panel 
Did you consider `JedisPooled`? It is available starting with Jedis 4 and helps to avoid a try-with-resources block for each command.

https://redis.io/docs/latest/develop/clients/jedis/connect/#connect-with-a-connection-pool

src/main/java/redis/clients/jedis/JedisPool.java
```diff
@@ -392,4 +394,17 @@ public void returnResource(final Jedis resource) {
...
+
+  public void withJedisPoolDo(Consumer<Jedis> consumer) {
+    try (Jedis jedis = this.getResource()) {
+      consumer.accept(jedis);
+    }
+  }
+
+  public <K> K withJedisPoolGet(Function<Jedis, K> function) {
+    try (Jedis jedis = this.getResource()) {
+      return function.apply(jedis);
```
ggivo, line 406: I suggest :
Rename withJedisPoolGet → withResource

src/main/java/redis/clients/jedis/JedisPool.java
```diff
@@ -392,4 +394,17 @@ public void returnResource(final Jedis resource) {
       }
     }
   }
+
+  public void withJedisPoolDo(Consumer<Jedis> consumer) {
```
ggivo: I suggest :
Rename withJedisPoolDo → withResource

oscar-besga-panel: The problem is if the two methods are renamed equally this won't compile
```java
        withResource(jedis -> jedis.set("c"));
        String x = withResource(jedis -> jedis.get("c")); 
```

as it will trigger ambiguous method call in both lines.
Even if I like your idea, Java won't allow 
(if you use a full body lambda it will make it, but that will reducefuture uses of the methods)

I'll suggest withResource and withResourceGet

src/main/java/redis/clients/jedis/JedisPool.java
oscar-besga-panel: See bellow

oscar-besga-panel:
> @oscar-besga-panel Did you consider `JedisPooled`? It is available starting with Jedis 4 and helps to avoid a try-with-resources block for each command.
> 
> https://redis.io/docs/latest/develop/clients/jedis/connect/#connect-with-a-connection-pool

Yes, but I use JedisPool in my library for now. 
(considering change as is Legacy but not deprecated)

JedisPooled is 99% the times better, except if you want to send multiple commands. then it will be getting/closing a pooled connection for every one. 
Well you can use MULTI/EXEC or pipeline, thougth

oscar-besga-panel:
I'm considering extending JedisPooled to be compatible with JedisPool  by making it return a Jedis object... but not for now; and I know it has serious challenges. 
If it is to be done, in another PR

src/main/java/redis/clients/jedis/JedisPool.java
ggivo: make sense. let's keep  withResource/withResourceGet

ggivo:
> I'm considering extending JedisPooled to be compatible with JedisPool by making it return a Jedis object... but not for now; and I know it has serious challenges. If it is to be done, in another PR

I wouldn't go that way. JedisPooled is based on `UnifiedJedis`. Let's discuss this separately.

ggivo:
@oscar-besga-panel 
Did a bit of unit test clean up; other than that, the change looks good to me. Let me know if you plan more changes or we are good to go and merge it once test's are green

oscar-besga-panel:
It looks good to me, merge when appropiate.
_edit: typo_

oscar-besga-panel:
Thanks

ggivo:
@oscar-besga-panel

Thanks a lot for the PR and the cleanup — appreciate the time and effort you put into this.Thanks for the contribution!

PR #4371, opened 2025-11-29
Make a bridge between JedisPool and JedisPooled

oscar-besga-panel:
This PR is intented to make a bridge between JedisPool and JedisPooled by making these two share a common interface **JedisProvider**.

**JedisProvider** provides methods to get and return a Jedis connection; and also helpers to use within a lambda.
Jedis is modified to return the connection when closed to the appropiate pool.
JedisPooled implements the method to get and return a Jedis object by providing the connection and the source pool, and to return this connection to the pool.
Also it has unit and funcional test to check everything.

I'm not sure if this is adecuate rigth now, if there's a neeed for it (in my [project ](https://github.com/oscar-besga-panel/JedisExtraUtils) I'm already migrating from JedisPool to JedisPooled).

But if it serves, or you have detected the need for it, I think this could be the way to do it.

Thanks for your time

### DIFF
--- a/src/test/java/redis/clients/jedis/JedisPoolTest.java
+++ b/src/test/java/redis/clients/jedis/JedisPoolTest.java
@@ -33,9 +33,11 @@ import static org.junit.jupiter.api.Assertions.fail;
 @Tag("integration")
 public class JedisPoolTest {
 
-  private static final EndpointConfig endpointStandalone0 = HostAndPorts.getRedisEndpoint("standalone0");
+  private static final EndpointConfig endpointStandalone0 = HostAndPorts
+      .getRedisEndpoint("standalone0");
 
-  private static final EndpointConfig endpointStandalone1 = HostAndPorts.getRedisEndpoint("standalone1");
+  private static final EndpointConfig endpointStandalone1 = HostAndPorts
+      .getRedisEndpoint("standalone1");
 
   private String testKey;
   private String testValue;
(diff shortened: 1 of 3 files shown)


oscar-besga-panel:
It gives me this error

> (default) on project jedis: File '/home/runner/work/jedis/jedis/src/test/java/redis/clients/jedis/JedisPooledUnitTest.java' has not been previously formatted.  Please format file and commit before running validation!

But I don't know how to fix it, I don't see how to propertly make it valid.

ggivo:
Hi @oscar-besga-panel 
Thanks for taking the time and sharing the idea.
The team is currently a bit overloaded finishing other tasks, but we will review this as soon as possible.

Regarding the formatting issue, please run:
`mvn formatter:format`

oscar-besga-panel:
Hi @ggivo 
No problem, this is not urgent at all, no  rush here.
Now tests look better.

Tks

oscar-besga-panel:
I'm thinking about closing this PR. Not because the code, but because I see it worthless.
Now, we have that JedisPooled, RedisClient, RedisClusterClient, RedisSentinelClient are childs of UnifiedRedis.   

So, making a bridge between JedisPool and JedisPooled doen't have much sense, the two are deprecated now.

It would be better to make something in order to have JedisPool compatible or usable with new developments; like a JedisPoolUnifiedAdapted, and make something like jedisPool.asUnifiedRedis() to ensure compatibility.

Also, this PR has been opened for some many time, and there are conflicts; which may be problematic to resolve.

What do you think ?

ggivo:
@oscar-besga-panel 
> I'm thinking about closing this PR. Not because the code, but because I see it worthless. Now, we have that JedisPooled, RedisClient, RedisClusterClient, RedisSentinelClient are childs of UnifiedRedis.
> 
> So, making a bridge between JedisPool and JedisPooled doen't have much sense, the two are deprecated now.
> 
> It would be better to make something in order to have JedisPool compatible or usable with new developments; like a JedisPoolUnifiedAdapted, and make something like jedisPool.asUnifiedRedis() to ensure compatibility.
> 
> Also, this PR has been opened for some many time, and there are conflicts; which may be problematic to resolve.
> 
> What do you think?

Hey, thanks for following this up.

I agree with your assessment. We’re planning to remove JedisPool and move forward by promoting the use of RedisClient, RedisClusterClient, and RedisSentinelClient. Given that direction, adding a bridge between JedisPool and JedisPooled doesn’t bring much value, especially since both are deprecated.

ggivo:
Closing it per above comments
@oscar-besga-panel thanks again for sharing the idea!

PR #4373, opened 2025-12-05
Expose method to add upstream driver libraries to CLIENT SETINFO payload

viktoriya-kutsarova:
This PR introduces support for attaching upstream driver library information to the CLIENT SETINFO command.

Changes:

Introduced a DriverInfo class with builder pattern to hold driver information and upstream drivers, providing validation and formatting. Updated ClientSetInfoConfig to use DriverInfo for appending upstream drivers, chaining on top of existing configurations, and optionally overriding the full driver name.

Why we need the change:

Some drivers, frameworks, or SDK layers are built on top of this client and need a way to report their own library identifiers to Redis. This helps improve observability, analytics, and troubleshooting by making the client stack visible through CLIENT LIST and related introspection commands.

### DIFF
--- a/src/main/java/redis/clients/jedis/DriverInfo.java
+++ b/src/main/java/redis/clients/jedis/DriverInfo.java
@@ -1,7 +1,3 @@
-/*
- * Copyright 2025-Present, Redis Ltd. and Contributors All rights reserved. Licensed under the MIT
- * License.
- */
 package redis.clients.jedis;
 
 import java.util.ArrayList;


src/main/java/redis/clients/jedis/DriverInfo.java
```diff
@@ -0,0 +1,292 @@
...
+
+    validateNoInvalidCharacters(driverName, "Driver name");
+
+    if (!driverName.matches(DRIVER_NAME_PATTERN)) {
+      throw new JedisValidationException(
+          "Upstream driver name must follow Maven artifactId naming conventions: "
+              + "lowercase letters, digits, hyphens, and underscores only, starting with a lowercase letter "
+              + "(e.g., 'spring-data-redis', 'lettuce-core', 'akka-redis_2.13')");
+    }
+  }
```
ggivo: Looking at the code, this is applied on the legacy code path, but the validation is more strict than existing one. Introduces maven naming convention (`DRIVER_NAME_PATTERN`), which, according to https://redis.io/docs/latest/commands/client-setinfo/ is not enforced. 

Some existing upstream driver names might fail (for example, because of using upper-case chars...)

src/main/java/redis/clients/jedis/ClientSetInfoConfig.java
```diff
@@ -1,68 +1,132 @@
...
+ * This class supports two modes of operation:
+ * <ul>
+ * <li>Legacy mode: Using {@link #withLibNameSuffix(String)} for simple suffix customization</li>
+ * <li>Driver info mode: Using {@link #ClientSetInfoConfig(DriverInfo)} for structured driver
+ * information with upstream drivers</li>
+ * </ul>
+ * <p>
+ * For backward compatibility, {@link #getUpstreamDrivers()} returns the upstream drivers string
+ * when using driver info mode.
+ * @see DriverInfo
```
tishun, line 19:
```suggestion
 * @see {@link DriverInfo}
```

src/main/java/redis/clients/jedis/ClientSetInfoConfig.java
```diff
@@ -1,68 +1,132 @@
...
 
-  public final String getLibNameSuffix() {
-    return libNameSuffix;
+  /**
+   * Returns the driver information.
+   * @return the driver information
+   */
+  public DriverInfo getDriverInfo() {
+    return driverInfo;
   }
```
tishun:
```suggestion
   * @return {@code true} if CLIENT SETINFO is disabled, {@code false} otherwise
   */
  public boolean isDisabled() {
return disabled;
}

  /**
   * @return the driver information
   */
  public DriverInfo getDriverInfo() {
    return driverInfo;
}
```

src/main/java/redis/clients/jedis/ClientSetInfoConfig.java
```diff
@@ -1,68 +1,132 @@
...
 
+/**
+ * Configuration for CLIENT SETINFO command behaviour.
+ * <p>
+ * This class supports two modes of operation:
+ * <ul>
+ * <li>Legacy mode: Using {@link #withLibNameSuffix(String)} for simple suffix customization</li>
+ * <li>Driver info mode: Using {@link #ClientSetInfoConfig(DriverInfo)} for structured driver
+ * information with upstream drivers</li>
+ * </ul>
```
tishun, line 15:
```suggestion
 * <li>Advanced mode: Using {@link #withLibNameSuffix(String)} for advanced suffix customization, where the provided string is already preformatted according to the rules of the `CLIENT SETINFO` command</li>
 * <li>Simple mode: Using {@link #ClientSetInfoConfig(DriverInfo)} used when the command parameter will be built by the driver based on the {@link DriverInfo} provided</li>
 * </ul>
```

src/main/java/redis/clients/jedis/DriverInfo.java
```diff
@@ -0,0 +1,292 @@
+package redis.clients.jedis;
```
tishun, line 1: Missing license header
```suggestion
/*
 * Copyright 2025-Present, Redis Ltd. and Contributors
 * All rights reserved.
 *
 * Licensed under the MIT License.
 */
package redis.clients.jedis;
```

src/main/java/redis/clients/jedis/DriverInfo.java
```diff
@@ -0,0 +1,292 @@
...
+
+/**
+ * Immutable class representing driver information for Redis client identification.
+ * <p>
+ * This class is used to identify the client library and any upstream drivers (such as Spring Data
+ * Redis or Spring Session) when connecting to Redis. The information is sent via the
+ * {@code CLIENT SETINFO} command.
+ * <p>
+ * The formatted name follows the pattern: {@code name(driver1_vVersion1;driver2_vVersion2)}
+ * @see ClientSetInfoConfig
```
tishun, line 19:
```suggestion
 * @see {@link ClientSetInfoConfig}
```

src/main/java/redis/clients/jedis/DriverInfo.java
```diff
@@ -0,0 +1,292 @@
...
+   * Examples:
+   * <ul>
+   * <li>{@code "spring-data-redis_v3.2.0"} - single upstream driver</li>
+   * <li>{@code "spring-session_v3.3.0;spring-data-redis_v3.2.0"} - multiple upstream drivers</li>
+   * </ul>
+   * @return the formatted upstream drivers string, or {@code null} if no upstream drivers are set
+   */
+  public String getUpstreamDrivers() {
+    if (upstreamDrivers.isEmpty()) {
+      return null;
```
tishun: Not a [huge fan of null](https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare/).
Would returning "" be a problem here?

tishun (approved):
LGTM

src/main/java/redis/clients/jedis/ClientSetInfoConfig.java
viktoriya-kutsarova: No need to use link here, the @see tag creates a separate "See also" section with a link to DriverInfo class.

src/main/java/redis/clients/jedis/ClientSetInfoConfig.java
viktoriya-kutsarova: Good catch, sounds like some duplication otherwise.

src/main/java/redis/clients/jedis/DriverInfo.java
viktoriya-kutsarova: I noticed most files in the codebase don't have copyright headers, so I followed that pattern. I can definitely add one - would be helpful to know the project's convention for when headers are expected.

src/main/java/redis/clients/jedis/DriverInfo.java
viktoriya-kutsarova: No need to use link here, the @see tag creates a separate "See also" section with a link to ClientSetInfoConfig class.

src/main/java/redis/clients/jedis/DriverInfo.java
viktoriya-kutsarova: Sure, we can do that.

src/main/java/redis/clients/jedis/ClientSetInfoConfig.java
viktoriya-kutsarova: Let's discuss when you are back how we define advance/simple modes, if there is "legacy" mode, etc.

src/main/java/redis/clients/jedis/DriverInfo.java
viktoriya-kutsarova: Agreed, I will revert it to the previous validation so that we keep the driver backwards compatible.

src/main/java/redis/clients/jedis/DriverInfo.java
ggivo: @tishun @viktoriya-kutsarova 
My understanding is that a license header is required if we incorporate a third-party code with a different license; otherwise, it is covered by the MIT license at the project level

ggivo (approved):
LGTM

ggivo (dismissed):

ggivo (approved):

PR #4379, opened 2025-12-17
Add configurable retry delay for Sentinel reconnection (#2864)

### LINKED ISSUE #2864: Sleeping time not configurable when reconnecting to Sentinel
### Expected behavior

`JedisSentinelPool`  listen to the master messages via a `MasterListener`. If a connection with a Sentinel is lost, the listener tries to reconnect after a configurable sleeping time.

This can be an issue when several clients get disconnected together. When they will try to re-connect, all of them after 5000ms,  to Redis they will create a spike in new connections.

### Actual behavior

The retry time is hardcoded to [5000ms](https://github.com/redis/jedis/blob/890b54178278798870831e5f80eb12c909830334/src/main/java/redis/clients/jedis/JedisSentinelPool.java#L321). 

### Steps to reproduce:

1. Setup a Redis cluster using Sentinels
2. Connect the Santinels in your app
3. Stop your sentinels
4. Check the logs. In the logs there will be a message like: "Lost connection to Sentinel at <sentinelhost:sentinelport>. Sleeping 5000ms and retrying"

The `5000ms` is not configgurable.

ggivo:
Adds configurable delay strategy for Sentinel subscription retries with exponential backoff and jitter support. Prevents all clients from reconnecting simultaneously when connection is lost.

Default: 5000ms constant delay (backward compatible)
New: Configurable via RedisSentinelClient.Builder.sentinelReconnectDelay()

Closes #2864

### DIFF
--- a/src/main/java/redis/clients/jedis/builders/SentinelClientBuilder.java
+++ b/src/main/java/redis/clients/jedis/builders/SentinelClientBuilder.java
@@ -19,15 +19,13 @@ import redis.clients.jedis.util.JedisAsserts;
 public abstract class SentinelClientBuilder<C>
     extends AbstractClientBuilder<SentinelClientBuilder<C>, C> {
 
-  private static final Delay DEFAULT_RESUBSCRIBE_DELAY = Delay.constant(Duration.ofMillis(5000));
-
   // Sentinel-specific configuration fields
   private String masterName = null;
   private Set<HostAndPort> sentinels = null;
   private JedisClientConfig sentinelClientConfig = null;
 
   // delay between re-subscribing to sentinel nodes after a disconnection
-  private Delay sentinelReconnectDelay = DEFAULT_RESUBSCRIBE_DELAY;
+  private Delay sentinelReconnectDelay = SentineledConnectionProvider.DEFAULT_RESUBSCRIBE_DELAY;
 
   /**
    * Sets the master name for the Redis Sentinel configuration.
--- a/src/main/java/redis/clients/jedis/providers/SentineledConnectionProvider.java
+++ b/src/main/java/redis/clients/jedis/providers/SentineledConnectionProvider.java
@@ -34,7 +34,10 @@ public class SentineledConnectionProvider implements ConnectionProvider {
 
   private static final Logger LOG = LoggerFactory.getLogger(SentineledConnectionProvider.class);
 
-  static final Delay DEFAULT_RESUBSCRIBE_DELAY = Delay.constant(Duration.ofMillis(5000));
+  protected static final long DEFAULT_SUBSCRIBE_RETRY_WAIT_TIME_MILLIS = 5000;
+
+  public static final Delay DEFAULT_RESUBSCRIBE_DELAY = Delay
+          .constant(Duration.ofMillis(DEFAULT_SUBSCRIBE_RETRY_WAIT_TIME_MILLIS));
 
   private static final Sleeper DEFAULT_SLEEPER = Thread::sleep;
 


src/main/java/redis/clients/jedis/builders/SentinelClientBuilder.java
```diff
@@ -16,11 +19,16 @@
 public abstract class SentinelClientBuilder<C>
     extends AbstractClientBuilder<SentinelClientBuilder<C>, C> {
 
+  private static final Delay DEFAULT_RESUBSCRIBE_DELAY = Delay.constant(Duration.ofMillis(5000));
```
atakavci: why do we need both of these `DEFAULT_RESUBSCRIBE_DELAY` ?

src/main/java/redis/clients/jedis/providers/SentineledConnectionProvider.java
```diff
@@ -73,33 +82,62 @@ public SentineledConnectionProvider(String masterName, final JedisClientConfig m
...
+        Delay.constant(Duration.ofMillis(subscribeRetryWaitTimeMillis)));
+  }
+
+  /**
+   * @deprecated use
+   *             {@link #SentineledConnectionProvider(String, JedisClientConfig, Cache, GenericObjectPoolConfig, Set, JedisClientConfig, Delay)}
+   */
+  @Experimental
+  @Deprecated
+  public SentineledConnectionProvider(String masterName, final JedisClientConfig masterClientConfig,
```
atakavci, line 130: what is the reason we deprecated this one but not the one above ?

src/main/java/redis/clients/jedis/providers/SentineledConnectionProvider.java
```diff
@@ -73,33 +82,62 @@ public SentineledConnectionProvider(String masterName, final JedisClientConfig m
...
     this.masterPoolConfig = poolConfig;
 
     this.sentinelClientConfig = sentinelClientConfig;
-    this.subscribeRetryWaitTimeMillis = subscribeRetryWaitTimeMillis;
+    this.resubscribeDelay = resubscribeDelay;
+
+    this.sentinelConnectionFactory = sentinelConnectionFactory != null ? sentinelConnectionFactory
+        : defaultSentinelConnectionFactory();
+
+    this.sleeper = sleeper != null ? sleeper : Thread::sleep;
```
atakavci: nit: a static sleeper would do the job.

src/main/java/redis/clients/jedis/providers/SentineledConnectionProvider.java
```diff
@@ -340,4 +388,22 @@ public void shutdown() {
       }
     }
   }
+
+  protected SentinelConnectionFactory defaultSentinelConnectionFactory() {
+    return (node, config) -> new Jedis(node, config);
+  }
+
+  @FunctionalInterface
+  protected interface Sleeper {
```
atakavci: assuming this exist only to improve testability, is that correct ?

src/main/java/redis/clients/jedis/builders/SentinelClientBuilder.java
ggivo: what you mean by both?

src/main/java/redis/clients/jedis/providers/SentineledConnectionProvider.java
ggivo: Yes. Correct.
Used in test, to inject custom sleeper so that we don't need to have time based tests

src/main/java/redis/clients/jedis/builders/SentinelClientBuilder.java
ggivo: ah. I see your point. Removed the one from the builder, and  made the one in provider public

src/main/java/redis/clients/jedis/providers/SentineledConnectionProvider.java
ggivo: missed that one. Now is deprecated.

PR #4385, opened 2025-12-22
Remove hard-coded endpoints from the tests and rely solely on endpoints.json

uglide:
Goals:
1. Skip integration test if the required endpoint is not configured (missing in endpoints.json)
2. Move as many tests as possible to use Endpoints, including sentinel and cluster tests
3. Strip down the "local" testing flow to just a few endpoints to avoid duplicated configs, but allow agile development based on the source-build redis server. The new responsibility of the local/legacy test infrastructure is to support development, not to run all tests. **All advanced integration flows (TLS, sentinel, cluster, etc) should now be tested on Docker-based infra only to make failures easier to reproduce.**

### DIFF
--- a/src/test/java/redis/clients/jedis/scenario/ClusterTopologyRefreshIT.java
+++ b/src/test/java/redis/clients/jedis/scenario/ClusterTopologyRefreshIT.java
@@ -15,6 +15,10 @@ import java.util.HashSet;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicLong;
 
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.greaterThan;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.not;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
@@ -49,15 +53,11 @@ public class ClusterTopologyRefreshIT {
         .socketTimeoutMillis(RecommendedSettings.DEFAULT_TIMEOUT_MS)
         .connectionTimeoutMillis(RecommendedSettings.DEFAULT_TIMEOUT_MS).build();
 
-    ClusterConnectionProvider provider = new ClusterConnectionProvider(jedisClusterNode, config,
-        RecommendedSettings.poolConfig);
-    ClusterConnectionProvider spyProvider = spy(provider);
-
-    try (RedisClusterClient client = RedisClusterClient.builder().connectionProvider(spyProvider)
-        .maxAttempts(RecommendedSettings.MAX_RETRIES)
+    try (RedisClusterClient client = RedisClusterClient.builder().nodes(jedisClusterNode)
+        .clientConfig(config).maxAttempts(RecommendedSettings.MAX_RETRIES)
         .maxTotalRetriesDuration(RecommendedSettings.MAX_TOTAL_RETRIES_DURATION).build()) {
-      assertEquals(1, client.getClusterNodes().size(),
-        "Was this BDB used to run this test before?");
+      Set<String> initialNodes = client.getClusterNodes().keySet();
+      assertEquals(1, initialNodes.size(), "Was this BDB used to run this test before?");
 
       AtomicLong commandsExecuted = new AtomicLong();
 
@@ -100,7 +100,12 @@ public class ClusterTopologyRefreshIT {
         assertTrue(client.exists(String.valueOf(i)));
       }
 
-      verify(spyProvider, atLeast(2)).renewSlotCache(any(Connection.class));
+      Set<String> afterReshardNodes = client.getClusterNodes().keySet();
+      assertThat("After set should have more nodes than initial set", afterReshardNodes.size(),
+        greaterThan(initialNodes.size()));
+
+      boolean hasNewNode = afterReshardNodes.stream().anyMatch(n -> !initialNodes.contains(n));
+      assertThat("After set should have a node not in initial set", hasNewNode, is(true));
     }
   }
 


uglide:
Reminder: Re-run all tests when Maven Central is back online.

src/test/resources/env/.env
```diff
@@ -3,6 +3,6 @@ REDIS_STACK_VERSION=8.4.0
 CLIENT_LIBS_TEST_IMAGE=redislabs/client-libs-test
 REDIS_ENV_CONF_DIR=./
 REDIS_MODULES_DIR=/tmp
-REDIS_ENV_WORK_DIR=/tmp/redis-env-work
+REDIS_ENV_WORK_DIR=/home/imalinovskyi/playground/jedis_package_fixes/env
```
ggivo: Looks like unintentional change?

Makefile
```diff
@@ -415,88 +47,24 @@ appendonly no
...
-export REDIS_STABLE_CLUSTER_NODE1_CONF
-export REDIS_STABLE_CLUSTER_NODE2_CONF
-export REDIS_STABLE_CLUSTER_NODE3_CONF
 export REDIS_UDS
 export REDIS_UNAVAILABLE_CONF
 
 
 start: cleanup compile-module
+	# Simple local test env that provides only "standalone-0" endpoint and an instance listening on Unix socket
 	export TEST_ENV_PROVIDER=local
```
ggivo: Since we are removing the full local test env and replace it with a simple one covering only module/UDS, we could probably consider renaming 

`start-test-env`/`stop-test-env`/`test-on-docker`  targets also to common ones start/stop/test, and use different ones for simple test env.?

src/test/java/redis/clients/jedis/util/EnvCondition.java
```diff
@@ -0,0 +1,143 @@
...
+ * public class DockerOnlyTest {
+ *   // Tests in this class only run in Docker environment
+ * }
+ *
+ * &#64;ExtendWith(EnvCondition.class)
+ * &#64;SkipOnEnv("re")
+ * public class NotOnEnterpriseTest {
+ *   // Tests in this class are skipped in Redis Enterprise environment
+ * }
+ * </pre>
```
ggivo, line 36: looks like  `SkipOnEnv`& `EnabledOnEnv` are mutually exclusive?

If both are provided, `SkipOnEnv` takes precedence & `EnabledOnEnv` is ignored. 
It's good to have it documented or probably better to merge them in single annotation
```

// Enable only in oss
@ConditionalOnEnv(value = "oss", enabled = true)

// Skip in Redis Enterprise  
@ConditionalOnEnv(value = "re", enabled = false)

// Multiple environments
@ConditionalOnEnv(value = {"docker", "local"}, enabled = true)
```

current logic :
```
If @SkipOnEnv annotation exists → evaluateDisabledOnEnv() is called
If current env NOT in skip list → ENABLED (test runs)
@EnabledOnEnv is completely ignored when @SkipOnEnv is present
```

src/test/java/redis/clients/jedis/util/RedisVersionCondition.java
```diff
@@ -7,36 +7,54 @@
...
+  }
+
   @Override
   public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {
+    try {
+      ensureInitialized();
+    } catch (TestAbortedException e) {
+      return ConditionEvaluationResult.disabled(e.getMessage());
+    }
+
```
ggivo: This way I think we will hide misconfiguration issues and mark tests as disabled instead of failing. I think in case we are not able to resolve the Redis endpoint version for any reason test should fail.

Same as in `EnabledOnCommandCondition`

src/test/java/redis/clients/jedis/util/TestEnvUtil.java
```diff
@@ -6,18 +6,35 @@ public class TestEnvUtil {
     // Redis servers running inside docker
     public static final String ENV_DOCKER = "docker";
 
+    public static final String ENV_LEGACY = "local";
+
```
ggivo: nit:  'docker' (my bad) is not the most descriptive name, 
don't have much better proposal either, just rising the topic if we can come up with something more meaningfull
```
// Redis OSS running in Docker containers
public static final String ENV_OSS_DOCKER = "oss-docker";

// Local Redis OSS processes (legacy environment for UDS tests)
public static final String ENV_OSS_LOCAL = "oss-local";

// Remote Redis Enterprise cluster
public static final String ENV_REDIS_ENTERPRISE = "redis-enterprise";
```

src/test/java/redis/clients/jedis/commands/jedis/ClusterCommandsTest.java
```diff
@@ -77,19 +86,22 @@ public static void resetRedisAfter() {
   }
 
   public static void removeSlots() {
+    if (endpoint == null)
+      return;
```
ggivo, line 90: In which case, it can be null?

uglide: If the test is skipped in the `@BeforeAll` hook, this method is still called as a cleanup, and the endpoint will be `null`.

src/test/java/redis/clients/jedis/util/RedisVersionCondition.java
uglide: I'm afraid this is the only way to do this because of:

1. Class is instantiated and all @RegisterExtension are applied
2. If the class doesn't have endpoint explicitly defined in the `@beforeAll` hooks, then Conditions are checked
3. `ConditionEvaluationResult` doesn't have an option to skip the test

If we don't handle TestAbortedException here, the test will be marked as failed, but we know by the exception type that the endpoint doesn't exist in the `endpoints.json` file. 

The only way to avoid it is to mark each and every test with tags in "allow list" fashion like "env_oss_docker", "env_oss_source", etc but it's quite overhelming because it will be on the each test method level for most of the test classes. On top of that, it's better to do the "mass renaming" of integration tests first, to remote the "magic" in pom first.

I suggest going in stages. Once we merge this, we need to do the "mass renaming" + extraction of "RE not compatible tests" to separate test classes. Then we can do tags on a class level.

src/test/resources/env/.env
uglide: Fixed

Makefile
uglide: Renamed

src/test/java/redis/clients/jedis/util/EnvCondition.java
uglide: Done

src/test/java/redis/clients/jedis/util/TestEnvUtil.java
uglide: I think I came up with better naming, see [my commit](https://github.com/redis/jedis/pull/4385/commits/72789e50c3a4d90e8ed8e64c4da879a84bdde50e).

Makefile
```diff
@@ -82,28 +82,28 @@ stop:
...
+deploy: | start-local mvn-deploy stop-local
 
 mvn-deploy:
 	mvn clean deploy
 
 format:
 	mvn java-formatter:format
 
-release: | start mvn-release stop
+release: | start-local mvn-release stop-local
```
ggivo: Do we need to `start-local`/`stop-local` for package/deploy/release targets? 
Now that we have separated unit/integration tests, I think we are running only unit tests there?

Makefile
ggivo, line 1: Need  to update also docs
https://github.com/redis/jedis/blob/0bb78cedfb6d078258f36c2a689cbb18a456417a/.github/CONTRIBUTING.md#L18-L18

Makefile
uglide: I think we can remove those targets completely

Makefile
uglide: Done

src/test/java/redis/clients/jedis/util/RedisVersionCondition.java
uglide: Fixed

PR #4394, opened 2026-01-12
Fix flaky AccessControlListCommandsTest

ggivo:
Introduce ACLTestUtil to isolate test client ACL log entries from concurrent activity.
- filterByClientId: filters List<AccessControlLogEntry>
- filterBinaryByClientId: filters List<byte[]> from aclLogBinary()


uglide (approved):
Nice! Thanks!

Issue #4416, opened 2026-01-30
RedisClient builder does not use credentials from JedisClientConfig


dawnstrider:
**Jedis Version**: 7.2.1

**Description**

We are testing against a REDIS standalone server 8.x using ACLs to restrict access. Therefore we want to provide user and password to the Jedis client, respectively the replacement class, RedisClient.

When creating a RedisClient object using the builder and passing a JedisClientConfig object WITH user and password to the builder, subsequent calls to the RedisClient object do not consider the usage of the credentials and send all commands with doing a previous "AUTH ...." command.

**Steps**

This does **not** work:

```

        JedisClientConfig jedisConfig = DefaultJedisClientConfig.builder().user("mark").password("secret").build();
        RedisClient client = RedisClient.builder().clientConfig(jedisConfig).fromURI(redisURL).build();

        client.info();
```

This **works**:
```
        JedisClientConfig jedisConfig = DefaultJedisClientConfig.builder().user("mark").password("secret").build();
        HostAndPort hAp = new HostAndPort(host, port);
        ConnectionProvider p = new PooledConnectionProvider(hAp, jedisConfig);
        RedisClient client = RedisClient.builder().connectionProvider(p).build();

        client.info();
```

**Expectation**

I would expect the 1. example to work as well since I am providing a RedisClientConfig with a set of credentials which should be used on the default ConnectionProvider as well (it is also the pooled provider)

ggivo:
@dawnstrider 
Thanks for reporting it!

The issue happens because fromURI(redisURL) does two things:

1. Sets the host and port from the URI.
2. Overwrites the clientConfig with values derived from the URI. 

If you already provide a clientConfig via clientConfig(jedisConfig) before calling fromURI(...), it gets overwritten, which breaks the configuration.
        
Possible workaorund is to use `hostAndPort()` with .clientConfig(jedisConfig) or change the order, first fromURI(...) then  clientConfig()

```java
public static void main(String[] args) {

        String redisURL = "redis://localhost:6379";
        JedisClientConfig jedisConfig = DefaultJedisClientConfig.builder().user("default").password("foobared").build();

        // This will work
        // fromUri invoked first, clientConfig second
        System.out.println("Try .fromURI(redisURL).clientConfig(jedisConfig)");
        try ( RedisClient client = RedisClient.builder().fromURI(redisURL).clientConfig(jedisConfig).build()) {
            client.info();
            System.out.println("OK");
        } catch (Exception e) {
[...]

PR #4424, opened 2026-02-10
RedisClient builder does not use credentials from JedisClientConfig

ggivo:
Addresses:  #4416

### DIFF
--- a/src/main/java/redis/clients/jedis/RedisClient.java
+++ b/src/main/java/redis/clients/jedis/RedisClient.java
@@ -127,6 +127,7 @@ public class RedisClient extends UnifiedJedis {
    * @return a new {@link RedisClient} instance
    */
   public static RedisClient create(final URI uri) {
+    JedisAsserts.notNull(uri, "Redis URI must not be null");
     JedisAsserts.isTrue(JedisURIHelper.isValid(uri), "Invalid Redis URI");
 
     JedisClientConfig clientConfig = DefaultJedisClientConfig.builder(uri).build();
--- a/src/main/java/redis/clients/jedis/util/JedisURIHelper.java
+++ b/src/main/java/redis/clients/jedis/util/JedisURIHelper.java
@@ -99,7 +99,7 @@ public final class JedisURIHelper {
 
     String[] pathSplit = uri.getPath().split("/", 2);
 
-    return pathSplit.length > 1;
+    return pathSplit.length > 1 && !pathSplit[1].isEmpty();
   }
 
   /**
--- a/src/test/java/redis/clients/jedis/builders/ClientBuilderTest.java
+++ b/src/test/java/redis/clients/jedis/builders/ClientBuilderTest.java
@@ -217,7 +217,7 @@ class ClientBuilderTest {
 
   @Test
   @SuppressWarnings("deprecation")
-  void fromURI_WithCredentials_overridesClientConfigCredentials() {
+  void fromURI_withCredentials_overridesClientConfigCredentials() {
     JedisClientConfig config = DefaultJedisClientConfig.builder().user("olduser")
         .password("oldpass").build();
 


src/main/java/redis/clients/jedis/builders/StandaloneClientBuilder.java
```diff
@@ -62,23 +63,150 @@ protected void validateSpecificConfiguration() {
...
+   *             will be removed in Jedis 8.0.0. See {@link #fromURI(String)} for migration example.
    */
+  @Deprecated
   public StandaloneClientBuilder<C> fromURI(URI uri) {
-    this.clientConfig = DefaultJedisClientConfig.builder().user(JedisURIHelper.getUser(uri))
-        .password(JedisURIHelper.getPassword(uri)).database(JedisURIHelper.getDBIndex(uri))
-        .protocol(JedisURIHelper.getRedisProtocol(uri)).ssl(JedisURIHelper.isRedisSSLScheme(uri))
-        .build();
+    JedisAsserts.notNull(uri, "Redis URI must not be null");
+    JedisAsserts.isTrue(JedisURIHelper.isValid(uri), "Invalid Redis URI");
```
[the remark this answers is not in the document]
ggivo: This concern should be addressed separately

src/main/java/redis/clients/jedis/builders/StandaloneClientBuilder.java
```diff
@@ -62,23 +63,144 @@ protected void validateSpecificConfiguration() {
...
+
+    RedisProtocol uriProtocol = JedisURIHelper.getRedisProtocol(uri);
+    if (uriProtocol != null) {
+      configBuilder.protocol(uriProtocol);
+    }
+
+    if (JedisURIHelper.isRedisSSLScheme(uri)) {
+      configBuilder.ssl(true);
+    } else if (JedisURIHelper.isRedisScheme(uri)) {
+      configBuilder.ssl(false);
```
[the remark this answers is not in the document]
ggivo: Consistent with the currently documented behaviour in JedisClientCOnfig


```
  /**
   * {@link JedisClientConfig#isSsl()}, {@link JedisClientConfig#getSslSocketFactory()} and
   * {@link JedisClientConfig#getSslParameters()} will be ignored if
   * {@link JedisClientConfig#getSslOptions() this} is set.
   * @return ssl options
   */
  default SslOptions getSslOptions() {
    return null;
  }
```

src/main/java/redis/clients/jedis/DefaultJedisClientConfig.java
```diff
@@ -147,6 +150,56 @@ public static Builder builder() {
...
+   * <ul>
+   * <li>Credentials (username/password) if present in URI</li>
+   * <li>Database index if specified in path</li>
+   * <li>SSL enabled if scheme is "rediss"</li>
+   * <li>Protocol version if specified in query parameters</li>
+   * </ul>
+   * @param redisUri the Redis URI to extract settings from
+   * @return a new Builder pre-initialized from the URI
+   */
+  public static Builder builder(URI redisUri) {
```
uglide, line 170: Why not provide `public Builder from(URI redisUri)` similar to https://github.com/redis/jedis/blob/9225756f0434335c9b2662b38e3af9e03aeb2d5a/src/main/java/redis/clients/jedis/DefaultJedisClientConfig.java#L354, then we can do:

```
DefaultJedisClientConfig.builder().from(this.clientConfig).from(redisUri)
```
instead of duplicating the logic of populating clientConfig based on URI

ggivo: Yes, this is a valid option,

My thinking was that once we remove the deprecated method ` StandaloneClientBuilder.fromURI,` duplication will be gone

I also tried to avoid adding another “bulk” method that changes multiple settings at once. That’s why the URI logic currently only lives at builder creation time (initialize with some defaults), and from there you’re expected to use the existing builder methods explicitly.

tishun: Wouldn't this approach share the same problem until we remove the duplicate information from the `clientConfig` - specifically which one takes precedent?

IMHO we should work in that direction and never allow in the same builder to set both URI and `clientConfig`

src/main/java/redis/clients/jedis/builders/StandaloneClientBuilder.java
```diff
@@ -62,23 +63,144 @@ protected void validateSpecificConfiguration() {
...
+
+    // Override with URI values only if URI provides them (non-null)
+    String uriUser = JedisURIHelper.getUser(uri);
+    String uriPassword = JedisURIHelper.getPassword(uri);
+
+    // If URI provides credentials, we need to clear the credentialsProvider
+    // so that the new user/password values are used instead
+    if (uriUser != null || uriPassword != null) {
+      configBuilder.credentials(new DefaultRedisCredentials(uriUser, uriPassword));
+    }
```
tishun, line 186: So, when merging, we overwrite both `username` and `password` even if only one was provided.

I think this is really messy. I am not sure what is the right thing to do here. Both scenarios seems possible to me:

**Scenario 1**
The `clientConfig` already had a username (e.g. `default`) and no password. The URI provided was supposed to add the password for this specific instance, but has not username. As a result we end up with an empty username, but this was not expected

**Scenario 2**
The 'clientConfig` already had a username and password (e.g. `user1` and `pass1`) and the URI only provides the a username - `default`. In this case - obviously - we need to clear the old password.

In such ambiguous situations perhaps we need to just throw an exception and let the user handle it?

ggivo: I think for clarity we should treat credentials as a single unit.

If credentials are provided in the URI, they should fully replace the existing credentials in the clientConfig, similar to how other URI settings override config values.

Partially merging username/password creates ambiguity and surprising edge cases.

src/main/java/redis/clients/jedis/DefaultJedisClientConfig.java
ggivo: I believe so  (that's why I prefer not adding fromUri in `DefaultJedisClientConfig.Builder`)  but use it only as conivinience for initialise the builder on creation

PR #4425, opened 2026-02-10
ci: add reusable run-tests action and support custom test image/tag

Peter-Sh:
# Main changes

* Created a new reusable action `.github/actions/run-tests/action.yml` that is intended to be run by workflows in this and external repositories.
* Moved all the steps from `test-on-docker.yml` to the new action, with modifications allowing the use of a custom image tag instead of a baked-in version.

## What was changed in `test-on-docker.yml`:

* Uses the run-tests action to run the tests
* Allows running tests against a custom image tag (this is indicated in the run-name, if used)

## Docker Compose

* Added the `CLIENT_LIBS_TEST_IMAGE_TAG` variable to Docker Compose.
I'm using a separate `CLIENT_LIBS_TEST_IMAGE_TAG` variable to clearly separate regular versions set with REDIS_VERSION from image tags.

## Changes to `Makefile`

* If `CLIENT_LIBS_TEST_IMAGE_TAG` is present, then the version argument is ignored.

## Changes to the workflow logic

* `CLIENT_LIBS_TEST_IMAGE_TAG`, along with `CLIENT_LIBS_TEST_IMAGE`, is passed to make start if any or both are present. Then Docker Compose gets them from the environment.

```
make start CLIENT_LIBS_TEST_IMAGE=redislabsdev/client-libs-test CLIENT_LIBS_TEST_IMAGE_TAG=custom-1235-debian-amd64
```

* Removed `CLIENT_LIBS_IMAGE_PREFIX` - it wasn't used anywhere. The run-tests action allows setting a different image name using `CLIENT_LIBS_TEST_IMAGE` (not used in `test-on-docker.yml`).

### Minor changes
* Maven is started with the `-B` option to reduce output noise from progress messages a bit.
* Docker Compose is started with `--quiet-pull` to reduce noise.
* Docker Compose is started with the `--wait` option, which is a more correct way to wait for readiness.

### DIFF
--- a/src/main/java/redis/clients/jedis/MultiDbConfig.java
+++ b/src/main/java/redis/clients/jedis/MultiDbConfig.java
@@ -12,7 +12,9 @@ import redis.clients.jedis.exceptions.JedisConnectionException;
 import redis.clients.jedis.exceptions.JedisValidationException;
 import redis.clients.jedis.mcf.ConnectionFailoverException;
 import redis.clients.jedis.mcf.PingStrategy;
+import redis.clients.jedis.util.JedisAsserts;
 import redis.clients.jedis.mcf.HealthCheckStrategy;
+import redis.clients.jedis.mcf.InitializationPolicy;
 
 /**
  * Configuration class for multi-database Redis deployments with automatic failover and failback
@@ -644,6 +646,22 @@ public final class MultiDbConfig {
   private int delayInBetweenFailoverAttempts;
 
   /**
+   * Initialization policy that determines when the multi-database connection is ready to be
+   * returned based on the availability of individual database connections.
+   * <p>
+   * The policy is evaluated based on the completion status of database health checks, and the
+   * decision to continue waiting, succeed, or fail is based on the number of available, pending,
+   * and failed connections.
+   * </p>
+   * <p>
+   * <strong>Default:</strong> {@link InitializationPolicy.BuiltIn#MAJORITY_AVAILABLE}
+   * </p>
+   * @see InitializationPolicy
+   * @see #getInitializationPolicy()
+   */
+  private InitializationPolicy initializationPolicy;
+
+  /**
    * Constructs a new MultiDbConfig with the specified database configurations.
    * <p>
    * This constructor validates that at least one database configuration is provided and that all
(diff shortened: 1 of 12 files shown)


src/test/resources/env/docker-compose.yml
```diff
@@ -149,7 +149,7 @@ services:
   jedis-stack:
     sysctls:
       - net.ipv6.conf.all.disable_ipv6=1
-    image: "${CLIENT_LIBS_TEST_IMAGE}:${REDIS_STACK_VERSION:-${REDIS_VERSION}}"
```
ggivo: I think the correct precedence here is
1.	CLIENT_LIBS_TEST_IMAGE_TAG → highest priority
2.	REDIS_STACK_VERSION → fallback
3.	REDIS_VERSION → last fallback

image: "${CLIENT_LIBS_TEST_IMAGE}:${CLIENT_LIBS_TEST_IMAGE_TAG:-${REDIS_STACK_VERSION:-${REDIS_VERSION:-}}}"

src/test/resources/env/docker-compose.yml
```diff
@@ -149,7 +149,7 @@ services:
...
+    image: "${CLIENT_LIBS_TEST_IMAGE}:${REDIS_STACK_VERSION:-${CLIENT_LIBS_TEST_IMAGE_TAG:-${REDIS_VERSION:-}}}"
```
ggivo: I think the correct precedence should be 

	1.	CLIENT_LIBS_TEST_IMAGE_TAG → highest priority
	2.	REDIS_STACK_VERSION → fallback
	3.	REDIS_VERSION → last fallback

Something like

image: "${CLIENT_LIBS_TEST_IMAGE}:${CLIENT_LIBS_TEST_IMAGE_TAG:-${REDIS_STACK_VERSION:-${REDIS_VERSION:-}}}"


-------
We can also extract it and use  references
x-client-libs-image: &client-libs-image
image: "${CLIENT_LIBS_TEST_IMAGE}:${CLIENT_LIBS_TEST_IMAGE_TAG:-${REDIS_VERSION:-}}"

x-client-libs-stack-image: &client-libs-stack-image
image: "${CLIENT_LIBS_TEST_IMAGE}:${CLIENT_LIBS_TEST_IMAGE_TAG:-${REDIS_STACK_VERSION:-${REDIS_VERSION:-}}}"

and then
cluster-stable:
<<: *client-libs-image
...
jedis-stack:
<<: *client-libs-stack-image
...

Makefile
```diff
@@ -3,8 +3,6 @@ PATH := ./redis-git/src:${PATH}
 # Supported test env versions
 SUPPORTED_TEST_ENV_VERSIONS := 8.6 8.4 8.2 8.0 7.4 7.2 6.2
 DEFAULT_TEST_ENV_VERSION := 8.4
-REDIS_ENV_WORK_DIR := $(or ${REDIS_ENV_WORK_DIR},/tmp/redis-env-work)
```
ggivo, line 6: hm, what's the reason for removing REDIS_ENV_WORK_DIR

now after` make stop` / `make cleanup`, might end up with the old state from the previous run used by some containers

Peter-Sh: My bad, I thought by mistake it's not required since there is already `mkdir REDIS_ENV_WORK_DIR` in the action, but missed the point of cleaning up.

src/test/resources/env/docker-compose.yml
Peter-Sh: Implemented with anchors, looks much better.

PR #4432, opened 2026-02-13
Remove cluster-stack endpoint

uglide:
The cluster consumes significant VM resources, so it's better to use the cluster-stable endpoint, since the test is guarded by a version check.


ggivo:
The intention was to use the cluster-stack endpoint for module tests in the cluster environment.
Including search module tests for older envs 7.x

PR #4435, opened 2026-02-18
[automatic failover] Add Runtime Weight Management for MultiDbClient Endpoints

atakavci:
### Summary
Adds the ability to get and set endpoint weights at runtime in `MultiDbClient`.

### Changes
- Added `getWeight(Endpoint)` and `setWeight(Endpoint, float)` methods to `MultiDbClient`
- Made `weight` field mutable in `MultiDbConnectionProvider.Database` with a setter
- Added unit tests for weight get/set operations

### Use Case
Allows dynamic adjustment of endpoint priority for failover operations without recreating the client.


atakavci:
> explicitly trigger re-selection of the best matching DB after the weights change

sound like a good improvement. i got other things in mind with the API; like improving accessibility with `MultiDbClient` instance for what is exposed from connection provider. 
Will work on those in a seperate PR. thanks @ggivo

PR #4436, opened 2026-02-19
Fix resp2 parsing of LibraryInfo

a-TODO-rov:
Closes #4436

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
 	<packaging>jar</packaging>
 	<groupId>redis.clients</groupId>
 	<artifactId>jedis</artifactId>
-	<version>7.3.0-SNAPSHOT</version>
+	<version>7.4.0-SNAPSHOT</version>
 	<name>Jedis</name>
 	<description>Jedis is a blazingly small and sane Redis java client.</description>
 	<url>https://github.com/redis/jedis</url>


src/main/java/redis/clients/jedis/resps/LibraryInfo.java
```diff
@@ -20,10 +20,6 @@ public class LibraryInfo {
   private final List<Map<String, Object>> functions;
   private final String libraryCode;
 
-  public LibraryInfo(String libraryName, String engineName, List<Map<String, Object>> functions) {
-    this(libraryName, engineName, functions, null);
-  }
-
```
ggivo, line 26: This change is backward incompatible. As it might be used externally, we should mark it as @Deprecated in Jedis 7.4 and schedule its removal for the next major version

/**
 * @deprecated Since 7.4. This method will be removed in the next major release.
 */

src/main/java/redis/clients/jedis/resps/LibraryInfo.java
```diff
@@ -54,37 +50,44 @@ public LibraryInfo build(Object data) {
       List list = (List) data;
       if (list.isEmpty()) return null;
 
+      String[] libname = { null };
+      String[] enginename = { null };
+      String[] librarycode = { null };
+      List[] functions = { null };
```
ggivo: Any reason not using mutable complex object instead?
```
    private static class LibraryInfoDTO {
        private String libraryName;
        private String engineName;
        private String libraryCode;
        private List<Object> functions;
    }
...


    private void processField(Object key, Object value, LibraryInfoDTO dto) {
        String field = BuilderFactory.STRING.build(key);

        switch (field) {
            case "library_name":
                dto.libraryName = BuilderFactory.STRING.build(value);
```

src/test/java/redis/clients/jedis/BuilderTest.java
ggivo, line 1: Also, this touches the functionality of LibraryInfo.LIBRARY_INFO builder.
I suggest creating a dedicated unit test class, `LibraryInfoTest` and moving the respective tests there

PR #4439, opened 2026-02-20
Fix Connection Close Exception Handling for commons-pool2 2.13.1

atakavci:
### Summary
Here the [bump attempt](https://github.com/redis/jedis/pull/4409)  for org.apache.commons:commons-pool2 from 2.12.1 to 2.13.1.
Those tests failing during connection close is due to behavioral change - the way invalidateObject tries addObject - in [GenericObjectPools](https://github.com/apache/commons-pool/blob/34a857ceb70198ce6d5ff7ef171d12d18f6a3355/src/main/java/org/apache/commons/pool2/impl/GenericObjectPool.java#L939) implementation. 
This PR suppresses those exceptions during connection close to accommodate the behavioral change.

### Context
Starting with `commons-pool2` 2.13.1, `GenericObjectPool.invalidateObject()` attempts to replace invalidated instances, which can fail when `ConnectionFactory.makeObject()` throws an exception. This change ensures the original exception is preserved while suppressing any secondary exceptions from connection cleanup.

---

 ⚠️ The _version  bump_ itself will also result propagating the behavioral change to Jedis users.. So it will be a **breaking change** for Jedis and i am not sure this is the right approach in this PR.
 I am willing to follow up with the commons-pool team to explore whether a built-in option or a solution can be introduced in the pool implementation. Will update here any progress with commons-pool team.

---

### Changes
- Refactored `MultiDbCommandExecutor.handleExecuteCommand()` to properly suppress exceptions during `connection.close()`
- Added `closeAndSuppress()` helper to capture initial exceptions and suppress close-time exceptions
- Updated `UnavailableConnectionTest` to handle expected exception from pool invalidation

---

> [!NOTE]
> **Medium Risk**
> Changes exception propagation around `Connection.close()` in `MultiDbCommandExecutor`, which can affect what callers see and when failover triggers. Limited scope but touches runtime error-handling paths and adds new wrapping/suppression behavior.
> 
> **Overview**
> Adjusts `MultiDbCommandExecutor` to **preserve the original command failure** when `connection.close()` also throws (a behavior that can occur with commons-pool2 2.13.1 pool invalidation), by capturing the command exception, suppressing close-time exceptions onto it, and rethrowing consistently (wrapping non-runtime causes in `JedisException`).
> 
> Refactors failover-related exception creation into helpers (`wrapIfFailover`, `isFailDuringFailover`) and updates `UnavailableConnectionTest` to tolerate the new close-time exception by catching `JedisException` and asserting on the expected cause message.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2f13dd7ae5a3d25ef0a43cdc698a578f2de06bfd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tishun:
Okay, I am missing a lot of stuff here:
* what is the reason to bump in the first place - is there some security / functional regression fix we are trying to consume, or just keeping the dependencies up-to-date
* not sure what the old flow used to be, compared to the new one, can we have a specific example that explains it?
* if we have a behavioral regression we can use the 8.x major release - @uglide please have this in mind, because if we release 8.x without this change then we might have to wait for 9.x

ggivo:
@tishun 
> * what is the reason to bump in the first place - is there some security / functional regression fix we are trying to consume, or just keeping the dependencies up-to-date

Keeping the dependencies up to date -> https://github.com/redis/jedis/pull/4409

> * not sure what the old flow used to be, compared to the new one. Can we have a specific example that explains it?

With the new behaviour of the pool, each time a connection is returned to the pool (e.g., after each command execution). If the number of idle connections is less than the minimum idle connection, it will try to create a new connection in the invoking thread, which can result in an error during creation and be propagated to the calling thread.

tishun:
Considering that:
- there is no functional or security reason to consume the new version
- there is a behavioral regression when consuming this new version

→ **I think we should consider this bump only as part of a major release.**

Besides that I have no additional concerns, if this is how the commons pool is going forward then we must consume their new approach at some point.

tishun (approved):

PR #4441, opened 2026-02-20
fix: Fix out-of-order responses in manual transactions with commands before MULTI

ggivo:
## PR Title
 
```
Fix out-of-order responses in manual transactions with commands before MULTI
```

## Problem

When using manual transactions (`doMulti=false`) and executing commands before calling `multi()`, responses were returned out of order. This occurred because all commands were being queued as pipelined commands, regardless of whether the transaction had started.

```markdown

 
**Example of the issue:**
```java
try (AbstractTransaction tx = jedis.transaction(false)) {
    Response<String> beforeMulti = tx.get("key");  // Queued, not executed
    tx.multi();
    Response<String> inMulti = tx.set("key", "value");  // Queued
    tx.exec();  // Both responses returned here, out of order
    beforeMulti.get();  // Wrong value!
}
```

## Solution

Modified `Transaction.appendCommand()` to check the `inMulti` flag:
- **Before `multi()`**: Commands are executed immediately and wrapped in `Response.DecodedResponse`
- **After `multi()`**: Commands are queued as pipelined commands (existing behavior)

This ensures commands before the transaction get immediate responses while maintaining proper transaction semantics for commands within the transaction block.

### Core Changes
- **`Transaction.appendCommand()`**: Added `inMulti` flag check to determine execution strategy (execute immediately vs queue in transaction)
- **`Transaction.execute()`**: New private method to execute commands immediately
- **`Response.DecodedResponse`**: New inner class to wrap pre-decoded responses
- **`Response.of()` / `Response.error()`**: Factory methods for creating decoded responses

## Testing
`RedisClientTransactionIT.transactionManualWithCommandsBeforeMulti()` validates the fix

## Backward Compatibility

✅ No breaking changes  
✅ Existing transaction behavior unchanged  
✅ Only affects manual transactions with commands before `multi()`
```

<!-- CURSOR_SUMMARY -->
---

> [!NOTE]
> **Medium Risk**
> Changes transaction execution semantics by running pre-`multi()` commands immediately and introduces a new `Response` variant, which could affect edge cases around error propagation and response handling in transaction flows.
> 
> **Overview**
> Fixes manual transactions (`transaction(false)`) so commands issued *before* `multi()` are executed immediately instead of being queued, preventing out-of-order `Response` values when `exec()` later runs.
> 
> Adds `Response.of()`/`Response.error()` backed by a new `Response.DecodedResponse` to return already-decoded results (or immediate `JedisDataException`) for those pre-`multi()` commands, while preserving existing queued behavior once `inMulti` is true.
> 
> Refactors and expands coverage by moving transaction-related tests into a new `RedisClientTransactionIT` (including new response-handling assertions) and adds a new `Transactions/Multi` user-guide doc page wired into `mkdocs.yml`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 32bed64dd4b784e5256c33f4f921564c083a0f16. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>
<!-- /CURSOR_SUMMARY -->

### DIFF
--- a/docs/transactions-multi.md
+++ b/docs/transactions-multi.md
@@ -97,11 +97,14 @@ jedis.close();
 
 The `WATCH` command monitors keys for changes. If any watched key is modified before `EXEC`, the transaction is aborted and `exec()` returns `null`.
 
+**Important:** `WATCH` must be executed through the transaction object to ensure it uses the transaction's dedicated connection.
+
 ```java
 try (AbstractTransaction tx = jedis.transaction(false)) {
+    // WATCH must be on the transaction's dedicated connection
     tx.watch("counter");
 
-    // Read current value (executed immediately)
+    // Read current value - can use the client directly
     String current = jedis.get("counter");
     int newValue = Integer.parseInt(current) + 1;
 
@@ -129,6 +132,10 @@ try (AbstractTransaction tx = jedis.multi()) {
 }
 ```
 
+**Important:** `WATCH` must be executed through the transaction object to ensure it uses the transaction's dedicated connection.
+
+
+
 ## Transaction Completion
 
 Complete a transaction by calling either:
@@ -139,4 +146,6 @@ Complete a transaction by calling either:
 ### Automatic Cleanup
 
 When using try-with-resources, `close()` automatically sends `DISCARD` (if in `MULTI` state) or `UNWATCH` (if in `WATCH` state) to ensure the connection is returned to the pool in a clean state.
+
+
 
--- a/src/main/java/redis/clients/jedis/Response.java
+++ b/src/main/java/redis/clients/jedis/Response.java
@@ -77,7 +77,7 @@ public class Response<T> implements Supplier<T> {
 
   static class DecodedResponse<T> extends Response<T> {
 
-    DecodedResponse(T response, JedisDataException exception) {
+   private DecodedResponse(T response, JedisDataException exception) {
       super(null);
       this.exception = exception;
       this.response = response;

### CODE REVIEW

File: pom.xml [hunk 1]
```diff
@@ -567,6 +567,7 @@
 						<include>**/*Hotkeys*.java</include>
 						<include>**/TestDataUtil*.java</include>
 						<include>**/*JedisClientConfig*.java</include>
+
```
atakavci, line 570: nit

src/main/java/redis/clients/jedis/Response.java
```diff
@@ -75,4 +75,34 @@ public String toString() {
     return "Response " + builder.toString();
   }
 
+  static class DecodedResponse<T> extends Response<T> {
+
+    DecodedResponse(T response, JedisDataException exception) {
```
atakavci: nit: never gonna be both at the same time

docs/transactions-multi.md
```diff
@@ -0,0 +1,142 @@
...
+    tx.exec();
+}
+```
+
+## Transaction Completion
+
+Complete a transaction by calling either:
+
+- **`exec()`** - Executes all queued commands atomically and returns a `List<Object>` with the results
+- **`discard()`** - Discards all queued commands without executing them
```
atakavci, line 144: since the example felts like using `transaction()` is an isolated way to issue commands, may be we should mention that all issued WATCH commands -whether they are issued via transaction object or not- would be discarded on a `discard`.
Indeed this is an extreme case, but worth to mention to make sure users understand the semantics of transactional commands.

ggivo: actually WATCH commands issued through the client directly (not using the transaction ) might land on a different connection, and will not get discarded on transaction DISCARD, let me try to clarify it in the docs,

src/main/java/redis/clients/jedis/Response.java
ggivo: kept both, but will make  it private so we use static factory methods with single param instead

src/test/java/redis/clients/jedis/commands/unified/client/RedisClientTransactionIT.java
```diff
@@ -0,0 +1,223 @@
...
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@ParameterizedClass
+@MethodSource("redis.clients.jedis.commands.CommandsTestsParameters#respVersions")
+public class RedisClientTransactionIT extends UnifiedJedisCommandsTestBase {
```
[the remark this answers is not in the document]
ggivo: Test with sufix IT  does not rqure integration tag

pom.xml
ggivo: missed that one, will fix with next change to pom.xml

PR #4411, opened 2026-01-21
Introduce request & response policy support

uglide:
This PR introduces support for **request/response policies** in Redis cluster command execution, enabling proper handling of keyless commands and multi-shard operations. The changes implement Redis's command tips specification for cluster deployments, allowing the Jedis client to intelligently route commands to appropriate nodes and aggregate responses based on command-specific policies.

Fixes: #990

---

> [!NOTE]
> **High Risk**
> High risk because it rewires Redis Cluster command routing/execution (including keyless and broadcast behaviors), adds response aggregation logic, and changes how hash slots are tracked for pipeline/node selection.
> 
> **Overview**
> Adds **Redis Cluster command tip support** by extending `CommandFlagsRegistry` with `RequestPolicy`/`ResponsePolicy` and updating the static registry generator output to include these policies.
> 
> Reworks cluster execution to route commands by policy in `ClusterCommandExecutor` (keyless commands use round-robin connections, `ALL_SHARDS`/`ALL_NODES` are broadcast with aggregated replies, and new multi-shard execution aggregates per-slot sub-commands), introducing new connection resolvers and reply aggregators plus clearer multi-node error reporting (`JedisClusterOperationException` now carries node info; new aggregation exceptions).
> 
> Introduces **multi-shard implementations** for cross-slot commands by splitting arguments by hash slot in `ClusterCommandObjects` and overriding `RedisClusterClient` methods like `del`/`exists`/`mget`/`mset`/`unlink`/`touch`/`msetex` to transparently execute across shards while preserving input order.
> 
> Refactors key tracking by removing `ClusterCommandArguments` and making `CommandArguments` record keys + cached hash slots (`addHashSlotKey(s)`), updates pipeline routing/validation to reject commands that can’t be routed to a single slot, and removes the old broadcast/round-robin config path from `UnifiedJedis`. Also expands CI branch triggers (`topic/**`) and updates `pom.xml` include patterns for coverage/validation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3d286f8adf5f69afa9223a28824ead4308e5ab59. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #4444, opened 2026-02-23
refactor: unify retry/redirect logic for keyless and keyed commands

ggivo:
Extract connection resolution strategy to share retry, follow-redirect, and read intent handling between keyless and keyed command execution.

- Add ConnectionResolver interface with SlotBased and RoundRobin implementations
- Add ConnectionIntent enum (READ/WRITE) for connection type selection
- Consolidate retry/redirect logic in doExecuteCommand()
- Eliminate code duplication from executeKeylessCommand()


src/main/java/redis/clients/jedis/ReadFrom.java
```diff
@@ -0,0 +1,5 @@
+package redis.clients.jedis;
+
+public enum ReadFrom {
+  ANY, UPSTREAM, REPLICA
```
uglide, line 4: Not sure where "Upstream" is comming from, but should be "PRIMARY" :)

src/main/java/redis/clients/jedis/executors/ClusterCommandExecutor.java
```diff
@@ -22,13 +22,27 @@ public class ClusterCommandExecutor implements CommandExecutor {
 
   private final Logger log = LoggerFactory.getLogger(getClass());
 
+  /**
+   * Connection resolver used for keyed commands, to acuire connection based on slot
```
uglide, line 26: Typo in "acuire"

src/main/java/redis/clients/jedis/executors/ClusterCommandExecutor.java
```diff
@@ -51,13 +70,118 @@ public ClusterCommandExecutor(ClusterConnectionProvider provider, int maxAttempt
...
+    this.replicaOnlyConnectionResolver = new ConnectionResolver.ReplicaOnlyConnectionResolver(provider);
   }
 
   @Override
   public void close() {
     this.provider.close();
   }
 
+
+  public enum ConnectionIntent {
```
uglide, line 86: Prob makes sense to make part of ConnectionResolver interface, since there is no usage outside of it.

src/main/java/redis/clients/jedis/executors/ClusterCommandExecutor.java
```diff
@@ -51,13 +70,118 @@ public ClusterCommandExecutor(ClusterConnectionProvider provider, int maxAttempt
...
+    READ, WRITE
+  }
+
+  public interface ConnectionResolver {
```
uglide, line 90: Let's move it to subpackage redis/clients/jedis/executors/cluster as standalone file, WDYT?

src/main/java/redis/clients/jedis/executors/ClusterCommandExecutor.java
```diff
@@ -254,14 +334,15 @@ private <T> T doExecuteCommand(CommandObject<T> commandObject, boolean toReplica
...
           connection = provider.getConnection(redirect.getTargetNode());
           if (redirect instanceof JedisAskDataException) {
             // TODO: Pipeline asking with the original command to make it faster....
             connection.executeCommand(Protocol.Command.ASKING);
           }
         } else {
-          connection = toReplica ? provider.getReplicaConnection(commandObject.getArguments())
-              : provider.getConnection(commandObject.getArguments());
+          ConnectionIntent intent = getIntent(commandObject);
+          connection = resolver.resolve(commandObject, intent);
```
uglide, line 345: Something is wrong about the interface of `ConnectionResolver`. We pass both `commandObject` and `intent`, but then here we are using `commandObject` to deduct the `intent`.  Why do we need `intent` in the resolve method?

Seems like `intent` should be removed and `getIntent()` moved into the interface as the default implementation.

src/main/java/redis/clients/jedis/executors/ClusterCommandExecutor.java
```diff
@@ -179,72 +303,28 @@ public final <T> T executeCommand(CommandObject<T> commandObject) {
...
-        "No more cluster attempts left.");
-    maxAttemptsException.addSuppressed(lastException);
-    throw maxAttemptsException;
+  private <T> T executeKeyedCommand(CommandObject<T> commandObject) {
+    return doExecuteCommand(commandObject, slotBasedConnectionResolver);
   }
 
-  public final <T> T executeCommandToReplica(CommandObject<T> commandObject) {
-    return doExecuteCommand(commandObject, true);
+  private ConnectionIntent getIntent(CommandObject<?> command) {
```
uglide, line 323: See my comment below about the `resolve` API

src/main/java/redis/clients/jedis/executors/ClusterCommandExecutor.java
```diff
@@ -51,13 +70,118 @@ public ClusterCommandExecutor(ClusterConnectionProvider provider, int maxAttempt
...
     this.provider.close();
   }
 
+
+  public enum ConnectionIntent {
+    READ, WRITE
+  }
+
+  public interface ConnectionResolver {
+    Connection resolve(CommandObject<?> cmd, ConnectionIntent intent);
```
uglide, line 91: See my comment on the `resolve` API in the block where its used.

uglide (changes requested):
Overall looks promising, but needs a clean-up.

src/main/java/redis/clients/jedis/ReadFrom.java
ggivo: UPSTREAM comes from lettuce, tryed to unify them ,agree PRIMARY makes more sense

src/main/java/redis/clients/jedis/executors/ClusterCommandExecutor.java
ggivo: make sense, will do!

src/main/java/redis/clients/jedis/executors/ClusterCommandExecutor.java
ggivo: Agree, but I suggest keeping it internal for now (e.g package private)

src/main/java/redis/clients/jedis/executors/ClusterCommandExecutor.java
ggivo: I was thinking that at some point to extend the ConnectionProvider interface with capabilities to request read/write capable connection.
`getConnection(ConnectionIntent intent) ` for example, if you need a dedicated pipeline/transaction connection. 

But till we reach there, I suggest tokeep it package private. What do you think?

uglide:
Partially incorporated in https://github.com/redis/jedis/pull/4411 except for ReadFrom support

PR #4473, opened 2026-03-25
[maintenance events] Support maintenance PUSH notification 

ggivo:
### Description

This PR enhances Jedis and UnifiedJedis with improved support for **RESP3 push messages** and introduces a configurable option for **relaxed timeout behavior**.

These changes are part of a broader effort to support more robust and controlled upgrade and maintenance scenarios. Specifically, they lay the groundwork for:
- Temporarily relaxing timeouts during planned maintenance windows.
- Gracefully re-binding pooled connections to updated endpoints without disrupting active operations.

This PR aims to provide support for reading arbitrary PUSH notifications  (such as maintenance events) and serve as an initial commit for introducing support for timeout relaxation on miantenace events

### Backround 
In the current state of the client:
- 	Jedis supports only known **Push** notifications like **invalidate**.  Push notifications are supported using dedicated client-side caching (CSC) connection  (CacheConnection)
- 	This leads to runtime errors if `CLIENT TRACKING ON `is enabled in a context other than CSC, or if other **RESP3** push events are received.

### Changes Introduced

- Added support for **processing and discarding all RESP3 push messages**, preventing errors in non-CSC scenarios.
- **Pub/Sub push messages** are still handled as before and **propagated to the client** for backward compatibility.
- Added a **configurable option to enable relaxed timeout behavior**, intended for usage during maintenance events or when expecting slow responses from the server.

---

> [!NOTE]
> **High Risk**
> Changes core connection lifecycle, pooling return/eviction, socket connect targets, and read timeouts—any bug can misroute traffic or drop pooled connections during maintenance.
> 
> **Overview**
> Adds **Smart Client Handoff** support for Redis maintenance RESP3 push events (`MOVING`, `MIGRATING`, `FAILING_OVER`, and their terminators), wired through pooled connections when `MaintenanceNotificationsConfig` is enabled or in AUTO mode.
> 
> Pooled clients negotiate `CLIENT MAINT_NOTIFICATIONS ON` (RESP3), register a maintenance push consumer, and run a **chained timeout model** so reads can temporarily use relaxed socket timeouts during migration/failover and restore them on `MIGRATED`/`FAILED_OVER` or a configurable max window. **MOVING** drives proactive endpoint rebind: post-DNS connect remapping, pool eviction on handoff, returning affected connections as broken, and expiring receivers that saw the event.
> 
> Configuration is exposed on **standalone** builders (`RedisClient` via `StandaloneClientBuilder`, default AUTO) and **per-database** in `MultiDbConfig` (default disabled). `ConnectionPool`, `ConnectionFactory`, `DefaultJedisSocketFactory`, `TrackingConnectionPool`, and `PooledConnectionProvider` thread the maintenance controller through factory construction.
> 
> `Connection` timeout handling is refactored around `TimeoutSource` overrides (with redundant `setSoTimeout` avoided via caching), blocking vs non-blocking reads, and optional connection expiry on pool return. Push consumer chains gain **remove** support for failed handshakes.
> 
> Broad **sch** unit/integration tests, shared abstract suites, and a manual `MaintenanceEventsExample` document handshake, relaxation, rebind, and multi-DB behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0b944a1c5620e6551277ec616fc24c3e24685752. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -666,4 +721,157 @@ protected boolean isTokenBasedAuthenticationEnabled() {
...
+      if (proactiveRebindEnabled) {
+        rebindRequested = true;
+      }
+      relaxTimeouts();
+
+      // notify owning pool to rebind factory and clear idle connections
+      if (memberOf != null) {
+        memberOf.onMoving(rebindTarget);
+      }
+    }
```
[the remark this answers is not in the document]
ggivo: by design

The special value none indicates that the MOVING message doesn’t need to contain an endpoint. Instead it contains the value null then. In such a corner case, the client is expected to schedule a graceful reconnect to its currently configured endpoint after half of the grace period that was communicated by the server (as part of the MOVING notification) is over.

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -114,6 +131,29 @@ protected Connection(Builder builder) {
     this.clientConfig = builder.getClientConfig();
   }
 
+  protected void initPushConsumers(JedisClientConfig config) {
```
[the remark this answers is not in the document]
ggivo: InitializeFromClientConfig sis expected to be called before the connection becomes operational, when using JedisClientConfig

src/main/java/redis/clients/jedis/util/NumberUtils.java
```diff
@@ -0,0 +1,13 @@
+package redis.clients.jedis.util;
+
+public final class NumberUtils {
+
+  public static int safeToInt(long millis) {
```
atakavci, line 5: if we like to keep it as utility across lib, lets make it aware of negative values.

src/main/java/redis/clients/jedis/ConnectionPool.java
```diff
@@ -89,4 +93,30 @@ private void postAuthentication(Token token) {
...
+   *
+   * @param target the new host and port to rebind to
+   */
+  void onMoving(HostAndPort target) {
+    if (target == null) {
+      return;
+    }
+
+
+    HostAndPort previous = rebindTarget.getAndSet(target);
```
atakavci, line 111: the way we are using `target` suggests it should be thread-safe, but there is no guard with calling `RebindAware.rebind`

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -211,22 +214,39 @@ private static List<KeyValue> processMapKeyValueReply(final RedisInputStream is)
...
-    Object unhandledPush = readPushes(is, cache, false);
-    return unhandledPush == null ? process(is) : unhandledPush;
+  public static Object read(final RedisInputStream is, PushConsumer pushConsumer) {
+    // read until we have a non-push event,
+    // or push-event is requested to be propagated propagated to caller as result
+    Object reply;
+    do {
+      reply = process(is, pushConsumer);
+
+    } while (isPush(reply) && !((PushConsumerContext) reply).isReturnToCaller());
```
atakavci: why do we defer the loop to this later stage ? isn't it more clear to run consecutive push messages inside a single call to `process` if we know we are not going to propagate the result anyway? 
lets think about shifting it to another spot where we already know (without `instanceof`) its a out-of-band msg or not ?

src/main/java/redis/clients/jedis/PushConsumerChain.java
```diff
@@ -0,0 +1,133 @@
...
+  @Override
+  public void accept(PushConsumerContext context) {
+    if (consumers.isEmpty()) {
+      return;
+    }
+
+    for (PushConsumer consumer : consumers) {
+      consumer.accept(context);
+    }
+  }
```
atakavci: as we discussed offline, this approach is more like a broadcast fan out. 
each item hits all consumers whether or not they r relevant.
i suggest a more constrained approach, like a conditional chaining of consumers; where order does matter as well as individual consumers can lock out others remaining in the chain and own the message (or even opt to mutate/swap the context).

atakavci: to better represent what i have in mind;

```
--> ConsumerChain
Context    -->   Consumer1 -->    Consumer2    -->  Consumer3   -->   Completed
                     │              │
                  (mutate &      (short-circuit,
                   pass on)       propagate back to caller)
```
	
- Consumer1 can also decide to mutate/reframe the context to pass on
- Consumer2 can short circuit and decide to propagate to caller or drop it.
- When chain completed, its always a NoOp, nothing to propagate, context/msg just dies

Each bit of necessary computing or processing should be a consumer instance, including pub/sub (depending on RESP2/RESP3 it either consumes itself or propagates back)

src/main/java/redis/clients/jedis/PushConsumer.java
```diff
@@ -0,0 +1,25 @@
...
+   * <li>Process the message and prevent it from being returned by not calling
+   * {@link PushConsumerContext#setReturnToCaller(boolean)}</li>
+   * </ul>
+   * Following handlers in the chain can override the decision of previous handlers. By default, if
+   * no handler sets the flag, the message will not be returned to the caller.
+   * @see PushConsumerContext#setReturnToCaller(boolean)
+   * @see PushConsumerChain
+   * @param context The context of the push message
+   */
+  void accept(PushConsumerContext context);
```
atakavci: just a suggestion; to return a context.
this will create a level of ownership and more control on context/message for a consumer.

atakavci (commented):
Here a set of comments and observations i have.
This round mostly focused on push messages.
Planning to go for another round tomorrow , and will try to cover relaxedTimeout this time.

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -84,18 +90,50 @@ public static Builder builder() {
...
+
   private boolean broken = false;
   private boolean strValActive;
   private String strVal;
   protected String server;
   protected String version;
   private AtomicReference<RedisCredentials> currentCredentials = new AtomicReference<>(null);
   private AuthXManager authXManager;
+  private boolean isBlocking = false;
+  private boolean isRelaxed = false;
```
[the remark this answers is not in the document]
ggivo: The Connection class in general, is considered not thread-safe with some exceptions

src/main/java/redis/clients/jedis/MaintenanceEvent.java
```diff
@@ -0,0 +1,232 @@
...
+    if (t == null) {
+      return null;
+    }
+    List<Object> c = msg.getContent();
+    if (Arrays.equals(t, PushMessageTypes.MOVING_BYTES)) { // [MOVING, seq, time_s, host:port]
+      if (!hasLongs(c, 1, 2)) {
+        return malformed(msg);
+      }
+      HostAndPort target = parseHostPort(c, 3); // logs on failure
+      return target != null ? new MovingEvent((Long) c.get(1), (Long) c.get(2), target) : null;
```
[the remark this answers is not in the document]
ggivo: NONE support will be implemented with separaete PR

src/main/java/redis/clients/jedis/MaintenanceEventController.java
```diff
@@ -0,0 +1,188 @@
...
+   * peer is the active rebind's affected node and the window is still open; else returns null (no
+   * remap).
+   */
+  @Override
+  public SocketAddress getSocketAddress(SocketAddress resolved) {
+    RebindState s = rebind.get();
+    if (s == null || s.deadlineNanos - clockNanos.getAsLong() <= 0) {
+      return null;
+    }
+    return resolved.equals(s.affected) ? s.target : null;
```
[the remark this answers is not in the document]
ggivo: This is by design, but we need to support the case where the same node resolves to multiple IP's. E.g IPv4 & IPv6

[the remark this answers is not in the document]
ggivo: Done

src/main/java/redis/clients/jedis/MaintenanceEvent.java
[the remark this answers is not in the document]
ggivo: Done

src/main/java/redis/clients/jedis/MaintenanceEventController.java
```diff
@@ -0,0 +1,188 @@
...
+    SocketAddress affected = c.getRemoteSocketAddress();
+    if (affected == null) {
+      return; // receiver socket already closed; no peer to register
+    }
+    SocketAddress target = new InetSocketAddress(e.target.getHost(), e.target.getPort());
+    long deadline = clockNanos.getAsLong() + TimeUnit.SECONDS.toNanos(e.ttlSeconds);
+    while (true) {
+      RebindState cur = rebind.get();
+      if (cur != null && e.seq <= cur.seq) {
+        return; // duplicate or out-of-order event
```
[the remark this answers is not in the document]
ggivo: Only affected connections will receive the MOVING. They need to relax timeouts  for subsequent commands. Once the triggering node is decommissioned, the connection is also closed

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -477,6 +670,7 @@ protected Object readProtocolWithCheckingBroken() {
       throw new JedisConnectionException("Attempting to read from a broken connection.");
     }
 
+    expireRelaxedTimeout();
```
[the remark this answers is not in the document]
ggivo: Outdated

src/main/java/redis/clients/jedis/DefaultJedisSocketFactory.java
```diff
@@ -77,7 +82,16 @@ private Socket connectToFirstSuccessfulHost(HostAndPort hostAndPort) throws Exce
...
+        InetSocketAddress resolved = new InetSocketAddress(host, hostAndPort.getPort());
+        SocketAddress target = resolved;
+        SocketAddressMapper mapper = socketAddressMapper;
+        if (mapper != null) {
+          SocketAddress mapped = mapper.getSocketAddress(resolved);
+          if (mapped != null) {
+            target = mapped;
+          }
+        }
+        socket.connect(target, connectionTimeout);
```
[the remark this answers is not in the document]
ggivo: That is the whole purpose of the feature, avoid creating new connection toward affected nodes (ip's) but keep creating them against not affected nodes (IP's) + newly suggested target node

src/main/java/redis/clients/jedis/ConnectionFactory.java
```diff
@@ -138,6 +169,15 @@ public ConnectionFactory(Builder builder) {
     initAuthXManager();
   }
 
+  JedisClientConfig getClientConfig() {
+    return clientConfig;
+  }
```
atakavci: seems redundant

src/main/java/redis/clients/jedis/ConnectionFactory.java
```diff
@@ -138,6 +169,15 @@ public ConnectionFactory(Builder builder) {
...
+
+  /** Visible for testing: the underlying connection builder (and via it, the socket factory). */
+  Connection.Builder getConnectionBuilder() {
+    return connectionBuilder;
+  }
```
atakavci: seems redundant

src/main/java/redis/clients/jedis/builders/StandaloneClientBuilder.java
```diff
@@ -49,7 +66,12 @@ protected StandaloneClientBuilder<C> self() {
...
   protected ConnectionProvider createDefaultConnectionProvider() {
     return new PooledConnectionProvider(this.hostAndPort, this.clientConfig, this.cache,
-        this.poolConfig);
+        this.poolConfig, effectiveMaintNotificationsConfig());
+  }
+
+  private MaintenanceNotificationsConfig effectiveMaintNotificationsConfig() {
+    return maintNotificationsConfig != null ? maintNotificationsConfig
+        : MaintenanceNotificationsConfig.DEFAULT;
   }
```
atakavci: possible leftover from experimental changes.. otherwise , why need this ?

src/main/java/redis/clients/jedis/ConnectionPool.java
```diff
@@ -50,6 +60,37 @@ public ConnectionPool(HostAndPort hostAndPort, JedisClientConfig clientConfig,
...
+    if (controller != null) {
+      setEvictionPolicy(new RebindAwareEvictionPolicy(controller, getEvictionPolicy()));
+      controller.addHandoffHook(handoff -> evictQuietly());
+    }
+  }
+
+  /** The pool's maintenance controller, or {@code null} when maintenance is off. */
+  MaintenanceEventController getMaintenanceController() {
+    return maintenanceController;
+  }
```
atakavci: seems redundant

src/main/java/redis/clients/jedis/JedisClientConfig.java
```diff
@@ -61,6 +75,28 @@ default int getBlockingSocketTimeoutMillis() {
...
+   * active. Use {@link #UNSET_TIMEOUT_MS} to fall back to
+   * {@link #getBlockingSocketTimeoutMillis()}.
+   * @return relaxed blocking socket timeout in milliseconds (default
+   *         {@link #DEFAULT_RELAXED_BLOCKING_SOCKET_TIMEOUT_MS}, i.e. unset)
+   */
+  default int getRelaxedBlockingSocketTimeoutMillis() {
+    return DEFAULT_RELAXED_BLOCKING_SOCKET_TIMEOUT_MS;
+  }
+
   /**
```
atakavci: if we are going to add these to interface and all, we need to make sure it is supported/ready to be used  end2end via core components like connection/factories etc. 
Otherwise it doesn't mean much and creates questions/confusion on user side of the world.

src/main/java/redis/clients/jedis/MaintenanceEvent.java
```diff
@@ -0,0 +1,220 @@
...
```
atakavci: it needs more structure here; lets resolve the type of event at once and leave the parse/construction to concrete event types. Also i would suggest to not fallback to null when its identified as moving event. Fundamentally its just a variation of `MovingEvent`. Even if there is nothing to do with it, let the parser do the job and make the listener responsible to decide on it.

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -855,4 +1147,142 @@ List<PushConsumer> getPushConsumers() {
   protected void addPushConsumer(PushConsumer consumer) {
     this.pushConsumers.add(consumer);
   }
+
+  @Internal
+  protected void removePushConsumer(PushConsumer consumer) {
+    this.pushConsumers.remove(consumer);
+  }
```
atakavci, line 1073: possible source of future issues since `pushConsumerChain` ops are not thread-safe.

atakavci (commented):
I left notes around a few thigns i see; a few nits, refactorings for the sake of readability and some runtime optimizations.  

I have the feeling that the approach to handle the timeout from different events/sources would be simplified into a timoutsupplie/timeoutconsumer couple, but i dont have the clear definition of the solution yet. Will ping you as soon as i have it drafted.

Another important point; IIRC the timeouts received in maintenance messages are containing the information around when the migration/failover would be commencing. With the current proposal in the PR, its perceived as the end of the process/flow instead of time of expected initiation. Lets clear if my understanding is wrong or not.

src/main/java/redis/clients/jedis/ConnectionPool.java
ggivo: It is used for testing.  Access the pool's maintenance controller and inject a custom clock for tests.

src/main/java/redis/clients/jedis/JedisClientConfig.java
ggivo: Current solution exposes on the Connection level
  ``` 
// use relaxed timeout for period of time
  @Experimental
  public void relaxTimeouts(Duration period)

// use standard timeouts
@Experimental
  public void resetRelaxedTimeouts() 
```
and use above setting to configure the relaxed timeouts.

Currently, we are going to trigger/reset relaxed timoeuts based on SCH events, but 
I was thinking that at some point, relaxation might be triggered by different triggers.

Another option would be to move them in MaintenanceNotificationsConfig and keep them dedicated to SCH.
WDYT?

File: src/main/java/redis/clients/jedis/MaintenanceEvent.java (hunk 19 above)
Comment (ggivo) [reply src/main/java/redis/clients/jedis/MaintenanceEvent.java]: agree about the structure and will try to address it., but I think if a syntax invalid event is detected, we should not perform any action but report it and discard it (including moving)

File: src/main/java/redis/clients/jedis/JedisClientConfig.java (hunk 18 above)
Comment (ggivo) [reply src/main/java/redis/clients/jedis/JedisClientConfig.java]: Will move the `relaxedBlockingSocketTimeoutMillis`/`relaxedSocketTimeoutMillis`  inside `MaintenanceNotificationsConfig`, and make Connection level methods related to relaxed timeout package private, so we don't expose them till we have a clear usecase

File: src/main/java/redis/clients/jedis/MaintenanceEvent.java (hunk 19 above)
Comment (ggivo) [reply src/main/java/redis/clients/jedis/MaintenanceEvent.java]: Addressed with [commit](https://github.com/redis/jedis/pull/4473/changes/d896f7dbe0fecc47fefae984fded875a3dd19435)
Now we first resolve the type from push notification, then parse and populate MaintenanceEvent based on the resolved type

File: src/main/java/redis/clients/jedis/builders/StandaloneClientBuilder.java [hunk 21]
```diff
@@ -41,6 +42,20 @@ public StandaloneClientBuilder<C> hostAndPort(HostAndPort hostAndPort) {
...
+   * Configures server-side maintenance notifications (MIGRATING, MIGRATED, FAILING_OVER,
+   * FAILED_OVER, MOVING).
+   * <p>
+   * If unset, defaults to ).
+   * @param config maintenance notifications configuration; {@code null} keeps the default
+   * @return this builder
+   */
+  public StandaloneClientBuilder<C> maintenanceNotifications(
+      MaintenanceNotificationsConfig config) {
+    this.maintNotificationsConfig = config;
```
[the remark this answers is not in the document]
ggivo: resolved with [
Set per-client maintenance-notifications defaults: standalone AUTO ](https://github.com/redis/jedis/pull/4473/commits/16fdfd7de43647661f66aa79c08d91f0c0fb7b3c)

src/main/java/redis/clients/jedis/Connection.java
ggivo: Agree, but Connection itself is not thread safe.

src/main/java/redis/clients/jedis/JedisClientConfig.java
ggivo: addressed with [Move relaxed-timeout config to the maintenance feature](https://github.com/redis/jedis/pull/4473/commits/3c7ad056668db290d53f51688288d0ab13b21a39)

src/main/java/redis/clients/jedis/MaintenanceEventConsumer.java
```diff
@@ -0,0 +1,41 @@
...
+      return context;
+    }
+
+    MaintenanceEvent event = MaintenancePushCodec.build(type, message);
+    if (event != null) {
+      for (MaintenanceEventListener listener : listeners) {
+        event.accept(listener, connection);
+      }
+    }
+    context.drop(); // a maintenance event is consumed even if malformed
```
atakavci: what are the exact cases that event type is not null but message is null ?

src/main/java/redis/clients/jedis/MaintenancePushCodec.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+    return event;
+  }
+
+  private static MaintenanceEvent moving(List<Object> c) { // [MOVING, seq, time_s, host:port]
+    if (c.size() < 3 || !(c.get(1) instanceof Long) || !(c.get(2) instanceof Long)) {
+      return null;
+    }
+    HostAndPort target = parseHostPort(c, 3);
+    return target == null ? null : new MovingEvent((Long) c.get(1), (Long) c.get(2), target);
```
atakavci: if this is just to drop the event, controller is the ideal decision maker to me, instead of codec/parser. 
not a strong opinion, but as-is it feels like codec decides and controller is just prepared for decision.

src/main/java/redis/clients/jedis/MaintenancePushCodec.java
```diff
@@ -0,0 +1,134 @@
...
+    PushType(byte[] token, Function<List<Object>, MaintenanceEvent> decoder) {
+      this.token = token;
+      this.decoder = decoder;
+    }
+
+    /**
+     * Resolves a push type token to its maintenance type, or {@code null} when it is not a
+     * maintenance push. Length-switch fast path: rejects unrelated pushes with one comparison.
+     */
+    static PushType resolve(byte[] type) {
```
atakavci, line 36: how about one single method to `resolve type` + `parse` ?

src/main/java/redis/clients/jedis/MaintenanceEventConsumer.java
ggivo: The if (event != null) guard is for a recognized type with a malformed payload

src/main/java/redis/clients/jedis/MaintenancePushCodec.java
ggivo: I may have misunderstood the question, but there is a missing functionality to support "null" for the target, there is a missing TODO. We still need to implement(logic to parse "_NULL_" and update the Moving Event domain object), and implement a "rebind" operation toward the original endpoint address

src/main/java/redis/clients/jedis/MaintenancePushCodec.java
ggivo: Kept `resolve` separate from `build` for two reasons: `resolve` gives a cheap classification step that identifies and skips non-maintenance pushes without any parsing, and the split still lets us distinguish a malformed maintenance frame from a valid one (`build` returns `null` for malformed, the event for valid).

The consumer branches on three outcomes:
- **not a maintenance push** → pass through (don't drop)
- **maintenance type but malformed body** → drop, no dispatch
- **maintenance type, valid** → dispatch + dropped

ggivo: Discussed and decided to keep both methods, but propagate errors for malformed maintence events via exception instead of returning `null`

src/main/java/redis/clients/jedis/MaintenancePushCodec.java
ggivo: refactored the code to propagate malformed maintence events  errors as excpetions

src/main/java/redis/clients/jedis/MaintenanceEventController.java
```diff
@@ -0,0 +1,277 @@
...
+  @Override
+  public void onMigrated(MigratedEvent e, Connection c) {
+    logger.debug("Migrated shards {} (seq={})", e.shardIds, e.seq);
+    c.resetRelaxedTimeouts();
+  }
+
+  @Override
+  public void onFailedOver(FailedOverEvent e, Connection c) {
+    logger.debug("Failed over shards {} (seq={})", e.shardIds, e.seq);
+    c.resetRelaxedTimeouts();
```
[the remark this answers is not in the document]
ggivo: It is per design,
MIGRATING/MIGRATED, FAILING_OVER/FAILED/OVER maintenance events received on the connection are handled per connection. They are received and processed sequentially in order of appearance, applying/disabling relaxation per connection. seq is used to deduplicate MOVING notification which is handle on the pool level, applying relaxation globally

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -907,4 +1064,79 @@ List<PushConsumer> getPushConsumers() {
...
+   * into effect once the window closes or {@link #resetRelaxedTimeouts()} is called. Calling this
+   * with a later deadline extends the window; an earlier one is ignored.
+   * @param period maximum duration of the relaxation window
+   */
+  @Experimental
+  void relaxTimeouts(long expiration) {
+    if (this.relaxedTimeoutSource == null) {
+      throw new IllegalStateException("Relaxed timeouts not activated");
+    }
+    relaxedTimeoutSource.setExpirationTime(expiration);
```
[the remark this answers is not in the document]
ggivo: handled with follow up [PR](https://github.com/redis/jedis/issues/4603)

src/main/java/redis/clients/jedis/MaintenanceEventController.java
```diff
@@ -0,0 +1,277 @@
...
+      }
+      if (e.seq == cur.seq) {
+        // Same handoff event from another affected connection: merge its peer into the set.
+        if (!target.equals(cur.target)) {
+          logger.warn("Ignoring MOVING with conflicting target for seq {}: have {}, got {}", e.seq,
+            cur.target, target);
+          return;
+        }
+        if (cur.affected.contains(affectedPeer)) {
+          return; // already merged
```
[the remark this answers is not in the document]
ggivo: It is by design. 
  - First MOVING per seq wins and sets the Pool level RebindState deadline.
  - expiresAT  is per connection
  - isAffected is pool level and used to apply pool level relaxation on top of conneciton ones.

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -394,7 +546,7 @@ public void close() {
     if (this.memberOf != null) {
       ConnectionPool pool = this.memberOf;
       this.memberOf = null;
-      if (isBroken()) {
+      if (isBroken() || expireAt < NanoClock.INSTANCE.getAsLong()) {
```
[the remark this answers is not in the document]
ggivo: By design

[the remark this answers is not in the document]
ggivo: accepted trade of some connection might get notification just after isAffected check, and before marked as idle in the pool

src/main/java/redis/clients/jedis/MaintenanceEventController.java
```diff
@@ -0,0 +1,277 @@
...
+    SocketAddress affectedPeer = c.getRemoteSocketAddress();
+    if (affectedPeer == null) {
+      return; // receiver socket already closed; no peer to register
+    }
+    SocketAddress target = new InetSocketAddress(e.target.getHost(), e.target.getPort());
+
+    while (true) {
+      RebindState cur = rebind;
+      if (e.seq < cur.seq) {
+        return; // older event
```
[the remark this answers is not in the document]
ggivo: Moving events on pool level are sequential, and connectio from previious seq should be already disconencted when next MOVING sseq is received

PR #4479, opened 2026-04-01
fix : [ClusterPipeline] ExecutorService/thread is created and destroyed too frequently in ClusterPipeline

### LINKED ISSUE #4141: [ClusterPipeline] ExecutorService/thread is created and destroyed too frequently in ClusterPipeline
The code at here 
https://github.com/redis/jedis/blob/1d1b268f9be5b093f6684bc814d2fdab376a1ee3/src/main/java/redis/clients/jedis/MultiNodePipelineBase.java#L88
It will create a ExecutorService for each pipeline sync, It will cause huge cost when there are many pipeline operations

ggivo:
### Problem
Multi-node cluster pipelines create a new `ExecutorService` on every `sync()` call, causing significant overhead in high-throughput scenarios.

### Solution
Added support for externally managed `ExecutorService` in cluster pipelines:
 
- **New API**: `cluster.pipelined(ExecutorService)` - allows users to provide a shared executor
- **Backward compatible**: `cluster.pipelined()` retains existing auto-managed behavior

### Implementation
- Modified `MultiNodePipelineBase` to accept optional `ExecutorService`
- Added `pipelined(ExecutorService)` methods to `RedisClusterClient` and `JedisCluster`

**Usage Example:**
```java
ExecutorService executor = Executors.newFixedThreadPool(3);
try (RedisClusterClient cluster = ...) {
    // Reuse executor across multiple pipelines
    try (ClusterPipeline pipeline = cluster.pipelined(executor)) {
        pipeline.set("key", "value");
        pipeline.sync();
    }
} finally {
    executor.shutdown(); // User controls lifecycle
}
```

Closes #4141

---

> [!NOTE]
> **Medium Risk**
> Changes the concurrency/executor lifecycle for multi-node cluster pipelining; incorrect executor management or shared-executor contention could impact performance or cause hangs, though behavior is covered by new integration tests.
> 
> **Overview**
> **Cluster pipelining can now reuse a caller-provided `ExecutorService` instead of creating/shutting down a thread pool on every multi-node `sync()`.**
> 
> `MultiNodePipelineBase` accepts an optional shared executor and only shuts down internally-created executors, while `JedisCluster` and `RedisClusterClient` add `pipelined(ExecutorService)` (with `pipelined()` delegating to it) to pass the executor through to `ClusterPipeline`. New integration tests verify the shared executor is not shut down and that many concurrent pipelines sharing a small executor complete without deadlock.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 38fa2a5593bfa7a0fd53cc6402cc6218cac32019. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
+++ b/src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
@@ -11,8 +11,6 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import java.util.concurrent.ThreadFactory;
-import java.util.concurrent.atomic.AtomicInteger;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
--- a/src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
+++ b/src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
@@ -11,6 +11,8 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -38,8 +40,7 @@ public abstract class MultiNodePipelineBase extends AbstractPipeline {
    * External executor service to use for {@code sync()}. If not set, a new executor service will be
    * created for each {@code sync()} call.
    */
-  private final ExecutorService executorService;
-  private final boolean sharedExecutor;
+  private final ExecutorService sharedExecutorService;
 
   public MultiNodePipelineBase(CommandObjects commandObjects) {
     this(commandObjects, StaticCommandFlagsRegistry.registry());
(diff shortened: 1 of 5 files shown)


src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
```diff
@@ -100,7 +112,7 @@ public final void sync() {
     Executor executor;
     ExecutorService executorService = null;
     if (multiNode) {
-      executorService = Executors.newFixedThreadPool(MULTI_NODE_PIPELINE_SYNC_WORKERS);
+      executorService = getPipelineExecutor();
```
atakavci, line 115: it is not this commit itself but this body of method looks like a bit jumbled, probably change after change in time.  
just to trigger thinking, if is it better to refactor it now or some other time. i ll let you call. 

removing items through iterator while iterating is also a pretty big deal here.

src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
```diff
@@ -144,12 +156,53 @@ public final void sync() {
...
+  private boolean isUsingSharedExecutor() {
+    return this.sharedExecutorService != null;
+  }
+
+  /**
+   * Creates a new dedicated executor for multi-node pipeline execution.
+   */
+  private ExecutorService createDedicatedPipelineExecutor() {
+    return Executors.newFixedThreadPool(MULTI_NODE_PIPELINE_SYNC_WORKERS);
+  }
```
atakavci, line 204: dont see benefit of having any of these.

atakavci (approved):
`MultiNodePipelineBase` is somewhat in need of care in general.
but approving if you like to proceed with this PR as is.

src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
ggivo: @atakavci  thanks for the review

I will move on as it is since,  immediate issue reported seems to be affecting multiple users, and prefer to  provide a resoluton

src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
ggivo: put them for readability only

PR #4489, opened 2026-04-16
Add out-of-band push notification handling

ggivo:
Handle unsupported push notifications gracefully instead of failing the connection.

Fixes errors occurring when push messages are received on connections that are not configured to process them. For example, enabling CLIENT TRACKING on a regular Connection may result in errors when invalidate push messages are delivered.

This change introduces initial support for out-of-band push notifications by:
- detecting push messages on the connection
- silently skipping unsupported push types

---

> [!NOTE]
> **Medium Risk**
> Changes the core reply parsing path (`Protocol.read`/`Connection`) to detect and handle RESP3 push frames, which can affect command-reply decoding and connection stability if edge cases are missed.
> 
> **Overview**
> Adds first-class handling of RESP3 out-of-band push messages by introducing a `PushConsumer`/`PushConsumerChain` API and wiring it into `Protocol.read` so `>` frames can be consumed (or selectively propagated) while reading normal command replies.
> 
> Updates `Connection` to maintain a default push-consumer chain (propagating pub/sub pushes, silently consuming other pushes) and lets `CacheConnection` register a new `PushInvalidateConsumer` to evict CSC keys on `invalidate` pushes instead of relying on protocol-layer cache hooks.
> 
> Expands unit/integration coverage with a new TCP-based mock server plus new tests/benchmarks validating push consumption/propagation and ensuring arbitrary pushes no longer break connections; adjusts existing mocks to the new `Protocol.read(is, chain)` signature.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 71d619b57282250671d799e52522ac6fc14e252c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/main/java/redis/clients/jedis/PushConsumerChainImpl.java
```diff
@@ -0,0 +1,124 @@
...
+    @Override
+    public PushConsumerContext handle(PushConsumerContext context) {
+      if (pubSubCommands.contains(context.getMessage().getType())) {
+        // Ensure pub/sub events are propagated to application
+        context.propagate();
+      }
+
+      return context;
+    }
+  };
```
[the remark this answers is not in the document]
ggivo: When in Pub/Sub mode with  RESP3, command reposnses (like ping) are not propagated as push messages but as standard command reponse.

References:
https://github.com/redis/jedis/blob/dad4cd7d25d1f2cdfaccbbc9b2c309f8f5aa0a5b/src/main/java/redis/clients/jedis/JedisPubSubBase.java#L78-L96

A passing test to validate PING with RESP3, available :
PublishSubscribeCommandsTest#pubSubChannelWithPingPong(), 
PublishSubscribeCommandsTest#pubSubChannelWithPingPongWithArgument()

src/main/java/redis/clients/jedis/csc/CacheConnection.java
```diff
@@ -63,6 +67,24 @@ public static Builder builder() {
...
+  static class PushInvalidateConsumer implements PushConsumer {
+    private final Cache cache;
+    public PushInvalidateConsumer(Cache cache) {
+      this.cache = cache;
+    }
+
+    @Override
+    public PushConsumerContext handle(PushConsumerContext context) {
+      if (context.getMessage() != null &&
+              "invalidate".equals(context.getMessage().getType())) {
```
uglide, line 79: Prob better to move it as a const to PushMessage

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -127,40 +127,50 @@ private static String[] parseTargetHostAndSlot(String clusterRedirectResponse) {
...
-      case GREATER_THAN_BYTE:
-        return processMultiBulkReply(is);
-      case MINUS_BYTE:
-        processError(is);
-        return null;
-      // TODO: Blob error '!'
-      default:
-        throw new JedisConnectionException("Unknown reply: " + (char) b);
-    }
+  private static Object process(final RedisInputStream is, PushConsumerChain pushConsumer) {
```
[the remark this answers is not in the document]
uglide: Seems legit and is related to my comment on CacheConnection

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -127,40 +127,50 @@ private static String[] parseTargetHostAndSlot(String clusterRedirectResponse) {
...
+        case COMMA_BYTE:
+          return is.readDoubleCrLf();
+        case LEFT_BRACE_BYTE:
+          return is.readBigIntegerCrLf();
+        case PERCENT_BYTE: // TODO: currently just to start working with HELLO
+          return processMapKeyValueReply(is);
+        case TILDE_BYTE: // TODO:
+          return processMultiBulkReply(is);
+        case GREATER_THAN_BYTE:
+          // return processMultiBulkReply(is)
```
uglide, line 157: Remove the commented line?

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -127,40 +127,50 @@ private static String[] parseTargetHostAndSlot(String clusterRedirectResponse) {
...
+          PushMessage message = processPush(is, pushConsumer);
+          if( message != null ) {
```
uglide, line 159: I'd add some comment explaining what is happening here, like: If the message is not consumed by PushConsumers, it's propagated to the application to preserve the old behavior.

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -127,40 +127,50 @@ private static String[] parseTargetHostAndSlot(String clusterRedirectResponse) {
...
+            return message.getContent();
```
uglide, line 160: BTW, do we really need to preserve the old behavior? I think atm we don't have a test that tests CSC enabled, with let's say a 10k command responses in the cache and then tests what happens if we receive an invalidate message with NULL instead of keys (all keys are invalid). Also a test with hundreds on invalidation messages pending.   I think it would be interesting to see how slow the new approach is, and if we break it, then we have a chance in the future releases to optimize this code path by offloading push messages processing to a separate thread, etc

src/main/java/redis/clients/jedis/PushConsumerChainImpl.java
```diff
@@ -0,0 +1,124 @@
...
+  static final PushConsumerChain PROPAGATE_ALL_CONSUMER_CHAIN = of(PROPAGATE_ALL_CONSUMER);
+
+  /**
+   * PushConsumer that marks pub/sub related events to be propagated to the caller
+   */
+  public static final PushConsumer PUBSUB_CONSUMER = new PushConsumer() {
+    final Set<String> pubSubCommands = new HashSet<>();
+
+    {
+      pubSubCommands.add("message");
```
uglide, line 41: Move to PushMessage as consts to have a single place where you can find all supported invalidation messages?

src/main/java/redis/clients/jedis/PushConsumerContext.java
```diff
@@ -0,0 +1,72 @@
...
+ * </ul>
+ */
+@Experimental
+public class PushConsumerContext {
+  private final PushMessage message;
+
+  private boolean propagate = false;
+  private boolean drop = false;
+
+  public PushConsumerContext(PushMessage message) {
```
atakavci, line 29: i cant imagine the use case where a context carries a null message. in theory, its not a push message until we identify what we read from buffer as push message.
i'd make it strict and validate in context ctor. And trim null checks of `PushMessage` everywhere else.

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -211,52 +221,69 @@ private static List<KeyValue> processMapKeyValueReply(final RedisInputStream is)
...
-        && Arrays.equals(INVALIDATE_BYTES, (byte[]) list.get(0))) {
-      cache.deleteByRedisKeys((List) list.get(1));
-      return null;
-    } else {
-      return list;
+    PushMessage message = new PushMessage(list);
+    if (consumer != null)  {
+      return consumer.process(message);
     }
+    return message;
```
atakavci, line 286: why a null `consumer`, what it means?
i guess it is semantically _'drop the the out-of-band message'_. or no-op consumer. 
it is a bit vague if the intention is to cover the case _'out-of-band message is not expected'_.  

how about one line and not letting a null chain?
```suggestion
    return consumer.process(message);
```

atakavci (approved):
LGTM! 
my comments are more like personal preference with the semantics.
also, i aggre with @uglide on the use of literals.

src/main/java/redis/clients/jedis/Protocol.java
ggivo: Semantic was if no consumer is provided just propagate the messages as before to client, agree that it is more conciser to always require a push consumer.

src/main/java/redis/clients/jedis/Protocol.java
ggivo: We need the prevoius behaviour to support Pub/Sub. Pub/Sub messages are propagated as before.  Will need to refacgore Pub/Sub implmentation in porder to drop push message propagation completely.

src/main/java/redis/clients/jedis/csc/CacheConnection.java
```diff
@@ -63,6 +67,24 @@ public static Builder builder() {
...
+        cache.deleteByRedisKeys((List) context.getMessage().getContent().get(1));
```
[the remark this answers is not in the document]
ggivo: key is requried

src/main/java/redis/clients/jedis/Protocol.java
```diff
@@ -193,7 +204,7 @@ private static List<Object> processMultiBulkReply(final RedisInputStream is) {
     final List<Object> ret = new ArrayList<>(num);
     for (int i = 0; i < num; i++) {
       try {
-        ret.add(process(is));
+        ret.add(process(is, null));
```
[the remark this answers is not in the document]
ggivo: push notifications can not be nested; they should appear only before/after commmand output

src/main/java/redis/clients/jedis/Protocol.java
ggivo: > with let's say a 10k command responses in the cache and then tests what happens if we receive an invalidate message with NULL instead of keys (all keys are invalid)

Probably not directly related to push handling, will add a test with a separate PR.

>Also, a test with hundreds on invalidation messages pending. I think it would be interesting to see how slow the new approach is, and if we break it.

Will do.

atakavci (approved):
LGTM

uglide (approved):

PR #4515, opened 2026-05-08
CommandObjects refactoring Part 3: Enforce protocol value and use RESP2 instead of the null value 

uglide:
> [!NOTE]
> **Medium Risk**
> Touches protocol selection/negotiation paths across `Jedis`, pipelines, transactions, and cluster pipeline, which can affect reply decoding and compatibility when protocol is unspecified or negotiation fails. Changes are small but impact core client behavior and are best validated against both RESP2/RESP3 servers.
> 
> **Overview**
> **Enforces a concrete RESP protocol for `CommandObjects` and removes reliance on `null` as a protocol sentinel.** `CommandObjects` now throws if constructed with `null`, and callers (`Jedis`, `Pipeline`, `Transaction`, `ReliableTransaction`, `UnifiedJedis`, `ClusterPipeline`) consistently use `RedisProtocol.orServerDefault(...)` to fall back to the server default (RESP2) when no protocol is resolved.
> 
> Updates a few command reply parsers (`hrandfieldWithValues`, `jsonType`) to branch explicitly on *RESP2 vs non-RESP2* rather than *non-RESP3*, and adjusts tests to stop parameterizing `CommandObjects` tests with a `null` protocol.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9006db1d0e4aa09850f7a7432943a193a78780ed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/main/java/redis/clients/jedis/Jedis.java
+++ b/src/main/java/redis/clients/jedis/Jedis.java
@@ -98,8 +98,6 @@ public class Jedis implements ServerCommands, DatabaseCommands, JedisCommands, J
 
   private static final Logger logger = LoggerFactory.getLogger(Jedis.class);
 
-  private static final RedisProtocol REDIS_SERVER_DEFAULT_PROTO = RedisProtocol.RESP2;
-
   protected final Connection connection;
   private final CommandObjects commandObjects;
   private int db = 0;
@@ -132,7 +130,7 @@ public class Jedis implements ServerCommands, DatabaseCommands, JedisCommands, J
 
   public Jedis() {
     connection = new Connection();
-    commandObjects = new CommandObjects(REDIS_SERVER_DEFAULT_PROTO);
+    commandObjects = new CommandObjects(RedisProtocol.REDIS_SERVER_DEFAULT_PROTO);
   }
 
   /**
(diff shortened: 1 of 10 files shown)


src/main/java/redis/clients/jedis/Jedis.java
```diff
@@ -157,7 +160,7 @@ public Jedis(final HostAndPort hostPort, final JedisClientConfig config) {
     JedisClientConfig effective = sanitize(config);
     connection = new Connection(hostPort, effective);
     RedisProtocol proto = effective.getRedisProtocol();
-    if (proto != null) commandObjects.setProtocol(proto);
+    commandObjects = new CommandObjects(proto != null ? proto : RedisProtocol.RESP2);
```
atakavci: it will be easier to manage with a central logic, when the time comes;
```java
public enum RedisProtocol {

  RESP2("2"),
  RESP3("3");

  public static RedisProtocol orDefualt(RedisProtocol proto) {
    return (proto == null) ? RESP2 : proto;
  }
...
```

atakavci (commented):
LGTM
(went through only the changes against [part2](https://github.com/redis/jedis/compare/im/refactor-command-objects-p2...im/refactor-command-objects-p3))

atakavci:
many test case impacted; 
`java.lang.IllegalArgumentException: protocol must not be null`

src/main/java/redis/clients/jedis/Jedis.java
uglide: Added with a slightly better name :)

ggivo (approved):
LGTM

PR #4519, opened 2026-05-08
Mitigate exception thrown on returningBroken connection to the pool

ggivo:
## Mitigate POOL-431 regression: `returnBrokenResource` throws when upstream is down

 A behavioral regression was introduced in commons-pool 2.13.0 as part of the fix for [POOL-424](https://issues.apache.org/jira/browse/POOL-424): `GenericObjectPool#invalidateObject` now unconditionally replaces the destroyed instance by internally calling `addObject()`. When the upstream Redis is unreachable, the implicit replacement fails and the exception propagates out of `Jedis#close()` / `pool.returnBrokenResource(...)`, breaking the pre-2.13 contract where returning a broken connection was a silent no-op on the failure path. This is tracked upstream as [POOL-431](https://issues.apache.org/jira/projects/POOL/issues/POOL-431).

Related PR's:
 - #4409 — initial Dependabot bump to commons-pool2 2.13.x; reverted because of this regression.
 - #4439 — narrow fix applied only inside `MultiDbClient`.

 This change moves the mitigation down to `Pool<T>` so every Jedis pool benefits. A thread-local guard set across `returnBrokenResource(...)` lets an `addObject()` override swallow failures originating from the implicit auto-replace

---

> [!NOTE]
> **Medium Risk**
> Touches core connection pooling behavior by changing how broken resources are invalidated under commons-pool2 2.13.x; mistakes could hide real pool/factory failures or affect pool liveness under load.
> 
> **Overview**
> Restores pre-commons-pool2-2.13 behavior where closing/returning a *broken* Jedis resource does not throw if Redis is down, by adding a thread-local guard in `Pool.returnBrokenResource` and overriding `addObject()` to suppress only the implicit auto-replacement failure triggered by `invalidateObject()`.
> 
> Upgrades `commons-pool2` to `2.13.1` and adds coverage: a new `PoolUnitTest` for suppression/propagation semantics, plus an updated `UnavailableConnectionTest` that uses Toxiproxy (new `redis-unavailable` endpoint + docker-compose port) to reliably simulate an unreachable Redis instance.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 62d001fbca53c29ebbed667ebfbb71932f4b574e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -90,7 +90,7 @@
 		<dependency>
 			<groupId>org.apache.commons</groupId>
 			<artifactId>commons-pool2</artifactId>
-			<version>2.12.1</version>
+			<version>2.13.1</version>
 		</dependency>
 		<dependency>
 			<groupId>org.json</groupId>
--- a/src/main/java/redis/clients/jedis/util/Pool.java
+++ b/src/main/java/redis/clients/jedis/util/Pool.java
@@ -3,9 +3,28 @@ package redis.clients.jedis.util;
 import org.apache.commons.pool2.PooledObjectFactory;
 import org.apache.commons.pool2.impl.GenericObjectPool;
 import org.apache.commons.pool2.impl.GenericObjectPoolConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import redis.clients.jedis.exceptions.JedisException;
 
 public class Pool<T> extends GenericObjectPool<T> {
+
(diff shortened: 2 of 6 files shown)


ggivo:
Discussed with the team and decided it introduces risks of future incompatibility with the library. Closing in favour of 
https://github.com/redis/jedis/pull/4439

PR #4521, opened 2026-05-12
test: verify CSC NULL invalidation clears 10k cached entries (CAE-2930)

ggivo:
## Summary

Adds two CSC integration tests that exercise large-scale **invalidation drain** behavior on the tracked connection. intended as a **baseline** for the upcoming push-notification refactor — pin down current correctness and timing so we can compare before/after.

 Also: `ClientSideCacheTestBase` control connection switched to `.resp2()` to avoid warning messages.

## Added tests

**`invalidateAllWithLargeCacheTest`**
 Warms 10 000 keys, `flushAll` on control, then a single `GET`. Asserts the **NULL invalidation** evicts everything except the just-fetched key (`size == 1`).

**`pendingInvalidationMessagesTest`**
 Caches 1 000 keys, mutates all 1 000 via control so the server queues 1 000 invalidation messages. A single `GET`  drains the backlog synchronously — asserts `CacheStats.getInvalidationCount() == 1000`.

### Example test log
invalidateAllWithLargeCacheTest: NULL invalidation flushed 10000 entries in 1038535 ns (9628948.47 keys/sec)
cache stats: CacheStats{hits=0, misses=10001, loads=10001, evicts=0, nonCacheable=0, flush=1, invalidationsByServer=0, invalidationMessages=0}

pendingInvalidationMessagesTest: 1000 pending invalidations processed in 2076278 ns (481631.07 invalidations/sec)
cache stats: CacheStats{hits=0, misses=1001, loads=1001, evicts=0, nonCacheable=1000, flush=0, invalidationsByServer=1000, invalidationMessages=1000}
```


<!-- CURSOR_SUMMARY -->
---

> [!NOTE]
> **Low Risk**
> Low risk since changes are test-only, but the new timing/await-based invalidation drain assertions could be slightly flaky on slow CI or different Redis configs.
> 
> **Overview**
> Adds two new CSC integration tests to cover *large-scale invalidation drain* behavior: one validates that a `flushAll` NULL invalidation clears a warmed 10k-entry client cache on the next command, and another verifies that a backlog of ~1k invalidation push messages is fully processed synchronously on a subsequent `GET` (asserting `CacheStats.getInvalidationCount()`).
> 
> Updates the test base control connection to use RESP2 (`.resp2()`) and adds logging/imports needed for the new performance/timing diagnostics.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 514623b62bb0fc51eedbba3f836771c15fac1fc7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>
<!-- /CURSOR_SUMMARY -->

### CODE REVIEW

File: src/test/java/redis/clients/jedis/csc/ClientSideCacheFunctionalityTest.java [hunk 1]
```diff
@@ -24,6 +24,8 @@
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 import org.mockito.Mockito;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import redis.clients.jedis.CommandObjects;
```
uglide, line 30: Add the `Protocol` import so the test can issue a raw `CLIENT ID` command on the tracked connection (used by the suggestion below).

```suggestion
import redis.clients.jedis.CommandObjects;
import redis.clients.jedis.Protocol;
```

src/test/java/redis/clients/jedis/csc/ClientSideCacheFunctionalityTest.java
```diff
@@ -55,6 +59,91 @@ public void flushAllTest() {
...
+      CacheStats stats = cache.getStats();
+      assertEquals(count, stats.getInvalidationCount());
+
+      double invalidationsPerSec = (count * 1_000_000_000.0) / processingElapsedNanos;
+
+      log.info("pendingInvalidationMessagesTest: {} pending invalidations processed in {} ns ({} invalidations/sec)", count,
+              processingElapsedNanos, String.format("%.2f", invalidationsPerSec));
+      log.info("cache stats: {}", stats);
+    }
+  }
```
uglide, line 160: Replace `control.ping()` with a server-side flush check.

`control.ping()` only proves the *control* connection round-tripped — it doesn't guarantee the server has finished writing the 1000 invalidation frames into the *tracked client's* output buffer. Under CI load or with larger N this can be racy.

This suggestion:

1. Pins the cached client to a single connection (`singleConnectionPoolConfig`) so `CLIENT ID` deterministically identifies the tracked connection.
2. Captures that connection's server-side id via `CLIENT ID`.
3. Polls `CLIENT LIST <id>` on the control connection until both `oll=0` (output list length) and `omem=0` (output buffer memory) — meaning the server has handed every queued invalidation byte to the kernel.
4. Also drops the unused `throws InterruptedException`.

Verified locally against Redis 8.6 / Azul JDK 8 — test passes; timing is comparable (~10 ms / ~91k invalidations/sec).

Note: `nonCacheable` in the stats bumps from 1000 to 1001 because the extra `CLIENT ID` command itself is non-cacheable.

```suggestion
  @Test
  public void pendingInvalidationMessagesTest() {
    final int count = 1000; // Hundreds of keys

    try (RedisClient jedis = RedisClient.builder().hostAndPort(hnp).clientConfig(clientConfig.get())
            .poolConfig(singleConnectionPoolConfig.get())
            .cacheConfig(CacheConfig.builder().maxSize(count).build()).build()) {
      Cache cache = jedis.getCache();

      // 1. Create and cache 1000 keys
      for (int i = 0; i < count; i++) {
        jedis.set("key" + i, "value" + i);
      }
      for (int i = 0; i < count; i++) {
        jedis.get("key" + i);
      }

      assertEquals(count, cache.getSize());
      assertEquals(0, cache.getStats().getInvalidationCount());

      // Capture the tracked connection's server-side id. Pool is pinned to 1 so this
      // is the same connection that holds the cached entries and will receive the
      // invalidation push messages.
      long trackedClientId = (Long) jedis.sendCommand(Protocol.Command.CLIENT, "ID");

      // 2. Use control connection to update ALL keys
      // This generates hundreds of invalidation messages
      for (int i = 0; i < count; i++) {
        control.set("key" + i, "newvalue" + i);
      }

      // 3. Wait until the server has flushed all queued output (the invalidation
      // frames) on the tracked client's connection. `oll` (output list length) and
      // `omem` (output buffer memory) report what the server still has buffered for
      // that client; once both are zero, every invalidation has been handed to the
      // kernel and is on its way / already in the tracked socket's receive buffer.
      await().atMost(5, TimeUnit.SECONDS).pollInterval(50, TimeUnit.MILLISECONDS)
              .until(() -> {
                String info = control.clientList(trackedClientId);
                return info != null && info.contains(" oll=0 ") && info.contains(" omem=0 ");
              });

      // 4. Now use the cached connection - it should process all pending invalidations synchronously
      long processingStartTime = System.nanoTime();
      jedis.get("key0");
      long processingElapsedNanos = System.nanoTime() - processingStartTime;

      // 5. Verify all invalidations were processed
      CacheStats stats = cache.getStats();
      assertEquals(count, stats.getInvalidationCount());

      double invalidationsPerSec = (count * 1_000_000_000.0) / processingElapsedNanos;

      log.info("pendingInvalidationMessagesTest: {} pending invalidations processed in {} ns ({} invalidations/sec)", count,
              processingElapsedNanos, String.format("%.2f", invalidationsPerSec));
      log.info("cache stats: {}", stats);
    }
  }
```

uglide (commented):
Suggestion: replace `control.ping()` with a server-side flush check via `CLIENT LIST` `oll`/`omem` polling. See inline comments.

uglide (commented):
See my suggestions for reducing the potential risk of flakiness of the test.

uglide (approved):

PR #4523, opened 2026-05-12
Add JMH benchmarks (CAE-2930)

ggivo:
## JMH benchmark migration & expansion

Migrates legacy manual benchmarks to a unified JMH suite and adds new workload and pub/sub coverage.

### Added benchmarks

- **`CRC16Benchmark`** — hash slot CRC16 computation (string + byte[] paths).
- **`SafeEncoderBenchmark`** — UTF-8 encode/decode hot paths.
- **`ProtocolReadBenchmark`** — RESP2/RESP3 parser throughput, including cache-aware and push-frame variants.
- **`JedisGetSetBenchmark`** — single-threaded GET/SET + pipelined GET/SET baseline using `Jedis`.
- **`RedisClientGetSetBenchmark`** — pooled GET/SET + pipelined GET/SET using `RedisClient` at T1 / T8 / T64.
- **`GetSetMixedR90W10Benchmark`** — fixed 90/10 read/write workload over a 1000-key working set; emits comparable troughput numbers for `Jedis` (T1), `JedisPool` (T1/T8), `RedisClient` (T1/T8), and `RedisClient` + client-side che (T1/T8). Replaces the legacy `PoolBenchmark` and `RedisClientCSCBenchmark`.
- **`PubSubPushBenchmark`** — end-to-end `publish → onMessage` round-trip throughput, complementing the protocolReadBenchmark` push-frame microbenchmark.

### Infrastructure

- `JmhMain` IDE launcher with per-suite include patterns.
- Maven `jmh` profile wires `src/test/jmh` as an additional test-source root and runs JMH with JSON result output.

### Example CI run
https://github.com/redis/jedis/actions/runs/25842479419/job/75930489878

---

> [!NOTE]
> **Low Risk**
> Changes are test-only and CI/docs wiring; production library code is untouched. CI grants `contents: write` for benchmark auto-push, and several benchmarks flush Redis—acceptable for dedicated CI Redis only.
> 
> **Overview**
> Adds a **JMH benchmark suite** under `src/test/jmh` and wires it into Maven and GitHub Actions so performance can be measured and tracked over time.
> 
> **Build & run:** `pom.xml` adds JMH test dependencies and a `-Pjmh` profile that registers `src/test/jmh`, skips normal unit/integration tests, and runs `org.openjdk.jmh.Main` with JSON written to `benchmarks.json` (per-benchmark annotations control mode/warmup; CLI does not override them). `JmhMain` supports IDE runs; `CyclingInputStream` reduces allocation noise in protocol microbenchmarks.
> 
> **Benchmark coverage (new):** RESP read/encode and CSC push paths (`ReadBenchmark`, `ReadPushesBenchmark`, `SendCommandBenchmark`); `CRC16Benchmark` / `SafeEncoderBenchmark`; live-Redis GET/SET and pipelining for `Jedis` and pooled `RedisClient` (1/8/64 threads); fixed **90% read / 10% write** workload comparing `Jedis`, `JedisPool`, `RedisClient`, and CSC (`GetSetMixedR90W10Benchmark`); end-to-end pub/sub round-trip (`PubSubPushBenchmark`). Several setups call **`FLUSHDB`** (documented as unsafe on shared Redis).
> 
> **CI & docs:** `.github/workflows/benchmarks.yml` replaces the placeholder with a nightly (and manual) job: Java 8, Docker Redis with password, `mvn -Pjmh clean test`, then `github-action-benchmark` **auto-pushes** results to the `benchmark-data` branch. `docs.yml` also builds on `benchmark-data` pushes and **copies** `benchmark-data/jedis/*` into the published site under `benchmarks/`.
> 
> Legacy `src/test/java/.../CRC16Benchmark.java` is only formatting; the JMH replacement lives under `util/CRC16Benchmark`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2ba6c500827caf326983ec32696dec91d47c57f9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/workflows/docs.yml
```diff
@@ -31,6 +31,15 @@ jobs:
...
           mkdocs build -d docsbuild
+      - name: Pull benchmark dashboard
+        uses: actions/checkout@v4
+        with:
+          ref: benchmark-data
+          path: benchmark-data
+      - name: Embed benchmark dashboard into site
+        run: |
+          mkdir -p docsbuild/benchmarks
+          cp -r benchmark-data/jedis/* docsbuild/benchmarks/
```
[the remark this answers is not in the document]
ggivo: benchmark-data branch  is created & exists

.github/workflows/benchmarks.yml
```diff
@@ -1,10 +1,67 @@
...
+          sleep 5
+          docker ps | grep redis-benchmark-test
+
+      - name: Maven offline dependencies
+        run: mvn -q dependency:go-offline
+
+      - name: Run JMH benchmarks
+        run: mvn -Pjmh clean test
+        env:
+          JVM_OPTS: -Xmx3200m
```
uglide: This config has no effect because you have hardcoded 2G value here https://github.com/redis/jedis/pull/4523/changes#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R783 and JMH reports `# VM options: -Xmx2G`

.github/workflows/benchmarks.yml
```diff
@@ -1,10 +1,67 @@
...
+        with:
+          java-version: '8'
+          distribution: 'temurin'
+          cache: 'maven'
+
+      - name: Start Redis server
+        run: |
+          docker run --name redis-benchmark-test \
+            -p 6379:6379 \
+            -d redis:8.6.3 \
```
uglide: Better to use without a patch :
```suggestion
            -d redis:8.6 \
```

src/test/jmh/redis/clients/jedis/benchmark/ProtocolReadBenchmark.java
```diff
@@ -0,0 +1,289 @@
...
+    // Create cache for testing
+    cache = CacheFactory.getCache(
+      CacheConfig.builder().maxSize(10000).cacheable(DefaultCacheable.INSTANCE).build());
+  }
+
+  /**
+   * Recreate streams before each benchmark invocation. This is necessary because streams get
+   * consumed during reading.
+   */
+  @Setup(Level.Invocation)
```
uglide: I'm not a big expert on JMH, but the docs say that:
> This level is only usable for benchmarks taking more than a millisecond per single [Benchmark](https://javadoc.io/static/org.openjdk.jmh/jmh-core/1.1.1/org/openjdk/jmh/annotations/Benchmark.html) method invocation. It is a good idea to validate the impact for your case on ad-hoc basis as well.


From: https://javadoc.io/doc/org.openjdk.jmh/jmh-core/1.1.1/org/openjdk/jmh/annotations/Level.html#Invocation

So, for ProtocolRead, we need some kind of batching.

uglide: Implemented the batching approach you described. Summary:

- Removed `@Setup(Level.Invocation) recreateStreams()` and the eight pre-allocated `RedisInputStream` fields.
- Pre-allocated only the source `byte[]`s (the cheap part). A small `stream(byte[])` helper wraps a fresh `RedisInputStream` per op — required because `RedisInputStream` retains internal `buf`/`count`/`limit` state and has no public reset.
- Each `@Benchmark` is now a tight `BATCH_SIZE`-loop with `@OperationsPerInvocation(BATCH_SIZE)` (`BATCH_SIZE = 100`).
- Applied the same batching to `encodeSetCommand` for consistency.

Trade-off worth flagging: stream construction now lives inside the measured op rather than excluded via Setup. Reported `ns/op` will be slightly higher than before, but the measurement is jitter-free and JMH no longer emits the sub-microsecond invocation-setup warning. I think that's the right call — Setup-excluded numbers were optimistic anyway since the JMH machinery transitions between setup and measurement aren't free.

Verified locally: compiles under `mvn -Pjmh test-compile` with JDK 8 (Azul Zulu 1.8.0_422). Formatter already applied.

GitHub can't render this as a one-click `suggestion` block here because the parent comment is anchored to a single diff position rather than a multi-line range — pasting the full new file body for review:

<details><summary><b>Full <code>ProtocolReadBenchmark.java</code> (lines 18 → EOF)</b></summary>

```java
/**
 * Comprehensive JMH Benchmark for Jedis Protocol operations. This benchmark covers: 1. Baseline
 * protocol operations (parsing, encoding) 2. Cache-aware protocol reads (push notification
 * overhead) 3. Push invalidation message processing at scale (1, 10, 100 pattern) 4. Realistic
 * mixed scenarios (push messages + regular responses) Run with: mvn -Pjmh clean test
 * <p>
 * Each benchmark wraps a {@code BATCH_SIZE} loop, with {@link OperationsPerInvocation} declaring
 * the per-invocation op count. Streams are constructed inside the loop from pre-allocated byte
 * arrays — {@link RedisInputStream} retains internal read state, so a fresh wrapper is required per
 * op. This pattern replaces {@code @Setup(Level.Invocation)}, which JMH flags as unreliable for
 * sub-microsecond operations.
 */
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@State(Scope.Benchmark)
@Fork(1)
@Warmup(iterations = 5, time = 2, timeUnit = TimeUnit.SECONDS)
@Measurement(iterations = 5, time = 2, timeUnit = TimeUnit.SECONDS)
@Threads(1)
public class ProtocolReadBenchmark {

  private static final int BATCH_SIZE = 100;

  // ========== BASELINE PROTOCOL DATA ==========

  private static final byte[] KEY = "123456789".getBytes();
  private static final byte[] VAL = "FooBar".getBytes();

  private static final byte[] SIMPLE_STRING_BYTES = "+OK\r\n".getBytes();
  private static final byte[] BULK_STRING_BYTES = "$5\r\nHello\r\n".getBytes();
  private static final byte[] ARRAY_BYTES = "*3\r\n$3\r\nfoo\r\n$3\r\nbar\r\n$3\r\nbaz\r\n"
      .getBytes();
  private static final byte[] MULTI_BULK_BYTES = "*4\r\n$3\r\nfoo\r\n$13\r\nbarbarbarfooz\r\n$5\r\nHello\r\n$5\r\nWorld\r\n"
      .getBytes();

  // ========== PUSH INVALIDATION DATA ==========

  // >2\r\n$10\r\ninvalidate\r\n*1\r\n$3\r\nkey\r\n
  private byte[] singlePushBytes;
  // >2\r\n$10\r\ninvalidate\r\n*100\r\n... (100 keys)
  private byte[] largePushBytes;

  // Mixed: N pushes followed by a regular response
  private byte[] mixedSinglePushBytes;
  private byte[] mixedTenPushBytes;
  private byte[] mixedHundredPushBytes;

  // Cache for testing
  private Cache cache;

  @Setup
  public void setup() {
    // Single push with 1 key
    singlePushBytes = ">2\r\n$10\r\ninvalidate\r\n*1\r\n$3\r\nkey\r\n".getBytes();

    // Large push with 100 keys invalidated at once
    StringBuilder largePush = new StringBuilder(">2\r\n$10\r\ninvalidate\r\n*100\r\n");
    for (int i = 0; i < 100; i++) {
      String key = "key:" + i;
      largePush.append("$").append(key.length()).append("\r\n").append(key).append("\r\n");
    }
    largePushBytes = largePush.toString().getBytes();

    // 1 push + response
    mixedSinglePushBytes = ">2\r\n$10\r\ninvalidate\r\n*1\r\n$3\r\nkey\r\n+OK\r\n".getBytes();

    // 10 pushes + response
    StringBuilder mixed10 = new StringBuilder();
    for (int i = 0; i < 10; i++) {
      mixed10.append(">2\r\n$10\r\ninvalidate\r\n*1\r\n$4\r\nkey").append(i).append("\r\n");
    }
    mixed10.append("+OK\r\n");
    mixedTenPushBytes = mixed10.toString().getBytes();

    // 100 pushes + response (heavy invalidation scenario)
    StringBuilder mixed100 = new StringBuilder();
    for (int i = 0; i < 100; i++) {
      mixed100.append(">2\r\n$10\r\ninvalidate\r\n*1\r\n$5\r\nkey").append(String.format("%02d", i))
          .append("\r\n");
    }
    mixed100.append("+OK\r\n");
    mixedHundredPushBytes = mixed100.toString().getBytes();

    cache = CacheFactory.getCache(
      CacheConfig.builder().maxSize(10000).cacheable(DefaultCacheable.INSTANCE).build());
  }

  private static RedisInputStream stream(byte[] bytes) {
    return new RedisInputStream(new ByteArrayInputStream(bytes));
  }

  // ========== BASELINE PROTOCOL OPERATIONS (NO CACHE) ==========

  /**
   * Baseline: Read simple string without cache.
   */
  @Benchmark
  @OperationsPerInvocation(BATCH_SIZE)
  public void readSimpleString(Blackhole blackhole) throws Exception {
    for (int i = 0; i < BATCH_SIZE; i++) {
      blackhole.consume(Protocol.read(stream(SIMPLE_STRING_BYTES)));
    }
  }

  /**
   * Baseline: Read bulk string without cache.
   */
  @Benchmark
  @OperationsPerInvocation(BATCH_SIZE)
  public void readBulkString(Blackhole blackhole) throws Exception {
    for (int i = 0; i < BATCH_SIZE; i++) {
      blackhole.consume(Protocol.read(stream(BULK_STRING_BYTES)));
    }
  }

  /**
   * Baseline: Read array without cache.
   */
  @Benchmark
  @OperationsPerInvocation(BATCH_SIZE)
  public void readArray(Blackhole blackhole) throws Exception {
    for (int i = 0; i < BATCH_SIZE; i++) {
      blackhole.consume(Protocol.read(stream(ARRAY_BYTES)));
    }
  }

  /**
   * Baseline: Read multi-bulk response without cache.
   */
  @Benchmark
  @OperationsPerInvocation(BATCH_SIZE)
  public void readMultiBulkResponse(Blackhole blackhole) throws Exception {
    for (int i = 0; i < BATCH_SIZE; i++) {
      blackhole.consume(Protocol.read(stream(MULTI_BULK_BYTES)));
    }
  }

  /**
   * Encode SET command (baseline protocol performance).
   */
  @Benchmark
  @OperationsPerInvocation(BATCH_SIZE)
  public void encodeSetCommand(Blackhole blackhole) throws Exception {
    for (int i = 0; i < BATCH_SIZE; i++) {
      RedisOutputStream out = new RedisOutputStream(new ByteArrayOutputStream(8192));
      Protocol.sendCommand(out, new CommandArguments(Protocol.Command.SET).key(KEY).add(VAL));
      blackhole.consume(out);
    }
  }

  // ========== CACHE-AWARE READS: Overhead of checking for push messages ==========

  /**
   * Cache-aware: Read simple string with cache (checks for push messages but none present). Compare
   * to readSimpleString to measure overhead.
   */
  @Benchmark
  @OperationsPerInvocation(BATCH_SIZE)
  public void cacheAwareReadSimpleString(Blackhole blackhole) throws Exception {
    for (int i = 0; i < BATCH_SIZE; i++) {
      blackhole.consume(Protocol.read(stream(SIMPLE_STRING_BYTES), cache));
    }
  }

  /**
   * Cache-aware: Read bulk string with cache (checks for push messages but none present). Compare
   * to readBulkString to measure overhead.
   */
  @Benchmark
  @OperationsPerInvocation(BATCH_SIZE)
  public void cacheAwareReadBulkString(Blackhole blackhole) throws Exception {
    for (int i = 0; i < BATCH_SIZE; i++) {
      blackhole.consume(Protocol.read(stream(BULK_STRING_BYTES), cache));
    }
  }

  /**
   * Cache-aware: Read array with cache (checks for push messages but none present). Compare to
   * readArray to measure overhead.
   */
  @Benchmark
  @OperationsPerInvocation(BATCH_SIZE)
  public void cacheAwareReadArray(Blackhole blackhole) throws Exception {
    for (int i = 0; i < BATCH_SIZE; i++) {
      blackhole.consume(Protocol.read(stream(ARRAY_BYTES), cache));
    }
  }

  /**
   * Cache-aware: Read multi-bulk response with cache (checks for push messages but none present).
   * Compare to readMultiBulkResponse to measure overhead.
   */
  @Benchmark
  @OperationsPerInvocation(BATCH_SIZE)
  public void cacheAwareReadMultiBulkResponse(Blackhole blackhole) throws Exception {
    for (int i = 0; i < BATCH_SIZE; i++) {
      blackhole.consume(Protocol.read(stream(MULTI_BULK_BYTES), cache));
    }
  }

  // ========== PUSH MESSAGE PROCESSING: Direct invalidation overhead ==========

  /**
   * Process single push invalidation message (1 key). Measures cost of processing one cache
   * invalidation.
   */
  @Benchmark
  @OperationsPerInvocation(BATCH_SIZE)
  public void processSinglePushInvalidation(Blackhole blackhole) throws Exception {
    for (int i = 0; i < BATCH_SIZE; i++) {
      blackhole.consume(Protocol.readPushes(stream(singlePushBytes), cache, true));
    }
  }

  /**
   * Process large push invalidation (100 keys in single message). Simulates mass invalidation event
   * (e.g., FLUSHDB on tracked keys).
   */
  @Benchmark
  @OperationsPerInvocation(BATCH_SIZE)
  public void processLargePushInvalidation(Blackhole blackhole) throws Exception {
    for (int i = 0; i < BATCH_SIZE; i++) {
      blackhole.consume(Protocol.readPushes(stream(largePushBytes), cache, true));
    }
  }

  // ========== MIXED STREAMS: Realistic scenarios with push + response ==========

  /**
   * Read response preceded by 1 push invalidation. Most common realistic scenario.
   */
  @Benchmark
  @OperationsPerInvocation(BATCH_SIZE)
  public void readWith1PushMessage(Blackhole blackhole) throws Exception {
    for (int i = 0; i < BATCH_SIZE; i++) {
      blackhole.consume(Protocol.read(stream(mixedSinglePushBytes), cache));
    }
  }

  /**
   * Read response preceded by 10 push invalidations. Moderate invalidation burst before response.
   */
  @Benchmark
  @OperationsPerInvocation(BATCH_SIZE)
  public void readWith10PushMessages(Blackhole blackhole) throws Exception {
    for (int i = 0; i < BATCH_SIZE; i++) {
      blackhole.consume(Protocol.read(stream(mixedTenPushBytes), cache));
    }
  }

  /**
   * Read response preceded by 100 push invalidations. Heavy invalidation scenario (worst-case for
   * read latency).
   */
  @Benchmark
  @OperationsPerInvocation(BATCH_SIZE)
  public void readWith100PushMessages(Blackhole blackhole) throws Exception {
    for (int i = 0; i < BATCH_SIZE; i++) {
      blackhole.consume(Protocol.read(stream(mixedHundredPushBytes), cache));
    }
  }
}
```

</details>

src/test/jmh/redis/clients/jedis/benchmark/workload/GetSetMixedR90W10Benchmark.java
```diff
@@ -0,0 +1,313 @@
...
+
+    protected JedisPool pool;
+
+    @Override
+    protected void setupClient() {
+      GenericObjectPoolConfig<Jedis> poolConfig = new GenericObjectPoolConfig<>();
+      poolConfig.setMaxTotal(POOL_SIZE);
+      poolConfig.setMaxIdle(POOL_SIZE);
+      pool = new JedisPool(poolConfig, new HostAndPort(endpoint.getHost(), endpoint.getPort()),
+          DefaultJedisClientConfig.builder().password(endpoint.getPassword()).build());
```
uglide: Please add `.serverDefaultProtocol()` to silence the warnings about Jedis not supporting auto-negotiation

uglide (commented):
Good initiative, but needs some polishing.

.github/workflows/benchmarks.yml
ggivo: Do you mean 8.6.0?
I think benchmarks should run against pinned version

uglide: Then its better to use 8.2, because it's an LTS version

ggivo: WDYT about pinned to  8.2.6 (latest 8.2), make sure we don't pick patch releases

.github/workflows/benchmarks.yml
ggivo: Will remove the one from the workflow.

src/test/jmh/redis/clients/jedis/benchmark/ProtocolReadBenchmark.java
ggivo: Will try to avoid recreating the stream on each read, so let me try an alternative approach.
Work on batches, but pre-populate the stream with BATCH*RESPONSES,

PR #4527, opened 2026-05-14
ci: declare workflow-level `contents: read` on 8 build/test workflows

arpitjain099:
Pins the default `GITHUB_TOKEN` to `contents: read` on the 8 workflows in `.github/workflows/` that don't actually need any write scope:

- `benchmarks.yml`: scheduled JMH benchmarks.
- `doctests.yml`: jbang doctest runner.
- `format_check.yml`: pmd / formatter checks.
- `integration.yml`: redis integration test matrix.
- `snapshot.yml`: snapshot deploy precheck. The actual Sonatype deploy uses a separate secret, not `GITHUB_TOKEN`.
- `spring-data-redis-integration.yml`: spring-data-redis sanity check.
- `test-on-docker.yml`: docker-based redis matrix.
- `unit-tests.yml`: junit run.

None call a GitHub API beyond the initial checkout.

`release-drafter.yml` and `version-and-release.yml` are intentionally left implicit since they need write scopes; those are best declared by maintainers who own the release flow.

## Why

CVE-2025-30066 (March 2025 `tj-actions/changed-files` supply-chain compromise) exfiltrated `GITHUB_TOKEN` from workflow logs and the leaked token retained whatever scope was issued at the workflow level. Pinning per workflow caps that runtime authority irrespective of the repo or org default, gives drift protection if the default ever widens, and is credited per-file by the OpenSSF Scorecard `Token-Permissions` check.

YAML validated locally with `yaml.safe_load` on each touched file.

---

> [!NOTE]
> **Low Risk**
> CI YAML only; narrows token scope with no change to build or test logic.
> 
> **Overview**
> Adds **workflow-level `permissions: contents: read`** to the docker-based Redis matrix workflow (`test-on-docker.yml`), so the default `GITHUB_TOKEN` is limited to checkout/read and cannot use broader repo scopes if the token is leaked or the org default widens.
> 
> This is part of a broader hardening pass (OpenSSF Scorecard **Token-Permissions**); the workflow still only checks out code and runs tests via `.github/actions/run-tests`—no GitHub API writes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d08ada551e35f98a66c643f0ff59b0d9c2d71cc6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ggivo:
Thanks — makes sense.

Before merging, we'll need to sanity-check that none of the flows rely on a write scope. One concrete flag: unit-tests.yml uses EnricoMi/publish-unit-test-result-action@v2.

arpitjain099:
Thanks for the catch @ggivo, you're right. `EnricoMi/publish-unit-test-result-action@v2` writes test results back to the PR (it needs `pull-requests: write` and `checks: write` to do that). Plain `contents: read` would break it.

Dropped `unit-tests.yml` from this PR. The 7 remaining files now in the diff (`benchmarks.yml`, `doctests.yml`, `format_check.yml`, `integration.yml`, `snapshot.yml`, `spring-data-redis-integration.yml`, `test-on-docker.yml`) really are read-only with respect to `GITHUB_TOKEN`. The `unit-tests.yml` permissions are best declared by a maintainer who can pin the exact write scopes the action needs.

arpitjain099:
Good sanity-check @ggivo. Took a closer look at all 7 files: 4 of them (`doctests`, `integration`, `snapshot`, `spring-data-redis-integration`) use `actions/cache@v4` directly, which needs `actions: write` on the post step to save the cache. As an external contributor I don't want to widen scopes on your behalf, so I reverted those 4 back to upstream and kept only the cache-free workflows in the PR: `benchmarks`, `format_check`, `test-on-docker`. `unit-tests.yml` you flagged isn't in this PR so no change there.

Force-pushed. Let me know if reduced scope works for you, or if you'd rather close.

ggivo:
@arpitjain099 
Thanks for the effort on this one.

Only one that can be safely hardened is `format_check`, and hardening on it is already applied with your previous PR #4538,

Closing without merging!

arpitjain099:
Thanks for taking a look @ggivo, and good to know #4538 already covered format_check. That makes sense - the rest aren't worth hardening. Appreciate the review!

PR #4504, opened 2026-04-28
TimeSeries: support multiple aggregators in TS.RANGE/TS.MRANGE (CAE-792)

ggivo:
Extend the existing tsRange/tsRevRange/tsMRange/tsMRevRange family to accept multiple aggregators in a single call, matching the new server syntax `AGGREGATION agg1,agg2,…,aggN bucketDur` (RedisTimeSeries 8.8).

Changes:
- TSElement: carry `List<Double> values` internally; existing`(long, double)` constructor preserved (wraps in a singleton list); `getValue()` returns `values.get(0)` for backward compatibility; new `getValues()` exposes the full positional list.
- TSRangeParams / TSMRangeParams: new `aggregation(AggregationType[],long)` setter; Wire format is identical to the previous single-aggregator output when length == 1.
- TimeSeriesBuilderFactory.TIMESERIES_ELEMENT_LIST: parse positional `[ts, v1, …, vN]` samples; size-2 takes the existing single-value path, size > 2 builds a multi-value TSElement.

No new public methods on RedisTimeSeriesCommands / RedisTimeSeriesPipelineCommands

---

> [!NOTE]
> **Medium Risk**
> Changes the `TSElement` data model and time series range parameter encoding to support multi-aggregator responses, which may affect equality/hashCode semantics and downstream code that assumes a single value per sample.
> 
> **Overview**
> Adds support for RedisTimeSeries multi-aggregator syntax (`AGGREGATION a,b,c bucket`) across `TS.RANGE`/`TS.REVRANGE` and `TS.MRANGE`/`TS.MREVRANGE` by allowing `TSRangeParams`/`TSMRangeParams` to accept `AggregationType[]` and encoding aggregators as a comma-joined token.
> 
> Updates `TSElement` to carry *multiple values per timestamp* (`getValues()`), while keeping `getValue()` returning the first value for backward compatibility, and extends response parsing to handle `[ts, v1, ..., vN]` samples.
> 
> Expands test coverage for wire-format validation and new multi-aggregator behaviors (including empty results) and adjusts cluster time series ITs to disable the new label-based multi-range tests; also updates `pom.xml` formatter includes for timeseries tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 23a26e9775446bcb35d179747fe1b68b7438934e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #4537, opened 2026-05-22
TimeSeries: reduce overhead on single-aggregator TSElement path (follow-up to #4504)

ggivo:
…

---

> [!NOTE]
> **Low Risk**
> Internal representation and builder wiring change with preserved public API and broad unit tests; no auth or security impact.
> 
> **Overview**
> Optimizes the common **single-value** Time Series sample path by storing a **`double`** on `TSElement` instead of allocating a **`List`**, while keeping **`getValues()`** behavior via **`Collections.singletonList`**.
> 
> Multi-aggregator samples are now built as an internal **`TSElement.MultiValueTSElement`** subclass (used from **`TimeSeriesBuilderFactory`** when a sample has more than one value). **`equals`/`hashCode`** are adjusted so single-value and singleton multi-value instances compare consistently, with **`doubleToLongBits`** for plain **`TSElement`** equality.
> 
> Adds **`TSElementTest`** covering single-value, multi-value, and cross-type equality and accessors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4cb5d1d854ac1061dd76f580742daa2d3c300f22. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


atakavci (approved):
LGTM

src/main/java/redis/clients/jedis/timeseries/TSElement.java
```diff
@@ -47,26 +41,64 @@ public double getValue() {
...
+      return values;
+    }
+
+    @Override
+    public int hashCode() {
+      return 31 * Long.hashCode(getTimestamp()) + values.hashCode();
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
atakavci: keep in mind this PR suggests user is not aware of type `MultiValueTsElement`.

`new TsElement(1L,1.0).equals(new MultiValueTsElement(1L, Collections.singletonList(1.0)) `
this gets surprising in user land. 

Right now factory does not let such `MultiValueTsElement` instance, but doesnt mean that the type we introducing just for the sake of performance holds other meanings.
i think that ☝️ `equals` should return true.

atakavci (commented):
Just one comment with the new type..

src/main/java/redis/clients/jedis/timeseries/TSElement.java
```diff
@@ -47,26 +41,61 @@ public double getValue() {
...
+
+    @Override
+    public List<Double> getValues() {
+      return values;
+    }
+
+    @Override
+    public int hashCode() {
+      return 31 * Long.hashCode(getTimestamp()) + values.hashCode();
+    }
```
atakavci, line 97: for equals/hashcode consistency;

```suggestion
    public int hashCode() {
	  if(values.size() == 1) {
        return 31 * Long.hashCode(timestamp) + 31 + Double.hashCode(values.get(0));
	  }
      return 31 * Long.hashCode(getTimestamp()) + values.hashCode();
    }
```

atakavci (changes requested):

src/main/java/redis/clients/jedis/timeseries/TSElement.java
atakavci: or else better `if(size==1) return base.hashCode()`

ggivo: @atakavci done!

atakavci (approved):
LGTM

Issue #4154, opened 2025-04-29
Best way to "renewSlotCache" in JedisCluster?


taowang487:
## Problem
Hi Jedis experts! I am testing Jedis when there is a master node failure. I noticed that when the master is down, there is a decent chance that Jedis will throw error `redis.clients.jedis.exceptions.JedisConnectionException: Failed to connect to XXX`. This can take an extremely long time to recover, and I could not find a convenient way to manually refresh the slot cache, or let Jedis automatically refreshes it.

Here are my thoughts of how to tackle it:  
1. We can expose `renewSlotCache()` function in `JedisCluster` : [GitHub diff](https://github.com/redis/jedis/compare/master...taowang487:taowang487/renewSlotCache?expand=1). So when the client runs into the `redis.clients.jedis.exceptions.JedisConnectionException`, client can run `JedisCluster.renewSlotCache ()` then retry
2. In [ClusterConnectionProvider. getConnectionFromSlot()](https://github.com/redis/jedis/blob/93dd26c992ed0de2eaa05cd4add7fe8f72b3867b/src/main/java/redis/clients/jedis/providers/ClusterConnectionProvider.java#L174), we can ping the returned connection to check the connection. However given that Redis master fail is rare, this can be too expensive and I prefer option 1.

Any advice of what can be the best way to renew the slot cache is greatly appreciated!

## My Redis Cluster setup
3 master and 3 replicas with these configs
```
Redis-server --port $PORT --protected-mode $PROTECTED_MODE --cluster-enabled yes --cluster-config-file nodes-${PORT}.conf --cluster-node-timeout $TIMEOUT --appendonly yes --appendfilename appendonly-${PORT}.aof --appenddirname appendonlydir-${PORT} --dbfilename dump-${PORT}.rdb --logfile ${PORT}.log --daemonize yes ${ADDITIONAL_OPTIONS} --cluster-require-full-coverage no --cluster-node-timeout 15000
```

## How to reproduce
This is a minimal code, 
- Write a key to slot `5139` and sleep for 50 seconds. My time is `--cluster-node-timeout 15000`, so 50s sleep should be more than enough for the fail over to finish.
- manually run `redis-cli -p 30001 cluster shards` to find the master node for this slot, then `Redis-cli -p XXX shutdown` to kill this master
- try `cluster.getConnectionFromSlot(slot);`
- try `cluster.getConnectionFromSlot(slot);` again
```java
package com.example;

import java.util.HashSet;
import java.util.Set;

[...]

PR #4544, opened 2026-05-28
Refresh slot cache when slot connection acquisition fails

agarwalbharat:
`getConnectionFromSlot()` previously refreshed the cluster slot cache only when no pool was mapped for the requested slot. If a slot did have a mapped pool but borrowing a connection from that pool failed with a `JedisConnectionException`, the exception was propagated immediately.
This made `getConnectionFromSlot()` behave differently from normal command execution, where connection failures trigger retry handling and slot cache renewal.

Implemented the fallback in `getConnectionFromSlot()` for `JedisConnectionException` to renew the slot cache after a connection failure from the mapped slot pool, then resolve the slot again and retry connection acquisition. Preserve the existing fallback to a random reachable node when the slot is still unmapped after renewal.

---

> [!NOTE]
> **Medium Risk**
> Changes cluster slot connection selection used by command routing and direct slot APIs; behavior is more resilient but alters failure/retry timing during topology or node outages.
> 
> **Overview**
> **`ClusterConnectionProvider.getConnectionFromSlot()`** now treats a failed borrow from the mapped slot pool like other cluster paths: on **`JedisConnectionException`**, it renews the slot cache, looks up the slot again, and retries (or falls back to a random reachable node if the slot is still unmapped). Renewal and retry logic live in a new **`getConnectionFromSlotAfterRenewing()`** helper, with the original connection failure attached as a suppressed exception when retries still fail.
> 
> A **unit test** (`ClusterConnectionProviderTest`) verifies cache renewal and a successful second borrow after a stale pool throws. The test class is added to the formatter plugin includes in **`pom.xml`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 32fcf362b116d001f8b682a45f8feff39fcd875b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -543,6 +543,7 @@
                         <include>src/test/java/redis/clients/jedis/commands/unified/cluster/ClusterStringValuesCommandsTest.java</include>
                         <include>src/test/java/redis/clients/jedis/ProtocolFallbackPropagationTest.java</include>
                         <include>src/test/java/redis/clients/jedis/ConnectionHelloAuthTest.java</include>
+						<include>src/test/java/redis/clients/jedis/providers/ClusterConnectionProviderTest.java</include>
 						<include>**/VectorSet*.java</include>
 						<include>**/VectorTestUtils.java</include>
 						<include>**/VAddParams.java</include>
--- a/src/main/java/redis/clients/jedis/providers/ClusterConnectionProvider.java
+++ b/src/main/java/redis/clients/jedis/providers/ClusterConnectionProvider.java
@@ -19,6 +19,7 @@ import redis.clients.jedis.JedisClusterInfoCache;
 import redis.clients.jedis.annots.Experimental;
 import redis.clients.jedis.csc.Cache;
 import redis.clients.jedis.exceptions.JedisClusterOperationException;
+import redis.clients.jedis.exceptions.JedisConnectionException;
 import redis.clients.jedis.exceptions.JedisException;
 
 import static redis.clients.jedis.RedisClusterClient.INIT_NO_ERROR_PROPERTY;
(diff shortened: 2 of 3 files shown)


ggivo:
@agarwalbharat                                                                                                                                                                              
  Thanks for spending time on this issue! As far as I see, this PR addresses https://github.com/redis/jedis/issues/4154.                                                                      
                                                                                                                                                                                              
  To get the Java Format Check green, the newly added test file needs to be opted into the formatter. Please add it to the `<includes>` block of the formatter-maven-plugin configuration in    
  pom.xml (https://github.com/redis/jedis/blob/master/pom.xml#L527):                                                                                                                          
                                                                                                                                                                                              
`  <include>src/test/java/redis/clients/jedis/providers/ClusterConnectionProviderTest.java</include>     `                                                                                      
                                             then to auto format it you can use                                                                                                                                                                                              
`  mvn formatter:format.`                                                                                                                                                                
                                                                                                                                                                                              
 I'll do a closer review pass shortly — thanks again!

agarwalbharat:
> To get the Java Format Check green, the newly added test file needs to be opted into the formatter. Please add it to the `<includes>` block of the formatter-maven-plugin configuration in

Thank you for the suggestion. I’ve added the file to the formatter includes and applied the formatting changes.

ggivo:
**Not merging this PR — superseded by #4555.**  Recording the rationale here for traceability.                                                                                                                               

**Why we're not taking this approach:**                                                                                                                                                                
`getConnectionFromSlot()` is reached via the command path (`doExecuteCommand → resolver.resolve → getConnection(args) → getConnectionFromSlot`). `ClusterCommandExecutor.handleConnectionProblem` deliberately throttles renewal — only after **2 consecutive failures + backoff** — specifically to avoid renewal storms.

This PR makes the *first* failed `getResource()` trigger `renewSlotCache()` immediately, with no backoff and no gate. `JedisClusterInfoCache.renewClusterSlots()` relies on `rediscoverLock.tryLock()`, which dedups *concurrent* renewals but not *serial* ones — so under concurrency a brief network hiccup can fan out into back-to-back renewals on the command path.                                  
                                                                                                                                                                                                                                                                                 
**Agreed direction** (see [issue discussion](https://github.com/redis/jedis/issues/4154#issuecomment-4577987773) :                                                                                     
  1. **Short term** — expose a manual `renewSlotCache()` for *direct* `getConnectionFromSlot()` callers (which bypass `ClusterCommandExecutor` and so don't get `handleConnectionProblem` today). **#4555**   
  2. **Longer term** — redesign topology refresh per the TODO at `ClusterCommandExecutor.java:338`: per-node success/failure tracking with a configurable grace period. Tracked separately as its own effort.  
  
  Thanks @agarwalbharat for the analysis that led here — closing this in favor of #4555. Will follow up on the topology-refresh redesign in a dedicated issue.

PR #4547, opened 2026-05-29
Fix double initialization of pooled Connection in TrackingConnectionPool

ggivo:
## What

When you get a `Connection` from a `TrackingConnectionPool` (i.e., every `MultiDbClient` database), it was being initialized twice: once by `Connection.Builder.build()`, and a second time by `FailFastConnectionFactory.makeObject()`.

Two visible effects, per pooled connection:
- An extra `HELLO`/`CLIENT` round-trip on the wire.
- The default `PUBSUB_CONSUMER` registered twice on the push-consumer chain.

Present since v7.0.0.

## Why it happened

`Connection.Builder.build()` has always initialized the connection it returns. The pool factory then ran initialization a second time on top, after `super.makeObject()` had already done it.

## Fix

Split construction from initialization so the pool factory can interpose without re-initializing:

- New protected `createConnection()` hook on `Connection.Builder` — subclasses (e.g. `CacheConnection.Builder`) override this to pick the concrete type.
- Package-private `buildUninitialized()` lets `ConnectionFactory` build without auto-init.
- `ConnectionFactory` exposes a protected `initialize()` hook that `FailFastConnectionFactory` overrides to wrap the (now sole) init in its existing tracking — which, as a side benefit, finally makes `forceDisconnect()` able to interrupt a connection mid-handshake.

Public API is unchanged.

## Tests

New `TrackingConnectionPoolInitTest` checks that a pool-obtained connection has exactly one `PUBSUB_CONSUMER` registered. Fails on the buggy version, passes here.

## Follow-up

`Connection.initializeFromClientConfig()` has no out-of-package caller after this — its visibility can be tightened in a separate PR for the next major release.

---

> [!NOTE]
> **Medium Risk**
> Touches core connection pooling and handshake initialization paths used by MultiDbClient; behavior fix is targeted but affects every pooled connection lifecycle.
> 
> **Overview**
> Fixes **double initialization** of pooled `Connection` instances from `TrackingConnectionPool` (used by multi-DB clients): `Connection.Builder.build()` and the pool factory both ran `initializeFromClientConfig()`, causing an extra HELLO/CLIENT handshake and **duplicate `PUBSUB_CONSUMER` registration**.
> 
> **Construction vs initialization** is split without changing public `build()` behavior: `Connection.Builder` adds `createConnection()` (subclass hook, e.g. `CacheConnection`) and package-private `buildUninitialized()`; `ConnectionFactory` builds uninitialized connections and calls a new protected `initialize()` that runs init once.
> 
> `FailFastConnectionFactory` drops the second init in `makeObject()` and **overrides `initialize()`** to add/remove connections in `factoryTrackedObjects` during the sole init, so `forceDisconnect()` can interrupt in-flight HELLO/AUTH/CLIENT handshakes.
> 
> Adds `TrackingConnectionPoolInitTest` to assert single init and a single pub/sub push consumer on borrowed connections.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3d1b06bb6dce7e12286ad948233434a4ea236386. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -54,9 +54,49 @@ public class Connection implements Closeable {
     }
 
     public Connection build() {
-      Connection conn = new Connection(this);
+      Connection conn = createConnection();
       conn.initializeFromClientConfig();
       return conn;
+    }
+
+    /**
+     * Construct a fresh, uninitialized {@link Connection} for this builder.
+     * <p>
+     * Template-method hook for subclasses to return their concrete {@code Connection} subtype
+     * (see {@link redis.clients.jedis.csc.CacheConnection.Builder} for an example). The
+     * returned instance has its socket factory and client config wired up, but the network
+     * handshake has not run yet.
+     * <p>
+     * Two callers drive initialization differently:
+     * <ul>
+     *   <li>{@link #build()} — the public entry point — invokes
+     *       {@link Connection#initializeFromClientConfig()} on the returned instance before
+     *       handing it back, so external callers receive a ready-to-use connection.</li>
+     *   <li>{@link #buildUninitialized()} — the package-private accessor used by pool
+     *       factories — returns the uninitialized instance so the factory can interpose
+     *       between construction and initialization (e.g. to track an in-flight handshake
+     *       so a forced disconnect can interrupt it).</li>
+     * </ul>
+     * <p>
+     * Implementations must be free of side effects: do not open sockets, send commands, or
+     * invoke {@link Connection#initializeFromClientConfig()} — running initialization here
+     * would break the {@code buildUninitialized()} contract that pool factories rely on.
+     *
+     * @return a freshly constructed, uninitialized {@code Connection}
+     */
+    protected Connection createConnection() {
+      return new Connection(this);
+    }
+
+    /**
+     * Package-private entry point for pool factories that need to interpose between construction
+     * and initialization (e.g. to track in-flight connections so a forced disconnect can
+     * interrupt the init handshake). Callers MUST invoke
+     * {@link Connection#initializeFromClientConfig()} exactly once before handing the connection
+     * out.
+     */
+    final Connection buildUninitialized() {
+      return createConnection();
     }
   }
 
--- a/src/main/java/redis/clients/jedis/ConnectionFactory.java
+++ b/src/main/java/redis/clients/jedis/ConnectionFactory.java
@@ -151,7 +151,19 @@ public class ConnectionFactory implements PooledObjectFactory<Connection> {
   }
 
   private Connection build() {
-    return connectionBuilder.build();
+    Connection conn = connectionBuilder.buildUninitialized();
+    initialize(conn);
+    return conn;
+  }
+
+  /**
+   * Initialize a freshly built {@link Connection}.
+   * <p>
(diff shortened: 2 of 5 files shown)


uglide (approved):
LGTM

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -54,10 +54,50 @@ public JedisClientConfig getClientConfig() {
     }
 
     public Connection build() {
-      Connection conn = new Connection(this);
+      Connection conn = createConnection();
```
atakavci, line 57: this `build()` gets no use anymore.

src/main/java/redis/clients/jedis/ConnectionFactory.java
```diff
@@ -151,7 +151,19 @@ private void initAuthXManager() {
   }
 
   private Connection build() {
-    return connectionBuilder.build();
+    Connection conn = connectionBuilder.buildUninitialized();
```
atakavci, line 154: i understand the intention is to seperate the steps to obtain an instance and let the steps to be customized,, but this behavioral decision and the code should belong to builder. Otherwise builder server no purpose other then defining the type of `Connection`.  Both the factories initializes the instances themselves.

atakavci (commented):
Looks good in terms of functionality and consistent with existing behaviour.
i have no comments other than the one related to the way we are changing the design and responsibilities. The factory is kind of taking on the role of builder.
Plus, i like to stay away from changing core types(as much as possible) for specific client setup.
Right now i am trying to explore the ways if we can achieve this with changes limited to `MultiDb`.

src/main/java/redis/clients/jedis/ConnectionFactory.java
ggivo: For the record, adding alternative approach suggested [here](https://github.com/redis/jedis/compare/master...atakavci:jedis:ali/failfastfactory-double-init?expand=1)
<details>
<summary>Example : </summary>

```
public Connection build() {
      Connection conn = createConnection();
      return initialize(conn);
    }
    protected Connection createConnection() {
      return new Connection(this);
    }

    protected Connection initialize(Connection conn) {
      conn.initializeFromClientConfig();
      return conn;
    }
...
  // Minimal wrapper to expose Connection.Builder's protected initialize() method
  private static class InitWrapper extends Connection.Builder {
    private void initConnection(Connection conn) {
      super.initialize(conn);
    }

    @Override
    protected Connection initialize(Connection conn) {
      return conn;
    }
  }
...
private FailFastConnectionFactory(ConnectionFactory.Builder factoryBuilder,
        InitWrapper wrapper) {
      super(factoryBuilder.connectionBuilder(wrapper));
      this.wrapper = wrapper;
    }

    private static InitWrapper createWrapper(ConnectionFactory.Builder factoryBuilder,
        JedisClientConfig clientConfig) {
      InitWrapper wrapper = factoryBuilder.getCache() == null ? new InitWrapper()
          : new InitWrapper() {
            @Override
            protected Connection createConnection() {
              return CacheConnection.builder(factoryBuilder.getCache()).build();
            }
          };

      wrapper.socketFactory(factoryBuilder.getJedisSocketFactory()).clientConfig(clientConfig);
      return wrapper;
    }
```
</details>

PR #4548, opened 2026-05-29
Tighten Connection.initializeFromClientConfig() visibility to package-private

ggivo:
## What

Tightens `Connection.initializeFromClientConfig()` (no-arg) from `public` to package-private. Follow-up to #4547.

## Why

After #4547, this method has **no out-of-package caller** — and importantly, no external caller could legitimately use it:

| Constructor / Builder path                                | Behaviour            |
|-----------------------------------------------------------|----------------------|
| `Connection.builder().build()`                            | Initializes for you  |
| `new Connection(socketFactory, clientConfig)`             | Initializes for you  |
| `new Connection(HostAndPort, JedisClientConfig)`          | Initializes for you  |
| `new Connection()` / `(host,port)` / `(HostAndPort)` etc. | Leaves `clientConfig == null` → calling `initializeFromClientConfig()` would NPE |
| `Connection.Builder.buildUninitialized()`                 | Package-private since #4547 |

There is no public construction path that produces an uninitialized-but-configured `Connection`. Keeping the method public means: an external caller can only reach it on a fully-initialized connection (where it would re-run the handshake — the exact bug #4547 fixed).

The `protected` overload `initializeFromClientConfig(JedisClientConfig)` stays as-is — `CacheConnection` overrides it.

## Breaking change

This is a binary-incompatible change.
- Realistically, an external caller of this method on 7.x was either (a) hitting the same connection lifecycle bug that #4547 fixed,

---

> [!NOTE]
> **Medium Risk**
> Binary-incompatible API change for any external caller of the no-arg initializer; runtime behavior for supported construction paths should be unchanged.
> 
> **Overview**
> **Narrows the API surface** by making `Connection.initializeFromClientConfig()` (no-arg) **package-private** instead of public, with Javadoc stating it is an internal lifecycle step for `Builder#build()` and pool `ConnectionFactory` only. The **protected** `initializeFromClientConfig(JedisClientConfig)` overload is unchanged for subclasses like `CacheConnection`.
> 
> **Strengthens the regression test** for `TrackingConnectionPool` double-init: drops Mockito construction/verify of `initializeFromClientConfig()` and instead borrows a real connection against `TcpMockServer`, asserting **`HELLO` is sent exactly once** per pooled connection.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8c607e26af46bc1c2a1248bb542ea3c76b204782. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/actions/run-tests/action.yml
+++ b/.github/actions/run-tests/action.yml
@@ -54,7 +54,7 @@ runs:
   using: 'composite'
   steps:
     - name: Checkout code
-      uses: actions/checkout@v4
+      uses: actions/checkout@v6
       with:
         repository: ${{ inputs.repository }}
         ref: ${{ inputs.ref }}
@@ -102,7 +102,7 @@ runs:
         fi
 
     - name: Set up Java
-      uses: actions/setup-java@v4
+      uses: actions/setup-java@v5
       with:
         java-version: ${{ inputs.java_version }}
         distribution: ${{ inputs.java_distribution }}
@@ -115,7 +115,7 @@ runs:
         make compile-module
 
     - name: Cache dependencies
-      uses: actions/cache@v4
+      uses: actions/cache@v5
       with:
         path: |
           ~/.m2/repository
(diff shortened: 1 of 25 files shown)


uglide (approved):
LGTM

atakavci (approved):
LGTM!

PR #4549, opened 2026-05-30
Mark connections broken on protocol I/O failures

### LINKED ISSUE #4278: Input Stream Not Cleaned Up on Errors May Lead to Buffer Corruption
## Summary

We encountered a data consistency issue after our application triggered multiple OutOfMemoryError exceptions due to an unrelated bug. Notably, these OOMs did not immediately crash the system—instead, the JVM continued running in a degraded state. 
This led to a situation where JedisCluster was left with a contaminated input buffer causing subsequent operations to read stale or incorrect responses from Redis.

## Technical Explanation
We traced the core pattern responsible for this issue to redis.clients.jedis.Connection:

```java
public Object executeCommand(final CommandArguments args) {
    sendCommand(args);
    return getOne();
}
```

If an OutOfMemoryError occurs after sending the command but before reading the response, the response from Redis remains in the input buffer. Subsequent commands may then read this leftover data instead of their own response. In our case, we saw multiple OOM exceptions over time, amplifying the buffer contamination effect.

```java
public Object executeCommand(final CommandArguments args) {
    sendCommand(args); // 1. Command sent successfully
    // 2. OutOfMemoryError thrown here!
    return getOne();  // 3. This never executes - response stays in buffer
}
```
### Example:
```java
// Operation 1
jedis.get("key-1");
// Redis responds: "data-1"
// OutOfMemoryError (or other exception) thrown before reading "data-1"
// "data-1" remains in input buffer

// Operation 2
jedis.get("key-2");
// Redis responds: "data-2"
// getOne() reads leftover "data-1" instead of "data-2"
```

## Questions
- Should JedisCluster proactively purge/reset its input stream after any non-JedisException failure (e.g. OOM during the command/response) or before issuing the next command?
- If not, how to prevent or recover from input buffer contamination?

## Steps to Reproduce
Reproducing this issue in a real-world scenario is challenging due to the unpredictable nature of the OutOfMemoryError. To make investigation easier, I created a small piece of code that reliably showcases the buffer contamination problem:

1. Clone [LeFilou/jedis-oom](https://github.com/LeFilou/jedis-oom).
2. Follow the instructions in the repository to run the reproduction scenario.
3. Buffer corruption does not occur on every single run, but it happens quite often when the scenario is executed multiple times.

## Redis/Jedis Configuration
- Jedis Version: 6.2.0
- Redis: AWS ElastiCache (Redis Cluster),

lh0156:
Fixes #4278.

## Summary
- Mark `Connection` broken when protocol send/read/push paths fail with `JedisConnectionException` or `Error`.
- Guard reads and writes after a connection has already been marked broken.
- Add focused unit coverage for serialization/write failures, protocol read failures, push read failures, pool invalidation, and Redis data errors that should not break the connection.

## Motivation
If a pooled connection fails while RESP bytes are being written or read, the connection can retain partial or stale protocol state. Returning that connection to the pool as healthy can allow a later borrower to read stale data.

Redis error replies are still valid protocol responses, so `JedisDataException` paths continue to leave the connection reusable.

## Notes
- The existing best-effort Redis error-line diagnostic in `sendCommand` is preserved.
- The connection is marked broken before the diagnostic read, so a failure during diagnostic enrichment cannot return a suspect connection to the pool.

## Testing
- `/Users/developseop/.m2/wrapper/dists/apache-maven-3.9.15/9925cc1d/bin/mvn -B -Dwith-param-names=true -Dtest=ConnectionErrorHandlingTest test`
- `/Users/developseop/.m2/wrapper/dists/apache-maven-3.9.15/9925cc1d/bin/mvn -B -Dwith-param-names=true -Dnet.bytebuddy.experimental=true test`

---

> [!NOTE]
> **High Risk**
> Changes behavior on every command path through core `Connection` and connection-pool invalidation; misclassification of broken vs healthy could return stale connections to the pool or over-invalidate.
> 
> **Overview**
> Hardens **`Connection`** so pooled clients are not reused after RESP I/O leaves the stream in an unknown state (fixes #4278).
> 
> **`Connection`** now records a **`brokenCause`** (first failure wins) and routes connect, send, flush, read, and push paths through **`markBroken`**, which sets the broken flag and rethrows the original exception. **`sendCommand`** and **`flush`** refuse further writes on a broken connection; read helpers refuse reads and attach **`brokenCause`** to guard **`JedisConnectionException`**s. **`JedisDataException`** from a fully parsed Redis error reply no longer marks the connection broken; transport/mid-read failures and **`Error`**s still do. Write failures still attempt the optional Redis error-line diagnostic after marking broken. **`executeCommand(CommandObject)`** only builds the reply after a successful read, so builder failures after a good read do not break the socket.
> 
> Adds **`ConnectionErrorHandlingTest`** (fake sockets, pool invalidation) and registers it in the formatter **`pom.xml`** include list.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 641e0b132f3d7e3764a5c30a6b6bc946180e7e6c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -277,7 +277,7 @@ public class Connection implements Closeable {
 
   public Object executeCommand(final CommandArguments args) {
     sendCommand(args);
-    return readReplyAfterCommandSent();
+    return getOne();
   }
 
   public <T> T executeCommand(final CommandObject<T> commandObject) {
@@ -286,25 +286,17 @@ public class Connection implements Closeable {
     sendCommand(args);
     final Object reply;
     if (!args.isBlocking()) {
-      reply = readReplyAfterCommandSent();
+      reply = getOne();
     } else {
       try {
         setTimeoutInfinite();
-        reply = readReplyAfterCommandSent();
+        reply = getOne();
       } finally {
         rollbackTimeout();
       }
     }
 
     return commandObject.getBuilder().build(reply);
-  }
-
-  private Object readReplyAfterCommandSent() {
-    try {
-      return getOne();
-    } catch (Error err) {
-      throw markBroken(err);
-    }
   }
 
   public void sendCommand(final ProtocolCommand cmd) {
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -277,21 +277,33 @@ public class Connection implements Closeable {
 
   public Object executeCommand(final CommandArguments args) {
     sendCommand(args);
-    return getOne();
+    return readReplyAfterCommandSent();
   }
 
   public <T> T executeCommand(final CommandObject<T> commandObject) {
     final CommandArguments args = commandObject.getArguments();
+
     sendCommand(args);
+    final Object reply;
     if (!args.isBlocking()) {
-      return commandObject.getBuilder().build(getOne());
+      reply = readReplyAfterCommandSent();
     } else {
       try {
         setTimeoutInfinite();
-        return commandObject.getBuilder().build(getOne());
+        reply = readReplyAfterCommandSent();
       } finally {
         rollbackTimeout();
       }
+    }
+
+    return commandObject.getBuilder().build(reply);
+  }
+
+  private Object readReplyAfterCommandSent() {
+    try {
+      return getOne();
+    } catch (Error err) {
+      throw markBroken(err);
     }
   }
 
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -556,7 +556,12 @@ public class Connection implements Closeable {
 
     try {
       return protocolRead(inputStream, pushConsumers);
-    } catch (JedisConnectionException exc) {
+    } catch (JedisDataException exc) {
+      // Redis error reply was fully parsed; the stream is aligned and the connection reusable.
+      throw exc;
+    } catch (RuntimeException exc) {
+      // Transport failures (JedisConnectionException) and any other unexpected failure mid-read
+      // leave the stream position indeterminate.
       throw markBroken(exc);
     } catch (Error err) {
       throw markBroken(err);
@@ -574,7 +579,7 @@ public class Connection implements Closeable {
       }
     } catch (IOException e) {
       throw markBroken(new JedisConnectionException("Failed to check buffer on connection.", e));
-    } catch (JedisConnectionException exc) {
+    } catch (RuntimeException exc) {
       throw markBroken(exc);
     } catch (Error err) {
       throw markBroken(err);
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -315,30 +315,37 @@ public class Connection implements Closeable {
   }
 
   public void sendCommand(final CommandArguments args) {
+    connect();
     try {
-      connect();
       Protocol.sendCommand(outputStream, args);
     } catch (JedisConnectionException ex) {
-      /*
-       * When client send request which formed by invalid protocol, Redis send back error message
-       * before close connection. We try to read it to provide reason of failure.
-       */
-      try {
-        String errorMessage = Protocol.readErrorLineIfPossible(inputStream);
-        if (errorMessage != null && errorMessage.length() > 0) {
-          ex = new JedisConnectionException(errorMessage, ex.getCause());
-        }
-      } catch (Exception e) {
-        /*
-         * Catch any IOException or JedisConnectionException occurred from InputStream#read and just
-         * ignore. This approach is safe because reading error message is optional and connection
-         * will eventually be closed.
-         */
-      }
-      // Any other exceptions related to connection?
       setBroken();
-      throw ex;
+      throw enrichWithRedisErrorLine(ex);
+    } catch (RuntimeException ex) {
+      throw markBroken(ex);
+    } catch (Error err) {
+      throw markBroken(err);
     }
+  }
+
+  private JedisConnectionException enrichWithRedisErrorLine(JedisConnectionException ex) {
+    /*
+     * When client send request which formed by invalid protocol, Redis send back error message
+     * before close connection. We try to read it to provide reason of failure.
+     */
+    try {
+      String errorMessage = Protocol.readErrorLineIfPossible(inputStream);
+      if (errorMessage != null && !errorMessage.isEmpty()) {
+        return new JedisConnectionException(errorMessage, ex.getCause());
+      }
+    } catch (Exception e) {
+      /*
+       * Catch any IOException or JedisConnectionException occurred from InputStream#read and just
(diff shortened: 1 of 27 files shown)


ggivo:
Thanks for the PR — the team will take a look as soon as we have free cycles.

In the meantime, to get the formatting checks passing, please add the new files to the formatter config in pom.xml and run mvn format over them. That should resolve the current CI failures.

lh0156:
Thanks for the guidance. I added the new test file to the formatter config, ran the formatter, and pushed the fix in 724d5ee8.

I also verified locally with:
- mvn formatter:validate
- mvn -Dtest=ConnectionErrorHandlingTest test

The GitHub Actions workflows on the latest commit currently show as action_required, so I’m waiting for them to be approved/rerun.

ggivo:
@lh0156 
Thanks. Just triggered the  CI

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -275,30 +275,37 @@ public void sendCommand(final ProtocolCommand cmd, final byte[]... args) {
   }
 
   public void sendCommand(final CommandArguments args) {
+    connect();
     try {
-      connect();
       Protocol.sendCommand(outputStream, args);
```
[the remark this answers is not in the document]
lh0156: Addressed in 16f174b7 by adding the broken-write guard after connect() and before Protocol.sendCommand(...), so reconnect can still clear the broken state. I also added focused coverage for direct sendCommand rejection, no bytes written, no argument serialization/diagnostic read, reconnect preservation, and pooled invalidation on close.

ggivo:
@lh0156 
Made one round of review, I think the review looks good. Want to make another round over the tests introduced before final approval, since it does not seem that we had good coverage over error propagations.

adding also @atakavci for another pair of review

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -275,30 +275,41 @@ public void sendCommand(final ProtocolCommand cmd, final byte[]... args) {
...
+    if (broken) {
+      throw new JedisConnectionException("Attempting to write to a broken connection.");
+    }
+
     try {
-      connect();
       Protocol.sendCommand(outputStream, args);
     } catch (JedisConnectionException ex) {
+      setBroken();
+      throw enrichWithRedisErrorLine(ex);
```
atakavci: nit:
```suggestion
      throw markBroken(enrichWithRedisErrorLine(ex));
```
and other uses of it same way.

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -275,30 +275,41 @@ public void sendCommand(final ProtocolCommand cmd, final byte[]... args) {
   }
 
   public void sendCommand(final CommandArguments args) {
+    connect();
```
atakavci, line 320: leaving `connect` outside of try-block rings the bells... Didnt investigated throguh the possible use-cases but it would leave the door open to an `Error` escaping from both `connect` and `sendCommand` without marking it broken.

atakavci (commented):
i suggest to investigate if `connect` outside of tyry-block 💯 safe

src/main/java/redis/clients/jedis/Connection.java
ggivo: I will second @atakavci  here.

lh0156: @atakavci @ggivo 
Good point, thank you. I agree that `connect()` can currently let unchecked failures escape without marking the connection broken.

After looking at this again, I handled this inside `connect()` rather than moving `connect()` into the `sendCommand()` try block. `sendCommand()` primarily owns writing a command to the Redis protocol stream, while `connect()` owns socket creation, timeout initialization, and input/output stream setup. Handling both in the same catch block could make the failure boundary less clear.

This covers the `sendCommand()` path as well as direct callers such as `Jedis.connect()`, `initializeFromClientConfig()`, and `setTimeoutInfinite()`. It also fits the existing cleanup flow in `connect()`, where the `finally` block closes the socket when `broken` is set during connection initialization failure.

I also added focused coverage for direct `connect()` failures, lazy `sendCommand()` connect failures, partial socket initialization cleanup, and pooled connection invalidation.

src/main/java/redis/clients/jedis/Connection.java
lh0156: Thanks, I addressed this.

I replaced the repeated `setBroken(); throw ...` patterns with `markBroken(...)` helpers and applied the same style to the other matching paths.

For the `sendCommand()` path, I kept the connection marked broken before the diagnostic Redis error-line read is attempted:

```java
throw enrichWithRedisErrorLine(markBroken(ex));
```

That keeps the connection safe even if the diagnostic read itself fails.

I also added focused coverage for direct `connect()` failures, lazy `sendCommand()` connection failures, command write failures where diagnostic read also fails, timeout setter failures, partial socket initialization cleanup, and pooled connection invalidation.

Verified locally with:
- `mvn formatter:validate`
- `mvn -Dwith-param-names=true -Dtest=ConnectionErrorHandlingTest test`
- `mvn -Dwith-param-names=true clean test`

ggivo:
@lh0156

Nice work!

I did another review pass, and I think the main remaining gap is hardening the `Error` (OutOfMemoryError example) handling described in the original issue.

Specifically, there is still a window in `executeCommand() `where an Error can occur after the command has been successfully sent (`sendCommand(args)`) but before the response is consumed (`getOne()`). In that case, the connection may be left in an unknown protocol state and later returned to the pool without being marked as broken.

lh0156:
Thanks for the review. I addressed the remaining `executeCommand()` window.

After a command is sent, reply consumption now goes through `readReplyAfterCommandSent()`, which marks the connection broken if an `Error` is thrown before or during reply consumption.

For `CommandObject` execution, the code now reads the reply first and only invokes the response builder after the reply has been consumed. That keeps the protocol-state boundary explicit: `Error`s while consuming the Redis reply mark the connection broken, but builder lookup/build failures after the reply is consumed do not.

I also added regression coverage for the builder lookup case and verified locally with:

```bash
mvn -B -Dwith-param-names=true -Dtest=ConnectionErrorHandlingTest test
```

ggivo:
@lh0156 

Did some tests with the provided reproduction example, suggested with the original issue,  and confirmed that the buffer corruption was not observed.

Thanks for taking care of the unstable tests. They don't seem related to the current change, so I will suggest extracting them in a dedicated PR.

If you don't mind, I will take it from here, perform a small cleanup, and move on to merging it.

ggivo:
tracking [issue]( https://github.com/redis/jedis/issues/4591)  for flaky `AllKindOfValuesCommandsTestBase.touch `

ggivo (approved):
LGTM

ggivo:
@lh0156 

Merged! Thanks for seeing this through, @lh0156 

I reverted `"Stabilize touch idletime assertions" `and opened a dedicated issue (#4591) to track that effort. Feel free to jump in there if you're interested.

PR #4561, opened 2026-06-08
Improve MultiDbTransaction functionality and test coverage

atakavci:
## Summary

Improves the behavior and functionality of `MultiDbTransaction` so it matches the semantics users expect from a Redis transaction, and adds the test coverage to back it.

`WATCH`, `UNWATCH` and any pre-`MULTI` commands are now first-class operations on the transaction object — they execute immediately on a lazily-acquired connection and return their real responses to the caller, while post-`MULTI` commands stay buffered until `exec()`. This enables the canonical `watch → read → multi → write → exec` flow through a single transaction handle.

## What changed in `MultiDbTransaction`

- **Lazy connection acquisition.** With `doMulti=true`, deferred until `exec()`. With `doMulti=false`, acquired on the first command that needs the server. The same connection is reused for the whole lifecycle.
- **Unified command path.** Everything goes through `appendCommand`: `Response.of` / `Response.error` for immediate commands, buffered `Response`s for queued ones.
- **`MULTI` reply validated** by byte-array equality against `OK`; a non-`OK` reply now surfaces as `JedisDataException` instead of being silently treated as success.
- **`watch(...)` inside `MULTI` is rejected** with `IllegalStateException`. `unwatch()` inside `MULTI` is buffered like any other queued command and returns `null` (its reply arrives via `exec()`).
- **`discard()` contacts the server only when needed** — sends `UNWATCH` if a connection was already held for `WATCH`, otherwise clears local state only. Never sends `DISCARD` for a transaction that was never started server-side.
- **Database-switch protection.** The active database is pinned at acquisition. If it changes mid-transaction, the next pre-`MULTI` command and the pre-`EXEC` check both throw `JedisException("Active database has changed since transaction started")`. On `exec()` a compensating `DISCARD` is sent on the original connection before aborting.
- **Pool poisoning prevented.** If `exec()` fails leaving the server in `MULTI` (e.g. `EXECABORT`, or the compensating `DISCARD` also fails), the held connection is marked broken before release so the pool invalidates it instead of returning a mid-transaction connection to idle.

## Supporting production change

- **`Response.of` / `Response.error` promoted to `public`** so `MultiDbTransaction` (in a different package) can build already-decoded responses for pre-`MULTI` commands instead of mutating internal state.

## Tests

`MultiDbPipeline` and `MultiDbTransaction` previously had no dedicated unit or integration suites. This PR introduces both.

### `MultiDbTransaction` — new coverage for the reworked behavior

**Unit** (mocked connection + pool, no I/O):
- Happy path: `MULTI` via `executeCommand(CommandObject)` returns `OK` `byte[]`, `getMany(...)` returns `QUEUED` `byte[]`s, `EXEC` formats replies through each buffered `Response`.
- `null`-reply (WATCH-abort) path.
- `WATCH → close → UNWATCH` on a single eagerly-acquired connection.
- Illegal-state guards on `exec()` / `discard()` without `MULTI`.
- Deferred-acquisition `discard()` / `close()` never touches the pool.
- Validation-failure path still returns the borrowed connection.

**Integration** (real Redis):
- Queued commands execute; `discard()` rolls them back.
- `close()` while in `MULTI` issues a discard.
- `exec()` and `discard()` return the connection to the active pool.
- `WATCH → MULTI → EXEC` flow.
- Queue-time server rejections propagate as errors with suppressed causes.
- Failed `exec()` does not poison the pooled connection on subsequent reuse.

### `MultiDbPipeline` — baseline coverage (no production change)

**Unit:**
- Connection accounting under retry / circuit-breaker — every connection borrowed during repeated validation failures is returned to the pool.
- Transient validation failures recovered by retry.
- Connection released when `sendCommand` or `getMany` throws mid-flush.
- Empty-sync / empty-close no-ops; happy path; `close()` triggers `sync()`; ping-failure propagation.

**Integration:**
- Queued execution with mixed reply types.
- Pool return on `sync()` / `close()`.
- Idempotency of repeated `sync()` on the same pipeline.
- `Response.get()` contract — `IllegalStateException` before `sync()`, real value after.

---

> [!NOTE]
> **Medium Risk**
> Changes experimental multi-DB transaction and connection-pool behavior during failover and failed EXEC paths; mistakes could leak mid-MULTI connections or complete work against the wrong endpoint.
> 
> **Overview**
> **Reworks experimental `MultiDbTransaction`** so multi-database clients get Redis-like transaction behavior: lazy connection pinning to the active database at acquisition, immediate execution for pre-`MULTI` / `WATCH` traffic, in-memory buffering after `multi()`, and `MULTI` only sent at `exec()`.
> 
> **Safety and lifecycle:** Rejects `WATCH` inside `MULTI`; validates `MULTI` replies; aborts with `JedisException` if failover changes the active DB (compensating `DISCARD` on `exec()`); marks connections broken when the server may still be in `MULTI` after failed `exec()`; `discard()`/`close()` avoid pointless server `DISCARD` when no server transaction was started. **`Response.of` / `Response.error` are now public** for immediate pre-`MULTI` responses.
> 
> **Docs** add `MultiDbTransaction` connection and database-switch notes in `docs/transactions-multi.md`.
> 
> **Tests:** New unit and integration suites for `MultiDbTransaction` and baseline coverage for `MultiDbPipeline` (pool return, retry, error paths)—no production changes to pipeline in this diff.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd3c201d23436627c1a1d1b579aeebc69c4c1fb8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/main/java/redis/clients/jedis/mcf/MultiDbTransaction.java
+++ b/src/main/java/redis/clients/jedis/mcf/MultiDbTransaction.java
@@ -13,6 +13,8 @@ import java.util.Queue;
 import redis.clients.jedis.*;
 import redis.clients.jedis.annots.Experimental;
 import redis.clients.jedis.exceptions.JedisDataException;
+import redis.clients.jedis.exceptions.JedisException;
+import redis.clients.jedis.mcf.MultiDbConnectionProvider.Database;
 import redis.clients.jedis.util.IOUtils;
 import redis.clients.jedis.util.KeyValue;
 import redis.clients.jedis.util.SafeEncoder;
@@ -44,24 +46,25 @@ public class MultiDbTransaction extends AbstractTransaction {
 
   private static final Builder<?> NO_OP_BUILDER = BuilderFactory.RAW_OBJECT;
 
-  private final MultiDbConnectionSupplier failoverProvider;
+  private final MultiDbConnectionSupplier connectionSupplier;
   private final Queue<KeyValue<CommandArguments, Response<?>>> commands = new LinkedList<>();
 
   private Connection connection;
+  private Database initialDatabase;
   private boolean inWatch = false;
   private boolean inMulti = false;
 
   /**
-   * A user wanting to WATCH/UNWATCH keys followed by a call to MULTI ({@link #multi()}) it should
-   * be {@code doMulti=false}.
-   * @param provider
+   * A user wanting to WATCH/UNWATCH keys followed by a call to {@link #multi()} should set
+   * {@code doMulti=false}.
+   * @param provider connection provider supplying connections from the active database
    * @param doMulti {@code false} should be set to enable manual WATCH, UNWATCH and MULTI
    * @param commandObjects command objects
    */
   public MultiDbTransaction(MultiDbConnectionProvider provider, boolean doMulti,
       CommandObjects commandObjects) {
     super(commandObjects);
-    this.failoverProvider = new MultiDbConnectionSupplier(provider);
+    this.connectionSupplier = new MultiDbConnectionSupplier(provider);
 
     if (doMulti) {
       multi();
@@ -110,6 +113,9 @@ public class MultiDbTransaction extends AbstractTransaction {
       commands.add(KeyValue.of(args, response));
       return response;
     }
+    if (initialDatabase != null && !connectionSupplier.isActiveDatabase(initialDatabase)) {
+      throw new JedisException("Active database has changed since transaction started");
+    }
     try {
       return Response.of(acquireConnection().executeCommand(commandObject));
     } catch (JedisDataException e) {
@@ -156,6 +162,10 @@ public class MultiDbTransaction extends AbstractTransaction {
         }
       });
 
+      if (!connectionSupplier.isActiveDatabase(initialDatabase)) {
+        throw new JedisException("Active database has changed since transaction started");
+      }
+
       conn.sendCommand(EXEC);
 
       List<Object> unformatted = conn.getObjectMultiBulkReply();
@@ -190,7 +200,7 @@ public class MultiDbTransaction extends AbstractTransaction {
     }
 
     try {
-      // MULTI was never sent to the server (acquisition is deferred until exec()).
+      // MULTI itself is only issued from exec(), so the server never started a transaction.
       // Buffered commands only exist locally, so there is nothing to roll back server-side
       // unless we have already acquired a connection for pre-MULTI traffic (e.g. WATCH).
       if (inWatch) {
@@ -215,7 +225,8 @@ public class MultiDbTransaction extends AbstractTransaction {
 
   private Connection acquireConnection() {
     if (connection == null) {
-      connection = failoverProvider.getConnection();
+      initialDatabase = connectionSupplier.provider.getDatabase();
+      connection = connectionSupplier.getConnection();
     }
     return connection;
   }
--- a/src/main/java/redis/clients/jedis/mcf/MultiDbTransaction.java
+++ b/src/main/java/redis/clients/jedis/mcf/MultiDbTransaction.java
@@ -41,7 +41,6 @@ import redis.clients.jedis.util.SafeEncoder;
  */
 @Experimental
 public class MultiDbTransaction extends AbstractTransaction {
-  private static final byte[] QUEUED_IN_BYTES = SafeEncoder.encode("QUEUED");
   private static final byte[] OK_IN_BYTES = SafeEncoder.encode("OK");
   private static final String OK_STR = "OK";
 
@@ -143,6 +142,7 @@ public class MultiDbTransaction extends AbstractTransaction {
       throw new IllegalStateException("EXEC without MULTI");
     }
 
+    boolean serverInMultiMode = false;
     try {
       Connection conn = acquireConnection();
 
@@ -153,6 +153,8 @@ public class MultiDbTransaction extends AbstractTransaction {
             : multiReply;
         throw new JedisDataException("Unexpected response: " + response);
       }
+
+      serverInMultiMode = true;
 
       commands.forEach((command) -> conn.sendCommand(command.getKey()));
       // following connection.getMany(int) flushes anyway, so no flush here.
@@ -165,6 +167,7 @@ public class MultiDbTransaction extends AbstractTransaction {
             "Active database has changed since transaction started");
         try {
           conn.executeCommand(new CommandArguments(DISCARD));
+          serverInMultiMode = false;
         } catch (Exception e) {
           dbSwitchException.addSuppressed(e);
         }
@@ -176,6 +179,7 @@ public class MultiDbTransaction extends AbstractTransaction {
       List<Object> unformatted;
       try {
         unformatted = conn.getObjectMultiBulkReply();
+        serverInMultiMode = false;
       } catch (JedisDataException jce) {
         // A command may fail to be queued, so there may be an error before EXEC is called
         // In this case, the server will discard all commands in the transaction and return the
@@ -205,7 +209,7 @@ public class MultiDbTransaction extends AbstractTransaction {
       inMulti = false;
       inWatch = false;
       commands.clear();
-      releaseConnection();
+      releaseConnection(serverInMultiMode);
     }
   }
 
@@ -228,7 +232,7 @@ public class MultiDbTransaction extends AbstractTransaction {
       inMulti = false;
       inWatch = false;
       commands.clear();
-      releaseConnection();
+      releaseConnection(false);
     }
   }
 
@@ -247,8 +251,11 @@ public class MultiDbTransaction extends AbstractTransaction {
     return connection;
   }
 
-  private void releaseConnection() {
+  private void releaseConnection(boolean setBroken) {
     if (connection != null) {
+      if (setBroken) {
+        connection.setBroken();
+      }
       IOUtils.closeQuietly(connection);
       connection = null;
     }
--- a/src/main/java/redis/clients/jedis/mcf/MultiDbTransaction.java
+++ b/src/main/java/redis/clients/jedis/mcf/MultiDbTransaction.java
@@ -149,53 +149,42 @@ public class MultiDbTransaction extends AbstractTransaction {
       Object multiReply = conn
           .executeCommand(new CommandObject<>(new CommandArguments(MULTI), NO_OP_BUILDER));
       if (!bytesEquals(OK_IN_BYTES, multiReply)) {
-        throw new JedisDataException(
-            "Unexpected response: " + SafeEncoder.encode((byte[]) multiReply));
+        Object response = multiReply instanceof byte[] ? SafeEncoder.encode((byte[]) multiReply)
+            : multiReply;
+        throw new JedisDataException("Unexpected response: " + response);
       }
 
-      // From here until EXEC is written, the server is in MULTI state on this connection.
-      // Any exception that escapes this block must be paired with a DISCARD so the
-      // connection can be safely returned to the pool; otherwise the next borrower would
-      // receive QUEUED replies instead of their commands being executed.
+      commands.forEach((command) -> conn.sendCommand(command.getKey()));
+      // following connection.getMany(int) flushes anyway, so no flush here.
+
+      // server replies QUEUED OR ERROR for each buffered command
+      List<Object> queuedCmdResponses = conn.getMany(commands.size());
+
+      if (!connectionSupplier.isActiveDatabase(initialDatabase)) {
+        JedisException dbSwitchException = new JedisException(
+            "Active database has changed since transaction started");
+        try {
+          conn.executeCommand(new CommandArguments(DISCARD));
+        } catch (Exception e) {
+          dbSwitchException.addSuppressed(e);
+        }
+        throw dbSwitchException;
+      }
+
+      conn.sendCommand(EXEC);
+
+      List<Object> unformatted;
       try {
-        commands.forEach((command) -> conn.sendCommand(command.getKey()));
-        // following connection.getMany(int) flushes anyway, so no flush here.
-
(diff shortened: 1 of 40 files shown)


src/main/java/redis/clients/jedis/mcf/MultiDbTransaction.java
```diff
@@ -115,28 +142,40 @@ public final List<Object> exec() {
...
-      // ignore QUEUED (or ERROR)
-      connection.getMany(commands.size());
+      // server replies QUEUED for each buffered command
+      List<Object> queuedCmdResponses = conn.getMany(commands.size());
+      queuedCmdResponses.forEach((rawReply) -> {
+        if (!bytesEquals(QUEUED_IN_BYTES, rawReply)) {
+          throw new JedisDataException(
+              "Unexpected response: " + SafeEncoder.encode((byte[]) rawReply));
+        }
+      });
```
ggivo: `getMany() `might return `JedisDataException` in`queuedCmdResponses` if the command is rejected by the server, this will end up with`ClassCastException` on  `SafeEncoder.encode((byte[]) rawReply))`  and masking the real     
  server error

 ``` java
public void exec_queuedCommandRejectedByServer_surfacesServerError() {
    ConnectionPool pool = activePool();

    try (MultiDbTransaction tx = client.multi()) {
      // GET with no arguments: Redis rejects it at queue time with an arity error instead of
      // replying QUEUED. Connection.getMany stores that error reply inline as a JedisDataException.
      tx.appendCommand(
        new CommandObject<>(new CommandArguments(Protocol.Command.GET), BuilderFactory.RAW_OBJECT));

      // the original server error must propagate, not a ClassCastException
      JedisDataException ex = assertThrows(JedisDataException.class, tx::exec);
      assertTrue(ex.getMessage().toLowerCase().contains("wrong number of arguments"),
        "expected arity error, got: " + ex.getMessage());
    }

    // connection is released back to the pool on the error path
    assertEquals(0, pool.getNumActive(), "no active connections after a failed exec");
  }
```

File: src/main/java/redis/clients/jedis/mcf/MultiDbTransaction.java [hunk 2]
```diff
@@ -115,28 +142,40 @@ public final List<Object> exec() {
...
 
-      // remove extra response builders
-      for (int idx = 0; idx < extraCommandCount.get(); ++idx) {
-        commands.poll();
+      if (!connectionSupplier.isActiveDatabase(initialDatabase)) {
+        throw new JedisException("Active database has changed since transaction started");
       }
 
-      connection.sendCommand(EXEC);
+      conn.sendCommand(EXEC);
```
ggivo, line 177: An exception escaping `exec()` after `MULTI` (but before `EXEC/DISCARD`) leaves the connection in `MULTI` state server-side. Returned to the pool as-is, it poisons the next borrower — their command gets  `+QUEUED` instead of being executed.


  ``` java
@Test
  public void failedExec_doesNotPoisonPooledConnection() {
    // A queue-time rejection aborts exec() before EXEC/DISCARD is sent, leaving the server
    // mid-MULTI on the borrowed connection.
    try (MultiDbTransaction tx = client.multi()) {
      tx.appendCommand(
        new CommandObject<>(new CommandArguments(Protocol.Command.GET), BuilderFactory.RAW_OBJECT));
      assertThrows(JedisDataException.class, tx::exec);
    }

    // The connection that handled the failed transaction is back in the pool. A subsequent command
    // must execute normally; if the connection were returned mid-MULTI, SET would reply QUEUED.
    assertEquals("OK", client.set("x", "y"));
    assertEquals("y", client.get("x"));
  }
```

File: src/main/java/redis/clients/jedis/mcf/MultiDbTransaction.java [hunk 3]
```diff
@@ -115,28 +143,64 @@ public final List<Object> exec() {
...
+        if (!conn.isBroken()) {
+          try {
+            conn.sendCommand(DISCARD);
+            conn.getStatusCodeReply();
+          } catch (RuntimeException discardEx) {
+            ex.addSuppressed(discardEx);
+            conn.setBroken();
+          }
+        }
+        throw ex;
```
ggivo: **Consider letting the server handle this via `EXEC`/`EXECABORT` instead of validating `QUEUED` + manual `DISCARD`.**

Since Jedis only supports Redis 6.x+ (well past 2.6.5), a command that fails to queue is *guaranteed* to abort the transaction at `EXEC` time and discard it server-side, leaving the connection clean — no `DISCARD` needed:

> "Starting with Redis 2.6.5, the server will detect an error during the accumulation of commands. It will then refuse to execute the transaction returning an error during `EXEC`, discarding the transaction." — [Redis Transactions docs](https://redis.io/docs/latest/develop/using-commands/transactions/)

This is how the canonical `Transaction.exec()` already works: it skips `QUEUED` validation, always sends `EXEC`, and on the resulting `EXECABORT` rethrows the `JedisDataException` enriched with the queued errors as `suppressed` (`Transaction.java:180-192`).

Similar handling tested against `Transaction` is available here — `TransactionV2Test.transactionPropagatesErrorsBeforeExec` covers the queue-time-error → `EXECABORT` path end-to-end.

atakavci: why do we need to issue the rest of the commands to the server in a situation where its absolute fact it will fail in the end?
only "the behavioral alignment" with others would be the argument against it,, but IMHO this should be something to fix in other places.

Another approach would be to stop issuing more commands in case of fail, and then EXEC. But the server provides/suggests use of `DISCARD`. `EXECABORT` is a security measure but not the actual flow designed.

ggivo: Reading the flow: by the time we hit a non-`QUEUED` reply, all commands have already been sent and `getMany` has already read every reply — there's no "rest of the commands" left to avoid issuing. The only open choice is the terminator that takes the connection out of `MULTI`: `EXEC` or `DISCARD`, both one command + one reply, both leaving the connection healthy.

I'd lean `EXEC`: per the [docs](https://redis.io/docs/latest/develop/using-commands/transactions/), EXECABORT is the server's *defined* response to a queue-time error since 2.6.5 — not a safety net. `DISCARD` is meant for a client-initiated abort. `EXEC` cleans the connection on both client and server and surfaces the failure in one round-trip (EXECABORT, with the queue errors as `suppressed`), matching the canonical `Transaction.exec()`.

ggivo: **Note:** I see your point for using `DISCARD` in the "Active database has changed since transaction started" case. For queue-time errors, though, I still think propagating the `EXEC` error (`EXECABORT`) is the 
  right approach

src/test/java/redis/clients/jedis/mcf/MultiDbPipelineIT.java
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  private ConnectionPool activePool() {
+    MultiDbConnectionProvider p = ClientTestUtil.getConnectionProvider(client);
+    return p.getDatabase().getConnectionPool();
+  }
+
+  @Test
+  public void pipeline_executesQueuedCommands_andReturnsExpectedValues() {
+    int count = 10;
```
ggivo, line 71: nit: What is the idea behind using count (10) and not a single invocation for functional verification?

src/main/java/redis/clients/jedis/mcf/MultiDbTransaction.java
ggivo, line 1: I will suggest adding some notes to  [transactions docs](https://redis.github.io/jedis/transactions-multi/), or documenting `MultiDbTransaction` specifically, noting semantics around connection lifecycle and discarding the connection if the DB has switched.

src/test/java/redis/clients/jedis/mcf/MultiDbPipelineIT.java
```diff
@@ -0,0 +1,155 @@
...
+
+  @Test
+  public void pipeline_closeOnEmptyPipeline_doesNotBorrowConnection() {
+    ConnectionPool pool = activePool();
+    int idleBefore = pool.getNumIdle();
+    int activeBefore = pool.getNumActive();
+
+    try (MultiDbPipeline p = client.pipelined()) {
+      // no commands appended
+    }
```
ggivo, line 119:
```suggestion
    try (MultiDbPipeline p = client.pipelined()) {
      // no commands appended
      assertEquals(activeBefore, pool.getNumActive());
    }
```

src/test/java/redis/clients/jedis/mcf/MultiDbTransactionIntegrationTest.java
```diff
@@ -0,0 +1,198 @@
...
+import redis.clients.jedis.util.ClientTestUtil;
+
+/**
+ * Integration tests for {@link MultiDbTransaction} against a running Redis standalone instance.
+ * Verifies MULTI/EXEC/DISCARD/WATCH semantics and that pool connections are returned to the pool
+ * after {@link MultiDbTransaction#exec()} / {@link MultiDbTransaction#discard()} /
+ * {@link MultiDbTransaction#close()}.
+ */
+@Tag("integration")
+public class MultiDbTransactionIntegrationTest {
```
ggivo: For newly added tests lets converge to default *IT failsafe schema
MultiDbTransactionIntegrationTest -> MultiDbTransactionIT

src/test/java/redis/clients/jedis/mcf/MultiDbPipelineIntegrationTest.java
```diff
@@ -0,0 +1,155 @@
...
+import redis.clients.jedis.Response;
+import redis.clients.jedis.util.ClientTestUtil;
+
+/**
+ * Integration tests for {@link MultiDbPipeline} against a running Redis standalone instance.
+ * Verifies command execution semantics and that pool connections are returned to the pool after
+ * {@link MultiDbPipeline#sync()} / {@link MultiDbPipeline#close()}.
+ */
+@Tag("integration")
+public class MultiDbPipelineIntegrationTest {
```
ggivo: for newly introduced tests lets converge to failsafe default *IT
MultiDbPipelineIntegrationTest-> MultiDbPipelineIT

src/test/java/redis/clients/jedis/mcf/MultiDbPipelineUnitTest.java
```diff
@@ -0,0 +1,243 @@
...
+
+    assertThrows(Exception.class, () -> pipeline.sync());
+  }
+
+  /**
+   * Demonstrates that a Resilience4j retry replays the validation ping. With a higher maxAttempts,
+   * an intermittent ping failure is recovered and the pipeline completes successfully.
+   */
+  @Test
+  public void sync_pingFailsThenSucceeds_pipelineRecoversWithRetry() throws Exception {
```
ggivo: nit: ping is just the mechanic the behavior is connection validation. Can not come up with much better naming but can suggest something  like  sync_transientConnectionValidationFailure_recoversViaRetry, keeping the PING detail in the Javadoc.

src/test/java/redis/clients/jedis/mcf/MultiDbPipelineUnitTest.java
```diff
@@ -0,0 +1,243 @@
...
+      }
+    }
+  }
+
+  /**
+   * When the validation ping fails repeatedly and the retry policy borrows a new connection on each
+   * attempt, every borrowed connection must be returned to the pool (i.e. closed).
+   */
+  @Test
+  public void sync_pingFailsMultipleTimes_allAcquiredConnectionsReturnedToPool() throws Exception {
```
ggivo: nit: same namin suggestion sync_connectionValidationAlwaysFails_allAcquiredConnectionsReturnedToPool

src/main/java/redis/clients/jedis/mcf/MultiDbTransaction.java
```diff
@@ -6,106 +6,134 @@
...
+ * <ul>
+ * <li>Commands issued before {@code MULTI} are executed immediately on the held connection and
+ * their responses are returned to the caller.</li>
+ * <li>Commands issued after {@code MULTI} (including {@code UNWATCH}) are buffered in memory and
+ * dispatched to the server in a single batch when {@link #exec()} is called.</li>
+ * </ul>
  */
 @Experimental
 public class MultiDbTransaction extends AbstractTransaction {
+  private static final byte[] QUEUED_IN_BYTES = SafeEncoder.encode("QUEUED");
```
ggivo: not used

ggivo (commented):
Logic LGTM — left a few syntax notes inline.                                                                                                                                                             
                                                                                                                                                                                                           
  - Database-switch coverage is missing — added it in a stacked PR but might be further extended: https://github.com/atakavci/jedis/pull/10. It also disables health checks in these ITs (no failover tested), removing the ~1s per-test 
  startup wait in toital ~20s.                                                                                                                                                                                            
  - We should review/update the transaction docs for the MultiDb case (WATCH/UNWATCH flow, deferred MULTI, fail-fast on database switch). 
  - PR title is a bit misleading since this change actually introduces an enhancement of the current MultiDBTransaction behavior

src/test/java/redis/clients/jedis/mcf/MultiDbPipelineIT.java
atakavci: it tries to populate more commands and more bits on to the wire to make sure multiple commands are sent and responses are read without any potential failure, just by issuing a small but large-enough number of commands. !0 is not a magic number but an attempt to compromise between 1 and 1000, in terms of multiplicity and speed.

ggivo (approved):
LGTM

Issue #2052, opened 2019-09-01
Getting JedisConnectionException: Could not get a resource from the pool When connecting to remote server


sohailzahid:
I am trying to connect from a Java App running on Windows 10 using Jedis client to a remote Redis instance. My program fails with exception JedisConnectionException.

### Environment
- Redis version 5.x running on Ubuntu 19.04 in a VM
- Redis is configured to listen on a specific IP with a password and with protected mode set to on/yes.
- I am running my dev environment on a Windows 10 Pro machine.
- Jedis version is 3.1.0
- JDK is OpenJDK 11.0.4 on both Windows & Linux.

### What I have tried

- I have tested connecting to the remote (running on VM) Redis server using local redis-cli and was successful, confirming that I am able to access, connect and use the redis-server.
- I ran my test program from Netbeans 11.1 IDE running on another Linux VM on the same host with different IP and that was also a success.
- I ran my test program (100% same code) from Netbeans 11.1 IDE running on the host Windows 10 machine and the program failed to connect and threw the exception.

### Actual behavior

Write here what happens instead ...

### Steps to reproduce:

Code Snippet 

pool = new JedisPool( config, "192.168.199.130", 6379, Protocol.DEFAULT_TIMEOUT,"foobar", Protocol.DEFAULT_DATABASE); 
Jedis jedis = pool.getResource(); // this statement throws the exception 
jedis.close();

JedisPool.getResource() works on Linux while same code fails from Windows.
Did anyone else run into this?
Any and all help appreciated. :-)

josearaujo:
Do you find some solution for your problem?

sohailzahid:
> Do you find some solution for your problem?

I did.  Thank you.

The issue was my environment configuration.  I maybe have made changes to the code but I do not recall for sure.

Appreciate you checking.  Thanks.

josearaujo:
I was able to find a solution changing this
- REDIS_MASTER_HOST=127.0.0.1
      

thanks :)

sazzad16:
Resolved

PR #4566, opened 2026-06-18
Add 8.10 preview build and adjust RTS align test

uglide:
> [!NOTE]
> **Low Risk**
> Changes are limited to CI/Makefile env wiring and test expectations; no production client API or runtime behavior is modified.
> 
> **Overview**
> Adds **Redis 8.10** to the Docker CI matrix, `Makefile` supported test versions, and a new `src/test/resources/env/.env.v8.10` profile (`8.10-m01-int`) so the client libs test image can run against the 8.10 preview stack.
> 
> **TimeSeries `TS.RANGE` ALIGN coverage** is updated: `align()` now uses a non-keyword key (`ts:align`) and asserts distinct bucket timestamps for default alignment, `ALIGN start`, `ALIGN end`, and numeric `ALIGN 5`. A new `alignKeyNotConsumedAsOption()` test (gated with `@SinceRedisVersion("8.10.0")`) uses the key `"align"` to lock in the RedisTimeSeries option-parsing fix (PR #2052), where the series key must not be mistaken for the `ALIGN` option.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 021c5dc9d687128e9a8e64271170d343ac0eeec7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/test-on-docker.yml
+++ b/.github/workflows/test-on-docker.yml
@@ -40,6 +40,7 @@ jobs:
       fail-fast: false
       matrix:
         redis_version:
+          - "8.10"
           - "8.8"
           - "8.6"
           - "8.4"
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 PATH := ./redis-git/src:${PATH}
 
 # Supported test env versions
-SUPPORTED_TEST_ENV_VERSIONS := 8.8 8.6 8.4 8.2 8.0 7.4 7.2 6.2
+SUPPORTED_TEST_ENV_VERSIONS := 8.10 8.8 8.6 8.4 8.2 8.0 7.4 7.2 6.2
 DEFAULT_TEST_ENV_VERSION := 8.8
 REDIS_ENV_WORK_DIR := $(or ${REDIS_ENV_WORK_DIR},/tmp/redis-env-work)
 TOXIPROXY_IMAGE := ghcr.io/shopify/toxiproxy:2.8.0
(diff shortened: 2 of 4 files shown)


ggivo (commented):
The whole `align()` test is now gated behind `@SinceRedisVersion("8.10.0")`, so on every server older than 8.10 we lose **all** basic ALIGN-mode coverage (start/end/explicit), not just the regression case. Those modes work fine on older servers — only the keyword-named-key case depends on the            
  RedisTimeSeries #2052 fix.                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                   
  Proposal: split into two tests so the generic coverage stays version-agnostic and only the regression guard is gated.                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                   
  Diff against this PR's branch: https://github.com/redis/jedis/compare/im/adjust-rts-tests-for-8_10...topic/ggivo/adjust-rts-tests-for-8_10                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                   
  - `align()` — non-keyword key (`ts:align`), **unguarded**, runs on all versions → keeps ALIGN-mode coverage everywhere.                                                                                                                                                                                          
  - `alignKeyNotConsumedAsOption()` — dedicated to the #2052 fix, key named `align`, keeps the `@SinceRedisVersion("8.10.0")` guard.

And I think the guard should be  against  module version vs the server version

PR #4572, opened 2026-06-24
[maintenance events] Alternative approach for handling timeout changes via TimeoutSupplier

atakavci:
> [!NOTE]
> **Medium Risk**
> Touches core connection read paths, pool return/eviction, and maintenance/MOVING behavior; behavior changes (e.g. relaxed blocking default, always applying timeout before reads) warrant careful regression testing.
> 
> **Overview**
> Replaces inline `SO_TIMEOUT` fields and `SoTimeoutSupplier` with a **chain-of-responsibility** model: `DefaultTimeoutSource` holds baseline timeouts, `ExpiringTimeoutSource` applies time-bounded maintenance relaxation, and a nested `RebindTimeoutSource` overlays pool-wide MOVING rebind timeouts. `Connection` resolves blocking vs non-blocking reads via `defaultTimeoutSource.get()` and applies the socket timeout on each protocol read.
> 
> **Maintenance setup** moves out of `Connection` into `MaintenanceAwareVisitor` (`InitVisitor`), which registers push consumers, runs `CLIENT MAINT_NOTIFICATIONS`, and wires the timeout chain. `Connection.Builder` / `ConnectionFactory` use visitors instead of injecting listeners and `SoTimeoutSupplier` directly.
> 
> **Pool behavior:** `ConnectionPool` uses a return hook to discard connections whose peer is in an active rebind (`isAffected(Connection)`). MOVING receivers are marked with `expireAt` instead of `requestRebind()`. `JedisClientConfig.UNSET_TIMEOUT_MS` is removed; relaxed blocking default is `0`. `MaintenanceNotificationsConfig` is immutable with renamed getters (`getRelaxedTimeout`, optional `eventListener`). Tests use a shared `NanoClock.INSTANCE` for time control.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fb58f865b9168d9d3668520675c81e8d3dd813d4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -495,16 +440,14 @@ public <T> T executeCommand(final CommandObject<T> commandObject) {
     final CommandArguments args = commandObject.getArguments();
     sendCommand(args);
     if (!args.isBlocking()) {
+    applySoTimeout(currentTimeoutInfo().timeout);
```
ggivo: Need to consider also other code paths like
`public Object executeCommand(final CommandArguments args)`
and aprobably also protocolReadPushes

Let's discuss if we should either gate them or have applyTimeout triggered before protocol read operations

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -381,41 +337,7 @@ public int getSoTimeout() {
    * @see #getSoTimeout()
    */
   public int getBlockingSoTimeout() {
```
ggivo, line 331: I think to preserve backword compatibility this should return the configured default blocking timeout and not actual applied one.

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -433,8 +355,36 @@ public int getRelaxedBlockingSoTimeout() {
...
+   * Pushes {@code timeout} to the socket only when it differs from the value already applied. The
+   * socket's SO_TIMEOUT is sticky, so caching the last-applied value lets the hot read path skip the
+   * syscall whenever the effective timeout is unchanged (the overwhelmingly common case).
+   */
+  private void applySoTimeout(int timeout) {
+    if (timeout == appliedSoTimeout) {
+      return;
+    }
+    try {
+      socket.setSoTimeout(timeout);
```
ggivo, line 363: should we check if socket is initialised

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -1231,23 +1103,17 @@ private void applyTimeout(int desired) {
...
    */
   @Experimental
-  public void relaxTimeouts(Duration period) {
-    long deadline = clockNanos.getAsLong() + period.toNanos();
-    if (relaxedUntilNanos == 0 || deadline - relaxedUntilNanos > 0) {
-      relaxedUntilNanos = deadline;
-    }
-    applyTimeout(effectiveSoTimeout());
+  public void relaxTimeouts(TimeoutInfo info) {
+    relaxedTimeouts.add(timeoutSupplier.push(info));
```
ggivo: if reading the code correctly, seems a timeout card can expire and get removed, but it will remain in relaxedTimeouts

src/main/java/redis/clients/jedis/MaintenanceEventController.java
```diff
@@ -227,16 +223,31 @@ public String toString() {
    * set is always immutable; updates swap the whole state.
    */
   private static final class RebindState {
+
+    static LongSupplier CLOCK_NANOS = System::nanoTime;
```
ggivo: I  would keep it per instance, it is  used for testing and it make sense different test to have different instance.

src/main/java/redis/clients/jedis/SimpleTimeoutSupplier.java
```diff
@@ -0,0 +1,26 @@
+package redis.clients.jedis;
+
+class SimpleTimeoutSupplier implements TimeoutSupplier {
```
ggivo: What if we use the AdvancedTimeoutSupplier  with defautls only?
Curious what will be the overhead on performance

src/main/java/redis/clients/jedis/Connection.java
ggivo: I think we still need applyTimeouts, to ensure timeouts for command output reads  upon receiving  maintence event.

ggivo (commented):
Direction looks good to me, will need some polish

I also forgot to add one additional requirement around maintenance events relaxation, 
we need to ensure we are not reducing the original configured on the connection in case relaxed timeout end's up to be lower then configured

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -1231,23 +1103,17 @@ private void applyTimeout(int desired) {
...
-    relaxedUntilNanos = 0;
-    applyTimeout(effectiveSoTimeout());
-  }
-
-  void setClockNanos(LongSupplier clockNanos) {
-    this.clockNanos = clockNanos;
+    TimeoutCard card = relaxedTimeouts.remove(0);
+    if(card != null) {
+      timeoutSupplier.remove(card);
+    }
```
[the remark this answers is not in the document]
atakavci: its fine; connection knows the exact moments when there is a need to consume value from timeoutSupplier

src/main/java/redis/clients/jedis/Connection.java
atakavci: now it knows about consuming the supplier
```java
  public Object executeCommand(final CommandArguments args) {
    sendCommand(args);
    applySoTimeout(currentTimeoutInfo().timeout);
    ...
```

src/main/java/redis/clients/jedis/Connection.java
atakavci: switched to single(last) reference instead of list/collection

src/main/java/redis/clients/jedis/SimpleTimeoutSupplier.java
atakavci: droppped simple one

src/main/java/redis/clients/jedis/Connection.java
atakavci: connection will know when to consume it from supplier instance

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -381,41 +337,7 @@ public int getSoTimeout() {
...
-   * <p>It is specifically used for blocking Redis commands (e.g. BLPOP, BRPOP) during
-   * maintenance periods where the connection may require different timeout behavior.</p>
-   *
-   * @return the relaxed timeout in milliseconds for blocking operations during maintenance
-   * @see #getRelaxedSoTimeout()
-   * @see #isRelaxedTimeoutActive()
-   */
-  public int getRelaxedBlockingSoTimeout() {
-    return relaxedBlockingTimeout;
+    return currentTimeoutInfo().blockingTimeout;
```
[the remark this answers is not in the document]
atakavci: dropped method

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -688,11 +636,11 @@ public List<Object> getUnflushedObjectMultiBulkReply() {
     return (List<Object>) readProtocolWithCheckingBroken();
   }
 
-  @SuppressWarnings("unchecked")
```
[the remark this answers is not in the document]
atakavci: `readProtocolWithCheckingBroken` should only use the socket with as-is settings. changing socket config (without the context it is running in) directly will be a potential source of issue.

src/main/java/redis/clients/jedis/MaintenanceEventController.java
atakavci: check `NanoClock` , i believe it looks good enough while keeping performance high as possible.

src/main/java/redis/clients/jedis/MaintenanceEventController.java
```diff
@@ -106,21 +100,27 @@ public void onMoving(MovingEvent e, Connection c) {
...
       return; // receiver socket already closed; no peer to register
     }
     SocketAddress target = new InetSocketAddress(e.target.getHost(), e.target.getPort());
-    long deadline = clockNanos.getAsLong() + ttlNanos;
 
     while (true) {
-      RebindState cur = rebind.get();
-      if (cur != null && e.seq < cur.seq) {
+      RebindState cur = rebind;
+      if (e.seq < cur.seq) {
```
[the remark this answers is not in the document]
atakavci: @ggivo 
we discussed it is not possible to have multiple MOVING operations happening at the same time. This suggests that multiple different nodes can not go into rebinding state anyway on server side. Could you confirm? If yes lets skip this one from AI.

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -94,12 +85,25 @@ List<MaintenanceEventListener> getMaintenanceEventListeners() {
...
     public Connection build() {
       Connection conn = createConnection();
+      // builder takes the initiative to wire up the maintenance-aware visitor if maintenance is
+      // configured
+      if (maintenanceConfig != null) {
+        if (visitors.stream().anyMatch(v -> v instanceof MaintenanceAwareVisitor)) {
+          visitors.add(
+            new MaintenanceAwareVisitor(this, MaintenanceEventController.from(maintenanceConfig)));
+        }
+      }
```
atakavci, line 95: @ggivo 
if user creates and uses same `connection.builder` to create its connections, it doesnt require a pool anymore. 
one thing that would not work is early cut-out of connections to the same rebinding node.

functionally it is pretty good, since maintenance events is the new default. on the other side, not very sure if connection.builder should take the responsilbilty. WDYT?

src/main/java/redis/clients/jedis/MaintenanceEventController.java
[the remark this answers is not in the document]
ggivo: @atakavci 
Yes, my reading is the same.
MOVING notifications are received sequentially after the previous rebind operation completes.
Handling an older MOVING sequence won't help, since the affected connection sockets are likely already dropped.

src/main/java/redis/clients/jedis/MaintenanceNotificationsConfig.java
```diff
@@ -94,6 +87,18 @@ public boolean isEnabledOrAuto() {
...
 
+  public int getRelaxedTimeout() {
+    return relaxedTimeout;
+  }
+
+  public int getRelaxedBlockingTimeout() {
+    return relaxedBlockingTimeout;
+  }
+
+  public MaintenanceEventListener getEventListener() {
```
ggivo, line 98: MaintenanceEventListener exposed outside it's visibility scope

src/main/java/redis/clients/jedis/MaintenanceNotificationsConfig.java
```diff
@@ -45,11 +54,12 @@ public enum Mode {
...
-  Mode mode = Mode.AUTO;
-  Duration relaxedWindowMaxDuration = DEFAULT_RELAXED_WINDOW_MAX_DURATION;
-  int relaxedSocketTimeoutMillis = DEFAULT_RELAXED_SOCKET_TIMEOUT_MS;
-  int relaxedBlockingSocketTimeoutMillis = DEFAULT_RELAXED_BLOCKING_SOCKET_TIMEOUT_MS;
+  private final EndpointType endpointType;
+  private final Mode mode;
+  private final Duration relaxedWindowMaxDuration;
+  private final int relaxedTimeout;
+  private final int relaxedBlockingTimeout;
+  private final MaintenanceEventListener listener;
```
ggivo, line 62: Do we need this or it is optional?

src/main/java/redis/clients/jedis/MaintenanceNotificationsConfig.java
```diff
@@ -122,47 +128,34 @@ public Builder mode(Mode mode) {
...
-     * {@link MaintenanceNotificationsConfig#DEFAULT_RELAXED_SOCKET_TIMEOUT_MS}.
-     */
-    public Builder relaxedSocketTimeoutMillis(int millis) {
-      this.relaxedSocketTimeoutMillis = millis;
+    public Builder relaxedBlockingTimeout(int millis) {
+      this.relaxedBlockingTimeout = millis;
+      return this;
+    }
+
+    public Builder eventListener(MaintenanceEventListener listener) {
```
ggivo, line 141: I assume providing a MaintenanceEventListener is optional, and we can handle it in separate PR, or MaintenanceEventListener need to be public also.

src/test/java/redis/clients/jedis/sch/AbstractRelaxedTimeoutBehaviorTest.java
```diff
@@ -150,47 +147,39 @@ public void testFailoverPushMessage() throws SocketException {
...
+      // The two configurations are genuinely different, so a mis-wiring between the sources would
+      // be
+      // observable rather than masked by equal values.
+      assertNotEquals(connection.getSoTimeout(), getRelaxedSoTimeout(connection));
+      assertNotEquals(connection.getBlockingSoTimeout(), getRelaxedBlockingSoTimeout(connection));
+
+      // The relaxed source is plugged into the default source as its override, but the window is
+      // not
+      // open anymore
+      assertTrue(isRelaxedTimeoutActive(connection));
```
ggivo, line 180: assertion seems to differ the comment claim

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -1165,95 +1010,54 @@ protected void removePushConsumer(PushConsumer consumer) {
...
-    long deadline = clockNanos.getAsLong() + period.toNanos();
-    if (relaxedUntilNanos == 0 || deadline - relaxedUntilNanos > 0) {
-      relaxedUntilNanos = deadline;
+  @Experimental
+  void relaxTimeouts(long expiration) {
+    if (this.relaxedTimeoutSource == null) {
+      throw new IllegalStateException("Relaxed timeouts not activated");
     }
-    applyTimeout(effectiveSoTimeout());
+    relaxedTimeoutSource.setExpirationTime(expiration);
```
ggivo, line 1047: Where do we enforce
`Calling this  with a later deadline extends the window; an earlier one is ignored`

src/test/java/redis/clients/jedis/MaintenanceEventControllerTest.java
```diff
@@ -201,7 +202,10 @@ public void handoffHook_firesOncePerAppliedHandoff() {
     moving(6L, TARGET_C, 100); // newer: fire
 
     assertEquals(2, fires.get());
-    assertTrue(receiver.isRelaxedTimeoutActive(), "receiver relaxed on MOVING");
+    // TODO: here it is not clear how controller is wired into connection;
+    // what is the reasoning around relaxed timeoux expectation here?
+    // assertTrue(ConnectionTestHelper.isRelaxedTimeoutActive(receiver), "receiver relaxed on
+    // MOVING");
```
ggivo, line 208: wiring - connection (receiver) is source of the event and is provided via onMoving
`controller.onMoving(new MovingEvent(seq, ttlSeconds, target), receiver);`

assertion was performed since previously we were also triggering relaxation on `receiver`, but this was redundand. I think this assertion can be dropped

ggivo (approved):
I suggest  addressing the comments for in follow up PR's
1. MaintenceEventListener exposed = in configuration 
2. Relaxation timeout on the connection level should only extend the expiration 

the scope of this one becoame to wide

PR #4583, opened 2026-06-29
Handle Pub/Sub PONG replies for Jedis ping

### LINKED ISSUE #4162: Redis PING command throws ClassCastException when client is in SUBSCRIBE mode
**Problem Description**

According to the official Redis documentation, the PING command is allowed while the client is in SUBSCRIBE mode, and it should return a PONG response. This applies to both RESP2 and RESP3 protocols.
[supporting Link](https://redis.io/docs/latest/develop/interact/pubsub/)

However, when using the Jedis library, calling ping() on a client that is currently in SUBSCRIBE mode results in a ClassCastException:

```java
java.lang.ClassCastException: java.util.ArrayList cannot be cast to byte[]
``` 

**Root Cause:**

The issue appears to be in the implementation of the **ping()** method which can be found in the File **Jedis.java** , which internally calls getStatusCodeReply():

```java
@Override
public String ping() {
  checkIsInMultiOrPipeline();
  connection.sendCommand(Command.PING);
  return connection.getStatusCodeReply();
}
``` 

The **getStatusCodeReply()** method is implemented as follows, following implementation can be found in **Connection.java** file:

```java
public String getStatusCodeReply() {
  flush();
  final byte[] resp = (byte[]) readProtocolWithCheckingBroken();
  if (null == resp) {
    return null;
  } else {
    return encode(resp);
  }
}
``` 

The **readProtocolWithCheckingBroken()** method is expected to return an Object as per the given implementation which can be found in Connection.java File. It returns an ArrayList when the client is in SUBSCRIBE mode, causing the cast to byte[] to fail with a ClassCastException. Proof for the Same which shows when in SUBSCRIBE Mode in redis-cli, the PING command returns an ArrayList consisting of an empty String and the message PONG 

Image

**Expected Behaviour:**

The ping() method should return "PONG" when the client is in SUBSCRIBE mode, just like the Redis CLI or any compliant Redis client implementation.

**Steps to Reproduce:**

1. Connect to Redis using Jedis.
2. Call subscribe() to enter subscription mode.
3. While subscribed, call ping().
4. Observe the exception.

**Jedis version:** 5.0.2
**Redis version:** 7.2.7

smf-h:
Fixes #4162.

This handles Redis PING while a connection is in subscribed mode, where Redis replies with a Pub/Sub-style PONG array instead of the normal status/bulk-string reply. Jedis#ping now accepts that PONG array path while preserving normal standalone PING behavior.

Changed files:
- src/main/java/redis/clients/jedis/Connection.java
- src/main/java/redis/clients/jedis/Jedis.java
- src/test/java/redis/clients/jedis/JedisMockTest.java

Validation:
- git diff --check
- diff-only debug/focused-test marker scan
- mvn "-Dformatter.skip=true" "-Dtest=JedisMockTest" "-DforkCount=0" test
- mvn "-Dformatter.skip=true" "-Dnet.bytebuddy.experimental=true" "-Dtest=JedisMockTest,JedisPubSubBaseTest" "-DforkCount=0" test

---

> [!NOTE]
> **Medium Risk**
> Changes low-level RESP reply handling on a core connection path used by health checks; scope is narrow and covered by new tests, but mis-parsing could affect any code calling ping on a subscribed connection.
> 
> **Overview**
> Fixes **PING** on connections that are already in **pub/sub subscribed mode**, where Redis returns a two-element array (`pong` + payload) instead of a simple status or bulk string.
> 
> `Connection` gains **`getPongReply` / `getBinaryPongReply`**, which read the next protocol object and, when it is a list, validate and normalize the pub/sub PONG shape (returning `"PONG"` or the echoed argument). **`Jedis#ping()`** and its string/binary overloads now use these helpers instead of **`getStatusCodeReply`** / **`getBulkReply`**, so standalone PING behavior is unchanged while subscribed-mode PING no longer fails parsing.
> 
> Adds **`JedisMockTest`** with a TCP mock server to assert all three PING variants against pub/sub-style replies.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 53550fe21a9b725a955966359e3b86bbc43a77f7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -435,6 +435,35 @@ public class Connection implements Closeable {
     }
   }
 
+  String getPongReply() {
+    final byte[] resp = getBinaryPongReply(false);
+    return resp == null ? null : encode(resp);
+  }
+
+  String getPongReply(final boolean hasArgument) {
+    final byte[] resp = getBinaryPongReply(hasArgument);
+    return resp == null ? null : encode(resp);
+  }
+
+  @SuppressWarnings("unchecked")
+  byte[] getBinaryPongReply(final boolean hasArgument) {
+    flush();
+    final Object resp = readProtocolWithCheckingBroken();
+    if (resp instanceof List) {
+      return getPubSubPongReply((List<Object>) resp, hasArgument);
+    }
+    return (byte[]) resp;
+  }
+
+  private static byte[] getPubSubPongReply(final List<Object> reply, final boolean hasArgument) {
+    if (reply.size() != 2 || !(reply.get(0) instanceof byte[])
+        || !Arrays.equals(Protocol.ResponseKeyword.PONG.getRaw(), (byte[]) reply.get(0))
+        || !(reply.get(1) instanceof byte[])) {
+      throw new JedisException("Unexpected PING reply: " + reply);
+    }
+    return hasArgument ? (byte[]) reply.get(1) : encode("PONG");
+  }
+
   public String getBulkReply() {
     final byte[] result = getBinaryBulkReply();
     if (null != result) {
--- a/src/main/java/redis/clients/jedis/Jedis.java
+++ b/src/main/java/redis/clients/jedis/Jedis.java
@@ -447,7 +447,7 @@ public class Jedis implements ServerCommands, DatabaseCommands, JedisCommands, J
   public String ping() {
     checkIsInMultiOrPipeline();
     connection.sendCommand(Command.PING);
-    return connection.getStatusCodeReply();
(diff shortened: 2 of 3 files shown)


ggivo:
@smf-h 
Thanks for the PR, team will review it shortly.

ggivo:
@smf-h 

Thanks for the contribution! After revisiting the issue, we've decided not to merge this — calling regular commands on a subscribed plain `Jedis` connection is a design limitation rather than a bug,   
  and normalizing the PONG reply can't make it safe; see #4162 for details.

PR #4592, opened 2026-07-06
escape backslash in RediSearchUtil.escape

insaf021:
RediSearchUtil.escapeQuery("\|evil")  ->  \\|evil

RediSearch reads `\\` as one literal backslash, so the trailing `|` stays live as a UNION operator and text passed through the library's own escaper still injects query syntax. The escape table covers every operator but not `\` itself, so a backslash already present in the value is copied through untouched and swallows the backslash the escaper prepends to the next character. Adding `\` to the set renders a literal backslash as `\\` and keeps the following operator inert; `escapeQuery` shares the same table so it is covered too.

---

> [!NOTE]
> **Medium Risk**
> Changes shared search query escaping behavior (possible query-string differences for inputs containing `\`); fix is narrowly scoped with tests and indexing path left as-is.
> 
> **Overview**
> Fixes **query escaping** so literal backslashes in user input no longer undo escaping of the next RediSearch operator.
> 
> `RediSearchUtil.escape(..., querying=true)` (used by `escapeQuery`) now prepends `\` before `\` as well as spaces, so values like `\|evil` become `\\\|evil` and `|` stays literal instead of acting as a UNION. **Indexing** via `escape(text)` without the querying flag is unchanged.
> 
> Unit tests cover standalone backslashes, embedded `\`, and the operator-injection case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5ea86218c31a414474dc78989d9de6f48e2b5147. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/main/java/redis/clients/jedis/search/RediSearchUtil.java
+++ b/src/main/java/redis/clients/jedis/search/RediSearchUtil.java
@@ -78,10 +78,10 @@ public class RediSearchUtil {
   }
 
   private static final Set<Character> ESCAPE_CHARS = new HashSet<>(Arrays.asList(//
-      '\\', ',', '.', '<', '>', '{', '}', //
-      '[', ']', '"', '\'', ':', ';', '!', //
-      '@', '#', '$', '%', '^', '&', '*', //
-      '(', ')', '-', '+', '=', '~', '|' //
+      ',', '.', '<', '>', '{', '}', '[', //
+      ']', '"', '\'', ':', ';', '!', '@', //
+      '#', '$', '%', '^', '&', '*', '(', //
+      ')', '-', '+', '=', '~', '|' //
   ));
 
   public static String escape(String text) {
@@ -98,7 +98,9 @@ public class RediSearchUtil {
     StringBuilder sb = new StringBuilder();
     for (char ch : chars) {
       if (ESCAPE_CHARS.contains(ch)
-          || (querying && ch == ' ')) {
+          // a backslash in the input would otherwise consume the escape prepended
+          // to the next special character, reactivating that query operator
+          || (querying && (ch == ' ' || ch == '\\'))) {
         sb.append("\\");
       }
       sb.append(ch);
--- a/src/main/java/redis/clients/jedis/search/RediSearchUtil.java
+++ b/src/main/java/redis/clients/jedis/search/RediSearchUtil.java
@@ -78,10 +78,10 @@ public class RediSearchUtil {
   }
 
   private static final Set<Character> ESCAPE_CHARS = new HashSet<>(Arrays.asList(//
-      ',', '.', '<', '>', '{', '}', '[', //
-      ']', '"', '\'', ':', ';', '!', '@', //
-      '#', '$', '%', '^', '&', '*', '(', //
-      ')', '-', '+', '=', '~', '|' //
+      '\\', ',', '.', '<', '>', '{', '}', //
+      '[', ']', '"', '\'', ':', ';', '!', //
+      '@', '#', '$', '%', '^', '&', '*', //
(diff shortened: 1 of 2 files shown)


ggivo:
@insaf021
Thanks for taking the time to look into this.  

However have some concerns regarding the change

 - RediSearchUtil serves different escaping contexts through one shared ESCAPE_CHARS table: escape() targets indexed text values (see toStringMap), while escapeQuery() targets query  
  strings — and RediSearch's rules differ between full-text values and tag queries. Modifying the shared table changes both paths at once.                                              
  - It's backward incompatible. Values containing `\` now index differently (`a\b` previously indexed as `ab`, now as `a\b`), so documents indexed by earlier Jedis versions would no longer match.
  versions may stop matching queries escaped by the new code until reindexed.                                                                                                           
  - The motivating example doesn't demonstrate a real scenario. "\|evil" isn't valid Java (\| is an illegal escape sequence), and no failing test or user report backs the described injection.
  - unescape() is also affected and becomes wrong unescape(escape("\\")) = ""

I'm going to hold off on this until there is a dedicated issue describing a concrete use case with example usage — ideally with an integration test reproducing it.

insaf021:
Fair points, and they all trace back to the shared table. I've narrowed the change so the backslash is escaped only in the query branch (`querying == true`); `escape()` and `ESCAPE_CHARS` are back to their master state.

That answers the concerns directly:

- Indexing is untouched, so `escape("a\\b")` still emits `a\b` and nothing needs reindexing. The two contexts no longer move together.
- `unescape()` pairs with the indexing `escape()`, and since neither moved the round-trip is unchanged; `unescape(escape("\\")) == ""` behaves exactly as on master.
- On the example: the runtime value is `\|evil`, written in Java source as `"\\|evil"`. The prose `escapeQuery("\|evil")` in the description was an illegal escape sequence, my mistake there.

Opened #4597 with the concrete use case and a failing integration test. Trace against a stock `redis:8`:

```
FT.SEARCH idx '\\|evil'    1 result   (today: the '|' union operator is still live)
FT.SEARCH idx '\\\|evil'   0 results  (with the fix: inert literal)
```

The unit test now asserts the query path escapes the backslash while the indexing path does not.

ggivo:
This silently changes query results for users who pre-escaped backslashes as a workaround for the old behavior — `escapeQuery(s.replace("\\", "\\\\"))` used to produce a correct query and now double-escapes, returning 0 results with no error.

insaf021:
That's true, and it's inherent to fixing the escaper in place: the pre-escape workaround bakes the old behaviour into caller code, so once `escapeQuery` handles `\` itself their `\\` becomes `\\\\` and the query quietly matches nothing. There's no way to complete the escaper inside the existing method without that failure mode.

The compatible shape would be to leave `escapeQuery` byte-for-byte as it is and put the backslash handling behind an explicit opt-in, an overload or a separate method, so pre-escaped callers see no change and new code gets the inert-operator behaviour deliberately. I can rework the branch that way if you want it; if you'd rather not grow the API for this, I'm fine closing this and leaving #4597 as the record of the behaviour.

ggivo:
Thanks for the thorough analysis and the offer to rework — you've captured the trade-off exactly: there's no way to fix this in place without breaking everyone who pre-escapes `\` today.

Honestly though, I'd rather not grow the API for this right now. Pre-escaping with `s.replace("\\", "\\\\")` works today, and #4597 documents the behaviour. If more people run into this, a separate method is what we'd add.

feel free to close it; happy to revisit if there's more demand.

insaf021:
Closing as discussed. #4597 stays as the record of the behaviour, and if demand builds an opt-in method is the shape I'd pick up again.

PR #4601, opened 2026-07-08
fix ConcurrentModificationException in AuthXManager token renewal

### LINKED ISSUE #4606: ConcurrentModificationException in AuthXManager.authenticateConnections when pruning GC'd connection references
### Expected behavior

On token renewal, `AuthXManager.authenticateConnections` walks the registered connection list, pushes the fresh credentials onto every live connection, and quietly prunes entries whose `Connection` has been garbage collected.

### Actual behavior

The walk aborts partway with `java.util.ConcurrentModificationException`. `authenticateConnections` iterates the `Collections.synchronizedList` of `WeakReference<Connection>` with an enhanced-for loop and calls `connections.remove(connectionRef)` in place when it meets a cleared reference. That structural modification bumps the backing `ArrayList`'s `modCount`, so the very next `iterator.next()` throws.

The knock-on effect is worse than the exception itself: every connection after the cleared entry never receives the renewed credentials, so they start failing auth once the old token expires. Cleared references are normal during pool churn (connections get collected after eviction), so an ordinary GC during token rotation is enough to trigger it.

### Steps to reproduce:

The failure shape reproduces standalone, without a server; this is exactly the pattern in `authenticateConnections`:

```java
List<WeakReference<Object>> connections = Collections.synchronizedList(new ArrayList<>());
connections.add(new WeakReference<>(null));           // connection collected during pool churn
connections.add(new WeakReference<>(new Object()));   // live
connections.add(new WeakReference<>(new Object()));   // live

for (WeakReference<Object> ref : connections) {
    if (ref.get() == null) {
        connections.remove(ref);                      // throws on the next iterator step
    }
}
```

```
Exception in thread "main" java.util.ConcurrentModificationException
	at java.base/java.util.ArrayList$Itr.checkForComodification(ArrayList.java:1095)
	at java.base/java.util.ArrayList$Itr.next(ArrayList.java:1049)
```

Against `AuthXManager` itself:

1. Seed the `connections` list with a cleared `WeakReference` ahead of two live connections (in production this is a pooled connection that has been GC'd; in a test you can plant `new WeakReference<>(null)` directly).
2. Call `authenticateConnections(token)`, as the renewal thread does on every token refresh.
3. Unpatched, it throws `ConcurrentModificationException` and the two live connections never get `setCredentials` for the new token.

PR #4601 has a unit test (`authenticateConnectionsPrunesClearedReferences`) doing exactly this: it errors with `ConcurrentModificationException` on current master and passes with the fix (cleared ref pruned, both live connections carrying the renewed credentials).

The `synchronizedList` javadoc requires synchronising on the list while iterating and mutating through the iterator, which is what the PR does: the loop moves under `synchronized (connections)` and prunes via `Iterator.remove()`. That also closes the narrower race with `addConnection` structurally modifying the list from another thread mid-walk.

#### Jedis version:

master (8.0.0-SNAPSHOT); the affected code is unchanged since `AuthXManager` was introduced

#### Redis version:

n/a, reproduces without a server

#### Java version:

any recent JDK (verified on OpenJDK 21); the fail-fast iterator behaviour isn't version specific

insaf021:
Fixes #4606

### Problem

`AuthXManager.authenticateConnections` runs on the token-renewal thread and walks a `Collections.synchronizedList` of `WeakReference<Connection>` to push the fresh credentials onto each pooled connection. Entries whose `Connection` has since been garbage collected turn up as cleared references during ordinary pool churn.

I ran into it exercising renewal against a pool that had recycled a connection: the renewal walk aborts with `ConcurrentModificationException`, so every connection after the cleared entry never receives the renewed credentials and starts failing auth once the old token expires.

### Root cause

The walk uses an enhanced-for loop and calls `connections.remove(connectionRef)` in place whenever it meets a cleared reference:

```java
for (WeakReference<Connection> connectionRef : connections) {
    Connection connection = connectionRef.get();
    if (connection != null) {
        connection.setCredentials(credentialsFromToken);
    } else {
        connections.remove(connectionRef);  // bumps modCount mid-iteration
    }
}
```

The structural modification bumps the backing `ArrayList`'s `modCount`, and the next `iterator.next()` throws. `Collections.synchronizedList` only guarantees safe iteration when you synchronise on the list yourself and mutate through the iterator.

### Fix

Move the loop under `synchronized (connections)` and prune cleared references via `Iterator.remove()`, per the `synchronizedList` javadoc. This also closes the narrower race where `addConnection` structurally modifies the list from another thread mid-walk. `setCredentials` just swaps an `AtomicReference`, so holding the monitor across the loop stays cheap.

### Testing

Added `authenticateConnectionsPrunesClearedReferences`: it seeds the connection list with a cleared `WeakReference` ahead of two live connections and drives `authenticateConnections` directly. On master it errors with `ConcurrentModificationException`; with the fix it passes, the cleared entry is pruned, and both live connections carry the renewed credentials. A standalone snippet showing the same failure shape is in #4606.

---

> [!NOTE]
> **Medium Risk**
> Changes authentication credential propagation on the token-renewal path; incorrect locking could affect renewal under pool churn, but the change follows synchronizedList usage and is narrowly scoped.
> 
> **Overview**
> Fixes token renewal aborting with **`ConcurrentModificationException`** when **`AuthXManager.authenticateConnections`** walks pooled connections and prunes garbage-collected **`WeakReference`** entries.
> 
> The renewal loop now runs under **`synchronized (connections)`** and drops cleared references with **`Iterator.remove()`** instead of mutating the list during an enhanced-for iteration. Several **`AuthXManager`** fields are marked **`final`**.
> 
> Adds **`authenticateConnectionsPrunesClearedReferences`** to assert cleared weak refs are pruned and live connections still receive renewed credentials.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2533458cb38a459e62efae2754899d0289240b1f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ggivo:
HI @insaf021 
Will take a look as soon as we find some free cycles. 

Meaning will appreciate opening an issue and providing a reproducer/code example to make sure we are dicussing same issue. It will help triage it faster.

 PR description is hard to read; some formatting will help reviewers navigate faster

insaf021:
Opened #4606 with a standalone reproducer: the same enhanced-for plus in-place remove pattern throws ConcurrentModificationException on any synchronizedList, no server needed, and the unit test in this PR errors the same way on current master. Also restructured the description into problem / root cause / fix / testing sections, should be easier to navigate now.

ggivo (approved):
LGTM

ggivo:
@insaf021 
Thanks for spotting & fixing it. Now merged.

PR #4617, opened 2026-07-13
Fix user/password split for percent-encoded Redis URI credentials

### LINKED ISSUE #4728: Percent-encoded colon in URI username shifts the user/password split
### Problem

`JedisURIHelper.getUser` and `getPassword` read `uri.getUserInfo()` and then `split(":", 2)`. `URI.getUserInfo()` returns the *decoded* form, so a percent-encoded colon (`%3A`) in the username has already become a literal `:` by the time the split runs, and the split lands on it instead of on the real separator.

For `redis://us%3Aer:pw@host:6379/0` the result is user `us` with password `er:pw`, where the URI means user `us:er` with password `pw`. The client then sends `AUTH us er:pw`, so the wrong ACL identity is presented and the connection is refused with an error that points at the wrong user.

ACL usernames aren't restricted to URI-safe characters, and namespaced names such as `svc:billing` are a natural fit for them. The only way to put one in a URI is to percent-encode the colon, which is exactly the input that breaks. The same happens for any URI assembled with `URLEncoder` or a URI builder, since those encode `:` in the username as a matter of course.

One constraint on any fix: userinfo is not form data, so a literal `+` in a password must stay a `+`. Decoding the split halves has to guard against `URLDecoder`'s plus-to-space rule.

No server needed:

```java
URI uri = new URI("redis://us%3Aer:pw@host:9000/0");
JedisURIHelper.getUser(uri);      // "us"     (expected "us:er")
JedisURIHelper.getPassword(uri);  // "er:pw"  (expected "pw")
```

As a unit test in `JedisURIHelperTest`:

```java
@Test
public void shouldKeepEncodedColonInsideUsername() throws URISyntaxException {
  URI uri = new URI("redis://us%3Aer:pw@host:9000/0");
  assertEquals("us:er", JedisURIHelper.getUser(uri));
  assertEquals("pw", JedisURIHelper.getPassword(uri));
}
```

On master this fails with `expected: <us:er> but was: <us>`.

### Fix

Split `uri.getRawUserInfo()` instead, where an encoded colon is still `%3A` and only the real separator is a literal `:`, then percent-decode each half. Encoded `@` or `:` in the password and a literal `+` behave as before. #4617 does this.

insaf021:
Fixes #4728

    redis://us%3Aer:pw@host  ->  getUser="us", getPassword="er:pw"
                                 (want user "us:er", password "pw")

`getUser`/`getPassword` read `uri.getUserInfo()`, which percent-decodes the userinfo before the `user:password` split. A `%3A` in the username decodes to a literal `:` first, so `split(":", 2)` lands on it and the credentials are cut at the wrong point, sending the wrong ACL identity to `AUTH`.

Splitting `uri.getRawUserInfo()` keeps the encoded colon as `%3A`, so only the real separator is a literal `:`, and each half is decoded after the split. `+` is kept literal since userinfo is not form data, so passwords already containing `+` are unaffected.

---

> [!NOTE]
> **Medium Risk**
> Changes how URI credentials are parsed for AUTH across Jedis connection paths; behavior shifts for percent-encoded userinfo but corrects mis-split identities.
> 
> **Overview**
> Fixes Redis URI credential parsing so **usernames and passwords with percent-encoded characters** (especially `%3A` in the username) split on the real `user:password` boundary instead of on decoded colons inside the username.
> 
> `getUser` and `getPassword` now split **`getRawUserInfo()`** and percent-decode each side via a new `decode()` helper; **`+` stays literal** in userinfo (guarded before `URLDecoder`). Tests cover encoded colons in usernames and common encoded password characters.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 164426bd881ba47945a90e407ae834dccb452c9d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ggivo:
Thanks for the fix — we'll take a closer look shortly.

insaf021:
Filed #4728 with the repro so this has a tracked issue. Also merged master in: the red SDR job was the `xinfo` expected 2 / got 1 failure that hit every PR between 4 and 7 August and has since been fixed in the workflow, nothing in this diff goes near streams.

atakavci (approved):
LGTM!

src/main/java/redis/clients/jedis/util/JedisURIHelper.java
```diff
@@ -53,9 +56,9 @@ public static HostAndPort getHostAndPort(URI uri) {
    * @return the user as a String, or null if user is empty or {@link URI#getUserInfo()} info is missing
```
atakavci: javadoc remains to reference to `getUserInfo`

src/test/java/redis/clients/jedis/util/JedisURIHelperTest.java
```diff
@@ -26,6 +26,22 @@ public void shouldGetNullUserFromURIWithCredentials() throws URISyntaxException
...
+    assertEquals("us:er", JedisURIHelper.getUser(uri));
+    assertEquals("pw", JedisURIHelper.getPassword(uri));
+  }
+
+  @Test
+  public void shouldDecodeEncodedCredentials() throws URISyntaxException {
+    // percent-encoded specials in the password decode, literal '+' stays literal
+    assertEquals("pa@ss", JedisURIHelper.getPassword(new URI("redis://user:pa%40ss@host:9000/0")));
+    assertEquals("pa:ss", JedisURIHelper.getPassword(new URI("redis://user:pa%3Ass@host:9000/0")));
+    assertEquals("pa+ss", JedisURIHelper.getPassword(new URI("redis://user:pa+ss@host:9000/0")));
```
ggivo: Also worth asserting the encoded form — together with the literal `+` it pins both sides of the `replace("+", "%2B")` guard in `decode()`:                                                     
                                                                                                                                                                                                 
  ```suggestion                                                                                                                                                                                  
      assertEquals("pa+ss", JedisURIHelper.getPassword(new URI("redis://user:pa+ss@host:9000/0")));                                                                                              
      assertEquals("pa+ss", JedisURIHelper.getPassword(new URI("redis://user:pa%2Bss@host:9000/0")));                                                                                            
  ```

ggivo (approved):
LGTM

PR #4620, opened 2026-07-14
[maintenance events] Support 'none' moving-endpoint-type (CAE-1559)

ggivo:
Adds the `none` value for `moving-endpoint-type` (CAE-1559).

With `EndpointType.NONE` the client sends `CLIENT MAINT_NOTIFICATIONS ON moving-endpoint-type none`, and the server's MOVING carries a RESP3 null endpoint. On such a MOVING the client performs no handoff: connections keep serving the current endpoint with relaxed timeouts and reconnect to the configured endpoint after half the grace period (`time_s/2`), giving DNS time to repoint.

### Behavior
- `MaintenancePushCodec` decodes an explicit RESP3 null MOVING target as a null-target `MovingEvent` (previously collapsed into malformed). A MOVING without the endpoint element remains malformed.
- Reconnect is a per-connection deadline (`Connection.expireAt`) — no background scheduler:
  - a target MOVING stamps the receiver for immediate discard (existing behavior); a `none` MOVING stamps `now + time_s/2`;
  - all connections affected by the same MOVING share the first notification's deadline;
  - expired connections are destroyed on return to the pool; borrow never blocks on expiry, so borrowed work always runs and a stale connection serves at most one more command;
  - `RebindAwareEvictionPolicy` stamps idle affected connections at notification time and evicts only those past their deadline;
  - `validateObject` reports expired connections invalid — `testOnBorrow`/`testWhileIdle` opt into stricter cleanup, mirroring how dead-while-idle connections are handled.
- Relax window and reconnect deadline are independent timers; supersession stays seq-gated for target and `none` alike.

### Testing
- `MaintenancePushCodecTest`: null-target decode + malformed frame shapes.
- `MaintenanceEventControllerTest`: `none` deadlines, shared same-seq deadline, supersession (`none` <-> target), handoff hook payload.
- `ConnectionPoolExpiryMockTest`: pool lifecycle — expired idle served then destroyed on return, validation semantics, `testOnBorrow`/`testWhileIdle` strictness, fail-fast termination bound.
- `AbstractRebindBehaviorTest` (RedisClient + MultiDbClient): `none` keeps connections through the window, no remap, replacement after half the grace period.

### Churn guard (second commit)
A reconnect that re-lands on a not-yet-repointed endpoint (JVM DNS cache, DNS lag, raw-IP config) would be re-stamped with the past shared deadline on every return and on the same-seq re-notification the moving node sends to every new connection - a new connection per command until the relax window closes. `RebindState.expireAtNanos` (the moment recycling starts) now also scopes the rebind: connections created before it are recycled on return; connections created after it are the rebind's own reconnects and are left alone. The target case stores the receipt time instead of the `EXPIRED` sentinel, which is removed.

### DIFF
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -172,7 +172,9 @@ public class Connection implements Closeable {
   private Set<InitVisitor> initVisitors = new HashSet<>();
 
   private static final long EXPIRE_NOT_SET = -1;
-  private long expireAt = EXPIRE_NOT_SET;
+  /** Reconnect-deadline sentinel: always expired without a clock read (MOVING receiver discard). */
+  static final long EXPIRED = 0;
+  private volatile long expireAt = EXPIRE_NOT_SET;
 
   /** Listeners notified synchronously of this connection's maintenance events (pool-injected). */
   private final Set<MaintenanceEventListener> maintenanceEventListeners =  ConcurrentHashMap.newKeySet();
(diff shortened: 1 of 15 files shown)


ggivo:
closing in favour of : #4625

PR #4622, opened 2026-07-15
synchronise the read path in AbstractCache.get

insaf021:
Every mutator in `AbstractCache` takes the `ReentrantLock`; the reader does not:

    set / delete / deleteByRedisKey / deleteByRedisKeys / flush  ->  lock.lock()
    get                                                          ->  getFromStore(...)   // no lock

`DefaultCache` backs the store with a plain `HashMap`, so `get()` runs `HashMap.get` concurrently with a `put`/`remove`/`clear` executing under the lock on another thread. The RESP3 server-push invalidation path (`PushInvalidateConsumer` -> `deleteByRedisKeys`) mutates that same map while application threads read it, so a `get` racing a rehash can observe a stale or foreign `CacheEntry`.

`get()` is a top-level lock acquirer like `set()` and `delete()` (it never runs while the eviction-policy monitor is held), so taking the same lock adds no ordering inversion. The regression test parks a reader inside `get()` and asserts a concurrent `set()` no longer proceeds until the read completes; it fails on the current code and passes with the lock.

---

> [!NOTE]
> **Medium Risk**
> Changes hot-path locking for client-side cache reads; correct for thread safety but may increase contention under heavy concurrent access.
> 
> **Overview**
> **`AbstractCache.get`** now acquires the same **`ReentrantLock`** used by **`set`**, **`delete`**, and invalidation paths, so reads no longer run **`getFromStore`** (and eviction **`touch`**) concurrently with mutations on the backing store (e.g. **`DefaultCache`**’s **`HashMap`**).
> 
> Adds **`AbstractCacheConcurrencyTest`**, which blocks a reader inside the store **`get`** and asserts a concurrent **`set`** cannot finish until the read completes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7049ac2b74c32b57643c2d6289e6121d71a62e62. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/main/java/redis/clients/jedis/csc/AbstractCache.java
+++ b/src/main/java/redis/clients/jedis/csc/AbstractCache.java
@@ -50,11 +50,16 @@ public abstract class AbstractCache implements Cache {
 
   @Override
   public CacheEntry get(CacheKey cacheKey) {
-    CacheEntry entry = getFromStore(cacheKey);
-    if (entry != null) {
-      getEvictionPolicy().touch(cacheKey);
+    lock.lock();
+    try {
+      CacheEntry entry = getFromStore(cacheKey);
+      if (entry != null) {
+        getEvictionPolicy().touch(cacheKey);
+      }
+      return entry;
+    } finally {
+      lock.unlock();
     }
-    return entry;
   }
 
   @Override
--- a/src/test/java/redis/clients/jedis/csc/AbstractCacheConcurrencyTest.java
+++ b/src/test/java/redis/clients/jedis/csc/AbstractCacheConcurrencyTest.java
@@ -1,0 +1,86 @@
+package redis.clients.jedis.csc;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import org.junit.jupiter.api.Test;
+
+import redis.clients.jedis.CommandObjects;
+import redis.clients.jedis.RedisProtocol;
+
+public class AbstractCacheConcurrencyTest {
+
+  /**
+   * A store whose read is made to block once, so a reader can be parked inside
+   * {@link AbstractCache#get(CacheKey)} while another thread tries to mutate the cache.
+   */
+  private static class BlockingReadCache extends HashMap<CacheKey, CacheEntry> {
+
+    private final CacheKey blockOn;
+    private final CountDownLatch enteredGet = new CountDownLatch(1);
+    private final CountDownLatch proceedGet = new CountDownLatch(1);
+
+    BlockingReadCache(CacheKey blockOn) {
+      this.blockOn = blockOn;
+    }
+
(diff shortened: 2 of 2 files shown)


ggivo:
Thanks for the PR. The data race is real: `AbstractCache.get()` reads the store outside the lock that guards the writers.

That said, since all writers are serialized by that lock, the unsynchronized read can't corrupt the map or return a wrong value — the realistic worst case is a spurious `null`, i.e. a cache miss that reissues the command and caches the fresh result. This is close to existing behavior: concurrent reads through `RedisClient` aren't serialized, so two threads can already miss the same key and both issue the command. An occasional extra round-trip from the race is hard to distinguish from that.

Before proceeding, could you share:

1. **Usage pattern** — pool size / concurrency, read-vs-write mix, key cardinality.
2. **Observed symptom** — the actual behavior you hit (exception + stack trace? wrong value? latency?), beyond occasional extra round-trips.

And we'll want to confirm the fix doesn't regress the read path, and locking `get()` adds contention with the invalidation path. A contended, read-heavy benchmark (baseline vs. fix, ideally vs. a `ConcurrentHashMap` store) would help confirm we're not trading a benign miss for a throughput hit.

Thanks!

insaf021:
1. There's no production workload behind this, so I can't give you a pool size or read/write mix. I found it tracing the push-invalidation path (`PushInvalidateConsumer` -> `deleteByRedisKeys`) and noticing `get()` reads the store outside the lock the writers take.

2. No observed symptom beyond what the test shows. Your read of the blast radius is fair: with writers serialised, `HashMap.get` can't return a foreign entry, so the realistic worst case is a spurious miss or a briefly stale hit ("foreign" in my description overstated it). What remains is JMM-level only: unlocked readers have no happens-before edge from an invalidation, and I can't show that misbehaving on real hardware.

Ran the benchmark you asked for. 1024 hot keys, 8 reader threads doing all-hit `get()`s, one unpaced thread looping `deleteByRedisKey` + `set` to model worst-case invalidation churn; 3 x 5s measured runs after warmup, JDK 21, 10-core Apple Silicon. Averages, stable across runs:

| variant | reads/s | invalidations/s |
|---|---|---|
| master (unlocked `get`, HashMap) | 6.85M | 217k |
| this PR (locked `get`) | 5.98M (-13%) | 97k (-55%) |
| unlocked `get`, ConcurrentHashMap store | 6.51M (-5%) | 243k |

So the concern is justified: the lock costs 13% on reads and roughly halves invalidation throughput, since every hit now contends with the writer's full critical section. The deltas would be larger still if the hit path weren't already serialising on the synchronised `LRUEviction.touch()`, which all three variants pay.

The ConcurrentHashMap store looks like the better trade: it closes the formal race and gives invalidations volatile visibility for about 5% on reads and nothing on the invalidation path. Happy to repoint this PR at swapping `DefaultCache`'s backing map to `ConcurrentHashMap` and dropping the lock from `get()`, if you'd take it that way. Can share the harness too.

ggivo:
Both the lock and the `ConcurrentHashMap` variant regress the common-case read path, and since the race only degrades to an occasional spurious miss (correct data after a reissue) with no reported real-world failure, I'd rather not pay that cost — closing for now, and happy to revisit if a concrete usage failure surfaces.

PR #4623, opened 2026-07-17
[maintenance events] Relax handshake reads for connections opened during a rebind window

### LINKED ISSUE #4621: [maintenance events] Relaxed timeouts must apply while creating new connections (handshake commands)
Pool-level timeout relaxation must also be effective while **creating** new connections — including the handshake commands: HELLO, AUTH, CLIENT SETNAME/SETINFO, READONLY, and the `CLIENT MAINT_NOTIFICATIONS` subscription itself.

### Current behavior
The pool-wide rebind relaxation (`RebindTimeoutSource`) is attached only by `MaintenanceAwareVisitor`, and init visitors run **last** in `Connection.initializeFromClientConfig`:

1. socket connect (base connect timeout)
2. `establishProtocol` — HELLO / AUTH (base socket timeout)
3. CLIENT SETNAME / SETINFO / READONLY (base socket timeout)
4. init visitors — only here the relaxation chain is wired and `CLIENT MAINT_NOTIFICATIONS` is sent

So a connection created **during an active maintenance window** — exactly the post-drain replacement or remapped-target connection, talking to the node at its busiest — handshakes with base, unrelaxed timeouts. A slow HELLO/AUTH then fails `makeObject`, i.e. connection creation fails at the moment resilience matters most.

### Suggested fix
Attach the controller-backed relaxation to the connection's timeout chain at **construction** instead of in the visitor: `Connection.Builder` / `ConnectionFactory` already receive the maintenance controller (used today as the post-DNS address mapper), so they can wire a rebind-gated timeout source into the default chain before connect/handshake. The visitor keeps only the per-connection relaxation source and the `MAINT_NOTIFICATIONS` handshake.

Open decision: whether to also relax the socket **connect** timeout (`DefaultJedisSocketFactory` also holds the controller).

### Test
With an active rebind window, `factory.makeObject()` against a mock server that delays HELLO beyond the base socket timeout but within `relaxedTimeout` must succeed; without a window it must time out with base semantics.

atakavci:
Closes #4621 
## Summary
Connections opened while a pool-wide maintenance rebind window is open now relax their read timeouts starting *before* the AUTH/HELLO handshake, rather than only after it completes. Previously the relaxed/rebind timeout overlay was installed as part of post-handshake init, so a slow HELLO during maintenance could exceed the tight configured socket timeout and fail the connection outright.

- `InitVisitor.visit` is split into `visitBeforeHandshake` and `visitAfterHandshake`. The pre-handshake hook installs state that must already be in effect while AUTH/HELLO runs; the post-handshake hook does the work that requires a negotiated protocol.
- Per-connection relaxed-timeout state (the `enableTimeoutRelaxing` / `disableTimeoutRelaxing` / `relaxTimeouts` / `resetRelaxedTimeouts` methods and the `relaxedTimeoutSource` field) is removed from `Connection`. The timeout chain is now the single source of truth, exposed via `getTimeoutSource()`.
- `ChainedTimeoutSource` gains an `identity` tag and `seekBy(identity)` so a specific override (the rebind source keyed by its controller, or the `ExpiringTimeoutSource` by class) can be located and mutated/removed within the chain.

- When maintenance is enabled, the rebind and relaxed-timeout overrides are added to the connection's timeout chain before the handshake. If the connection turns out not to support maintenance notifications (non-RESP3, or the server rejects `CLIENT MAINT_NOTIFICATIONS ON`), both overrides are torn down again after the handshake.
- `MaintenanceEventController` now drives relaxation by setting the expiration time on the `ExpiringTimeoutSource` it locates in the chain (via `seekBy`) instead of calling the removed `Connection.relaxTimeouts` / `resetRelaxedTimeouts` methods.

## Testing
Adds a test that opens a connection against a mock server whose HELLO reply is delayed beyond the configured socket timeout but within the relaxed timeout: without a rebind window the handshake times out, and after opening a pool-wide rebind window a fresh connection's handshake completes. Test helpers were updated to reach the `ExpiringTimeoutSource` through the timeout chain rather than the removed `Connection` methods.

---

> [!NOTE]
> **Medium Risk**
> Touches connection initialization and maintenance timeout wiring on the critical handshake path; behavior changes when opening connections during rebind, though unsupported maintenance paths roll back overrides.
> 
> **Overview**
> Connections opened while a **pool-wide maintenance rebind window** is active now use relaxed read timeouts **before** AUTH/HELLO, so a slow handshake during maintenance is less likely to fail on the tight configured `SO_TIMEOUT`.
> 
> **Init lifecycle:** `InitVisitor` is split into `visitBeforeHandshake` and `visitAfterHandshake`. `MaintenanceAwareVisitor` installs `RebindTimeoutSource` and `ExpiringTimeoutSource` on the timeout chain pre-handshake; post-handshake it enables RESP3 maintenance push handling and tears down both overrides if notifications are unsupported (non-RESP3 or server rejects `CLIENT MAINT_NOTIFICATIONS ON`).
> 
> **Timeout model:** Per-connection `relaxedTimeoutSource` and `relaxTimeouts` / `resetRelaxedTimeouts` on `Connection` are removed in favor of `getTimeoutSource()` and `ChainedTimeoutSource.seekBy(identity)` so overrides can be found and updated. `MaintenanceEventController` relaxes/clears timeouts by mutating the chain’s `ExpiringTimeoutSource`.
> 
> **Tests:** Adds a mock-server test where delayed HELLO fails without a rebind window and succeeds with one; helpers now observe relaxation via the override chain. Minor test harness tweaks (`CopyOnWriteArrayList`, logback WARN for `MaintenanceAwareVisitor`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fe2d7565e8a44e41eb0da3dd77f619270a889bb3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/main/java/redis/clients/jedis/MaintenanceAwareVisitor.java
+++ b/src/main/java/redis/clients/jedis/MaintenanceAwareVisitor.java
@@ -27,61 +27,91 @@ public class MaintenanceAwareVisitor implements InitVisitor {
     this.controller = maintenanceController;
   }
 
+  /**
+   * Installs the pool-wide MOVING rebind timeout overlay before the handshake runs, so a connection
+   * opened while a rebind window is open already relaxes its AUTH/HELLO handshake reads. The
+   * overlay is torn down again in {@link #visit(Connection)} if the feature turns out to be
+   * unsupported on this connection.
+   */
   @Override
-  public void visit(Connection connection) {
-    RedisProtocol protocol = connection.getRedisProtocol();
-    Set<MaintenanceEventListener> maintenanceEventListeners = connection
-        .getMaintenanceEventListeners();
-    MaintenanceNotificationsConfig maintenanceConfig = builder.getMaintenanceConfig();
-
-    if (maintenanceConfig == null
-        || maintenanceConfig.getMode() == MaintenanceNotificationsConfig.Mode.DISABLED) {
-      return;
-    }
-    boolean strict = maintenanceConfig.getMode() == MaintenanceNotificationsConfig.Mode.ENABLED;
-
-    // Maintenance push frames require RESP3.
-    if (protocol != RedisProtocol.RESP3) {
-      String reason = "RESP3 is required but the established protocol is "
-          + (protocol == null ? "RESP2" : protocol);
-      if (strict) {
-        throw new JedisConnectionException("Maintenance notifications: " + reason);
-      }
-      logger.debug("Maintenance notifications disabled: {}.", reason);
+  public void visitBeforeHandshake(Connection connection) {
+    MaintenanceNotificationsConfig mConfig = builder.getMaintenanceConfig();
+    if (!isMaintenanceEnabled(mConfig)) {
       return;
     }
 
-    maintenanceEventListeners.add(controller);
+    ChainedTimeoutSource dts = connection.getTimeoutSource();
+    dts.addOverride(new RebindTimeoutSource(controller));
 
-    // The server must accept CLIENT MAINT_NOTIFICATIONS ON. Pre-register the consumer so a
-    // push
-    // frame the server emits immediately on accepting the subscription cannot race ahead.
-    MaintenanceEventConsumer consumer = new MaintenanceEventConsumer(connection,
-        maintenanceEventListeners);
-    connection.addPushConsumer(consumer);
-    ExpiringTimeoutSource relaxedTimeoutSource = new ExpiringTimeoutSource(new TimeoutInfo(
-        maintenanceConfig.getRelaxedTimeout(), maintenanceConfig.getRelaxedBlockingTimeout()));
-    connection.enableTimeoutRelaxing(relaxedTimeoutSource);
+    TimeoutInfo relaxedTimeout = new TimeoutInfo(mConfig.getRelaxedTimeout(),
+        mConfig.getRelaxedBlockingTimeout());
+    dts.addOverride(new ExpiringTimeoutSource(relaxedTimeout));
+  }
 
-    ChainedTimeoutSource rebindTimeoutSource = new RebindTimeoutSource(controller);
-    relaxedTimeoutSource.addOverride(rebindTimeoutSource);
+  @Override
+  public void visitAfterHandshake(Connection connection) {
+    MaintenanceNotificationsConfig mConfig = builder.getMaintenanceConfig();
+    if (!isMaintenanceEnabled(mConfig)) {
+      return;
+    }
 
-    connection.sendCommand(Command.CLIENT, "MAINT_NOTIFICATIONS", "ON", "moving-endpoint-type",
-      resolveEndpointType(maintenanceConfig.getEndpointType()));
+    boolean strict = mConfig.getMode() == MaintenanceNotificationsConfig.Mode.ENABLED;
+    boolean keepOverrides = false;
+
     try {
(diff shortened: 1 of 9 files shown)


src/main/java/redis/clients/jedis/MaintenanceAwareVisitor.java
```diff
@@ -27,63 +27,93 @@ public class MaintenanceAwareVisitor implements InitVisitor {
...
+          "Maintenance notifications disabled: server rejected CLIENT MAINT_NOTIFICATIONS ({}).",
+          e.getMessage());
+      }
+    } finally {
+      if (!keepOverrides) {
+        // Undo the rebind overlay installed before the handshake — this connection does NOT support
+        // maintenance notifications.
+        ChainedTimeoutSource dts = connection.getTimeoutSource();
+        dts.removeOverride(dts.seekBy(controller));
+        dts.removeOverride(dts.seekBy(ExpiringTimeoutSource.class));
```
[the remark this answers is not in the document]
ggivo: By design, timeouts are calculated & applied just before read operations only if it differ from socket SO_TIMEOUT.

src/main/java/redis/clients/jedis/MaintenanceEventController.java
```diff
@@ -177,28 +177,33 @@ private void fireHandoffHook(MovingEvent e) {
...
-    c.resetRelaxedTimeouts();
+    relaxConnectionTimeoutsFor(c, 0);
+  }
+
+  private void relaxConnectionTimeoutsFor(Connection c, long expirationTime) {
+    ChainedTimeoutSource source = c.getTimeoutSource().seekBy(ExpiringTimeoutSource.class);
+    if (source != null) {
+      ((ExpiringTimeoutSource) source).setExpirationTime(expirationTime);
+    }
+    c.applyCurrentTimeout();
```
ggivo, line 206: `applyCurrentTimeout` will be invoked before read from socket and will calcualte & apply timeout on socket before read. 

In which case do we need to need to invoke it explicitly? 

If not required to invoke it, we can reduce `applyCurrentTimeout` visibility back to private.

ggivo (commented):
LGTM 

posted one question regarding need for invoking `c.applyCurrentTimeout` on conenction explicitly from MaintenceEventController

src/main/java/redis/clients/jedis/MaintenanceEventController.java
atakavci: Assume the case where user requested a command and bytes are issued to server, while there is already a maintenance push message pending in the buffer since past read. Attempting an immediate tcp read will not see the command response yet but the maintenance notification. With `applyCurrentTimeout()` in place, the command result will read and timeout applied before next read attempt which is for command response.

This could be considered as an improvement and i was thinking about not implemented it until the test case has failed. So i got the idea that you like to cover the case and introduced test cases around it, like [testRelaxedBlockingTimeoutAppliedDuringBlockingCommand()](https://github.com/atakavci/jedis/blob/fe2d7565e8a44e41eb0da3dd77f619270a889bb3/src/test/java/redis/clients/jedis/sch/AbstractRelaxedTimeoutBehaviorTest.java#L197).

ggivo (approved):

PR #4625, opened 2026-07-17
[maintenance events] Support 'none' moving-endpoint-type via connection registry (CAE-1559)

ggivo:
Adds the `none` value for `moving-endpoint-type` (CAE-1559). Alternative to #4620 — marks affected connections via a connection registry and a scheduled marking pass instead of per-connection deadlines.

With `EndpointType.NONE` the client sends `CLIENT MAINT_NOTIFICATIONS ON moving-endpoint-type none`, and the server's MOVING carries a RESP3 null endpoint. On such a MOVING the client performs no handoff: connections keep serving the current endpoint with relaxed timeouts and reconnect to the configured endpoint after half the grace period (`time_s/2`), giving DNS time to repoint.

### Behavior
- `MaintenancePushCodec` decodes an explicit RESP3 null MOVING target as a null-target `MovingEvent` (previously collapsed into malformed). A MOVING without the endpoint element remains malformed.
- The controller tracks every pool-managed connection in a weak-reference `ConnectionRegistry`; the factory registers each connection **before socket init**, so a connect racing a MOVING commit is either visible to the marking pass or remapped via the committed rebind.
- Every applied MOVING transition (new seq or merged source) runs one marking pass against the snapshot it produced — inline for a target MOVING, scheduled at `time_s/2` for `none`. The single daemon scheduler thread exists only when the config's endpoint type is `NONE`; the controller owns it and `ConnectionPool#destroy` closes it.
- The marking pass flips an advisory per-connection flag and runs the handoff hooks — the pool evicts marked idles. Marked in-use connections are destroyed on return; `validateObject` reports them invalid (`testOnBorrow`/`testWhileIdle` opt into stricter cleanup). Borrow never blocks on the mark, so borrowed work always runs.
- Connections created after the pass are never visited: a reconnect that re-lands on a not-yet-repointed endpoint (and the same-seq re-notification it receives) stays unmarked by construction — no creation churn.
- A null-target MOVING without `EndpointType.NONE` configured is an unsupported setup: warn and ignore.
- Relax window and reconnect instant are independent; supersession stays seq-gated for target and `none` alike.

### Testing
- `MaintenancePushCodecTest`: null-target decode + malformed frame shapes.
- `MaintenanceEventControllerTest`: seq guard and merge semantics, handoff hook fan-out, relax-on-borrow scope.
- `MaintenanceMarkingTest` (new): marking scheduling and coverage against a deterministic scheduler stub, churn immunity, factory registration order, supersession/stale fires, scheduler lifecycle.
- `AbstractRebindBehaviorTest` (RedisClient + MultiDbClient): `none` keeps connections through the window, no remap, replacement created against the configured endpoint after half the grace.

---

> [!NOTE]
> **Medium Risk**
> Changes core connection-pool lifecycle during MOVING maintenance (marking, eviction, validation, scheduled passes); behavior is well-tested but affects all maintenance-enabled pools.
> 
> **Overview**
> Adds **`EndpointType.NONE`** so clients can request `moving-endpoint-type none` and treat a RESP3 **null** MOVING target as “no remap”—connections stay on the configured endpoint and reconnect after **half the server grace period** (`time_s/2`).
> 
> **MOVING handling** moves from per-connection `expireAt` deadlines and return-time `isAffected` checks to a **`ConnectionRegistry`** (weak refs) plus **marking passes** that set an advisory **`retire()`** flag on affected peers. Real targets mark **inline**; `none` marks on a **lazy daemon scheduler** at the reconnect instant. The pool **evicts retired idles**, **destroys on return/validation**, and **`RebindAwareEvictionPolicy`** now keys off `isRetired()` instead of live rebind peer matching.
> 
> **Protocol/config:** `MaintenancePushCodec` accepts null MOVING targets (requires four frame fields); MOVING relax windows use the **full server `time_s`** (backstop cap applies only to MIGRATING/FAILING_OVER). Handoff hooks collapse to a single **`Runnable`**; **`MaintenanceHandoff`** is removed. **`ConnectionPool#destroy`** closes the maintenance controller.
> 
> Tests cover null-target decode, marking scheduling/merges/supersession/churn immunity, and end-to-end `none` grace behavior in rebind mock tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 79db224909dd3cc6cb07dc3ce947ed6f1ff501a6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -171,8 +171,12 @@ public static Builder builder() {
...
   private Set<InitVisitor> initVisitors = new HashSet<>();
 
-  private static final long EXPIRE_NOT_SET = -1;
-  private long expireAt = EXPIRE_NOT_SET;
+  /**
+   * Maintenance mark, written by the controller's marking passes: this connection
+   * must be recycled instead of re-pooled. Advisory — enforced on return and by validation; never
+   * interrupts in-flight work.
+   */
+  private volatile boolean markedForReconnect;
```
atakavci: name `markedForReconnect` in connection object suggests complete different idea.

src/main/java/redis/clients/jedis/MaintenanceEventController.java
```diff
@@ -97,81 +146,119 @@ public boolean isRebindActive() {
...
+  /**
+   * The marking pass: mark every registered connection whose peer is one of the snapshot's sources,
+   * then run the handoff hooks (the pool evicts marked idles). Mark-only — never any I/O; only the
+   * pool destroys. Every state transition (new seq or merged source) marks against the exact
+   * snapshot it produced, so every source is covered by the transition that admitted it;
+   * overlapping passes are harmless (marking is an idempotent one-way write). A stale fire
+   * (superseded epoch) is a no-op via the seq guard.
+   */
+  private void markAffected(RebindState snapshot) {
+    if (rebind.seq != snapshot.seq) {
```
atakavci, line 226: in case of endpoint type `none`, this leaves all the connections alive from seq1 if there is a seq2 received before the half time (t/2). keep in mind its possible to read a previous notification later than next one in the timeline.

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -1077,8 +1073,12 @@ void removePushConsumer(PushConsumer consumer) {
     this.pushConsumers.remove(consumer);
   }
 
-  void expireAt(long expireAt) {
-    this.expireAt = expireAt;
+  void markForReconnect() {
+    this.markedForReconnect = true;
+  }
+
+  boolean isMarkedForReconnect() {
```
atakavci: i d take this as generic validity state/flag and expose it in such way as well, like `setInvalid`, `isValid`

src/main/java/redis/clients/jedis/ConnectionFactory.java
```diff
@@ -171,10 +173,21 @@ private void initAuthXManager() {
 
   private Connection build() {
     Connection conn = connectionBuilder.buildUninitialized();
+    registerForMaintenanceEvents(conn);
```
atakavci: move it into visitor if there is no signtificant reason to keep here.

src/main/java/redis/clients/jedis/ConnectionPool.java
```diff
@@ -129,6 +142,17 @@ public void close() {
     }
   }
 
+  @Override
+  public void destroy() {
+    try {
+      super.destroy();
+    } finally {
+      if (maintenanceController != null) {
```
atakavci, line 149: adding a hook by pool to controller and having controller that can receive multiple hooks gives the idea that controller has a lifecycle not bound to pool. This will end up controllers killed by other pools if controller gets shared at any point.

src/main/java/redis/clients/jedis/MaintenanceEventController.java
```diff
@@ -97,81 +140,119 @@ public boolean isRebindActive() {
...
+    return registry;
+  }
+
+  /** Idempotent: stops the maintenance scheduler, dropping any queued marking pass. */
+  @Override
+  public void close() {
+    synchronized (schedulerLock) {
+      closed = true;
+      if (scheduler != null) {
+        scheduler.shutdownNow();
```
atakavci, line 248: closing without checking if shared or not

src/main/java/redis/clients/jedis/MaintenanceEventController.java
```diff
@@ -97,81 +140,119 @@ public boolean isRebindActive() {
...
             cur.target, target);
           return;
         }
         if (cur.affected.contains(affectedPeer)) {
-          return; // already merged
+          // Re-delivery from an already-known source — e.g. the moving node notifying a
+          // connection created after the reconnect instant. No state change, no pass: such
+          // connections
+          // stay unmarked (temporal churn immunity).
+          return;
```
atakavci, line 177: in case this message hit after t/2 duration passed, and no `handleRebind` occurs anymore?

src/main/java/redis/clients/jedis/Connection.java
ggivo: Agreed — renamed to `retire`()/`isRetired()` in f97bcd1b9: a generic, one-way, advisory "leave pool service" flag; the pool disposes of it on return/validation/eviction.

src/main/java/redis/clients/jedis/Connection.java
ggivo: Made it generic as suggested, but kept "`retire`" over `isValid`/`setInvalid` — Connection already has `isBroken`/`isConnected`

src/main/java/redis/clients/jedis/MaintenanceEventController.java
ggivo: The scheduler is intentionally not shareable: the production path creates it lazily inside the controller (one per controller, no accessor), so `close()` always owns what it shuts down. The package-private `from(cfg, scheduler)` exists only as a test seam; the controller shuts the injected scheduler down on `close()`. Shared-executor support is out of scope here.

src/main/java/redis/clients/jedis/ConnectionPool.java
ggivo: Agreed the hook list over-signaled a shared, free-standing controller. The design is strictly 1:1 — the owner creates the controller, registers its reaction, and closes it; Tightened the API  in d3b2521ef: single-slot `setHandoffHook`

src/main/java/redis/clients/jedis/ConnectionFactory.java
ggivo: Deliberate: registration must precede socket init, and the visitor runs after the socket connects. `connect()` consults the rebind mapper with the resolved requested address (no live socket needed), so register-before-connect covers both races with a concurrent MOVING:

```
Case A: pass after connect()  — registered first → pass sees peer node-a → retires it ✓
Case B: pass before connect() — no peer yet, skipped → connect() sees the applied rebind → node-b ✓
```

Registering in the visitor adds a third case: `connect()` picks node-a before the MOVING is applied, the pass runs before the visitor registers → an unretired connection on the old node that no later pass visits.

src/main/java/redis/clients/jedis/MaintenanceEventController.java
ggivo: Correct — both cases are real limitations of the single-slot model here: seq2 arriving before seq1's t/2 orphans seq1's pending pass (the seq guard no-ops it), and a seq1 delivery read after seq2 is dropped as a stale replay.

Addressed in #4636, which builds on this PR: the single slot becomes a per-identity operation store (`MovingEventId` = seq + endpoint) where operations never supersede each other, they only expire. seq1's pass fires at its own t/2 against its own operation, a late-read earlier notification merges into its own operation instead of being dropped.

src/main/java/redis/clients/jedis/ConnectionFactory.java
atakavci: Visitor is capable of doing it before/after `connect()`;
```java
interface InitVisitor {

  void visitBeforeHandshake(Connection connection);

  void visitAfterHandshake(Connection connection);
}
```

src/main/java/redis/clients/jedis/MaintenanceEventController.java
ggivo: Partially by design, partially fixed in #4636. A MOVING from a new peer still triggers handleRebind after t/2 — it is scheduled with a delay that already passed, so it runs immediately. A MOVING read late (or with a lower seq) also reacts in #4636: events no longer replace each other, each gets its own operation and its own pass.

The only case with no reaction is a MOVING from a peer already in the affected set. That connection landed on the old node after the pass already ran; retiring it would only trigger another reconnect to the same node and the same notification again, so we leave it alone — the server closes it at time_s

src/main/java/redis/clients/jedis/MaintenanceEventController.java
```diff
@@ -206,39 +282,6 @@ private void relaxConnectionTimeoutsFor(Connection c, long expirationTime) {
...
-    public long getSeq() {
-      return seq;
-    }
-
-    /** The new endpoint connections will be routed to during the handoff window. */
-    public HostAndPort getTarget() {
-      return target;
-    }
-
-    /** Window the handoff is active for, after which the configured endpoint is used again. */
```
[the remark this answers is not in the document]
ggivo: see https://github.com/redis/jedis/pull/4625#discussion_r3645855590

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -544,10 +540,10 @@ public void close() {
     if (this.memberOf != null) {
       ConnectionPool pool = this.memberOf;
       this.memberOf = null;
-      if (isBroken() || markedForReconnect) {
+      if (isBroken()) {
         pool.returnBrokenResource(this);
```
atakavci, line 544: i see we stopped gating for retired conditions, this is on purpose?

atakavci: ah sorry, now pool handles it with `returnHook`

src/main/java/redis/clients/jedis/ConnectionFactory.java
ggivo: @atakavci 

Forgot about `visitBeforeHandshake()`!

Addressed now. Moved registration inside `MaintenanceAwareVisitor'.

src/main/java/redis/clients/jedis/MaintenanceEventController.java
```diff
@@ -97,81 +144,114 @@ public boolean isRebindActive() {
...
+          // connections
+          // stay unmarked (temporal churn immunity).
+          return;
         }
         RebindState next = cur.merge(affectedPeer);
         if (REBIND.compareAndSet(this, cur, next)) {
           logger.debug("Merged source {} into rebind seq={} (sources={})", affectedPeer, cur.seq,
             next.affected.size());
-          fireHandoffHook(e);
+          handleRebind(next);
```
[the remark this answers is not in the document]
ggivo: By design

Common case — one node address, several connections ('none', ttl=10s):

```
t=0.0  conn1 reads MOVING(seq=1)  → peer A is new → one pass scheduled for t=5
t=0.1  conn2 reads the same MOVING → peer A already known → ignored, nothing scheduled
t=0.2  conn3 reads the same MOVING → ignored
t=5.0  the single pass retires every connection on A → hook runs once
```

Merge case — same node reachable via two addresses (IPv4 A, IPv6 B):

```
t=0  conn1 (peer A) reads MOVING seq=1 → pass#1 scheduled for t=5, capturing {A}
t=2  conn2 (peer B) reads MOVING seq=1 → B is new → pass#2 scheduled for t=5, capturing {A,B}
t=5  pass#1 retires connections on A — it captured {A}, it cannot see B
     pass#2 retires A (idempotent no-op) and B
```

So an extra pass appears only when a delivery brings a new affected peer; re-deliveries from known peers never schedule anything.

src/main/java/redis/clients/jedis/MaintenanceEventController.java
ggivo: Opened a follow-up issue to discuss/track implementing repeating reconnecting if still receving same MOVING notification after t/2 on new connection 
https://github.com/redis/jedis/issues/4672

ggivo: Verified in #4707 (eb2cd1705): the TLS handoff completes under default FULL verification against a live RE deployment; propagating the mapped hostname into TLS socket creation for hostname-specific-cert/SNI-routing deployments remains tracked in #4708; covered by MaintNotificationsIT#connectionHandoffOnMovingOverTls.

PR #4629, opened 2026-07-19
fix: clear credential byte arrays in JedisSafeAuthenticator after use

insaf021:
## Summary

`JedisSafeAuthenticator.safeReAuthenticate()` encodes token credentials into byte arrays but does not zero them after use, unlike the other authentication paths (`Connection.authenticate()`, `Connection.hello()`, `RedisRestAPI.getAuthenticationHeader()`), which all clear credential bytes in a `finally` block. This aligns `JedisSafeAuthenticator` with that pattern.

The arrays are method-local and GC'd after the method returns, but until then the credential bytes remain readable in heap memory (e.g., in a heap dump). Zeroing them shrinks that window, not a functional bug.

## Testing

- Existing `TokenBasedAuthenticationUnitTests` pass (no behavioural change).

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisSafeAuthenticator.java
+++ b/src/main/java/redis/clients/jedis/JedisSafeAuthenticator.java
@@ -1,5 +1,6 @@
 package redis.clients.jedis;
 
+import java.util.Arrays;
 import java.util.Queue;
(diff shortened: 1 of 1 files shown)


ggivo:
Thanks for the contribution, will take it from here and merge ti once tests are green

PR #1573, opened 2017-08-23
Modified Tuple.compareTo()

sazzad16:
Tuple.compareTo() was not handled correctly. For example, following code used to output `0`, but it should not.

```
    Tuple tuple1 = new Tuple(user1, 1d);
    Tuple tuple2 = new Tuple(user2, 1d);
    System.out.println(tuple1.compareTo(tuple2));
```

This issue is _attempted_ to be solved.

Furthermore, SortedSet<Tuple> will maintain the default order of Redis SortedSet as long as the precision issue does not come into play.

Issue #1534, opened 2017-06-21
Tuple compareTo method incorrectly implemented for use in SortedSet


tfeak:
### Expected behavior
When placing Tuples into a SortedSet, they should not be considered "equal" by compareTo() if equals() would return false.

From the SortedSet JavaDoc.
"Note that the ordering maintained by a sorted set (whether or not an explicit comparator is provided) must be consistent with equals if the sorted set is to correctly implement the Set interface.  (See the Comparable interface or Comparator interface for a precise definition of consistent with equals.)  This is so because the Set interface is defined in terms of the equals operation, but a sorted set performs all element comparisons using its compareTo (or compare) method, so two elements that are deemed equal by this method are, from the standpoint of the sorted set, equal.  The behavior of a sorted set is well-defined even if its ordering is inconsistent with equals; it just fails to obey the general contract of the Set interface."

### Actual behavior
CompareTo() currently short circuits to returning 0 (implying equality) if either the score or element member are equal. This is decidedly different than the implementation of the equals() method. As a result, you can't use Tuple with a SortedSet.

### Steps to reproduce:

This source code demonstrates the issue. Even though the 2 Tuple instances are not equal, the Set has a size of 1 after inserting both objects.

    public static void main(String[] args) {
        Tuple tuple1 = new Tuple("user1", 10.0);
        Tuple tuple2 = new Tuple("user2", 10.0);

        Set<Tuple> tuples = new TreeSet<>();
        tuples.add(tuple1);
        tuples.add(tuple2);

        System.out.println(tuples.size());
        System.out.println(tuple1.equals(tuple2));
    }

#### Jedis version:
2.9
#### Redis version:
N/A
#### Java version:
1.8

sazzad16:
PR #1573

sazzad16:
Resolved by #1573

PR #4632, opened 2026-07-20
compare byte arrays as unsigned in ByteArrayComparator

insaf021:
ByteArrayComparator.compare(new byte[]{ (byte) 0x80 }, new byte[]{ 0x7F }) == -1

Java bytes are signed, so a byte with the high bit set reads as negative: 0x80 becomes -128 and sorts before 0x7F. The comparator is meant to order raw bytes lexicographically, which is unsigned, so every value in 0x80..0xFF currently orders before 0x00..0x7F rather than after. That disagrees with Redis' own unsigned byte ordering and shows up for any non-ASCII UTF-8 or binary content.

Tuple implements Comparable<Tuple> and falls back to this comparator to break score ties, so Collections.sort / TreeSet over sorted-set members comes out wrong for those members. The existing tests only cover ASCII (all bytes < 0x80), which is why it went unnoticed.

Fixed by comparing each byte as unsigned with Integer.compare(a & 0xFF, b & 0xFF); the -1/0/1 result and the length tie-break stay as they were. Added a regression for the high-bit and multi-byte UTF-8 cases.

---

> [!NOTE]
> **Medium Risk**
> Changes sort order for non-ASCII/binary sorted-set members and any other ByteArrayComparator callers; behavior aligns with Redis but may reorder existing client-side collections that relied on the buggy ordering.
> 
> **Overview**
> Fixes **lexicographic byte ordering** in `ByteArrayComparator` so each byte is compared as **unsigned** (`0..255`) via `Integer.compare(a & 0xFF, b & 0xFF)` instead of signed Java `byte` comparison. Length tie-breaking is unchanged.
> 
> That wrong ordering affected anything using this helper for raw keys— notably **`Tuple` score ties** in sorted-set sorting— for UTF-8 or binary members with bytes `>= 0x80`. Adds **`testHighBitBytesOrderUnsigned`** for `0x80` vs `0x7F` and a multi-byte UTF-8 case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9468374ba0de653b6fcd6de37470973c10977391. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/main/java/redis/clients/jedis/util/ByteArrayComparator.java
+++ b/src/main/java/redis/clients/jedis/util/ByteArrayComparator.java
@@ -11,10 +11,8 @@ public final class ByteArrayComparator {
     int lmin = Math.min(len1, len2);
 
     for (int i = 0; i < lmin; i++) {
-      byte b1 = val1[i];
-      byte b2 = val2[i];
-      if (b1 < b2) return -1;
-      if (b1 > b2) return 1;
+      int result = Integer.compare(val1[i] & 0xFF, val2[i] & 0xFF);
+      if (result != 0) return result;
     }
 
     if (len1 < len2) return -1;
--- a/src/test/java/redis/clients/jedis/util/ByteArrayComparatorTest.java
+++ b/src/test/java/redis/clients/jedis/util/ByteArrayComparatorTest.java
@@ -24,4 +24,19 @@ public class ByteArrayComparatorTest {
     assertTrue(ByteArrayComparator.compare(foo, fooo) < 0);
     assertTrue(ByteArrayComparator.compare(fooo, foo) > 0);
   }
+
+  @Test
+  public void testHighBitBytesOrderUnsigned() {
+    // Bytes must order as unsigned (0..255) to match Redis' own byte ordering.
+    byte[] low = { (byte) 0x7F }; // 127
+    byte[] high = { (byte) 0x80 }; // 128 unsigned
+    assertTrue(ByteArrayComparator.compare(high, low) > 0);
+    assertTrue(ByteArrayComparator.compare(low, high) < 0);
+
+    // ASCII 'e' (0x65) sorts before the UTF-8 bytes of U+00E9 (0xC3 0xA9).
+    byte[] e = { (byte) 0x65 };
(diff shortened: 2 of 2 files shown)


ggivo:
Thanks — the signed-vs-unsigned observation is technically correct. However, the change is itself behavior-breaking: any existing application sorting Tuples (or using `ByteArrayComparator` directly) with bytes ≥ 0x80 will see its ordering silently change on upgrade. To weigh that risk we need evidence of actual impact:

1. **Who is affected.** Is there a report, linked issue, or real application where the current ordering caused a problem?
2. **An end-to-end scenario** (commands, data, expected vs. actual outcome) where it produces a wrong observable result for a supported use — not a client-side sort the application performs itself.

Without that, the compatibility risk outweighs the fix, so we'll hold this until the evidence is provided. Thanks!

insaf021:
Fair questions. I traced the callers first: ByteArrayComparator's only consumer in the tree is Tuple.compareTo, and the zset builders (TUPLE_LIST, TUPLE_ZSET's LinkedHashSet) all preserve server order, so Jedis never sorts Tuples internally. That means I can't show an end-to-end path where Jedis alone produces a wrong observable result, and I don't have a report of the high-bit case in the wild either. The ordering only surfaces through Tuple's public Comparable contract when the application sorts.

On who relies on that contract: client-side sorting is why this code exists at all. #1052 and #1534 were user reports about sorting and merging tuples from several zsets, and #1573, which introduced ByteArrayComparator, states the intent as "SortedSet<Tuple> will maintain the default order of Redis SortedSet". For bytes >= 0x80 it doesn't: after ZADD k 0 cafz 0 café, ZRANGE returns [cafz, café] (unsigned memcmp, 0x7A vs 0xC3), while sorting the same tuples via compareTo gives [café, cafz]. So the signed compare breaks the ordering #1573 promised, but only for applications doing their own sort, which I understand is outside the bar you've set.

Given that, holding it is reasonable. If you'd ever want it at a major boundary instead, I can add a note under docs/migration-guides.

insaf021:
any thoughts?

PR #4636, opened 2026-07-22
[maintenance events] Support overlapping MOVING events at pool level (CAE-3395)

ggivo:
## Summary

Builds on #4625. Replaces the controller's single active MOVING slot (`volatile RebindState` + CAS) with a per-operation store so distinct MOVING events that overlap in time on one pool no longer fight over one slot.

A pool talks to multiple nodes (or one node across a re-plan), so distinct MOVING events — different `seq` **or** different target — can be unexpired simultaneously. The single-slot model wholesale-replaced on a higher `seq`, dropped a lower `seq` as a stale replay, and rejected same-`seq`/different-target — in each case orphaning the earlier event's remap, marking pass, and relaxation window.

## Change

- **`MovingOperations`** — a `ConcurrentHashMap<MovingEventId, MovingOperation>` keyed by `(seq, original endpoint)`, deduplicating the per-connection deliveries of one server-side operation into a single pool-wide entry. Admission (dedup + peer merge) is atomic per key via `compute`; expired entries are removed by readers on sight, so the post-maintenance hot path returns to `isEmpty()`.
- **Per-peer remap** — `getSocketAddress` resolves the active operation for the connecting peer across all operations, at connect time (a mid-window DNS repoint is honored).
- **Per-operation marking** — each applied transition marks its own affected set; a newer event never orphans an earlier unexpired one.
- **Pool-wide relaxation** — relaxed timeouts stay in effect while any operation is unexpired and lift once all have expired.

## Behavior deltas

1. Same-`seq`/different-target is a distinct concurrent event, no longer a rejected conflict.
2. Lower-`seq` events are admitted (identity, not ordering); an earlier still-active event keeps remapping/relaxing after a newer one arrives and even after the newer one expires first.
3. A pending `none` marking pass runs after a newer event arrives instead of no-oping.
4. Relaxation ends at the **last** active event's deadline, not the latest-arrived event's.
5. The remap target is resolved at connect time, not at event admission.

## Tests

- Unit: `MaintenanceEventControllerTest`, `MaintenanceMarkingTest`, `MaintenanceEventIdentityTest` — overlapping scenarios on distinct peers (per-peer remap, relaxation outlives the shorter event, pending `none` pass survives a newer event, identity dedup).
- Integration: `AbstractRebindBehaviorTest`, `AbstractRelaxedTimeoutBehaviorTest` — chained/overlapping MOVING via mock servers, for both `RedisClient` and `MultiDbClient`.

---

Draft — opened for review; not ready to merge.

---

> [!NOTE]
> **Medium Risk**
> Touches core connection-pool maintenance, remap, and retirement during live server handoffs; behavior changes are broad but heavily covered by unit and SCH integration tests.
> 
> **Overview**
> Replaces the maintenance controller’s **single active MOVING slot** (`RebindState` + CAS) with **`MovingOperations`**, a concurrent store keyed by event **`identity()`** — `(seq, target)` for MOVING — so multiple unexpired MOVING announcements can coexist instead of superseding or rejecting each other.
> 
> **Per-operation behavior:** each admitted operation gets its own affected-peer set, remap window, and marking schedule; **`getSocketAddress`** picks the active operation for a peer and resolves the target **at connect time**. Pool-wide **relaxed timeouts** stay on while **any** operation is unexpired. Connections use **`retireAt(deadline)`** instead of a one-way boolean so `'none'` MOVINGs can retire at half the grace while targeted moves retire immediately; duplicate deliveries can still stamp retirement without re-running a full pass.
> 
> **Semantics changes:** same `seq`/different target and lower `seq` are distinct concurrent events; newer events no longer cancel earlier pending `'none'` passes; handoff/eviction runs on the maintenance scheduler (fixes async eviction / evictor livelock scenarios covered in tests).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5500c4aeebd96ce03cb2980459c7392e51fe88be. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/main/java/redis/clients/jedis/MaintenanceEvent.java
```diff
@@ -13,6 +14,42 @@ abstract class MaintenanceEvent {
...
+  /**
+   * Identity of the server-side operation this event announces; equality is the dedup rule for
+   * folding per-connection deliveries into one pool-wide operation.
+   */
+  MaintenanceEventId identity() {
+    return new MaintenanceEventId(getClass(), seq);
+  }
+}
+
+class MaintenanceEventId {
```
atakavci: does this means there would be `MOVING` and `MIGRATING` with same seq?

src/main/java/redis/clients/jedis/MaintenanceEventController.java
```diff
@@ -56,7 +53,7 @@ private MaintenanceEventController(MaintenanceNotificationsConfig config,
 
     TimeoutInfo relaxedTimeoutInfo = new TimeoutInfo(config.getRelaxedTimeout(),
         config.getRelaxedBlockingTimeout());
-    this.timeoutSupplier = () -> rebind.isValid() ? relaxedTimeoutInfo : null;
+    this.timeoutSupplier = () -> movingOperations.hasActive() ? relaxedTimeoutInfo : null;
```
atakavci, line 56: `hasActive` is on hot-path and not sure if `isEmpty` is that cheap operation, it should be more than a `size == 0` on a concurrent map.

src/main/java/redis/clients/jedis/MaintenanceEventController.java
```diff
@@ -127,88 +124,51 @@ Runnable getHandoffHook() {
...
-   * is one of the active rebind's affected sources and the window is still open; else returns null
-   * (no remap).
+   * Post-DNS address mapper: remaps the resolved peer to its active operation's endpoint when the
+   * peer is an affected source of an unexpired MOVING operation; else returns null (no remap). The
+   * endpoint is resolved here, at connect time, so a DNS repoint mid-window is honored.
    */
   @Override
   public SocketAddress getSocketAddress(SocketAddress resolved) {
-    return (rebind.isValid() && rebind.affected.contains(resolved)) ? rebind.target : null;
+    MovingOperation op = movingOperations.findActive(o -> o.affected.contains(resolved));
```
atakavci, line 133: do we need to consider if same resolved address somehow gets into two seperate rebind operations?

src/main/java/redis/clients/jedis/MaintenanceEventController.java
```diff
@@ -219,12 +179,12 @@ private void handleRebind(RebindState snapshot) {
...
-   * runs the handoff hooks. No I/O happens here — the pool destroys retired connections; retiring
-   * is idempotent, so overlapping passes are harmless.
+   * runs the handoff hook. No I/O happens here — the pool destroys retired connections; retiring is
+   * idempotent, so overlapping passes are harmless.
    */
-  private void retireAffected(RebindState snapshot) {
-    if (rebind.seq != snapshot.seq) {
-      return; // superseded; the new epoch's transitions run their own marking passes
+  private void retireAffected(MovingOperation snapshot) {
+    if (!snapshot.isValid()) {
```
atakavci: possible to see late execution of scheduled task would leave relevant connections untouched.

src/main/java/redis/clients/jedis/MaintenanceEventController.java
ggivo: A brief check of the performance of isEpty shows no locking/CAS used during the check.
Assuming it is good enough, will follow up with checking benchmark results.

src/main/java/redis/clients/jedis/MaintenanceEventController.java
ggivo: discussed offline and decided  not needed

src/main/java/redis/clients/jedis/MaintenanceEvent.java
ggivo: Droped completely MaintenanceEventId for now, use Object identity instead

src/main/java/redis/clients/jedis/MaintenanceEventController.java
ggivo: Discussed and decided to open a [follow-up task](https://github.com/redis/jedis/issues/4669) for improvement where marking is performed in the calling thread and eviction in the scheduled one.  For this to work need to keep the retireAt timestamp while marking

in addition dropping the` if (!snapshot.isValid())` to trigger IDLE connection eviction   even if the scheduled task is  delayed

PR #4644, opened 2026-07-24
test: Fix flaky MultiDbConnectionProviderTest caused by shared test key pollution

### LINKED ISSUE #4643: test: Flaky `MultiDbConnectionProviderTest` — WRONGTYPE failures caused by shared test key pollution
## Symptom

`MultiDbConnectionProviderTest` fails sporadically in CI with:

```
[ERROR]   MultiDbConnectionProviderTest.userCommand_connectionExceptions_thenMultipleTemporary_thenPermanent_inOrder:293 » JedisData WRONGTYPE Operation against a key holding the wrong kind of value
[ERROR]   MultiDbConnectionProviderTest.userCommand_firstTemporary_thenPermanent_inOrder:251 » JedisData WRONGTYPE Operation against a key holding the wrong kind of value
```

Example failure run: https://github.com/redis/jedis/actions/runs/30079395156/job/89437290034?pr=4625

## Root cause

The failing assertions are plain `GET foo` calls against the shared `standalone0`/`standalone1` test endpoints. The key `foo` is a well-known name written by many other test classes with non-string types (`TupleSortedSetTest` via `zadd`, `TransactionV2Test`/`ReliableTransactionTest` via `sadd`/`zadd`, list/geo command test bases, ...).

`JedisCommandsTestBase` calls `flushAll()` only in `@BeforeEach` — never after — so whatever the **last** test of a class writes stays on the server after the class finishes. In the failing run, `TupleSortedSetTest` ran before `MultiDbConnectionProviderTest` in the same fork and left `foo` as a zset on `standalone0`; the subsequent `GET foo` then failed with WRONGTYPE. The failure is order-dependent, which is why it appears sporadic.

Reproducible locally against the unmodified test:

```
redis-cli -a foobared -p 6379 zadd foo 1 a
mvn test -Dtest=MultiDbConnectionProviderTest -DexcludedGroupsForUnitTests=scenario
```

## Fix

Two complementary changes:

1. **Stop the pollution at the source:** `JedisCommandsTestBase` now also calls `flushAll()` in `@AfterEach`, so no subclass leaks its last test's keys to the rest of the suite.
2. **Make readers robust against leftover keys:** a reusable `TestKeyRegistry` test utility generates unique per-test key names (`ClassName.methodName:name`, auto-registered for cleanup, cluster hash-tag aware). `MultiDbConnectionProviderTest` is migrated to it; other test classes reading well-known keys on the shared endpoints can adopt it incrementally.

ggivo:
## What
Fix sporadic WRONGTYPE failures in `MultiDbConnectionProviderTest` caused by test-key pollution on the shared Redis endpoints, and add a reusable `TestKeyRegistry` utility to prevent this class of flakiness.

## Why
`MultiDbConnectionProviderTest` reads the well-known key `foo` on the shared `standalone0`/`standalone1` endpoints. `JedisCommandsTestBase` subclasses flush only in `@BeforeEach`, so the last test of each class leaks its keys — e.g. `TupleSortedSetTest` leaves `foo` as a zset, and a later `GET foo` fails with `WRONGTYPE` depending on class execution order.

Fixes #4643

## Changes
- **`TestKeyRegistry`** (new test utility): generates unique, human-readable keys namespaced by `ClassName.methodName`; every key is auto-registered and deletable via `cleanup()` with a per-slot `DEL` fallback on cluster `CROSSSLOT` errors.
  - Class-level scope: a registry created in `@BeforeAll` derives a `ClassName` identifier, so class-level keys survive between tests and are cleaned once in `@AfterAll`.
  - Binary keys: `bKey(pattern)` mirrors `key(pattern)` via `SafeEncoder`; keys are stored internally as bytes so string- and binary-created keys share one entry and cleanup is a single binary `DEL`.
- **`MultiDbConnectionProviderTest`**: migrated the three tests using `"foo"` to registry-generated per-test keys.
- **`JedisCommandsTestBase`**: also `flushAll()` in `@AfterEach`, so no subclass leaks keys to the rest of the suite.

Verified by planting `foo` as a zset on both endpoints: the unmodified test reproduces the exact CI failure; with this change all tests pass.

---

> [!NOTE]
> **Low Risk**
> Test-only changes (new utility, integration test keys, and teardown flush); no production client behavior is modified.
> 
> **Overview**
> Addresses sporadic **WRONGTYPE** failures when integration tests reuse shared Redis endpoints and collide on keys like `"foo"`.
> 
> Introduces **`TestKeyRegistry`**, a test helper that builds per-test namespaced keys (`ClassName.methodName:…`), tracks them, and deletes them via `cleanup()` (with per-slot `DEL` when cluster clients reject cross-slot multi-key deletes). **`MultiDbConnectionProviderTest`** switches its failover scenarios from `"foo"` to registry keys and tears them down through dedicated control `Jedis` clients on both standalones.
> 
> **`JedisCommandsTestBase`** also runs `flushAll()` in `@AfterEach` (errors ignored) so subclasses are less likely to leave data on the shared endpoint after a class finishes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4cb571b485eaa150e06f1d35a44716aeefc8ec68. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


atakavci (approved):
LGTM!

Issue #4557, opened 2026-06-04
Potential deadlock with pipelines


snago:
Imagine the following (simplified) scenario:
Using a `JedisCluster` on a cluster with 2 shards (A and B).
Connection pools configured for max 4 connections per shard.

- Open 8 pipelines.
- 4 pipelines take a connection to shard A.
- 4 pipelines take a connection to shard B.
- The first 4 pipelines then need a connection to shard B and block because all connections in B's pool are already in use.
- The other 4 pipelines then need a connection to shard A and block because all connections in A's pool are already in use.

At this point:
- Each pipeline is holding a connection to one shard.
- Each pipeline is waiting for a connection to the other shard.
- No pipeline can make progress and release its held connection.

This appears to create a circular wait condition and a classic deadlock.

We suspect this is what (rarely) happenes to us in production (but in our case it's 9 shards and connection pools with max 48 connections).
We use Jedis 5.2.0 and are in the progress of upgrading, but from briefly looking at the code it looks like this can still happen.
Pipelines will take connections lazily and the order depends on the order that keys are used, and no connection is released (except for errors) until the pipeline is closed.

Questions:
1. Is this scenario possible with the most recent Jedis version?
2. Is there any protection against this type of circular wait condition?
3. If this is a known limitation, are there recommended pool sizing or usage patterns to avoid it?

ggivo:
@snago 
 Thanks for the write-up — your analysis is correct.

 1. Yes, this is still possible on the latest release. `MultiNodePipelineBase.appendCommand()` acquires per-shard connections lazily and holds them until `sync()`/`close()`. With shared per-shard pools and unordered acquisition, both hold-and-wait and circular-wait conditions can be satisfied.

 2. A few possible mitigations come to mind:

 - Size per-shard pools such that `maxTotal >= maximum` concurrent pipelines (eliminates waiting).
 - Configure a finite maxWaitMillis on the pool  (fail-fast instead of deadlocking).
 - Keep pipelines single-shard by grouping keys by slot client-side.

 As a potential future improvement, it may be worth exploring a pipeline implementation that defers connection acquisition until `sync()`. Instead of borrowing connections in` appendCommand()`, commands could be buffered per node and all required connections acquired during `sync()` in a deterministic order (for example, sorted by HostAndPort). This would eliminate both hold-and-wait and circular-wait conditions, at the cost of additional client-side memory for buffered commands and the loss of overlap between command submission and network I/O.

snago:
[...]

PR #4645, opened 2026-07-26
fix(cluster): prevent lock-inversion deadlock in MultiNodePipelineBase (#4557)

anshullakra007:
## What does this PR do?
Fixes a confirmed lock-inversion deadlock in `ClusterPipeline` (and any pipelines extending `MultiNodePipelineBase`) when accessing multiple shards concurrently under heavy load.

## Root Cause Analysis
Resolves Issue #4557. 
When `MultiNodePipelineBase#appendCommand()` processes commands targeting different shards, it used a **Hold-and-Wait** locking pattern. Specifically, it borrowed the connection from the target node's connection pool immediately and held it in an internal map until `pipeline.sync()` or `pipeline.close()` was executed. 

If two pipelines executed concurrently with different target sequences (e.g., Pipeline 1 appends Command for Shard A then B; Pipeline 2 appends Command for Shard B then A), they permanently deadlocked waiting for the other pipeline to release its connection pool lock.

## Architectural Fix
To break this circular wait condition, this PR eliminates eager connection acquisition during the command-building phase:
1. `MultiNodePipelineBase#appendCommand()` now only buffers the `CommandObject` in memory grouped by `HostAndPort`.
2. Connection borrowing and network I/O have been completely deferred to the `sync()` phase.
3. During `sync()`, each worker thread securely borrows one connection, streams the entire buffered batch to the socket, reads the responses, and immediately releases the connection. 

A worker thread never holds a connection to Shard A while blocking on Shard B's pool. Deadlock is architecturally eliminated.

## Testing & Verification
Added a deterministic JUnit integration test (`MultiNodePipelineDeadlockTest.java`).
The test configures a strict 1-connection pool across two mock shards and uses a `CyclicBarrier` to enforce an exact lock-inversion trace scenario across two threads. 

- **Before this PR:** The test hangs indefinitely and throws a `TimeoutException`.
- **After this PR:** The test passes instantly (`~250ms`).

---

> [!NOTE]
> **Medium Risk**
> Changes core cluster pipeline connection lifecycle and error propagation for all `MultiNodePipelineBase` subclasses; behavior should be equivalent for success paths but failure handling and timing under load differ from eager connection holding.
> 
> **Overview**
> Fixes a **lock-inversion deadlock** in `MultiNodePipelineBase` (including `ClusterPipeline`) when multiple pipelines append to different shards in opposite order under a small connection pool.
> 
> **`appendCommand`** no longer borrows or keeps per-shard connections; it only queues `CommandObject`s and `Response`s by `HostAndPort`. **Network I/O moves to `sync()`**: each worker gets a connection, sends the buffered batch, reads replies, then closes the connection. **`sync()`** also aggregates the first worker failure via an atomic reference, clears pipeline state in `finally`, and **`close()`** relies on that path instead of holding a connection map across the build phase.
> 
> Adds **`MultiNodePipelineDeadlockTest`**, which uses single-slot pools and a `CyclicBarrier` to reproduce the A-then-B vs B-then-A scenario and assert both threads finish within a timeout.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e722177b8587eefec8d94b1ff9742b922cd42554. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
+++ b/src/main/java/redis/clients/jedis/MultiNodePipelineBase.java
@@ -90,14 +90,7 @@ public abstract class MultiNodePipelineBase extends AbstractPipeline {
 
   @Override
   public void close() {
-    try {
-      sync();
-    } finally {
-      // Connections are now managed cleanly within sync() and returned to their pools.
-      // We clear the state to ensure pipeline reuse won't carry over stale buffers.
-      pipelinedResponses.clear();
-      pipelinedCommands.clear();
-    }
+    sync();
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


ggivo:
@anshullakra007 
The team is stretched, working on the 8.10 server release currently; it might take some time before looking into this PR in detail.

Looking at the suggested approach is a breaking change and changes the behavior of `MultiNodePipelineBase`. I would rather try to preserve the existing  `ClusterPipeline` behaviour unchanged, and check how we can introduce  new implementation of it.

anshullakra007:
> @anshullakra007 The team is stretched, working on the 8.10 server release currently; it might take some time before looking into this PR in detail.
> 
> Looking at the suggested approach is a breaking change and changes the behavior of `MultiNodePipelineBase`. I would rather try to preserve the existing `ClusterPipeline` behaviour unchanged, and check how we can introduce new implementation of it.

@ggivo Thank you for the architectural guidance. I completely agree with preserving the legacy socket-writing behavior to avoid any breaking changes for existing ClusterPipeline users.

To address this, I have:

Completely reverted the edits to MultiNodePipelineBase and ClusterPipeline.

Introduced a parallel, opt-in architecture (DeferredMultiNodePipelineBase and DeferredClusterPipeline) that implements the deferred connection acquisition to safely eliminate the lock-inversion deadlock.

Exposed this safely via deferredPipelined() on JedisCluster and RedisClusterClient.

Addressed all AI bot reviews regarding charset encoding (SafeEncoder) and corrected the test routing logic to deterministically verify the deadlock fix.

I understand the team is heavily focused on the 8.10 server release right now, so please take your time. Let me know if the deferredPipelined naming convention aligns with the team's vision!

githubname1024:
This PR looks great! 
I do have one related question. When a Redis cluster failover, pipelines currently don't retry on MOVED exceptions. Would it be possible to fix this in this PR as well? (Note: I don't know why not contains the retry logic, considering the consistent?).

anshullakra007:
> This PR looks great! I do have one related question. When a Redis cluster failover, pipelines currently don't retry on MOVED exceptions. Would it be possible to fix this in this PR as well? (Note: I don't know why not contains the retry logic, considering the consistent?).

Hi @githubname1024,

Thank you so much for taking the time to review this PR and for your very kind words! I truly appreciate your feedback.

You raise an excellent point regarding the MOVED exceptions during cluster failovers. It is absolutely a valid and important feature request!

Because adding automatic retry logic for pipelines requires some complex architectural considerations (such as ensuring that partially executed pipeline commands are handled safely without breaking transaction guarantees), it would significantly expand the scope of this current PR.

To ensure we can get this specific deadlock fix merged smoothly and without delay, would you be open to creating a separate issue for the MOVED exception retry logic? I think having a dedicated space to discuss the design and implementation for that feature would be incredibly beneficial for the project.

Thank you again for your wonderful suggestion and your help in improving the code!

PR #4647, opened 2026-07-26
fix: reject malformed RESP length fields in RedisInputStream

### LINKED ISSUE #4646: RESP length field is not validated, desynchronising the connection
### Expected behavior

A bulk length or multibulk count that is not a valid, in-range decimal number is a protocol error. `Protocol.read` should raise `JedisConnectionException` so the connection is dropped rather than reused.

### Actual behavior

`RedisInputStream.readIntCrLf()` narrows `readLongCrLf()` with a plain cast, and `readLongCrLf()` accepts any byte as a digit (`value = value * 10 + b - '0'`) with no overflow check. Three ways a length field parses to something other than what is on the wire:

* `$4294967296` (2^32) truncates to `0`. The header is consumed as an empty bulk reply and the remainder of the payload stays in the buffer, framed as a fresh reply that the *next* command on that connection picks up.
* `$0:` parses as length 10, because `':'` is `'0' + 10`. Ten bytes of whatever follows come back as the value.
* A 20-digit length wraps modulo 2^64 back into range, e.g. `$18446744073709551621` becomes 5.

The guard in `Protocol.processBulkReply` (`len < skipBytes`) only rejects the sign-flip case, `$2147483648`, which truncates to a negative int. It does not catch any of the above.

### Steps to reproduce

No server needed, this reads straight from a byte array:

```java
RedisInputStream in = new RedisInputStream(
    new ByteArrayInputStream("$4294967296\r\n\r\n+INJECTED\r\n".getBytes()));

Object first = Protocol.read(in);   // reply to the command we actually sent
Object second = Protocol.read(in);  // reply to the next command on this connection
```

On master:

```
first  -> byte[0] ""
second -> byte[8] "INJECTED"
```

The same happens with `"$0:\r\n0123456789\r\n"` (returns `0123456789`) and with `"*4294967297\r\n..."` for a multibulk count.

### Why it matters

Every one of these paths returns silently wrong data instead of failing. Because the parser stays out of step with the stream, the misframing is not confined to the command that hit it: the connection goes back to the pool and hands the leftover bytes to unrelated commands as their replies. Anything that can corrupt or truncate a reply, a proxy sitting in front of Redis, a half-closed socket, or a byte-level fault on the wire, turns into wrong values returned to the application rather than a `JedisConnectionException` that makes the pool discard the connection.

Validating the digits, the accumulation overflow and the int range in `RedisInputStream` keeps this contained to the connection that saw it. No valid Redis reply changes: real lengths are short decimal digits well inside int range, and `:9223372036854775807` / `:-9223372036854775808` still parse.

I have a patch with unit tests for each case and can open a PR if this is accepted.

#### Jedis version:

8.0.0-SNAPSHOT (master), same code on all current branches

#### Redis version:

n/a, this is parser-level

#### Java version:

8

insaf021:
Feeding `$4294967296\r\n\r\n+INJECTED\r\n` to `Protocol.read` twice, on master:

    reply 1 -> byte[0] ""
    reply 2 -> byte[8] "INJECTED"

`readIntCrLf()` narrows `readLongCrLf()` with a plain cast, so 2^32 lands as 0: the header is consumed as an empty bulk reply and the remainder stays framed as a fresh reply, which the next command on that pooled connection collects. The `len < skipBytes` guard in `processBulkReply` only catches the sign-flip case, `$2147483648`.

`readLongCrLf` widens it. The accumulator takes any byte as a digit, so `$0:` reads as length 10, and with no overflow check a 20-digit length wraps modulo 2^64 back into range.

Validating the digits, the accumulation overflow and the int range turns all three into `JedisConnectionException`, so the pool drops the connection rather than serving misframed replies from it. Accumulation is negated to keep `Long.MIN_VALUE` representable, so `:-9223372036854775808` and `:9223372036854775807` parse exactly as before.

Closes #4646

---

> [!NOTE]
> **Medium Risk**
> Touches low-level protocol parsing on every Redis read; behavior changes for previously accepted malformed input, but the change is defensive and should only surface bad streams as connection errors.
> 
> **Overview**
> Hardens RESP numeric parsing in **`RedisInputStream`** so malformed bulk/array lengths and integer replies fail fast instead of desynchronizing the byte stream on pooled connections.
> 
> **`readIntCrLf()`** now checks the parsed value fits in `int` before casting, so lengths like `4294967296` (2³²) no longer truncate to `0` and leave trailing bytes to be read as the next reply.
> 
> **`readLongCrLf()`** only accepts decimal digits, enforces overflow bounds while accumulating (using negated accumulation so **`Long.MIN_VALUE`** still parses), and rejects out-of-range values. Non-digit characters (e.g. `$0:` misread as length 10) and overflow now raise **`JedisConnectionException`**, which should cause the pool to drop the bad connection.
> 
> **`ProtocolTest`** adds coverage for out-of-int-range bulk/multi lengths, non-digit length fields, and long boundary integer replies.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f04191fd2e482647d2c1d2377e0e3aa6cc6db1f9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/main/java/redis/clients/jedis/util/RedisInputStream.java
+++ b/src/main/java/redis/clients/jedis/util/RedisInputStream.java
@@ -29,6 +29,8 @@ public class RedisInputStream extends FilterInputStream {
       System.getProperty("jedis.bufferSize.input",
           System.getProperty("jedis.bufferSize", "8192")));
 
+  private static final long MIN_VALUE_DIV_10 = Long.MIN_VALUE / 10;
+
   protected final byte[] buf;
 
   protected int count, limit;
@@ -194,7 +196,11 @@ public class RedisInputStream extends FilterInputStream {
   }
 
   public int readIntCrLf() {
-    return (int) readLongCrLf();
+    final long value = readLongCrLf();
+    if (value < Integer.MIN_VALUE || value > Integer.MAX_VALUE) {
+      throw new JedisConnectionException("Number is out of int range: " + value);
+    }
+    return (int) value;
   }
 
   public long readLongCrLf() {
@@ -207,6 +213,8 @@ public class RedisInputStream extends FilterInputStream {
       ++count;
     }
 
+    // Accumulate negated so that Long.MIN_VALUE stays representable.
+    final long limit = isNeg ? Long.MIN_VALUE : -Long.MAX_VALUE;
     long value = 0;
     while (true) {
       ensureFill();
@@ -221,11 +229,22 @@ public class RedisInputStream extends FilterInputStream {
 
         break;
       } else {
-        value = value * 10 + b - '0';
+        final int digit = b - '0';
+        if (digit < 0 || digit > 9) {
+          throw new JedisConnectionException("Unexpected character!");
+        }
+        if (value < MIN_VALUE_DIV_10) {
+          throw new JedisConnectionException("Number is out of range.");
+        }
+        value *= 10;
+        if (value < limit + digit) {
+          throw new JedisConnectionException("Number is out of range.");
+        }
+        value -= digit;
       }
     }
 
-    return (isNeg ? -value : value);
+    return (isNeg ? value : -value);
   }
 
   public double readDoubleCrLf() {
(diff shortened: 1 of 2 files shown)


ggivo:
Thanks for the patch, but we've decided not to take this change; see the reasoning in #4646.

PR #4650, opened 2026-07-27
[Redis 8.10] Support TS.NRANGE and TS.NREVRANGE time series commands

atakavci:
## Summary
Adds client support for the multi-key pivot time series commands `TS.NRANGE` and `TS.NREVRANGE`. These query an explicit, ordered list of keys over a timestamp range and return a timestamp-major ("pivoted") reply — one row per distinct timestamp, with one value column per key (or per aggregator). The commands are wired through the full command matrix (`CommandObjects`, `UnifiedJedis`, `PipeliningBase`, and the `RedisTimeSeries` command/pipeline interfaces) and are marked `@Experimental`.

- Keys are passed to the command methods, not carried in params. The new `TSNRangeParams` holds only `fromTimestamp`/`toTimestamp` plus optional args (`LATEST`, `FILTER_BY_TS`, `FILTER_BY_VALUE`, `COUNT`, and the aggregation clause).
- Aggregation is per-key: exactly one `AGGREGATION` token is emitted per key. Convenience overloads accept a single aggregator (single-key only), one aggregator per key, or multiple comma-joined aggregators per key. Matching the token count to `numkeys` is enforced by the server, not the client.
- Key order and duplicate keys are significant and preserved; treated as a single-shard, key-routed command (all keys must share a hash slot in cluster mode).
- Reuses existing `TSElement`/`MultiValueTSElement` types and `AggregationType.joinRaw`.

- A missing sample or missing aggregation bucket for a key at a given row is surfaced as `NaN`, indistinguishable from a stored/aggregated `NaN`.
- Rows are returned in server order: increasing timestamps for `TS.NRANGE`, decreasing for `TS.NREVRANGE`. A new `TIMESERIES_PIVOT_ELEMENT_LIST` builder parses the pivoted reply for both RESP2 and RESP3, yielding a single-value `TSElement` for one column or a `MultiValueTSElement` for several.

## Testing
Adds unit coverage for wire-argument construction and `TSNRangeParams` (option ordering, `equals`/`hashCode`, aggregation validation) across `CommandObjects`, `UnifiedJedis`, and pipeline surfaces, plus an integration test base exercising the commands against a live server.

## Notes
- New public API is annotated `@since 8.0` and `@Experimental`; signatures may change.
- The `pom.xml` change only adds `TSNRangeParams.java` to the formatter include list — no new dependency.

---

> [!NOTE]
> **Low Risk**
> Additive public API and read-only commands with validation and tests; no changes to auth, persistence, or existing command behavior.
> 
> **Overview**
> Adds client support for **TS.NRANGE** and **TS.NREVRANGE**, which query an explicit ordered key list over a time range and return **timestamp-major pivot rows** (one value column per key, `NaN` for missing cells).
> 
> New **`TSNRangeParams`** carries range bounds and optional flags (`LATEST`, filters, `COUNT`, per-key **aggregation** with client-side validation that spec count matches `numkeys`). Keys stay on the command methods; empty key lists are rejected before send.
> 
> Commands are wired through **`CommandObjects`**, **`UnifiedJedis`**, **`PipeliningBase`**, and the time-series command interfaces. A dedicated **`TIMESERIES_PIVOT_ELEMENT_LIST`** builder maps pivoted replies to `TSElement` / `MultiValueTSElement`. **`DefaultCacheable`** treats both commands as cacheable reads.
> 
> Coverage includes **`TSNRangeParamsTest`**, integration tests (gated on command availability), mocked unified/pipeline tests, and test harness cleanup via **`TestKeyRegistry`** where needed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dbdeb5477a30e2420be4dfe92a6828f4982a5a6d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/test/java/redis/clients/jedis/commands/unified/timeseries/TimeSeriesCommandsTestBase.java
```diff
@@ -1937,4 +1739,123 @@ public void readErrorsPropagate() {
...
+    List<TSElement> rows = jedis.tsNRange(seriesKeys, 0L, 60000L);
+
+    List<Long> timestamps = new ArrayList<>();
+    for (TSElement row : rows) {
+      timestamps.add(row.getTimestamp());
+    }
+    assertEquals(Arrays.asList(1000L, 2000L, 3000L), timestamps);
+
+    assertEquals(2, rows.get(0).getValues().size());
+    assertEquals(10.0, rows.get(0).getValues().get(0), 0.001);
```
ggivo, line 1982: What is the reason for using delta/closeto (0.001) in most asserts?
I

src/test/java/redis/clients/jedis/commands/commandobjects/CommandObjectsTimeSeriesCommandsTest.java
```diff
@@ -43,10 +53,33 @@
...
+  public void setUpKeys(TestInfo testInfo) {
+    keys = TestKeyRegistry.create(testInfo);
+  }
+
+  @AfterEach
+  public void cleanUpKeys() {
+    // This low-level test drives commands through the private executor and has no reusable
+    // KeyBinaryCommands client, so open a short-lived Jedis to delete the registered keys.
+    try (Jedis cleanupClient = new Jedis(endpoint.getHostAndPort(),
+        endpoint.getClientConfigBuilder().protocol(protocol).build())) {
```
ggivo, line 78: any reason not to move it  directly in CommandObjectsTestBase

src/main/java/redis/clients/jedis/timeseries/TimeSeriesProtocol.java
```diff
@@ -16,6 +16,8 @@ public enum TimeSeriesCommand implements ProtocolCommand {
     READ("TS.READ"),
     RANGE("TS.RANGE"),
     REVRANGE("TS.REVRANGE"),
+    NRANGE("TS.NRANGE"),
+    NREVRANGE("TS.NREVRANGE"),
```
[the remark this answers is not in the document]
ggivo: seems ligit

ggivo (commented):
LGTM in general 

Posted some minor questions; wondering why assertions are using delta?

src/test/java/redis/clients/jedis/commands/unified/timeseries/TimeSeriesCommandsTestBase.java
atakavci: this is the case, i guess, AI takes measures against rounding errors with double operations, which is right approach most of the time, though there is no double operations this time.

src/test/java/redis/clients/jedis/commands/commandobjects/CommandObjectsTimeSeriesCommandsTest.java
atakavci: i kept it in class where it s explicitly needed,,, thinking it would be another task to clear those classes and introduce the use of registry to others.

src/test/java/redis/clients/jedis/commands/unified/timeseries/TimeSeriesCommandsTestBase.java
ggivo: I would say it make sense for aggreagated  values, but for single values I think test should assert actual value directly

ggivo (approved):

PR #4661, opened 2026-07-30
decode ScanParams.match with SafeEncoder, not the JVM default charset

### LINKED ISSUE #4660: Cluster SCAN routes by a default-charset decode of the MATCH pattern, not the bytes it sends
### Expected behavior

In cluster mode a `SCAN` is routed by the hash tag of its `MATCH` pattern, so it should reach the node that owns the tag it actually sends, whatever the JVM default charset happens to be.

### Actual behavior

`ScanParams.match(String)` encodes the pattern with `SafeEncoder`, but the `ScanParams.match()` getter reads it back with `new String(byte[])`, which decodes using the JVM default charset. `ClusterCommandObjects.scan(String, ScanParams)` takes the routing slot from that getter while `addParams(params)` puts the original `SafeEncoder` bytes on the wire, so the slot and the pattern disagree whenever the default charset is not the one `SafeEncoder` encoded with.

A cluster `SCAN` whose hash tag holds a non-ASCII character then goes to a node that does not own the tag. `SCAN` is not a keyed command server side, so no `MOVED` comes back to correct it: the scan walks the wrong shard and returns nothing. The binary overload is unaffected, it reads `binaryMatch()`.

Two ways to end up there:

- The JVM default charset is not UTF-8. Before JEP 400 landed in JDK 18 it followed the platform locale, so on Windows it resolves to the ANSI code page (windows-1252, GBK under zh-CN, Shift_JIS under ja-JP), and on a Linux JVM started without a locale it resolves to US-ASCII. Jedis targets Java 8, so this covers most supported runtimes.
- `SafeEncoder.DEFAULT_CHARSET` is set to something other than the platform default, which is the documented way to choose a wire charset (#448, #2658). That reproduces on any JDK, including 21.

This is the case CONTRIBUTING.md warns about under "type <-> byte array conversion", where the rule is *string <-> byte array : use SafeEncoder.encode()* with the caution that going direct *will break GBK support*. The getter is the decode half of that same rule.

### Steps to reproduce:

No server needed. This prints the slot Jedis routes to next to the slot the pattern it actually sends hashes to.

```java
import java.nio.charset.Charset;
import redis.clients.jedis.ClusterCommandObjects;
import redis.clients.jedis.RedisProtocol;
import redis.clients.jedis.params.ScanParams;
import redis.clients.jedis.util.JedisClusterCRC16;

public class ScanSlotRepro {
    System.out.println("JVM default charset : " + Charset.defaultCharset());

    ScanParams params = new ScanParams().match("{café}*");   // hash tag {café}
    ClusterCommandObjects cco = new ClusterCommandObjects(RedisProtocol.RESP3);

    System.out.println("params.match()      : " + params.match());
    System.out.println("slot of MATCH sent  : " + JedisClusterCRC16.getSlot(params.binaryMatch()));
    System.out.println("routed, String scan : "
        + cco.scan("0", params).getArguments().getKeyHashSlots());
    System.out.println("routed, binary scan : "
        + cco.scan(SafeEncoder.encode("0"), params).getArguments().getKeyHashSlots());
  }
}
```

1. Run it on a UTF-8 JVM, routing agrees with the pattern:

```
$ java -Dfile.encoding=UTF-8 ScanSlotRepro
JVM default charset : UTF-8
params.match()      : {café}*
slot of MATCH sent  : 5735
routed, String scan : [5735]
routed, binary scan : [5735]
```

2. Run it with a non-UTF-8 default charset, routing no longer matches what is sent:

```
$ java -Dfile.encoding=US-ASCII ScanSlotRepro
JVM default charset : US-ASCII
params.match()      : {caf??}*
slot of MATCH sent  : 5735
routed, String scan : [13531]
routed, binary scan : [5735]
```

Against a real cluster the second run scans whichever node owns slot 13531 and never sees the keys under `{café}`.

#### Jedis version:

8.0.0-SNAPSHOT (master), and earlier versions carrying this getter

#### Redis version:

Any Redis Cluster; the routing decision is entirely client side

#### Java version:

Reproduced on 11, 17 and 21. Applies to 8 through 17 wherever the platform locale is not UTF-8, and to any JDK if `SafeEncoder.DEFAULT_CHARSET` is changed.

### Suggested fix

Read the pattern back through `SafeEncoder` in `ScanParams.match()`, so the getter decodes with the charset `match(String)` encoded with. It is a no-op when the default charset is UTF-8, so no configuration that works today changes behaviour.

The four `SCAN`/`HSCAN`/`SSCAN`/`ZSCAN` cursor decodes in `BuilderFactory` use the same `new String(byte[])` construct, sitting next to `SafeEncoder.encode` calls for the sibling values in the same builder. Cursors are ASCII digits so nothing misbehaves there today, but they belong on the same helper.

Happy to raise the PR with a regression test if this looks right.

insaf021:
`java -Dfile.encoding=US-ASCII`, cluster SCAN with MATCH `{café}*`:

    params.match()      : {caf??}*
    slot of MATCH sent  : 5735
    routed, String scan : [13531]
    routed, binary scan : [5735]

`match(String)` encodes the pattern through `SafeEncoder`, while the `match()` getter reads it back with `new String(byte[])` on the JVM default charset. `ClusterCommandObjects.scan(String, ScanParams)` takes the routing slot from that getter and `addParams` sends the original bytes, so a non-ASCII hash tag lands on a node that does not own it. `SCAN` is not keyed server side, so no `MOVED` comes back and the scan quietly walks the wrong shard.

Decoding through `SafeEncoder` is a no-op when the default charset is UTF-8, so no configuration that works today changes behaviour.

The four `BuilderFactory` cursor decodes are the same construct, and `ScanResult` re-encodes that cursor through `SafeEncoder` on the way back out. Cursors are ASCII digits so nothing misbehaves there today, but both halves of the round trip belong on the same helper.

Regression test asserts the String and binary overloads route to the slot of the bytes actually sent.

Closes #4660

---

> [!NOTE]
> **Medium Risk**
> Touches cluster routing and RESP string decoding; behavior is unchanged when default charset is UTF-8, but fixes a silent data-access bug for other encodings.
> 
> **Overview**
> **Fixes wrong cluster SCAN routing** when `MATCH` patterns contain non-ASCII hash tags and the JVM default charset (e.g. US-ASCII) differs from `SafeEncoder.DEFAULT_CHARSET`. `ScanParams.match()` now decodes stored pattern bytes via `SafeEncoder.encode` instead of `new String(byte[])`, so slot selection in `ClusterCommandObjects.scan` matches the bytes actually sent on the wire.
> 
> The same `SafeEncoder` decode replaces `new String` for scan cursors in `BuilderFactory` (`SCAN`, `HSCAN`, `SSCAN`, `ZSCAN` response builders), aligning decode with how cursors are re-encoded elsewhere.
> 
> Regression tests cover non-ASCII `match()` round-tripping and cluster SCAN routing parity between String and binary overloads.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ef012c1baae28b99a95a541b8a9a92d347a70bec. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


renechoi (commented):
Verified this locally against `master` (87e9df25) on JDK 21.0.9. Your repro reproduces exactly, same slot numbers.

**master, `-Dfile.encoding=US-ASCII`:**

```
params.match()      : {caf??}*
slot of MATCH sent  : 5735
routed, String scan : [13531]
routed, binary scan : [5735]
```

**with this PR:** all four route to `5735`.

### Why the fix is exact, not just narrower

`addHashSlotKey(String)` resolves through `JedisClusterCRC16.getSlot(String)`, which re-encodes via `getCRC16(String)` -> `SafeEncoder.encode(key)`. Once the getter decodes with `SafeEncoder` as well, the decode and the re-encode are inverse, so the routing slot is by construction the slot of the bytes `addParams` puts on the wire. That is what makes this a fix rather than a shift to a different wrong answer.

Swept `SafeEncoder.DEFAULT_CHARSET` with the JVM default left at UTF-8, comparing the String-overload slot against `getSlot(binaryMatch())`:

| `SafeEncoder.DEFAULT_CHARSET` | wire slot | master | this PR |
| --- | --- | --- | --- |
| UTF-8 | 5735 | 5735 | 5735 |
| ISO-8859-1 | 9182 | **2149** | 9182 |
| GBK | 12376 | **9140** | 12376 |
| windows-1252 | 9182 | **2149** | 9182 |
| Shift_JIS | 2149 | 2149 | 2149 |
| US-ASCII | 2149 | 2149 | 2149 |
| UTF-16 | 2836 | **6474** | **5029** |

GBK is worth calling out: it is the charset `.github/CONTRIBUTING.md` names under *type <-> byte array conversion* ("use String.toBytes() directly will break GBK support"), and it is broken on master and correct here.

UTF-16 stays mismatched after the fix, but the residual is not in your change. `getSlot(String)` extracts the hash tag on chars and then encodes that tag on its own, picking up a BOM, while `getSlot(byte[])` CRCs the byte slice in place; those agree only when `{` and `}` are standalone bytes. I would not chase it, because `Protocol.BYTES_ASTERISK` and `Command.getRaw()` also go through `SafeEncoder`, so a UTF-16 `DEFAULT_CHARSET` would put the RESP framing and every command name on the wire in UTF-16 and nothing would work at all. Still worth stating that the guarantee is "every ASCII-compatible charset", since the test asserts String/binary parity as though it were unconditional.

### What the test actually pins

- Reverting only `ScanParams.java`: exactly 1 of 34 in `ClusterCommandObjectsTest` fails, the new one (`expected: <[9182]> but was: <[2149]>`).
- Reverting only `BuilderFactory.java`: 34/34 still pass.

So the regression guard covers the `ScanParams` half only, which matches what you wrote about cursors being ASCII today. Two things I would still add:

**1. The 3-arg overload is on the same path and is unasserted.** `ClusterCommandObjects:47-48` (`scan(String, ScanParams, String type)`) also reads `params.match()`. Measured: `13531` on master, `5735` with your change. One more `assertEquals` covers it.

**2. The substitute charset has to stay ASCII-compatible, and the reason is not visible at the call site.** `Protocol.Command` and `Protocol.Keyword` freeze `raw = SafeEncoder.encode(name())` in the enum constructor, so whichever charset is live the first time those enums initialise is baked in for the life of the JVM. The `finally` cannot undo it. Forcing the enum to initialise inside the window:

```
UTF-16 substituted : SCAN raw = [-2, -1, 0, 83, 0, 67, 0, 65, 0, 78]   (identical after restore)
ISO-8859-1 (yours) : SCAN raw = [83, 67, 65, 78]
```

Your ternary only ever picks UTF-8 or ISO-8859-1, so it is safe as written, and mutating that global is the only way to write this test in-process, since `Charset.defaultCharset()` is fixed once the JVM starts. But this is the first place in the test tree that writes `SafeEncoder.DEFAULT_CHARSET`, so a one-line comment saying the substitute must be ASCII-compatible would stop someone from "strengthening" it to UTF-16 later and silently poisoning every test that runs after it in the same fork.

### The `BuilderFactory` hunks are more than tidying

Worth having this ready if you are asked to split them out. `ScanResult(String, ...)` immediately does `SafeEncoder.encode(cursor)`, so on master the cursor is decoded with one charset and re-encoded with another and `getCursorAsBytes()` is not byte-transparent. With `DEFAULT_CHARSET=ISO-8859-1` and a JVM default of UTF-8:

```
wire [49, 55, -23] -> master  getCursorAsBytes [49, 55, 63]     (0xE9 became '?')
wire [49, 55, -23] -> this PR getCursorAsBytes [49, 55, -23]
```

Real cursors are ASCII decimal so nothing observable changes, but after this the builders are byte-transparent for any cursor rather than only for the ones Redis happens to send.

### Other checks

- `javac --release 8` is clean on all three files. Worth noting the pom pins `<source>`/`<target>` 1.8 rather than `<release>`, so the build alone would not flag a newer API.
- Ran `ClusterCommandObjectsTest` (34/34) and `ScanParamsTest` (5/5) on this branch. I did not get a usable full-suite number: the `mocked.*` classes error under JDK 21 in my environment (154/154 in `PipeliningBaseSortedSetCommandsTest`, all Mockito), identically on `master` and on this branch, so there is no signal there either way. CI runs these on JDK 8.

insaf021:
Both added in f2a5069: the typed scan(String, ScanParams, String) overload gets its own assert (it reads params.match() the same way, so it showed the same 13531 on master here), and the DEFAULT_CHARSET substitution now carries a comment spelling out why the substitute has to stay ASCII-compatible, with the Protocol.Command/Keyword raw-bytes freeze as the reason. ClusterCommandObjectsTest still passes 34/34 and the formatter is clean on the test file.

Agreed on the UTF-16 residual: getSlot(String) re-encodes the extracted tag on its own and picks up a BOM, so the parity the test asserts really holds for ASCII-compatible charsets, which the new comment states. A UTF-16 DEFAULT_CHARSET would break RESP framing before routing ever mattered, so I'm not chasing that path either.

On the BuilderFactory hunks, I'd keep them here since ScanResult already does the SafeEncoder half of the cursor round trip, but happy to split them out if that's preferred.

insaf021:
Good sweep, and the GBK row ties it straight back to the CONTRIBUTING.md conversion rule. Pushed both test changes: the typed scan(cursor, params, type) overload now has its own assertion, and the charset substitution carries a comment that the stand-in must stay ASCII-compatible, since Protocol.Command/Keyword freeze their raw bytes through SafeEncoder at enum initialisation and the finally can't unwind that. Also scoped the guarantee in the test javadoc to ASCII-compatible charsets rather than leaving the parity looking unconditional.

Agreed the UTF-16 residual sits in JedisClusterCRC16, not this path: getSlot(String) extracts the tag on chars and re-encodes it standalone, picking up a BOM, while getSlot(byte[]) CRCs the slice in place. Since a UTF-16 DEFAULT_CHARSET would already put the RESP framing itself on the wire in UTF-16, I've left that alone.

Keeping the BuilderFactory hunks in for now, since they make the cursor round trip byte-transparent to pair with the SafeEncoder.encode that ScanResult already applies on the way back, but happy to split them into a separate commit if the maintainers prefer.

ggivo:
LGTM — the getter now decodes through `SafeEncoder`, so the String SCAN path hashes the same bytes it sends and matches the binary overload. Verified the routing parity locally.              
                                                                                                                                                                                                 
  I've added a small `ScanParamsTest` covering the getter/setter charset round-trip returns same value. The `BuilderFactory` cursor changes are hygiene (SCAN cursors are ASCII), fine to keep. Approving.

ggivo (approved):

insaf021:
Thanks for merging, and for adding the round-trip test on top.

PR #4675, opened 2026-08-06
register source keys via key() in zrangestore, geosearchStore, eval

### LINKED ISSUE #4674: ZRANGESTORE, GEOSEARCHSTORE and EVAL source keys bypass the key pre-processor
`CommandObjects` with a `PrefixedKeyArgumentPreProcessor("tenantA:")` installed, no server needed:

```
zunionstore : [ZUNIONSTORE, tenantA:dst, 1, tenantA:src]
zrangestore : [ZRANGESTORE, tenantA:dst, src, 0, -1]
eval (List) : [EVAL, return 1, 1, tenantA:k, v]
eval (varar): [EVAL, return 1, 1, k, v]

slots seen by the cluster provider:
  zunionstore dst src : [9394, 13222]
  zrangestore dst src : [9394]
```

`ZRANGESTORE` writes inside the namespace and reads from outside it. `ZUNIONSTORE` over the same two keys is correct, so this is a gap in a few builders rather than a documented limit of the pre-processor.

### Reproducer

```java
CommandObjects co = new CommandObjects(RedisProtocol.RESP3);
co.setKeyArgumentPreProcessor(new PrefixedKeyArgumentPreProcessor("tenantA:"));

// prints the raw argv of each CommandObject
System.out.println(wire(co.zunionstore("dst", "src")));
System.out.println(wire(co.zrangestore("dst", "src", new ZRangeParams(0, -1))));
System.out.println(wire(co.eval("return 1", Arrays.asList("k"), Arrays.asList("v"))));
System.out.println(wire(co.eval("return 1", 1, "k", "v")));

static List<String> wire(CommandObject<?> c) {
  List<String> out = new ArrayList<>();
  for (Rawable r : c.getArguments()) out.add(SafeEncoder.encode(r.getRaw()));
  return out;
}
```

### Root cause

`CommandArguments.key(Object)` is the only place `keyPreProc.actualKey()` runs, and the only place a key is recorded in `keys` for hash slot computation. Three families of builders in `CommandObjects` put a key on the wire without it:

- `zrangestore(dest, src, ...)` source key, 2 overloads, `.key(dest).add(src)`
- `geosearchStore` / `geosearchStoreStoreDist` source key, 12 overloads, same shape
- `eval` / `evalsha` `(script, keyCount, params...)`, 4 overloads. These use `addObjects(params)` plus `addHashSlotKeys(Arrays.copyOf(params, keyCount))`. The second call records the slot but does not transform the argument that is actually sent, so the untransformed key goes out as `KEYS[1]`. The `List<String> keys` overloads next to them use `.keys(keys)` and are correct.

### Why it matters

Namespacing is the visible half. A client configured for tenant isolation sends a correctly prefixed destination and an unprefixed source, so `ZRANGESTORE` copies another tenant's sorted set into its own keyspace, and `EVAL` hands the script an un-namespaced key.

Cluster routing is the quieter half. The source key never reaches `CommandArguments.getKeys()`, so `ClusterConnectionProvider.getConnection` sees a single slot for a two key command and the multi slot guard cannot fire. The command is dispatched to the node owning `dest` and rejected by the server, rather than raising `JedisClusterOperationException` client side the way the sibling commands do.

### Fix

Route these key arguments through `key()` like every other key in the class. Happy to raise the PR if that shape is acceptable.

insaf021:
With a `PrefixedKeyArgumentPreProcessor("tenantA:")` installed:

    zunionstore : [ZUNIONSTORE, tenantA:dst, 1, tenantA:src]
    zrangestore : [ZRANGESTORE, tenantA:dst, src, 0, -1]
    eval (List) : [EVAL, return 1, 1, tenantA:k, v]
    eval (varar): [EVAL, return 1, 1, k, v]

`CommandArguments.key(Object)` is the only place the pre-processor runs and the only place a key is recorded for slot computation, so a key routed through `add()` reaches the wire untransformed and never enters `getKeys()`. `ZRANGESTORE` then writes inside the namespace and reads from outside it, and the cluster provider sees one slot for a two key command so its multi slot guard cannot fire.

Eighteen sites, all the same shape: the source key of `zrangestore` and `geosearchStore`/`geosearchStoreStoreDist`, and the first `keyCount` varargs of `eval`/`evalsha`. The eval varargs overloads already called `addHashSlotKeys`, which records the slot but leaves the outgoing argument alone; the `List<String> keys` overloads beside them were always correct and are untouched.

Without a pre-processor, the `eval`/`evalsha` varargs overloads behave exactly as before — their slot keys were already registered via `addHashSlotKeys`, so this PR only changes which argument value goes out when a pre-processor rewrites keys.

 For `ZRANGESTORE` and `GEOSEARCHSTORE` there is one behaviour change even without a pre-processor: the source key now enters `getKeys()`, so on cluster a cross-slot call fails client-side with `JedisClusterOperationException` instead of being dispatched to the destination's node and rejected by the server with a CROSSSLOT error. Standalone behaviour is unchanged.

Tests: cluster tests assert the cross-slot rejection for every `zrangestore`/`geosearchStore` overload and pin `eval`/`evalsha` varargs slot registration.

Closes #4674

---

> [!NOTE]
> **Medium Risk**
> Fixes cluster slot tracking and key preprocessing for multi-key commands; behavior is unchanged without a key preprocessor, but cluster routing and prefixed keys now correctly include source keys.
> 
> **Overview**
> **Fixes source keys bypassing `key()` registration** in `zrangestore`, `geosearchStore`/`geosearchStoreStoreDist`, and `eval`/`evalsha` varargs overloads.
> 
> Those keys were previously passed with `add()`, so they skipped key preprocessors and were omitted from cluster slot computation. They now go through `key()`, matching commands like `zunionstore`. The `eval`/`evalsha` varargs paths also stop using `addHashSlotKeys` alone (which recorded slots but left wire args untransformed).
> 
> Cluster tests assert cross-slot rejection for the fixed commands. Without a key preprocessor, wire behavior is unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2a37c6020296afdcf3fcc4ec97de5518be9e9fef. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ggivo:
Thanks for the PR, @insaf021.

Team capacity is limited so PR reviews can take some time. There's a corresponding issue (#4674) that still needs triage and acceptance by the team before the PR proposal can be considered.

insaf021:
Understood, no rush. I've answered the usage question over on #4674. Also rebased onto master to pick up the redis:8.6.2 pin from #4676, so the unrelated SDR failure on the previous run should clear.

PR #4677, opened 2026-08-08
Fix FTSearchParams.params(Map) dropping values when parameters are already set

### LINKED ISSUE #4679: FTSearchParams.params(Map) drops supplied values when parameters are already set
## Describe the bug

`FTSearchParams.params(Map)` is meant to merge the supplied values into the already-set parameters, but its merge branch executes `this.params.putAll(params)` — `params` resolves to the field itself (the argument is `paramValues`), so the call is a self-merge no-op and the supplied map is silently dropped.

Affected usage patterns:
- `searchParams().addParam("min", 1).params(map)` — the map values never reach the wire;
- `searchParams().params(first).params(second)` — the second map is discarded.

The failure is silent on the client: the `PARAMS` clause is emitted without the dropped names, so `FT.SEARCH` either errors on the unresolved `$placeholder` (`No such parameter`) or returns wrong results. The common single-`params(map)` call takes the working null branch, which is why this went unnoticed.

## Reproduction

```java
Map<String, Object> map = new HashMap<>();
map.put("max", 2);
jedis.ftSearch(INDEX, "@numval:[$min $max]",
    FTSearchParams.searchParams().addParam("min", 1).params(map).dialect(2));
// -> "No such parameter `max`" — the map was dropped
```

## Affected versions

All releases since the method was introduced in the FT.SEARCH rework (#3138, Jedis 4.0.0).

## Fix

#4677 — one-line fix (`putAll(paramValues)`), plus unit tests pinning the merge and last-wins semantics, a mixed-path integration test, and javadoc for `addParam`/`params(Map)`.

HDPark95:
### What

`FTSearchParams.params(Map<String, Object> paramValues)` is meant to merge the
given values into any parameters already set. The merge branch copies the field
onto itself instead of the argument:

```java
public FTSearchParams params(Map<String, Object> paramValues) {
  if (this.params == null) {
    this.params = new HashMap<>(paramValues);
  } else {
    this.params.putAll(params);   // uses the field, not the argument
  }
  return this;
}
```

`params` here resolves to the field, so `this.params.putAll(params)` is
`this.params.putAll(this.params)` — a no-op, and the supplied `paramValues` are
dropped.

### Impact

Whenever parameters are already present — e.g. `addParam(...)` followed by
`params(map)`, or two `params(map)` calls — the second map is silently
discarded. Those values never reach the wire (the `PARAMS` clause omits them),
so an `FT.SEARCH` query that references the corresponding `$name` placeholders
does not get them substituted. When the parameter map starts empty the first
branch is taken and the value is preserved, which is why this went unnoticed.

### Fix

Use the `paramValues` argument in the merge branch.

### Test

Added `FTSearchParamsTest.paramsMapMergesInsteadOfDroppingValues`, a server-free
unit test that sets one parameter via `addParam`, then merges a second via
`params(Map)`, and asserts both are emitted in the built command arguments. It
fails on the current code and passes with the fix.

---

> [!NOTE]
> **Low Risk**
> Small, localized fix to RediSearch query parameter assembly with dedicated unit and integration tests; no auth or data-path changes beyond correct PARAMS emission.
> 
> **Overview**
> Fixes **`FTSearchParams.params(Map)`** so a second map actually merges when parameters already exist (e.g. after **`addParam`** or an earlier **`params`** call). The merge branch incorrectly called **`putAll`** on the internal map with itself, so extra **`paramValues`** were dropped and never appeared in the **`PARAMS`** clause for **`FT.SEARCH`**.
> 
> Documents merge/overwrite behavior on **`addParam`** and **`params`**. Adds unit tests in **`FTSearchParamsTest`** and an integration test **`testQueryParamsMixedAddParamAndMap`** for mixed **`addParam`** + **`params`** usage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7500187ca6049dad25506fba4bff161af71f792e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/main/java/redis/clients/jedis/search/FTSearchParams.java
+++ b/src/main/java/redis/clients/jedis/search/FTSearchParams.java
@@ -412,7 +412,7 @@ public class FTSearchParams implements IParams {
     if (this.params == null) {
       this.params = new HashMap<>(paramValues);
     } else {
-      this.params.putAll(params);
+      this.params.putAll(paramValues);
     }
     return this;
   }
--- a/src/main/java/redis/clients/jedis/search/FTSearchParams.java
+++ b/src/main/java/redis/clients/jedis/search/FTSearchParams.java
@@ -392,12 +392,15 @@ public class FTSearchParams implements IParams {
   }
 
   /**
-   * Parameters can be referenced in the query string by a $ , followed by the parameter name,
-   * e.g., $user , and each such reference in the search query to a parameter name is substituted
-   * by the corresponding parameter value.
+   * Adds a query parameter. Parameters are referenced in the query string by {@code $} followed
+   * by the parameter name, e.g. {@code $user}, and each such reference is substituted by the
+   * corresponding parameter value.
+   * <p>
+   * Adding a parameter with an existing name overwrites its previous value. Parameters
+   * accumulate across {@code addParam} and {@link #params(Map)} calls.
    *
-   * @param name
-   * @param value can be String, long or float
+   * @param name the parameter name, referenced in the query as {@code $name}
+   * @param value the parameter value; can be String, long or float
    * @return the query object itself
    */
   public FTSearchParams addParam(String name, Object value) {
@@ -408,6 +411,15 @@ public class FTSearchParams implements IParams {
     return this;
   }
 
+  /**
+   * Adds all entries of the given map as query parameters, merging them with any parameters
+   * already set via {@link #addParam(String, Object)} or previous {@code params} calls. An entry
+   * whose name is already set overwrites its previous value.
+   *
+   * @param paramValues parameter name to value; values can be String, long or float
+   * @return the query object itself
+   * @see #addParam(String, Object)
+   */
   public FTSearchParams params(Map<String, Object> paramValues) {
     if (this.params == null) {
       this.params = new HashMap<>(paramValues);
--- a/src/test/java/redis/clients/jedis/commands/unified/search/SearchWithParamsCommandsTestBase.java
+++ b/src/test/java/redis/clients/jedis/commands/unified/search/SearchWithParamsCommandsTestBase.java
@@ -741,6 +741,23 @@ public abstract class SearchWithParamsCommandsTestBase extends UnifiedJedisComma
   }
 
   @Test
+  public void testQueryParamsMixedAddParamAndMap() {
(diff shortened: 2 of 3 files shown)


kdelay (commented):
The one-line fix looks correct: in the merge branch `params` resolves to the field, so `this.params.putAll(params)` was a self-copy and the argument was discarded. I grepped the rest of `src/main/java` for the same shape and this is the only occurrence — the other `putAll` hits are the `JedisByteMap`/`JedisByteHashMap` overrides and the executor map aggregators, so it is a one-off rather than a pattern.

Two notes on the test. I ran `mvn -B -Dwith-param-names=true -Dtest=FTSearchParamsTest -DfailIfNoTests=false test` on a clean checkout of master with this PR applied (Corretto 17): 1 test, 0 failures. Reverting only the production line reproduces the failure, with the emitted args being `[FT.SEARCH, PARAMS, 2, existing, 1]`.

1. Notice that the expected `2` in that assertion is satisfied by the `PARAMS` arity token, not by the value of `added` — `containsArguments` is presence-only and position-agnostic, so the only thing actually catching the regression is the `added` key. That also means the arity, which is the part the server parses, is never asserted. To check how much that costs, I mutated line 146 to `args.add(PARAMS).add(params.size())` (arity 2 instead of 4 for two parameters, which the server would reject) and re-ran the same command: still 1 test, 0 failures. An `assertThat(args, hasArgument(2, RawableFactory.from(4)))` alongside the existing assertion would pin the pair count and make the test fail on that class of change too.

2. The description names two ways into the merge branch — `addParam` then `params(map)`, and two `params(map)` calls — but only the first is covered. The second one is a couple of extra lines in the same test and is the path a caller building a query from several maps would hit.

Unrelated to this PR, but adjacent enough to be worth a separate issue: `AggregationBuilder.params(Map)` appends the `PARAMS` clause into `aggrArgs` eagerly instead of accumulating into a map, so a second call emits a second clause rather than merging. Built with the same style of unit check, `new AggregationBuilder("*").params({one=1}).params({two=2})` produces `FT.AGGREGATE * PARAMS 2 one 1 PARAMS 2 two 2`. Different failure mode from this one (duplicate clause instead of dropped values), same API shape.

One practical note if you touch the test: `formatter-maven-plugin:validate` runs before `test`, so an unformatted new file fails the build before any test executes.

Issue #4685, opened 2026-08-15
Cluster slot for String keys ignores byte-level hash tag under a DBCS default charset


insaf021:
### Problem

`JedisClusterCRC16.getSlot(String)` locates the `{hash tag}` by scanning for the `{` and `}` **characters** (through `JedisClusterHashTag.getHashTag`) and only encodes to bytes afterwards. The `byte[]` overload and the server's `keyHashSlot()` scan for `0x7B`/`0x7D` in the **encoded** bytes.

Under a DBCS `SafeEncoder.DEFAULT_CHARSET` (GBK, Shift_JIS, Big5) a trail byte can be `0x7B`/`0x7D` while the corresponding char is not a brace. The String path then picks a different tag (or none) than the wire bytes and computes a slot the server does not own. The same logical key also routes to two different slots depending on whether the String or the binary API is used.

`getSlot(String)` backs `CommandArguments.getKeyHashSlots()`, `ClusterCommandObjects` key routing and sharded pub/sub, so this affects routing for every keyed command, not just SCAN.

### Minimal reproducible example

```java
SafeEncoder.DEFAULT_CHARSET = Charset.forName("GBK");
// GBK bytes: 81 7B 41 81 7D  -> a lead byte + '{' , 'A', a lead byte + '}'
String key = new String(new byte[]{(byte)0x81,0x7B,0x41,(byte)0x81,0x7D}, SafeEncoder.DEFAULT_CHARSET);

JedisClusterCRC16.getSlot(key);                       // 14821  (whole key hashed, no '{' char)
JedisClusterCRC16.getSlot(SafeEncoder.encode(key));   // 16212  (== server keyHashSlot)
```

- Single-key ops survive only via a `MOVED` round-trip.
- Multi-key commands are mis-grouped: keys the server co-locates are rejected client-side (`Cannot get connection for command with multiple hash slots`), and keys the client groups get a server-side `CROSSSLOT`.

Under UTF-8 there is no discrepancy (no multibyte sequence contains `0x7B`/`0x7D`), so a fix is a no-op for the default configuration.

### Practical benefit

Non-UTF-8 default charsets are a supported knob (`SafeEncoder.DEFAULT_CHARSET`, used for GBK/Shift_JIS/Big5 deployments). Making `getSlot(String)` scan the encoded bytes brings it in line with the binary overload and the server, so hash tags co-locate correctly on those deployments.

### Proposed fix

[...]

PR #4686, opened 2026-08-15
Fix cluster slot mismatch between String and byte[] keys under non-UTF-8 charsets (#4685)

insaf021:
getSlot("<GBK 81 7B 41 81 7D>")                     -> slot 14821
    getSlot(SafeEncoder.encode("<GBK 81 7B 41 81 7D>")) -> slot 16212   (== server keyHashSlot)

`getSlot(String)` locates the `{hash tag}` by scanning for the `{` and `}` characters and only encodes to bytes afterwards. The `byte[]` overload and the server's `keyHashSlot()` scan for `0x7B`/`0x7D` in the encoded bytes. Under a DBCS `SafeEncoder.DEFAULT_CHARSET` (GBK, Shift_JIS, Big5) a trail byte can be `0x7B`/`0x7D` while the char is not a brace, so the String path picks a different tag (or none) and routes to a slot the server does not own. It also disagrees with the binary API for the same logical key, which reaches routing through `CommandArguments.getKeyHashSlots()`.

Delegating to the byte overload moves brace scanning onto the wire bytes. No-op under UTF-8 since no multibyte sequence contains `0x7B`/`0x7D`; the added regression covers both the GBK mismatch and the UTF-8 parity.

Companion issue #4685.

---

> [!NOTE]
> **Medium Risk**
> Changes core cluster slot calculation for String keys; impact is localized and aligns with server routing, but any app relying on the old String-level hashtag parsing under DBCS could see different slots.
> 
> **Overview**
> Fixes **cluster slot routing** when `getSlot(String)` disagreed with Redis and with `getSlot(byte[])` under DBCS charsets (GBK, Shift_JIS, Big5).
> 
> **`getSlot(String)`** no longer extracts the hash tag via `JedisClusterHashTag` on Java characters before hashing. It **encodes the key first** and delegates to **`getSlot(byte[])`**, so `{`/`}` are found on **wire bytes** the same way as the server’s `keyHashSlot()`. That avoids false brace matches when a multibyte trail byte is `0x7B`/`0x7D` but the decoded character is not a brace—behavior that could send commands to the wrong shard. UTF-8 behavior is unchanged.
> 
> Adds a **GBK regression test** asserting `getSlot(String)` matches `getSlot(SafeEncoder.encode(key))` for a key whose encoded form contains byte-level `0x7B`/`0x7D` delimiters.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 758b5ae7d19b719b5fb4049d9d87c42432317833. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/main/java/redis/clients/jedis/util/JedisClusterCRC16.java
+++ b/src/main/java/redis/clients/jedis/util/JedisClusterCRC16.java
@@ -37,9 +37,11 @@ public final class JedisClusterCRC16 {
       throw new NullPointerException("Slot calculation of null is impossible");
     }
 
-    key = JedisClusterHashTag.getHashTag(key);
-    // optimization with modulo operator with power of 2 equivalent to getCRC16(key) % 16384
-    return getCRC16(key) & (16384 - 1);
+    // Scan for the hash tag on the encoded bytes so the '{' and '}' delimiters are located the same
+    // way the server does in keyHashSlot(). Under a DBCS default charset (GBK, Shift_JIS, Big5) a
+    // trail byte can be 0x7B/0x7D while the corresponding char is not a brace, so scanning the
+    // String would pick a different tag (or none) than the wire bytes and route to the wrong slot.
+    return getSlot(SafeEncoder.encode(key));
   }
 
   public static int getSlot(byte[] key) {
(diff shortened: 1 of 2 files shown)


ggivo:
Reopening per the discussion in #4685.

One change before merge: the test leaks the global `SafeEncoder.DEFAULT_CHARSET` — the trailing UTF-8 assertion sets it outside any try/finally and never restores the original.

PR #4692, opened 2026-08-20
refresh primaryNodesCache in JedisClusterInfoCache.discoverClusterSlots

### LINKED ISSUE #4691: Cluster primary node map is not refreshed on slot renewal, so primary-only commands go to demoted nodes after failover
### Expected behavior

After a cluster failover (or any resharding / node replacement) that the client picks up through its normal slot renewal, `RedisClusterClient` / `JedisCluster` should route primary-only work to the current primaries: `ALL_SHARDS` broadcasts (`FLUSHALL`, `FUNCTION LOAD/DELETE/FLUSH`, `SCRIPT LOAD`, `WAIT`, ...), keyless write commands, and the "any primary" path used by `getConnection()`.

### Actual behavior

The primary set is frozen at whatever `CLUSTER SLOTS` returned when the client was built. Against the repo's `cluster-stable` docker env (Redis 8.6.1, three primaries 7479/7480/7481 each with one replica), promoting the replica of :7479 and then issuing a keyed write into the moved slot range (so the client gets `MOVED` and renews its slot cache) gives:

```
primaries at start : [localhost:7479, localhost:7480, localhost:7481]
  :7479 = master, :7483 = replica of :7479
CLUSTER FAILOVER on :7483 -> OK
  :7479 = replica of :7483, :7483 = master
primaries after     : [localhost:7479, localhost:7480, localhost:7481]
FLUSHALL -> redis.clients.jedis.exceptions.JedisBroadcastException: A failure occurred while broadcasting the command. (2 succeeded, 1 failed; first error: READONLY You can't write against a read only replica.)
   localhost:7480 -> OK
   localhost:7479 -> redis.clients.jedis.exceptions.JedisDataException: READONLY You can't write against a read only replica.
   localhost:7481 -> OK
```

Keyed commands are fine (the slot table was renewed and `set` landed on :7483). The primary map was not: `getPrimaryNodes()` still lists the demoted :7479 and never learns about :7483, so `flushAll()` hits the replica and never reaches the new primary. The same stale map feeds `ClusterConnectionProvider.getConnection()` and `RoundRobinConnectionResolver` for keyless writes. In the node-replacement case (rolling restart with new addresses, the usual Kubernetes shape) the sweep in `discoverClusterSlots` destroys the old pools but the primary map keeps referencing them, so `getConnection()` iterates only closed pools and ends in `No reachable node in cluster` on a healthy cluster.

### Root cause

`JedisClusterInfoCache.primaryNodesCache` (introduced in #4219, so 6.2.0 onwards) is written in exactly one place, `discoverClusterNodesAndSlots()`, which `ClusterConnectionProvider` runs once at construction. Every later refresh goes through `renewClusterSlots()` -> `discoverClusterSlots()` (the `MOVED` handling in `ClusterCommandExecutor`, the topology-refresh timer, `refreshClusterTopology()`), which rebuilds `nodes` / `slots` / `slotNodes` / `replicaSlots` and removes dead nodes, but does not touch `primaryNodesCache`. It is only ever cleared by `reset()`.

The existing `JedisClusterInfoCacheTest.getPrimaryNodesAfterMasterReplicaFailover` exercises the failover through `discoverClusterNodesAndSlots` twice, which is not the path a running client takes; the same scenario through `renewClusterSlots` fails on master.

### Steps to reproduce:

1. `docker compose --env-file src/test/resources/env/.env -f src/test/resources/env/docker-compose.yml up -d --wait cluster-stable`
2. Run the program below (it is in package `redis.clients.jedis` only to read `provider`; the observable part is `flushAll()`). It promotes :7483, triggers the renewal with a keyed write, then fails back so the env is left as it was.
3. `flushAll()` fails with `READONLY` from :7479; `getPrimaryNodes()` never changes.

```java
package redis.clients.jedis;

import java.util.Collections;
import java.util.TreeSet;
import redis.clients.jedis.exceptions.JedisBroadcastException;
import redis.clients.jedis.providers.ClusterConnectionProvider;

public class StalePrimariesRepro {

  static final JedisClientConfig CTRL = DefaultJedisClientConfig.builder().password("cluster")
      .autoNegotiateProtocol(false).build();

  static String primaries(RedisClusterClient client) {
    return new TreeSet<>(((ClusterConnectionProvider) client.provider).getPrimaryNodes().keySet()).toString();
  }

  static String role(int port) {
    try (Jedis j = new Jedis(new HostAndPort("localhost", port), CTRL)) {
      String role = null, masterPort = null;
      for (String line : j.info("replication").split("\r\n")) {
        if (line.startsWith("role:")) role = line.substring(5);
        if (line.startsWith("master_port:")) masterPort = line.substring(12);
      }
      return "slave".equals(role) ? "replica of :" + masterPort : role;
    }
  }

  static boolean synced(int port) {
    try (Jedis j = new Jedis(new HostAndPort("localhost", port), CTRL)) {
      String info = j.info("replication");
      return info.contains("master_link_status:up") && info.contains("master_sync_in_progress:0");
    }
  }

  static void failover(int replicaPort) throws Exception {
    for (int i = 0; i < 300 && !synced(replicaPort); i++) Thread.sleep(100);
    try (Jedis replica = new Jedis(new HostAndPort("localhost", replicaPort), CTRL)) {
      System.out.println("CLUSTER FAILOVER on :" + replicaPort + " -> " + replica.clusterFailover());
    }
    for (int i = 0; i < 300 && !"master".equals(role(replicaPort)); i++) Thread.sleep(100);
    Thread.sleep(1000); // let the demoted node settle as a replica
  }

  public static void main(String[] args) throws Exception {
    JedisClientConfig cfg = DefaultJedisClientConfig.builder().password("cluster").build();
    int masterPort = 7479, replicaPort = 7483; // :7483 replicates :7479 in the cluster-stable env
    try (RedisClusterClient client = RedisClusterClient.builder()
        .nodes(Collections.singleton(new HostAndPort("localhost", masterPort))).clientConfig(cfg).build()) {

      System.out.println("primaries at start : " + primaries(client));
      System.out.println("  :" + masterPort + " = " + role(masterPort) + ", :" + replicaPort + " = " + role(replicaPort));

      failover(replicaPort);
      System.out.println("  :" + masterPort + " = " + role(masterPort) + ", :" + replicaPort + " = " + role(replicaPort));

      // keyed write into the moved slot range: MOVED -> renewSlotCache(); then make sure it finished
      client.set("b", "1"); // slot 3300, owned by the failed-over shard
      client.refreshClusterTopology();
      System.out.println("primaries after     : " + primaries(client));

      try {
        System.out.println("FLUSHALL -> " + client.flushAll());
      } catch (JedisBroadcastException e) {
        System.out.println("FLUSHALL -> " + e);
        e.getReplies().forEach((k, v) -> System.out.println("   " + k + " -> " + v));
      }

      failover(masterPort); // fail back so the env is as we found it
    }
  }
}
```

A unit-level version of the same thing, next to the existing test:

```java
when(mockConnection.executeCommand(argThat(commandWithArgs(CLUSTER, "SLOTS"))))
    .thenReturn(masterReplicaSlotsResponse(MASTER_HOST, REPLICA_1_HOST))
    .thenReturn(masterReplicaSlotsResponse(REPLICA_1_HOST, MASTER_HOST));
cache.discoverClusterNodesAndSlots(mockConnection);
cache.renewClusterSlots(mockConnection);
assertThat(cache.getPrimaryNodes(), hasEntry(equalTo(getNodeKey(REPLICA_1_HOST)), ...));
// master: map was [127.0.0.1:7000 -> ...]
```

### Fix

Rebuild `primaryNodesCache` inside the write-locked block of `discoverClusterSlots`, the same two lines `discoverClusterNodesAndSlots` already has (clear alongside `resetSlots()`, `put` for the `MASTER_NODE_INDEX` entry). With that, the run above ends in `primaries after : [localhost:7480, localhost:7481, localhost:7483]` and `FLUSHALL -> OK`. Happy to raise the PR.

#### Jedis version:

master (3b81dfe), affects 6.2.0 and later

#### Redis version:

8.6.1 (`redislabs/client-libs-test:8.6.1`, `cluster-stable` service)

#### Java version:

17.0.19

insaf021:
`JedisClusterInfoCacheTest`, failover replayed through `renewClusterSlots` instead of `discoverClusterNodesAndSlots`:

    Expected: map containing ["127.0.0.1:7001"->...]
         but: map was [<127.0.0.1:7000=...>]

And against the `cluster-stable` docker env after `CLUSTER FAILOVER` on :7483:

    primaries after     : [localhost:7479, localhost:7480, localhost:7481]
    FLUSHALL -> JedisBroadcastException ... (2 succeeded, 1 failed; first error: READONLY You can't write against a read only replica.)
       localhost:7479 -> JedisDataException: READONLY You can't write against a read only replica.

`primaryNodesCache` is only written by `discoverClusterNodesAndSlots`, which runs once at construction. Every later refresh (MOVED handling, the topology-refresh timer, `refreshClusterTopology()`) goes through `discoverClusterSlots`, which rebuilds slots and prunes dead nodes but leaves the primary map as it was, so `ALL_SHARDS` broadcasts, keyless writes and `getConnection()` keep targeting the pre-failover primaries. Rebuild the map in the same write-locked block, as the initial discovery already does. With the change the run above ends in `[localhost:7480, localhost:7481, localhost:7483]` and `FLUSHALL -> OK`.

Fixes #4691.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 36ed7667efd82f9149ff87a5ea35b97d2f368192. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
+++ b/src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
@@ -249,6 +249,7 @@ public class JedisClusterInfoCache {
     w.lock();
     try {
       resetSlots();
+      primaryNodesCache.clear();
       if (clientSideCache != null) {
         clientSideCache.flush();
       }
@@ -274,6 +275,7 @@ public class JedisClusterInfoCache {
           hostAndPortKeys.add(getNodeKey(targetNode));
           setupNodeIfNotExist(targetNode);
           if (i == MASTER_NODE_INDEX) {
+            primaryNodesCache.put(getNodeKey(targetNode), getNode(targetNode));
             assignSlotsToNode(slotNums, targetNode);
           } else if (clientConfig.isReadOnlyForRedisClusterReplicas()) {
             assignSlotsToReplicaNode(slotNums, targetNode);
--- a/src/test/java/redis/clients/jedis/JedisClusterInfoCacheTest.java
+++ b/src/test/java/redis/clients/jedis/JedisClusterInfoCacheTest.java
@@ -156,6 +156,33 @@ public class JedisClusterInfoCacheTest {
             hasEntry(equalTo(getNodeKey(REPLICA_1_HOST)), equalTo(cache.getNode(REPLICA_1_HOST))));
   }
 
+  @Test
+  public void getPrimaryNodesAfterMasterReplicaFailoverOnRenew() {
+    JedisClientConfig clientConfig = DefaultJedisClientConfig.builder()
+            .readOnlyForRedisClusterReplicas().build();
+
+    Set<HostAndPort> startNodes = new HashSet<>();
+    startNodes.add(MASTER_HOST);
+
+    JedisClusterInfoCache cache = new JedisClusterInfoCache(clientConfig, startNodes);
+
+    when(mockConnection.executeCommand(argThat(commandWithArgs(CLUSTER, "SLOTS"))))
+            .thenReturn(masterReplicaSlotsResponse(MASTER_HOST, REPLICA_1_HOST))
+            .thenReturn(masterReplicaSlotsResponse(REPLICA_1_HOST, MASTER_HOST));
+
+    cache.discoverClusterNodesAndSlots(mockConnection);
+    assertThat(cache.getPrimaryNodes(),
+            hasEntry(equalTo(getNodeKey(MASTER_HOST)), equalTo(cache.getNode(MASTER_HOST))));
+
+    // Failover picked up through the slot renewal path (MOVED / topology refresh)
+    cache.renewClusterSlots(mockConnection);
+    assertThat(cache.getPrimaryNodes(), aMapWithSize(1));
+    assertThat(cache.getPrimaryNodes(),
+            hasEntry(equalTo(getNodeKey(REPLICA_1_HOST)), equalTo(cache.getNode(REPLICA_1_HOST))));
+    assertThat(cache.getShuffledPrimaryNodesPool(), equalTo(
+            Collections.singletonList(cache.getNode(REPLICA_1_HOST))));
+  }
+
   private List<Object> masterReplicaSlotsResponse(HostAndPort masterHost, HostAndPort replicaHost) {
     return createClusterSlotsResponse(
             new SlotRange.Builder(0, 16383).master(masterHost, masterHost.toString() + "-id")


ggivo (approved):
LGTM

PR #4697, opened 2026-08-21
Fix cluster slot calculation for String keys with DBCS charsets

### LINKED ISSUE #4685: Cluster slot for String keys ignores byte-level hash tag under a DBCS default charset

ansn366:
## Summary
- Make `JedisClusterCRC16.getSlot(String)` delegate to the byte-array slot calculation after `SafeEncoder` encoding
- Add a GBK regression test to ensure String and binary APIs compute the same cluster slot when encoded bytes contain hash tag delimiters

Fixes #4685.

## Verification
- Compiled the affected Jedis utility classes with JDK 8
- Ran an offline GBK slot verification; result: `OK 16212`
- Maven test command was attempted, but dependency download was blocked by the sandbox network policy

---

> [!NOTE]
> **Medium Risk**
> Touches cluster slot routing, so a mistake can send keys to the wrong node. The change is small and covered by a charset-specific test.
> 
> **Overview**
> Fixes Redis Cluster slot calculation for `String` keys when a non-UTF-8 charset (e.g. GBK) is in use.
> 
> `JedisClusterCRC16.getSlot(String)` no longer extracts hash tags on Java characters. It encodes with `SafeEncoder` and reuses the byte-array path, so `{`/`}` are detected on the same bytes Redis would see. A GBK regression test checks String and binary APIs agree (slot `16212`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3b86b5c9ca70e5c30afbffe8ac0f17b9e958bfc6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/main/java/redis/clients/jedis/util/JedisClusterCRC16.java
+++ b/src/main/java/redis/clients/jedis/util/JedisClusterCRC16.java
@@ -37,9 +37,7 @@ public final class JedisClusterCRC16 {
       throw new NullPointerException("Slot calculation of null is impossible");
     }
 
-    key = JedisClusterHashTag.getHashTag(key);
-    // optimization with modulo operator with power of 2 equivalent to getCRC16(key) % 16384
-    return getCRC16(key) & (16384 - 1);
+    return getSlot(SafeEncoder.encode(key));
   }
 
   public static int getSlot(byte[] key) {
--- a/src/test/java/redis/clients/jedis/util/JedisClusterCRC16Test.java
+++ b/src/test/java/redis/clients/jedis/util/JedisClusterCRC16Test.java
@@ -3,6 +3,7 @@ package redis.clients.jedis.util;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotEquals;
 
+import java.nio.charset.Charset;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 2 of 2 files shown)


ggivo:
Duplicate of #4686, which makes the same change (getSlot(String) delegating to
getSlot(SafeEncoder.encode(key))) and predates this PR. Closing in favor of #4686 to keep
the discussion and history in one place. The direction is right — follow-up continues on
#4686 and issue #4685.

PR #4701, opened 2026-08-24
Do not deliver pub/sub push messages as replies outside an active subscription

uglide:
### Problem

A RESP3 pub/sub push arriving while no pub/sub read loop is running — e.g. a `message` frame the server delivers **after** the unsubscribe confirmation, as observed behind the Redis Enterprise DMC proxy — was propagated by the default push consumer, and `Protocol.process` returned it as the next command's reply. The reply stream desyncs and the command typically fails with:

```
java.lang.ClassCastException: java.util.ArrayList cannot be cast to [B
    at redis.clients.jedis.BuilderFactory$18.build(BuilderFactory.java:283)
```

This is the root cause of the flaky `advancedPubsubWithClientCache` failures in the Redis Enterprise nightly CI (any client that reuses a pooled connection after subscribe/unsubscribe is exposed).

### Fix

- `Connection` tracks a volatile `activeSubscription` flag, set/cleared around the subscribe/process loops of `JedisPubSubBase` (`proceed`, `proceedWithPatterns`) and `JedisShardedPubSubBase` (`proceed`) — cleared in `finally`, so pooled connections are always returned with the flag off.
- The default pub/sub push consumer is now created per connection via `PushConsumerChainImpl.pubSubConsumer(BooleanSupplier)` and only propagates pub/sub-typed pushes while the flag is set. Outside a subscription, the push is debug-logged and left to the rest of the consumer chain (consumed by default), so `Protocol.process` continues to the real reply.

### Behavior notes

- During a subscription, behavior is byte-for-byte identical to before; the pub/sub loops keep receiving propagated frames via `getUnflushedObject()`.
- RESP2 is unaffected: its pub/sub frames arrive as regular multi-bulk replies and never reach the push consumer chain.
- Custom push consumers keep their propagate/drop contract, and now also observe out-of-subscription pub/sub pushes that the default consumer previously short-circuited.
- The public `PUBSUB_CONSUMER` constant is unchanged.

### Testing

- New `ConnectionMockTest` cases deterministically reproduce the bug with `TcpMockServer`: a `message` push injected before a `PING` corrupts the reply without the fix (verified failing) and is dropped with it; a second test verifies propagation still works while the flag is set.
- New `PushConsumerChainImplTest` cases cover the gated consumer for all pub/sub push types and verify chain fall-through semantics when inactive.
- Full unit suite passes locally.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Changes core RESP3 read/push handling on every connection; low risk for normal pub/sub API usage but can affect hand-rolled subscribe + `getUnflushedObject()` code paths.
> 
> **Overview**
> Fixes RESP3 reply-stream corruption when a **pub/sub push** (e.g. a `message` after `UNSUBSCRIBE`, as seen behind Redis Enterprise) arrives on a pooled connection **without** an active subscribe loop. Those frames were propagated by the default push consumer and returned as the next command reply, often causing `ClassCastException`.
> 
> **`Connection`** now tracks **`activeSubscription`**, set in **`finally`** around **`JedisPubSubBase`** / **`JedisShardedPubSubBase`** `proceed` loops. The default consumer is **`pubSubConsumer(this::isActiveSubscription)`** instead of always-on **`PUBSUB_CONSUMER`**: pub/sub pushes propagate only during a real read loop; otherwise they are **debug-logged** and **not** surfaced as command replies. Later **`PushConsumer`**s can still see those pushes when the default consumer does not propagate. **`PUBSUB_CONSUMER`** remains unchanged for unconditional propagation.
> 
> Release notes document the behavior change. Tests add mock-server regression coverage and gated-consumer unit tests; auth integration tests use a **delegating mock** so connection state is not forked by **`spy()`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 344f2e2d188271621c681bf08acac2c6fcf35b3c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -195,12 +195,10 @@ public class Connection implements Closeable {
    * Marks whether a pub/sub read loop is currently driving this connection. Set by the
    * pub/sub implementations around their subscribe/process loop.
    */
-  @Internal
   void setActiveSubscription(boolean active) {
     this.activeSubscription = active;
   }
 
-  @Internal
   boolean isActiveSubscription() {
     return activeSubscription;
   }
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -15,7 +15,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 
 import redis.clients.jedis.Protocol.Command;
@@ -130,11 +129,10 @@ public class Connection implements Closeable {
   private int infiniteSoTimeout = 0;
   private boolean broken = false;
   private volatile Throwable brokenCause = null;
-  // True only while a pub/sub read loop is driving this connection; gates whether pub/sub
-  // pushes are propagated as read results or dropped as stray frames. Held in a shared
-  // mutable holder (not a plain field) so the push consumer chain observes the flag even
-  // through shallow copies of this connection (e.g. Mockito spies wired via AuthXManager).
-  private final AtomicBoolean activeSubscription = new AtomicBoolean(false);
+  // True only while a pub/sub read loop is driving this connection; set by the pub/sub
+  // loops and read by the push consumer chain to decide whether pub/sub pushes are
+  // propagated as read results or dropped as stray frames.
+  private volatile boolean activeSubscription = false;
   private boolean strValActive;
   private String strVal;
   protected String server;
@@ -190,9 +188,7 @@ public class Connection implements Closeable {
        * active on this connection; stray pub/sub frames (e.g. a message delivered after the
        * unsubscribe confirmation) must not surface as a regular command's reply.
        */
-      // Capture the holder, not `this`: shallow copies of this connection share the holder,
-      // so the chain sees flag updates made through any copy.
-      addPushConsumer(PushConsumerChainImpl.pubSubConsumer(activeSubscription::get));
+      addPushConsumer(PushConsumerChainImpl.pubSubConsumer(this::isActiveSubscription));
   }
 
   /**
@@ -201,12 +197,12 @@ public class Connection implements Closeable {
    */
   @Internal
   void setActiveSubscription(boolean active) {
-    this.activeSubscription.set(active);
+    this.activeSubscription = active;
   }
 
   @Internal
   boolean isActiveSubscription() {
-    return activeSubscription.get();
+    return activeSubscription;
   }
 
   @Override
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -129,6 +129,9 @@ public class Connection implements Closeable {
   private int infiniteSoTimeout = 0;
   private boolean broken = false;
   private volatile Throwable brokenCause = null;
+  // True only while a pub/sub read loop is driving this connection; gates whether pub/sub
+  // pushes are propagated as read results or dropped as stray frames.
+  private volatile boolean activeSubscription = false;
   private boolean strValActive;
   private String strVal;
   protected String server;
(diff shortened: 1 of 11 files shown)


src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -180,9 +186,27 @@ protected void initPushConsumers(JedisClientConfig config) {
...
+      // Capture the holder, not `this`: shallow copies of this connection share the holder,
+      // so the chain sees flag updates made through any copy.
+      addPushConsumer(PushConsumerChainImpl.pubSubConsumer(activeSubscription::get));
+  }
+
+  /**
+   * Marks whether a pub/sub read loop is currently driving this connection. Set by the
+   * pub/sub implementations around their subscribe/process loop.
+   */
+  @Internal
```
atakavci: nit: already package-private, no need for annotation

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -129,6 +130,11 @@ public static Builder builder() {
   private int infiniteSoTimeout = 0;
   private boolean broken = false;
   private volatile Throwable brokenCause = null;
+  // True only while a pub/sub read loop is driving this connection; gates whether pub/sub
+  // pushes are propagated as read results or dropped as stray frames. Held in a shared
+  // mutable holder (not a plain field) so the push consumer chain observes the flag even
+  // through shallow copies of this connection (e.g. Mockito spies wired via AuthXManager).
+  private final AtomicBoolean activeSubscription = new AtomicBoolean(false);
```
atakavci: just had a quick look for possible scenarios and how it [failed](https://github.com/redis/jedis/actions/runs/32702556686/job/97356906276); trade off between a boolean against atomic does **not** seem fair... One test case failing on mockito-copied instances does not justify use of atomics in core implementation.

atakavci (commented):
left a note around atomics, which i feel needs change. other than that, looks good to me.

src/main/java/redis/clients/jedis/Connection.java
uglide: Fair point — reworked in 969dfe67. The flag is back to a plain `volatile boolean`. The test is fixed instead: `testJedisPubSubReauth` now wraps the pooled connection in a delegating mock (`mock(Connection.class, delegatesTo(original))`) rather than `spy(...)`. The spy is a shallow field-copy, so the pub/sub loop was setting the flag on the copy while the push consumer chain read it from the original — that state fork is what broke the test. The delegating mock still records invocations for the AUTH-count `verify(...)` but forwards every call to the single real `Connection`, so no state is duplicated. Verified locally against a real Redis (`TokenBasedAuthenticationIntegrationTests` 5/5). Also dropped the mock test that encoded the shallow-copy support.

src/main/java/redis/clients/jedis/Connection.java
uglide: Fixed

src/main/java/redis/clients/jedis/Connection.java
```diff
@@ -129,6 +129,10 @@ public static Builder builder() {
   private int infiniteSoTimeout = 0;
   private boolean broken = false;
   private volatile Throwable brokenCause = null;
+  // True only while a pub/sub read loop is driving this connection; set by the pub/sub
+  // loops and read by the push consumer chain to decide whether pub/sub pushes are
+  // propagated as read results or dropped as stray frames.
+  private volatile boolean activeSubscription = false;
```
ggivo: Do we need it to be volatile? Is it expected to be accessed multithreaded?

atakavci (approved):
LGTM

ggivo (approved):
@igor.malinovskiy
LGTM and resolve the issue.
One caveat is that pub/sub messages will no longer be propagated by getUnflushedObject().

An alternative solution: if we could actually flip the logic and drop all pub/sub push messages by default in Connection, and propagate them only  in getUnflushedObject() (or a new dedicated method used by JedisPubSub)

src/main/java/redis/clients/jedis/Connection.java
uglide: removed

PR #4703, opened 2026-08-24
Log connection identity for maintenance events and applied socket timeouts

ggivo:
Maintenance handling is per connection, but the debug logs did not say which connection an event or timeout change applied to. Every controller event log now carries the connection identity, and each SO_TIMEOUT actually pushed to a socket logs value, blocking flag, and connection. Extracted from the SCH scenario-test work, where these logs are used to correlate per-connection behavior.

---

> [!NOTE]
> **Low Risk**
> Mostly debug logging plus a small reconnect-path tweak to timeout/identity cache. Does not change auth, protocol, or public API behavior.
> 
> **Overview**
> Debug logs for maintenance handling now identify **which connection** an event or SO_TIMEOUT change applied to, so per-connection rebind/relax behavior can be correlated.
> 
> `MaintenanceEventController` event logs include `conn=` via `toIdentityString()`, gated with `isDebugEnabled()`. `applyCurrentTimeout()` logs millis, blocking flag, and the same identity.
> 
> On reconnect, `broken` and the cached identity string are cleared immediately after the new socket is created (before logging), and the applied-timeout cache is invalidated so the initial timeout goes through `applyCurrentTimeout()` instead of a one-off `setSoTimeout`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6b685e7d92bdfa56c229563c04bfcf34613bc2e3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/main/java/redis/clients/jedis/Connection.java
+++ b/src/main/java/redis/clients/jedis/Connection.java
@@ -20,6 +20,9 @@ import java.util.function.Supplier;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicReference;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import redis.clients.jedis.Protocol.Command;
 import redis.clients.jedis.Protocol.Keyword;
 import redis.clients.jedis.TimeoutSource.TimeoutInfo;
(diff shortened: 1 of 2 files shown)


atakavci (commented):
only interesting bit in this PR is applying `broken` before completing `connect` .
1) this might end up with a stale `strVal` is accessed, which is no big deal i believe
2) `setTimeoutInfinite` calls `connect`, do we have any policy mentioning `setTimeoutInfinite` is thread safe or not ?

PR #4711, opened 2026-08-27
parse StreamEntryID components as unsigned 64-bit integers

### LINKED ISSUE #4710: StreamEntryID cannot parse stream IDs in the upper half of the unsigned 64-bit range
Redis stream IDs are a pair of **unsigned** 64-bit integers (`<ms>-<seq>`); the server parses and stores them with `string2ull` and accepts explicit IDs across the whole range via `XADD key <id> ...`. `StreamEntryID` parses them with signed `Long.parseLong`:

```java
public StreamEntryID(String id) {
  String[] split = id.split("-");
  this.time = Long.parseLong(split[0]);
  this.sequence = Long.parseLong(split[1]);
}
```

Anything above `2^63-1` throws `NumberFormatException`. `BuilderFactory.STREAM_ENTRY_ID` and every stream reply builder (`STREAM_ENTRY`, `STREAM_ENTRY_LIST`, `STREAM_PENDING_ENTRY_LIST`, group/consumer full-info) construct the ID through this constructor, so once such an entry exists the stream becomes unreadable from Jedis: `XRANGE`/`XREVRANGE`/`XREAD`/`XPENDING`/`XINFO` all die in the builder rather than returning data.

`toString()` (`time + "-" + sequence`) and `compareTo` (`Long.compare`) have the same signed assumption, so an ID in the upper half also re-serialises as a negative decimal and sorts incorrectly.

### Realistic scenario

Applications that assign explicit stream IDs — for idempotency, or to carry an external/ordering value in the `ms` field — can legitimately land in the upper half of the range, and a max-sequence entry (`<ms>-18446744073709551615`) is a valid single write. Jedis writes the ID happily (`XAddParams.id(String)` passes it through) but can no longer read the stream back.

Pure client-side (no server):

```java
new StreamEntryID("0-18446744073709551615");
// java.lang.NumberFormatException: For input string: "18446744073709551615"
```

Server-backed:

```java
jedis.xadd("s", new XAddParams().id("0-18446744073709551615"), Collections.singletonMap("f", "v")); // OK
jedis.xrange("s", "-", "+"); // NumberFormatException in BuilderFactory.STREAM_ENTRY
```

### Suggested fix

Parse/format/compare the two components as unsigned (`Long.parseUnsignedLong`, `Long.toUnsignedString`, `Long.compareUnsigned`). This is a strict superset: every ID in the signed range parses, prints and orders exactly as before, and the upper half now round-trips instead of throwing.

I have a patch with a regression test and can open a PR.

insaf021:
java.lang.NumberFormatException: For input string: "18446744073709551615"
        at java.base/java.lang.Long.parseLong(Long.java:721)
        at redis.clients.jedis.StreamEntryID.<init>(StreamEntryID.java:24)

Redis stream IDs are two unsigned 64-bit integers, but the `String` constructor parses them with signed `Long.parseLong`, and `toString`/`compareTo` carry the same signed assumption. An entry whose `ms` or `sequence` is above `2^63-1` is a valid write (`XADD key <id>`), yet every stream reply builder funnels the echoed ID back through this constructor, so once such an entry exists the stream is unreadable from Jedis and a high ID also re-serialises negative and sorts wrong.

`Long.parseUnsignedLong`/`Long.toUnsignedString`/`Long.compareUnsigned` cover the whole range; every ID in the signed half parses, prints and orders exactly as before, so the change is a strict superset. Regression test added.

Closes #4710.

---

> [!NOTE]
> **Low Risk**
> Localized streams ID handling with backward-compatible behavior for typical IDs; callers comparing or printing upper-half `long` values need unsigned APIs, which is documented.
> 
> **Overview**
> Fixes streams becoming unreadable in Jedis when an entry ID has an `ms` or sequence above `2^63-1` (valid in Redis but previously blew up in reply parsing with `NumberFormatException`).
> 
> **`StreamEntryID`** now treats both components as unsigned 64-bit values: string parsing uses `Long.parseUnsignedLong`, `toString()` uses `Long.toUnsignedString`, and `compareTo` uses `Long.compareUnsigned`. Javadoc on `getTime()` / `getSequence()` clarifies they return raw unsigned bits (upper-half values look negative as signed `long`). IDs in the normal signed range behave the same as before.
> 
> **`XAddParams.id(long, long)`** and **`id(long)`** format components unsigned so `XADD` does not send negative decimal IDs Redis rejects.
> 
> Release notes document the behavior change for 8.1.0. Unit tests cover round-trip, ordering, and common IDs; integration tests cover `XADD` / `XRANGE` / `XREAD` round-trips through reply builders.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1095f65f34e95baae8aeb73f788c3cebccc405be. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/release-notes/8.1.0.md
+++ b/docs/release-notes/8.1.0.md
@@ -26,8 +26,12 @@ components as unsigned 64-bit integers (`Long.parseUnsignedLong` /
 `Long.toUnsignedString` / `Long.compareUnsigned`). A stream ID above `2^63-1`
 (a valid `XADD` id, e.g. a max-sequence entry `<ms>-18446744073709551615`)
 previously threw `NumberFormatException` in the reply builder, making the stream
-unreadable; it now round-trips. Every ID in the signed range parses, prints and
-orders exactly as before.
+unreadable; it now round-trips. `XAddParams.id(long, long)` and
+`XAddParams.id(long)` likewise format their components unsigned instead of
+serialising upper-half values as negative decimals the server rejects. Every ID
+in the signed range parses, prints and orders exactly as before. Note that
+`StreamEntryID.getTime()`/`getSequence()` return the raw unsigned bits, so
+upper-half values appear negative under signed interpretation.
 
 **Action required:** none.
 
--- a/docs/release-notes/8.1.0.md
+++ b/docs/release-notes/8.1.0.md
@@ -19,6 +19,18 @@ This guide helps you upgrade from Jedis 8.0.0.
 Changes that may affect existing applications even though they are
 not breaking API changes.
 
+### Stream IDs parsed as unsigned 64-bit ([#4711](https://github.com/redis/jedis/pull/4711))
+
+`StreamEntryID` now parses, formats and compares its `ms` and `sequence`
+components as unsigned 64-bit integers (`Long.parseUnsignedLong` /
+`Long.toUnsignedString` / `Long.compareUnsigned`). A stream ID above `2^63-1`
+(a valid `XADD` id, e.g. a max-sequence entry `<ms>-18446744073709551615`)
+previously threw `NumberFormatException` in the reply builder, making the stream
+unreadable; it now round-trips. Every ID in the signed range parses, prints and
+orders exactly as before.
+
+**Action required:** none.
+
 ### Key pre-processor applied to all key arguments ([#4675](https://github.com/redis/jedis/pull/4675))
 
 The source key of `ZRANGESTORE`/`GEOSEARCHSTORE` and the script `KEYS` passed to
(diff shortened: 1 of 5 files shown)


ggivo (commented):
The core change looks right. 

Some follow up items spotted:

1. `XAddParams.id(long, long)` and `id(long)` still format the components as signed longs — an upper-half value serialises as a negative decimal the server rejects. Are these affected the same way?
2. `getTime()`/`getSequence()` now return upper-half values as negative longs — nothing tells callers the returned `long` carries unsigned raw bits, so signed arithmetic on it silently goes wrong. How do we make users aware of this?
3.  Suggest adding IT. The unit test doesn't exercise the reply builders, which is where the reported failure happens — a simple server round-trip test (`xadd` with an explicit upper-half id, then reads it back) would cover it.

PR #4719, opened 2026-08-29
Route keyless cluster commands to any node instead of throwing

### LINKED ISSUE #4720: Keyless commands throw NoSuchElementException on the cluster replica path
### Expected behavior

A keyless command sent to a replica runs on an arbitrary replica.

### Actual behavior

`ClusterConnectionProvider.getReplicaConnection(CommandArguments)` calls `getKeyHashSlots().iterator().next()` without checking the set is non-empty, so it throws before acquiring a connection:

```
java.util.NoSuchElementException
  at ClusterConnectionProvider.getReplicaConnection(ClusterConnectionProvider.java:145)
  at ReplicaOnlyConnectionResolver.resolve(ReplicaOnlyConnectionResolver.java:24)
```

`getConnection(CommandArguments)` has the same line. `executeCommand` is unaffected: keyless commands use `RoundRobinConnectionResolver`.

### Steps to reproduce

1. Call `JedisCluster.executeCommandToReplica` with a keyless command such as `PING`.

renechoi:
Fixes #4720

## Summary

Both `CommandArguments` overloads call `slots.iterator().next()` after checking only for more than one slot, so a keyless command (PING, INFO) throws `NoSuchElementException`.

## Key Decisions

The primary path returns `getConnection()`, what the dead `slot >= 0` ternary was for (`getSlot` never returns -1). The replica path picks from the replicas the slot cache recorded, which it does only with `readOnlyForRedisClusterReplicas` on. `getSlotReplicaPools` now separates the two no-replica cases: null when the client records none, empty when the slot has none yet. Neither replica path renews the cache for a result the renewal cannot change, and the keyed one no longer throws `NullPointerException`.

## Testing

Unit tests cover keyless and keyed routing, both fallbacks and disabled tracking; each new one fails without its fix. 3306 tests, release note updated.

---

> [!NOTE]
> **Medium Risk**
> Changes core Redis Cluster connection selection for keyless and replica reads; fixes exceptions but alters which node handles those commands.
> 
> **Overview**
> Fixes cluster routing when a command has **no key** (`PING`, `INFO`, etc.). `ClusterConnectionProvider` previously called `slots.iterator().next()` on an empty hash-slot set and threw **`NoSuchElementException`** before opening a connection.
> 
> **Primary path:** empty slots now use **`getConnection()`** (any reachable primary). **Replica path:** keyless commands use a random recorded replica via new **`getAnyReplicaConnection()`**, or fall back to a primary when the topology has no replicas.
> 
> **Replica tracking:** `JedisClusterInfoCache.getSlotReplicaPools(int)` now distinguishes clients that **do not record replicas** (`null` when `readOnlyForRedisClusterReplicas` is off) from slots with **no replica known yet** (empty list). Keyed replica routing skips **`renewSlotCache()`** when renewal cannot add replicas, avoiding **`NullPointerException`** and repeated topology refresh on keyless replica calls. A **`replicaNodesCache`** and public **`getReplicaNodes()`** support picking any replica for keyless replica commands.
> 
> Release notes for 8.1.0 document the behavior change; unit and integration tests cover cache semantics and routing fallbacks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f2a81f91395d4acbcfac593c907bdcc14fba261d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/main/java/redis/clients/jedis/providers/ClusterConnectionProvider.java
+++ b/src/main/java/redis/clients/jedis/providers/ClusterConnectionProvider.java
@@ -3,6 +3,7 @@ package redis.clients.jedis.providers;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -131,8 +132,11 @@ public class ClusterConnectionProvider implements ConnectionProvider {
       throw new JedisClusterOperationException("Cannot get connection for command with multiple hash slots");
     }
 
-    int slot = slots.iterator().next();
-    return slot >= 0 ? getConnectionFromSlot(slot) : getConnection();
+    if (slots.isEmpty()) {
+      return getConnection(); // a keyless command can run on any node
+    }
+
+    return getConnectionFromSlot(slots.iterator().next());
   }
 
   public Connection getReplicaConnection(CommandArguments args) {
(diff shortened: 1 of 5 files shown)


src/main/java/redis/clients/jedis/providers/ClusterConnectionProvider.java
```diff
@@ -142,8 +146,39 @@ public Connection getReplicaConnection(CommandArguments args) {
...
+   * the same way {@link #getReplicaConnectionFromSlot(int)} does for a slot whose replicas are
+   * unknown.
+   */
+  private Connection getReplicaConnection() {
+    Connection replica = getAnyReplicaConnection();
+    if (replica != null) {
+      return replica;
+    }
+
+    renewSlotCache();
```
atakavci: the issue with this new method is,, in case of no replica in topology, it keeps falling back to slot cache renewal which is pretty expensive operation.

src/main/java/redis/clients/jedis/providers/ClusterConnectionProvider.java
```diff
@@ -142,8 +146,39 @@ public Connection getReplicaConnection(CommandArguments args) {
...
+    replica = getAnyReplicaConnection();
+    return replica != null ? replica : getConnection();
+  }
+
+  private Connection getAnyReplicaConnection() {
+    Map<String, ConnectionPool> replicas = new HashMap<>(getNodes());
+    replicas.keySet().removeAll(getPrimaryNodes().keySet());
```
atakavci: this doesn't respect to the `JedisClientConfig.isReadOnlyForRedisClusterReplicas()` flag, which excludes the replicas from slot cache when it is set to `false`

atakavci (commented):
hi @renechoi,

i got comments below and i believe both stands important.

PR #4721, opened 2026-09-01
Stabilize search-on-timeout FAIL tests against worker-thread race

atakavci:
## Summary
The `searchOnTimeoutFailReturnsError` / `aggregateOnTimeoutFailReturnsError` tests [fail nightly against Redis unstable](https://github.com/redis-developer/redis-oss-release-automation/actions/runs/33448618352/job/99673236515) since 2026-08-27. As of RediSearch/RediSearch#10961, the `FAIL` on-timeout policy is enforced by a blocked-client timeout callback racing the worker thread running the query — confirmed as intended behavior in [RediSearch#11274](https://github.com/RediSearch/RediSearch/issues/11274): when the worker wins, an over-budget query returns full results instead of an error. With 10k documents the query overran the 1ms timeout by only a few milliseconds, leaving the outcome to scheduling jitter. Raising `ON_TIMEOUT_DOC_COUNT` to 100k gives a ~200x runtime-to-timeout margin so the callback deterministically wins.

- Keep testing the realistic workers-enabled path with `TIMEOUT 1`, rather than forcing `search-workers 0` (inline execution sidesteps the race).
- Flakiness was measured below ~10x margin; the constant's javadoc records this threshold.

## Testing
Test-only change, verified via redis-cli against the failing image (`unstable-33278213642-debian`): at 100k docs the FAIL error returned 50/50 (search) and 20/20 (aggregate), versus 1/10 at 10k. The shared fixture also serves the RETURN-policy warning tests, which remain valid.

---

> [!NOTE]
> **Low Risk**
> Test-only changes to document count, comments, and version filtering; no production client or runtime behavior changes.
> 
> **Overview**
> Stabilizes flaky **search-on-timeout** integration tests that assert `FAIL` vs `RETURN` behavior for `FT.SEARCH` and `FT.AGGREGATE`.
> 
> **`ON_TIMEOUT_DOC_COUNT`** is raised from **10k to 100k** so timed-out queries run far past the 1ms limit (~200× margin). With `search-workers > 0`, RediSearch’s `FAIL` policy depends on a race between the blocked-client timeout callback and the worker thread; a small overrun could still return full results, which made nightly runs flaky at 10k docs. The fixture javadoc now documents that race and the ~10× flakiness threshold.
> 
> All four on-timeout tests gain **`@SinceRedisVersion("8.10.0", message = "search-on-timeout policy")`** so they are skipped on older Redis where the unified `search-on-timeout` policy is not in scope. Existing `assumeTrue` checks on the Search module version remain unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 28211db04e034d31bb9dabff540729b1a47dd05d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/test/java/redis/clients/jedis/commands/unified/search/SearchWithParamsCommandsTestBase.java
+++ b/src/test/java/redis/clients/jedis/commands/unified/search/SearchWithParamsCommandsTestBase.java
@@ -1561,11 +1561,14 @@ public abstract class SearchWithParamsCommandsTestBase extends UnifiedJedisComma
   }
 
   /**
-   * Number of documents indexed by the on-timeout tests. Large enough that scanning, scoring and
-   * sorting them cannot complete within the 1ms per-query timeout, so the query engine's on-timeout
-   * policy is guaranteed to kick in.
+   * Number of documents indexed by the on-timeout tests. With {@code search-workers > 0} the
+   * {@code FAIL} policy is enforced by a blocked-client timeout callback racing the worker thread
+   * that runs the query: the error is only returned when the callback wins, so a query that only
+   * slightly overruns its timeout may still return full results (intended server behavior). The
+   * query runtime must therefore exceed the 1ms per-query timeout by a wide margin — outcomes were
+   * measured flaky below ~10x, while 100k documents keep the margin around ~200x.
    */
-  private static final int ON_TIMEOUT_DOC_COUNT = 10_000;
+  private static final int ON_TIMEOUT_DOC_COUNT = 100_000;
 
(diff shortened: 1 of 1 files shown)


ggivo (commented):
LGTM, 
Will suggest leaving test running for all Redis versions, if not, causing a delay in CI. Using @SinceRedisVersion implies it is availble/compatible with the given version

 Otherwise, it's probably better to introduce a dedicated "slow" test job

Issue #4722, opened 2026-09-03
`UnifiedJedis.subscribe()` returns a still-subscribed connection to the pool when the subscribe loop exits abnormally


bironran:
**Title:** `UnifiedJedis.subscribe()` returns a still-subscribed connection to the pool when the subscribe loop exits abnormally

---

### Expected behavior

When the subscribe loop in `JedisPubSubBase.process()` terminates while the connection is still subscribed, the connection should be marked broken so that `Connection.close()` routes it to `ConnectionPool.returnBrokenResource()`. A connection that is still in subscriber mode on the server must never be handed to the next borrower.

### Actual behavior

`UnifiedJedis.subscribe()` borrows a pooled connection in a try-with-resources:

```java
// UnifiedJedis.java:3819
public void subscribe(final JedisPubSub jedisPubSub, final String... channels) {
  try (Connection connection = this.provider.getConnection()) {
    jedisPubSub.proceed(connection, channels);
  }
}
```

`proceed()` only rolls back the timeout on the way out:

```java
// JedisPubSubBase.java:87
public final void proceed(Connection client, T... channels) {
  this.client = client;
  this.client.setTimeoutInfinite();
  try {
    subscribe(channels);
    process();
  } finally {
    this.client.rollbackTimeout();   // <- broken is never set
  }
}
```

and the loop has two exits that leave the subscription live:

```java
// JedisPubSubBase.java:178
} while (!Thread.currentThread().isInterrupted() && isSubscribed());
```

1. **A callback throws.** Anything escaping `onMessage` / `onPMessage` / `onSubscribe` / `onUnsubscribe` propagates out of `process()`. `rollbackTimeout()` succeeds (the socket is still open), so `broken` stays `false`.
2. **The listener thread's interrupt flag is set.** The loop exits after the current message, still subscribed, and `subscribe()` returns *normally*. Because the interrupt check is evaluated **before** `isSubscribed()`, user code cannot even send an `UNSUBSCRIBE` and let the loop drain the confirmations — the loop will not run another iteration.
3. `process()` itself throws `JedisException("Unknown message type: ...")` (lines 121, 166, 176), or a `JedisDataException` comes off the read path.

In all three cases `Connection.broken` remains `false` — `readProtocolWithCheckingBroken()` (line 386) only sets it on `JedisConnectionException` — so `Connection.close()` (line 255) takes the `returnResource()` branch and a connection that is still `SUBSCRIBE`d goes back into the shared pool.

[...]

PR #4723, opened 2026-09-04
Discard pooled connections after aborted pub/sub loops

dawnwang3692:
Refs #4722.

Implements the pooled-connection cleanup suggested in the issue for `subscribe`, `psubscribe`, and `ssubscribe`, including both `JedisCluster` and `RedisClusterClient`.

Mark the connection broken if the subscription call aborts or returns while still subscribed. Normal unsubscription still allows reuse. This is conservative: a fully parsed server error also causes the connection to be discarded if the call aborts.

The regression tests cover callback exceptions and errors, unknown replies, failures before the first subscription reply, interrupted listeners, and normal unsubscribe cleanup. The integration tests use a single-connection pool and check both the next GET and whether the original connection was replaced.

Validation on Corretto 8:

- Targeted unit tests: 28 passed.
- Redis 8.6.1 integration tests: 24 passed across RESP2/RESP3, using standalone and a three-primary-node cluster.
- Full unit suite: 3,449 tests, no failures or errors, 7 skipped.
- Formatter validation and `git diff --check` passed.
- The full Docker version matrix, TLS, and failover tests were not run locally.

Connection detachment, authentication-hook teardown, and interruption of blocking socket reads remain outside this patch.

---

> [!NOTE]
> **Medium Risk**
> Changes connection lifecycle for all pub/sub subscription paths; pooled clients are safer but may discard connections more often after listener errors or interrupts.
> 
> **Overview**
> Fixes a pool safety bug where pub/sub connections could be returned while still in subscriber mode after a callback exception, interrupt, or other aborted read loop.
> 
> **`JedisPubSubBase` and `JedisShardedPubSubBase`** now wrap the subscription `process()` loop in `try`/`finally` and call `setBroken()` when the loop does not finish cleanly or exits while `isSubscribed()` is still true. Pooled clients then discard the connection via `returnBrokenResource` instead of reusing it; a normal unsubscribe confirmation leaves the connection healthy. The same marking applies to non-pooled `Jedis`, which rejects later commands until the instance is closed.
> 
> Release notes for 8.1.0 document the behavior change ([#4722](https://github.com/redis/jedis/issues/4722)). New unit and integration tests cover callback failures, interrupts, unknown replies, early read failures, and successful unsubscribe across `subscribe`/`psubscribe`/`ssubscribe` (standalone, `JedisCluster`, `RedisClusterClient`, RESP2/RESP3).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 955f8c88abe91928fe41bdcda62d168e9b358b30. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/main/java/redis/clients/jedis/JedisPubSubBase.java
+++ b/src/main/java/redis/clients/jedis/JedisPubSubBase.java
@@ -107,10 +107,15 @@ public abstract class JedisPubSubBase<T> {
     authenticator.registerForAuthentication(client);
     authenticator.client.setTimeoutInfinite();
     authenticator.client.setActiveSubscription(true);
+    boolean completed = false;
     try {
       subscribe(channels);
       process();
+      completed = true;
     } finally {
+      if (!completed || isSubscribed()) {
+        authenticator.client.setBroken();
+      }
       authenticator.client.setActiveSubscription(false);
       authenticator.client.rollbackTimeout();
     }
@@ -120,10 +125,15 @@ public abstract class JedisPubSubBase<T> {
     authenticator.registerForAuthentication(client);
     authenticator.client.setTimeoutInfinite();
     authenticator.client.setActiveSubscription(true);
+    boolean completed = false;
     try {
       psubscribe(patterns);
       process();
+      completed = true;
     } finally {
+      if (!completed || isSubscribed()) {
+        authenticator.client.setBroken();
+      }
       authenticator.client.setActiveSubscription(false);
       authenticator.client.rollbackTimeout();
     }
--- a/docs/release-notes/8.1.0.md
+++ b/docs/release-notes/8.1.0.md
@@ -62,13 +62,16 @@ pub/sub frames from a connection without going through the `JedisPubSub`/
 `getUnflushedObject()`) will no longer receive those frames; use the pub/sub APIs or
 register a custom `PushConsumer` that propagates them.
 
-### Discard connections after an aborted subscription ([#4722](https://github.com/redis/jedis/issues/4722))
+### Pooled connections are discarded after aborted subscriptions ([#4722](https://github.com/redis/jedis/issues/4722))
+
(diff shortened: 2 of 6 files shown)


src/main/java/redis/clients/jedis/JedisPubSubBase.java
```diff
@@ -107,10 +107,15 @@ public final void proceed(Connection client, T... channels) {
     authenticator.registerForAuthentication(client);
     authenticator.client.setTimeoutInfinite();
     authenticator.client.setActiveSubscription(true);
+    boolean completed = false;
     try {
       subscribe(channels);
       process();
+      completed = true;
     } finally {
+      if (!completed || isSubscribed()) {
```
atakavci: @dawnwang3692 
subscription operations should be safe in terms of handling a broken connection.
is there any particular reason to not to fix it in single place in `process()?`

atakavci:
@dawnwang3692 lets consider the impact with the legacy non-pooled type; `Jedis` also leverages `JedisPubSub`.

dawnwang3692:
@atakavci I added tests for plain `Jedis` in 955f8c88. Normal unsubscribe still leaves the connection reusable. If a callback throws, or the loop exits on interruption while still subscribed, the connection is marked broken and further commands throw `JedisConnectionException`.

The caller needs to close the affected instance and use a new one; there's no automatic close or reconnect. The original exception and interrupt flag are preserved. I've updated the release notes to make this clear.

The 12 new cases cover `subscribe`/`psubscribe` on RESP2 and RESP3, including normal reuse and recovery with a new instance. All 36 cleanup integration cases passed locally on JDK 8 / Redis 8.2.1. The unit suite passed too, with 6 environment-dependent skips. CI is still waiting for approval.

Does that look right for the non-pooled case?

PR #4724, opened 2026-09-04
[maintenance events - cluster] Decode SMIGRATING/SMIGRATED cluster maintenance pushes into events

atakavci:
## Summary
Adds RESP3 push decoding for the OSS-cluster maintenance notifications `SMIGRATING` and `SMIGRATED`, producing typed `SMigratingEvent` / `SMigratedEvent` domain objects alongside the existing standalone maintenance events. This PR is parsing and dispatch only: the events reach `MaintenanceEventListener`, but no cluster-aware handling (timeout relaxation, slot-map update, connection retirement) is implemented yet.

- Cluster events live in their own `ClusterMaintenanceEvent` hierarchy but still extend `MaintenanceEvent`, so the codec and consumer keep a single event type and dispatch path. The two families are treated as exclusive contracts: a connection receives one or the other.
- The slots-or-ranges wire field (`123,456,789-1000`) is modeled by a new package-private `HashSlotRanges` value type that keeps wire order, exposes membership, count and iteration, and validates every slot against `[0, 16384)`.
- `SMIGRATED` node addresses are assumed to be bare `host:port` strings parsed via `HostAndPort.from`.
- Events dedup by `seq` alone, like the non-MOVING standalone events.

- `PushMessageTypes` gains public `SMIGRATING` / `SMIGRATED` constants (`@since 8.1`).
- The push type resolver now handles a length collision: `MIGRATING` and `SMIGRATED` are both 9 bytes, so that case does two comparisons instead of one.
- Malformed cluster frames (wrong arity or types, unparseable slots or addresses, one bad entry in an `SMIGRATED` list) throw `MalformedMaintenanceEventException`, so the consumer drops them without dispatch, matching existing behavior.
- `MaintenanceEventListener` gains `onSMigrating` / `onSMigrated`. The standalone `MaintenanceEventController` implements them as no-ops that log a warning and leave pool-wide state untouched.

## Testing
New unit tests cover `HashSlotRanges` parsing and rejection rules, the event hierarchy (dispatch, read-only delta, rendering), and codec resolution and field extraction, including the exhaustive malformed-frame matrix. The consumer, identity and controller tests are extended to confirm routing to the new callbacks, seq-based identity, and that cluster events do not disturb an active MOVING window.

## Notes
- `HashSlotRanges.parse` accepts a lone or trailing comma (`","`, `"1,"`) silently because `String.split` drops trailing empty tokens. Worth tightening if the wire format should be strict.
- The `MaintenancePushCodec` class javadoc still references a removed `buildCluster` method.
- Breaking for any external `MaintenanceEventListener` implementations, but the interface is package-private, so none are expected.

---

> [!NOTE]
> **Low Risk**
> Changes are additive parsing and listener callbacks; cluster events are explicitly ignored by the standalone controller, with no slot-map or connection lifecycle impact yet.
> 
> **Overview**
> Adds **decode-and-dispatch** for OSS cluster maintenance pushes `SMIGRATING` and `SMIGRATED`, alongside the existing standalone maintenance events. Typed events (`SMigratingEvent`, `SMigratedEvent`, `SlotMigration`) live under a new `ClusterMaintenanceEvent` hierarchy but still extend `MaintenanceEvent`, so the existing codec/consumer path stays unified.
> 
> **`MaintenancePushCodec`** gains `SMIGRATING`/`SMIGRATED` decoders, a `HashSlotRanges` parser for the `slots-or-ranges` wire field, and updated token resolution (including disambiguating 9-byte `MIGRATING` vs `SMIGRATED`). Malformed cluster frames still throw `MalformedMaintenanceEventException` and are dropped without listener dispatch.
> 
> **`MaintenanceEventListener`** adds `onSMigrating` / `onSMigrated`. The standalone **`MaintenanceEventController`** implements them as warn-only no-ops so cluster pushes do not alter MOVING rebind, timeout relaxation, or remap behavior. **No cluster-aware handling** (slot-map updates, timeout relax on SMIGRATING, connection retirement) is in this PR.
> 
> Unit tests cover `HashSlotRanges`, event dispatch/identity, codec malformed matrices, consumer routing, and controller inertness.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 50615c11be8036f5c3b2ee57b6a46af6812cad1e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/test/java/redis/clients/jedis/ClusterMaintenanceEventTest.java
+++ b/src/test/java/redis/clients/jedis/ClusterMaintenanceEventTest.java
@@ -1,0 +1,89 @@
+package redis.clients.jedis;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Unit coverage for the {@link ClusterMaintenanceEvent} hierarchy: each event dispatches to its own
+ * listener callback with the owning connection, SMIGRATED exposes its delta read-only, and
+ * {@link SlotMigration} renders as {@code src -> dest [slots]}.
+ */
+@Tag("unit")
+public class ClusterMaintenanceEventTest {
+
+  private static final HostAndPort SRC = new HostAndPort("10.0.0.1", 7000);
+  private static final HostAndPort DEST = new HostAndPort("10.0.0.2", 7001);
+
+  private final MaintenanceEventListener listener = mock(MaintenanceEventListener.class);
+  private final Connection connection = mock(Connection.class);
+
+  @Test
+  public void sMigratingDispatchesToOnSMigrating() {
+    SMigratingEvent e = new SMigratingEvent(3L, HashSlotRanges.parse("0-100"));
+
+    e.accept(listener, connection);
+
+    verify(listener).onSMigrating(e, connection);
+    verifyNoMoreInteractions(listener);
+  }
+
+  @Test
+  public void sMigratedDispatchesToOnSMigrated() {
+    SMigratedEvent e = new SMigratedEvent(4L,
+        Collections.singletonList(new SlotMigration(SRC, DEST, HashSlotRanges.parse("0-100"))));
+
+    e.accept(listener, connection);
+
+    verify(listener).onSMigrated(e, connection);
+    verifyNoMoreInteractions(listener);
+  }
+
+  @Test
+  public void clusterEventsAreMaintenanceEvents() { // the codec/consumer handle one
+                                                    // MaintenanceEvent type
+    assertInstanceOf(MaintenanceEvent.class, new SMigratingEvent(1L, HashSlotRanges.parse("1")));
+    assertInstanceOf(MaintenanceEvent.class, new SMigratedEvent(1L, Collections.emptyList()));
+    assertEquals(1L, new SMigratingEvent(1L, HashSlotRanges.parse("1")).seq);
+    assertEquals(2L, new SMigratedEvent(2L, Collections.emptyList()).seq);
+  }
+
+  @Test
+  public void sMigratingExposesSlots() {
+    HashSlotRanges slots = HashSlotRanges.parse("123,456,789-1000");
+    assertSame(slots, new SMigratingEvent(1L, slots).slots);
+  }
+
+  @Test
+  public void sMigratedMigrationsAreReadOnly() {
+    List<SlotMigration> source = new ArrayList<>(
+        Arrays.asList(new SlotMigration(SRC, DEST, HashSlotRanges.parse("0-100")),
+          new SlotMigration(DEST, SRC, HashSlotRanges.parse("200"))));
+    SMigratedEvent e = new SMigratedEvent(1L, source);
+
+    assertEquals(source, e.migrations);
+    assertThrows(UnsupportedOperationException.class, () -> e.migrations.clear());
+    assertThrows(UnsupportedOperationException.class,
+      () -> e.migrations.add(new SlotMigration(SRC, DEST, HashSlotRanges.parse("5"))));
+  }
+
+  @Test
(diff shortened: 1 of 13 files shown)


src/test/java/redis/clients/jedis/ClusterMaintenanceEventTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Unit coverage for the {@link ClusterMaintenanceEvent} hierarchy: each event dispatches to its own
+ * listener callback with the owning connection, SMIGRATED exposes its delta read-only, and
+ * {@link SlotMigration} renders as {@code src -> dest [slots]}.
+ */
+@Tag("unit")
```
ggivo, line 24: Nit: drop @Tag("unit") from ClusterMaintenanceEventTest and HashSlotRangesTest — untagged already means unit here (the build selects unit tests by excluding integration/scenario), so the tag is inert. It was introduced by mistake in the first SCH drop; should consider dropping them from other SCH unit tests (this or separate review).

src/test/java/redis/clients/jedis/ClusterMaintenanceEventTest.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  @Test
+  public void sMigratedMigrationsAreReadOnly() {
+    List<SlotMigration> source = new ArrayList<>(
+        Arrays.asList(new SlotMigration(SRC, DEST, HashSlotRanges.parse("0-100")),
+          new SlotMigration(DEST, SRC, HashSlotRanges.parse("200"))));
+    SMigratedEvent e = new SMigratedEvent(1L, source);
+
+    assertEquals(source, e.migrations);
```
ggivo: consider a public accessor instead package private filed migrations

src/test/java/redis/clients/jedis/HashSlotRangesTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Unit coverage for {@link HashSlotRanges}: the slots-or-ranges wire format
+ * ({@code 123,456,789-1000}) of cluster maintenance pushes — parsing, membership, counting,
+ * iteration order and rejection of malformed input.
+ */
+@Tag("unit")
```
ggivo: nit: unit test tag is not needed.

My bad, seems I accidentally introduced them with the previous dorp of SCH, we should be able to drop them alslo form other SCH unit tests

ggivo: nit : unit test tag is not needed.

My bad, seems I accidentally introduced them with the previous drop of SCH. We should be able to drop them also from other SCH unit tests

ggivo (approved):
LGTM

added a small nit comment for unit test tags,( I have introduced by mistake with previous drop of SCH), and one question

src/test/java/redis/clients/jedis/MaintenanceEventIdentityTest.java
```diff
@@ -54,4 +58,29 @@ public void movingIdentity_neverEqualsSeqOnlyIdentity() {
...
+
+    assertEquals(new SMigratedEvent(5L, delta).identity(),
+      new SMigratedEvent(5L, Collections.emptyList()).identity(),
+      "same seq: same operation regardless of the slot delta");
+    assertNotEquals(new SMigratedEvent(5L, delta).identity(),
+      new SMigratedEvent(6L, delta).identity());
+
+    assertEquals(new SMigratingEvent(5L, slotsA).identity(),
+      new SMigratedEvent(5L, delta).identity(),
+      "an SMIGRATED terminates the SMIGRATING with the same seq");
```
ggivo: Hm, can you confirm this?
My understanding is that each family of maintence events comes with it's own seqId? 

E.g I would expect SMIGRATED paired with the given SMIGRATING to have a higher seqId, not the same.

ggivo (commented):
@atakavci 
And got a question. Going to test it, but probably you already have the answer

src/test/java/redis/clients/jedis/MaintenanceEventIdentityTest.java
atakavci: this line of assertion escaped my review, it holds a wrong assumption. 
The simple rule i have in my design is each family and the message-type in that family holds different `seqId`, and the one with `SMIGRATED` is bigger. 
I will investigate server behaviour if it is possible to have overlapping families in terms of `seqId`s, like in a timeline;
```
T1  MIGRATING seqId=1
T2  MIGRATING seqId=2
T3  MIGRATED  seqId=3   (closes seqId=1)
T4  MIGRATED  seqId=4   (closes seqId=3)
```

PR #4729, opened 2026-09-09
[maintenance events - cluster] Add cluster maintenance coordinator and atomic SMIGRATED slot-delta application

atakavci:
**This PR built on top of #4724** For the diff against it, [here](https://github.com/redis/jedis/compare/atakavci:ali/parseClusterMsg...atakavci:jedis:ali/clusterCoordinator). 

## Summary
Builds the reaction layer for OSS-cluster maintenance pushes on top of the SMIGRATING/SMIGRATED parsing: a per-client `ClusterMaintenanceCoordinator` that folds the per-node broadcast into one client-wide operation and drives a shared timeout-relax gate, and a `JedisClusterInfoCache.applySlotMigration` path that applies slot deltas directly from the push — no `CLUSTER SLOTS` round-trip — atomically against the topology-refresh lifecycle.

- Seq ids are assumed ordered but **not** assumed shared between an SMIGRATING/SMIGRATED pair: a closing SMIGRATED discards any open window with a lower seq, so the relax gate can't dangle on a lost closer (TTL backstop remains).
- Deltas queue and apply only under a dedicated `slotDeltaLock`; a refresh holds it for its whole query→apply span, so deltas never interleave with and always order after a refresh. Read threads never block or spin — a failed `tryLock` walks away because every lock releaser re-checks the queue.
- Delta application and the slotless-source scan happen under a single write-lock hold to avoid interleaving with `renewClusterSlots`; slotless sources leave `primaryNodesCache`, destinations join it, stale replica-slot entries drop until the next refresh.

- Timeouts stay relaxed while any migration window is open; overlapping migrations unrelax only when the last closes. Standalone events reaching the coordinator only log a warning.
- `MaintenanceEventController.retireAll` is added for cluster Case-2 retirement but has no caller yet. It might get cut off if the wiring of coordinator into maintenance events does not leverage it.

## Notes
- Retirement wiring (slotless pool teardown) and client construction wiring follow in later PRs.

_______________________________________________________________________________

### Problem
`SMIGRATED` push notifications are broadcast to the client's connections on every cluster node, so
the same client receives multiple copies of each event — and copies from a slow or stalled
connection can arrive **after** newer events were already processed. Sequence ids order the events:
with migrations M1 < M2 < M3 < M4, a client may process M3 and M4, apply their slot deltas, and
only then receive M1 from a lagging connection.

Applying a late delta as-is is wrong: it would overwrite slot assignments that newer migrations
already changed. Merging it only against the **newest** processed operation is also insufficient:
for the slots where the late delta overlaps an **intermediate** operation (M1 ∩ M3, with M4 not
covering them), a newest-only merge would resurrect the older assignment — M3's already-applied
truth would be silently rolled back.

### Decision

The coordinator keeps the recently completed SMIGRATED operations, ordered by seq (bounded
retention, oldest dropped first — `MAX_COMPLETED_MIGRATIONS`). When an unknown SMIGRATED arrives
with a seq **below** the newest completed operation:

1. **Unambiguous context → merge and apply.** If retention still reaches down to the late seq
   (`firstRetainedSeq <= lateSeq`) and **no other completed operation lies in the open interval
   `(lateSeq, newestSeq)`**, the late delta is folded into the newest operation (each contribution
   kept isolated under its own seq) and the freshly built combined delta is applied: contributions
   laid down in ascending seq order, higher seq overriding overlapping slots — equivalent to
   applying each one separately in seq order.
2. **Ambiguous context → drop.** If an intermediate completed operation exists in that interval,
   or the late seq lies below the retained range (intermediates may have been evicted, so the
   interval cannot be verified), the delta is **discarded**. The seq stays recorded so
   re-deliveries are treated as duplicates; stale SMIGRATING windows are still discarded and the
   timeout-relax gate re-evaluated; the drop is logged with the retained range.

Duplicates (any already-recorded seq) never re-apply a delta. Full topology renewals are a
separate mechanism and are explicitly out of scope for this rule.

### Arguments

- **Correctness over completeness.** Re-applying or re-deriving an old delta against an uncertain
  ordering context risks overwriting newer topology — an error the client cannot detect. Dropping
  is always safe: the slot table is merely *less up to date*, never *wrong by our own hand*.
- **The safe case covers the common case.** The typical reordering is a single straggler closer —
  exactly the adjacent case 1, which is handled fully and provably: with no intermediate, the
  newest operation and its prior contributions are the complete set of newer authorities.
- **Dropped information is cheap to lose.** Slots covered only by a dropped delta stay stale until
  the built-in self-healing corrects them: the first command touching such a slot gets a `MOVED`
  redirect (transparently retried and followed by a topology refresh), and periodic/error-triggered
  refreshes restore ground truth regardless of traffic.
- **The ambiguous case is rare.** It requires two or more out-of-order closers inside the retained
  window; the retention floor additionally converts the "history was evicted" corner from a
  correctness risk into a deliberate drop.
- **Simplicity.** No cross-history reconstruction machinery; one interval-emptiness check on an
  ordered map decides the path, and every structure involved stays lock-free.

### Consequences

- Within retained history, an older delta can never overwrite newer topology.
- Case 2 trades freshness for safety: bounded staleness on the dropped slots, self-healed by
  `MOVED` redirects or the next refresh; the drop is observable in the log.
- Accepted narrow races (shared with normal processing): two closers recorded concurrently can
  apply in either order; a delta racing a stale full refresh can be transiently overridden. Both
  self-heal via `MOVED`/refresh.

---

> [!NOTE]
> **High Risk**
> Changes cluster slot routing, refresh/delta ordering, and timeout behavior during live migrations—incorrect ordering or locking could cause transient wrong-node routing until MOVED/refresh.
> 
> **Overview**
> Adds a per-cluster-client **`ClusterMaintenanceCoordinator`** that deduplicates per-node **SMIGRATING/SMIGRATED** pushes, keeps a client-wide **timeout relax** gate open while migration windows or pending slot deltas exist, and routes applied deltas into the topology cache without a **CLUSTER SLOTS** round-trip.
> 
> **`JedisClusterInfoCache`** gains queued **`applySlotMigration`**: deltas drain under **`slotDeltaLock`**, never interleaving with **`renewClusterSlots`** (refresh holds the lock for query→apply, then drains the queue). Read threads enqueue and **`tryLock`** only—no blocking on refresh. **`processSlotDelta`** updates slot routing and **`primaryNodesCache`** for slotless sources in one write-lock hold.
> 
> **`MaintenanceEventController.retireAll`** is added for future cluster “node owns no slots” retirement; standalone events hitting the cluster coordinator only log warnings. Push codec docs now treat cluster events as part of the unified **`build`** path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7daefa81a2937b2831caf16f56c995176d3fab4c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/main/java/redis/clients/jedis/ClusterMaintenanceCoordinator.java
+++ b/src/main/java/redis/clients/jedis/ClusterMaintenanceCoordinator.java
@@ -1,0 +1,209 @@
+package redis.clients.jedis;
+
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.function.Supplier;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import redis.clients.jedis.TimeoutSource.TimeoutInfo;
+
+/**
+ * Cluster maintenance coordinator — one per cluster client, the single dedup/apply point for the
+ * per-node SMIGRATING/SMIGRATED broadcast. Owns the seq-keyed operations table that (a) folds the
+ * N-connection broadcast into one client-wide operation and (b) drives the shared relax gate:
+ * timeouts stay relaxed while ANY operation is open, so overlapping migrations unrelax only when
+ * the last one closes, and a connection created mid-event relaxes from the moment its overlay is
+ * installed. On SMIGRATED it applies the slot delta atomically through
+ * {@link JedisClusterInfoCache#applyMigrationDelta} — queued and drained after completion when a
+ * full topology refresh is in flight, never blocking a read thread — and retires the connections
+ * of nodes left without slots via each pool's own {@link MaintenanceEventController}, resolved at
+ * event time (no node map here). The pools remain their controllers' sole owners; this class only
+ * calls.
+ */
+final class ClusterMaintenanceCoordinator implements MaintenanceEventListener {
+
+  private static final Logger logger = LoggerFactory.getLogger(ClusterMaintenanceCoordinator.class);
+
+  private final JedisClusterInfoCache cache;
+  /** Backstop for an SMIGRATING whose SMIGRATED is lost, and retention of closed entries. */
+  private final long maxRelaxedDurationNanos;
+  private final Supplier<TimeoutInfo> timeoutSupplier;
+
+  /** Seq-keyed operations: open entries gate the relax; closed entries absorb late duplicates. */
+  private final ConcurrentHashMap<Object, MigrationOperation> operations = new ConcurrentHashMap<>();
+
+  /** SMIGRATED deltas received while a full refresh was in flight; drained post-refresh. */
+  private final ConcurrentLinkedQueue<SMigratedEvent> pendingDeltas = new ConcurrentLinkedQueue<>();
+
+  ClusterMaintenanceCoordinator(JedisClusterInfoCache cache,
+      MaintenanceNotificationsConfig config) {
+    this.cache = cache;
+    this.maxRelaxedDurationNanos = config.getRelaxedWindowMaxDuration().toNanos();
+    TimeoutInfo relaxedTimeoutInfo = new TimeoutInfo(config.getRelaxedTimeout(),
+        config.getRelaxedBlockingTimeout());
+    this.timeoutSupplier = () -> hasActiveMigration() ? relaxedTimeoutInfo : null;
+    cache.setPostRefreshHook(this::drainPendingDeltas);
+  }
+
+  /** The client-wide relax gate consulted by every cluster connection's timeout overlay. */
+  Supplier<TimeoutInfo> getTimeoutSupplier() {
+    return timeoutSupplier;
+  }
+
+  /**
+   * True while any migration window is open (SMIGRATING seen, SMIGRATED not yet, TTL unexpired).
+   */
+  boolean hasActiveMigration() {
+    if (operations.isEmpty()) {
+      return false;
+    }
+    boolean active = false;
+    for (MigrationOperation op : operations.values()) {
+      if (op.isExpired()) {
+        operations.remove(op.id, op);
+      } else if (!op.closed) {
+        active = true;
+      }
+    }
+    return active;
+  }
+
+  @Override
+  public void onSMigrating(SMigratingEvent e, Connection c) {
+    if (logger.isDebugEnabled()) {
+      logger.debug("Slot migration starting: {} (seq={}) conn={}", e.slots, e.seq,
+        c.toIdentityString());
+    }
+    long deadline = NanoClock.INSTANCE.getAsLong() + maxRelaxedDurationNanos;
+    operations.computeIfAbsent(e.identity(), k -> new MigrationOperation(k, deadline));
+    c.applyCurrentTimeout(); // the gate just opened; push the relax to the receiving socket now
+  }
+
+  @Override
+  public void onSMigrated(SMigratedEvent e, Connection c) {
+    long deadline = NanoClock.INSTANCE.getAsLong() + maxRelaxedDurationNanos;
+    boolean[] firstDelivery = { false };
+    operations.compute(e.identity(), (k, cur) -> { // atomic per identity
+      if (cur == null) {
+        // SMIGRATED without a preceding SMIGRATING (e.g. connected mid-event): still applied
+        firstDelivery[0] = true;
+        return MigrationOperation.closed(k, deadline);
+      }
+      if (!cur.closed) {
+        firstDelivery[0] = true;
+        return cur.close();
+      }
+      return cur; // duplicate broadcast delivery; retained to absorb the rest
+    });
+    c.applyCurrentTimeout(); // unrelax the receiving socket if the gate just shut
+    if (!firstDelivery[0]) {
+      return;
+    }
+    logger.debug("Slot migration done (seq={}, entries={})", e.seq, e.migrations.size());
+    if (cache.isRenewInFlight()) {
+      // Never block the read thread on a running full refresh; apply after it completes. If the
+      // refresh finished between the check and the enqueue, drain immediately — poll() makes a
+      // double drain harmless.
+      pendingDeltas.add(e);
+      if (!cache.isRenewInFlight()) {
+        drainPendingDeltas();
+      }
+      return;
+    }
+    applyDelta(e);
+  }
+
+  private void drainPendingDeltas() {
+    SMigratedEvent e;
+    while ((e = pendingDeltas.poll()) != null) {
+      applyDelta(e);
+    }
+  }
+
+  /**
+   * Applies the delta and retires the connections of nodes left without slots (Case 2). A node
+   * still owning slots after the delta keeps its connections untouched (Case 1). Retirement goes
+   * through the node pool's own controller, resolved now — a pool destroyed by a racing refresh
+   * simply no longer resolves.
+   */
+  private void applyDelta(SMigratedEvent e) {
+    Set<HostAndPort> slotless = cache.applyMigrationDelta(e.migrations);
+    if (slotless.isEmpty()) {
+      return;
+    }
+    long now = NanoClock.INSTANCE.getAsLong();
+    Map<String, ConnectionPool> nodes = cache.getNodes();
+    for (HostAndPort node : slotless) {
+      ConnectionPool pool = nodes.get(JedisClusterInfoCache.getNodeKey(node));
+      MaintenanceEventController controller = pool == null ? null : pool.getMaintenanceController();
+      if (controller != null) {
+        logger.debug("Node {} owns no slots after migration (seq={}); retiring its connections",
+          node, e.seq);
+        controller.retireAll(now);
+      }
+    }
+  }
+
+  /** One client-wide migration operation, folded from the per-node broadcast; keyed by seq. */
+  private static final class MigrationOperation {
+    final Object id;
+    final long deadlineNanos;
+    final boolean closed;
+
+    private MigrationOperation(Object id, long deadlineNanos, boolean closed) {
+      this.id = id;
+      this.deadlineNanos = deadlineNanos;
+      this.closed = closed;
+    }
+
+    MigrationOperation(Object id, long deadlineNanos) {
+      this(id, deadlineNanos, false);
+    }
+
+    static MigrationOperation closed(Object id, long deadlineNanos) {
+      return new MigrationOperation(id, deadlineNanos, true);
+    }
+
+    MigrationOperation close() {
+      return new MigrationOperation(id, deadlineNanos, true);
+    }
+
+    boolean isExpired() {
+      return deadlineNanos - NanoClock.INSTANCE.getAsLong() <= 0;
+    }
+  }
+
+  @Override
+  public void onMoving(MovingEvent e, Connection c) {
+    logger.warn("Standalone maintenance events are not supported by this controller: {} conn={}", e,
+      c);
+  }
+
+  @Override
+  public void onMigrating(MigratingEvent e, Connection c) {
+    logger.warn("Standalone maintenance events are not supported by this controller: {} conn={}", e,
+      c);
+  }
+
+  @Override
+  public void onMigrated(MigratedEvent e, Connection c) {
+    logger.warn("Standalone maintenance events are not supported by this controller: {} conn={}", e,
+      c);
+  }
+
+  @Override
+  public void onFailingOver(FailingOverEvent e, Connection c) {
+    logger.warn("Standalone maintenance events are not supported by this controller: {} conn={}", e,
+      c);
+  }
+
+  @Override
+  public void onFailedOver(FailedOverEvent e, Connection c) {
+    logger.warn("Standalone maintenance events are not supported by this controller: {} conn={}", e,
+      c);
+  }
+}
--- a/src/main/java/redis/clients/jedis/ClusterMaintenanceCoordinator.java
+++ b/src/main/java/redis/clients/jedis/ClusterMaintenanceCoordinator.java
@@ -19,8 +19,8 @@ import redis.clients.jedis.TimeoutSource.TimeoutInfo;
  * the last one closes, and a connection created mid-event relaxes from the moment its overlay is
  * installed. On SMIGRATED it applies the slot delta atomically through
  * {@link JedisClusterInfoCache#applyMigrationDelta} — queued and drained after completion when a
- * full topology refresh is in flight, never blocking a read thread — and retires the connections
- * of nodes left without slots via each pool's own {@link MaintenanceEventController}, resolved at
+ * full topology refresh is in flight, never blocking a read thread — and retires the connections of
+ * nodes left without slots via each pool's own {@link MaintenanceEventController}, resolved at
  * event time (no node map here). The pools remain their controllers' sole owners; this class only
  * calls.
  */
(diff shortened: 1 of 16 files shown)


src/main/java/redis/clients/jedis/ClusterMaintenanceCoordinator.java
```diff
@@ -0,0 +1,171 @@
...
+    c.applyCurrentTimeout(); // the gate just opened; push the relax to the receiving socket now
+  }
+
+  @Override
+  public void onSMigrated(SMigratedEvent e, Connection c) {
+    long deadline = NanoClock.INSTANCE.getAsLong() + maxRelaxedDurationNanos;
+    boolean[] firstDelivery = { false };
+    operations.compute(e.identity(), (k, cur) -> { // atomic per identity
+      if (cur == null) {
+        // SMIGRATED without a preceding SMIGRATING (e.g. connected mid-event): still applied
```
ggivo: looking at the code it looks. like  just misleading comment but want to confirm we  don;t expect seqId of SMIGRATED  to match opening  SMIGRATING seqId?

src/main/java/redis/clients/jedis/ClusterMaintenanceCoordinator.java
```diff
@@ -0,0 +1,171 @@
...
+
+  /**
+   * Seq ids are ordered, so a closing SMIGRATED also concludes any operation opened under a lower
+   * seq whose own SMIGRATED never arrived (lost or reordered) — the stale opener must not keep the
+   * relax gate held. Discarded entries are removed outright; a late lower-seq SMIGRATED would then
+   * re-apply as a fresh delta (the MOVED fallback self-heals the topology if it was stale).
+   */
+  private void discardOutdatedMigrationWindows(long closingSeq) {
+    for (MigrationOperation op : operations.values()) {
+      if (op.seq < closingSeq) {
```
ggivo: cleaning up an older seqId's will lead to reapplying them if observed on different connection (e.g. probably applying a stale deltas).

related to previous comment

src/main/java/redis/clients/jedis/ClusterMaintenanceCoordinator.java
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  @Override
+  public void onSMigrating(SMigratingEvent e, Connection c) {
+    if (logger.isDebugEnabled()) {
+      logger.debug("Slot migration starting: {} (seq={}) conn={}", e.slots, e.seq,
+        c.toIdentityString());
+    }
+    long deadline = NanoClock.INSTANCE.getAsLong() + maxRelaxedDurationNanos;
+    operations.computeIfAbsent(e.identity(), k -> new MigrationOperation(k, e.seq, deadline, true));
```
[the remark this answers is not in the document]
ggivo: Agree with the observation, not with the proposed solution. It can lead to opposite e.g. miss unprocessed older events.

Also need to verify in practice hoe likely is to end up with multiple stacked maintenance events

src/main/java/redis/clients/jedis/ClusterMaintenanceCoordinator.java
```diff
@@ -0,0 +1,171 @@
...
+    });
+    discardOutdatedMigrationWindows(e.seq);
+    c.applyCurrentTimeout(); // unrelax the receiving socket if the gate just shut
+    if (!firstDelivery[0]) {
+      return;
+    }
+    logger.debug("Slot migration done (seq={}, entries={})", e.seq, e.migrations.size());
+    // the cache queues and applies atomically against its refresh lifecycle; a delta racing a
+    // refresh may be applied later, on the refreshing/draining thread
+    cache.applySlotMigration(e.migrations);
```
ggivo: Discarding the older seqId means that if the same delta is later observed on another connection, it will be treated as new and re-applied — so we can end up applying a stale delta.

Assumption here is that applying same delta multiple time end in same stat, but what bothers me are side effect of creating connection pools for missing nodes

A couple of rough ideas if we decide to harden this later:
1. Keep the old seqId around for a while so a late duplicate isn't re-applied.
2. Track the last applied seqId per slot and apply a delta only if its seqId is newer.

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -365,6 +393,99 @@ public void assignSlotsToNode(List<Integer> targetSlots, HostAndPort targetNode)
...
+   * in separate acquisitions leaves an interleaving window against {@link #renewClusterSlots}. Only
+   * ever called under {@link #slotDeltaLock} (see {@link #drainSlotDeltas}).
+   * @return the source nodes that no longer own any slot (candidates for connection retirement);
+   *         their pools stay in place — removal belongs to the full-refresh dead-pool cleanup
+   */
+  private Set<HostAndPort> processSlotDelta(List<SlotMigration> migrations) {
+    w.lock();
+    try {
+      for (SlotMigration migration : migrations) {
+        ConnectionPool destPool = setupNodeIfNotExist(migration.dest);
```
ggivo, line 478: Hm. Applying older deltas might lead to creating obsolete nodes?

This relates to comment for hardening guard of applying an already applied delta. E.g.

src/main/java/redis/clients/jedis/ClusterMaintenanceCoordinator.java
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  /**
+   * Seq ids are ordered, so a closing SMIGRATED also concludes any operation opened under a lower
+   * seq whose own SMIGRATED never arrived (lost or reordered) — the stale opener must not keep the
+   * relax gate held. Discarded entries are removed outright; a late lower-seq SMIGRATED would then
+   * re-apply as a fresh delta (the MOVED fallback self-heals the topology if it was stale).
```
ggivo: hm. Not sure about that.
is it possible applying an older delta as side effects kick in creating  connection pools for missing nodes?

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -365,6 +393,99 @@ public void assignSlotsToNode(List<Integer> targetSlots, HostAndPort targetNode)
...
+        primaryNodesCache.put(getNodeKey(migration.dest), destPool);
+        migration.slots.forEachSlot(slot -> {
+          slots[slot] = destPool;
+          slotNodes[slot] = migration.dest;
+          if (replicaSlots != null) {
+            // the delta carries no replica placement; drop stale entries until the next refresh
+            replicaSlots[slot] = null;
+          }
```
ggivo: What is the reason to drop them? Should we keep the existing replica?

src/main/java/redis/clients/jedis/MaintenanceEventController.java
```diff
@@ -234,6 +234,16 @@ ConnectionRegistry registry() {
     return registry;
   }
 
+  /**
+   * Retires every registered connection at the given instant and runs the handoff hook — the
+   * cluster Case-2 reaction, invoked by the cluster coordinator when this pool's node no longer
+   * owns any slot. Caller-only access: the pool remains this controller's sole owner.
+   */
+  void retireAll(long retireAtNanos) {
```
ggivo, line 242: Unused?

ggivo (commented):
Done one review pass 
and have a few open question want to confirm. 

Adding some comments to doublecheck/discuss.

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
atakavci: will drop the reset operation and investigate if replicas are exposed with OSS API enabled setup.

src/main/java/redis/clients/jedis/MaintenanceEventController.java
atakavci: yes this will be dropped, i think it is dropped in next PR.

src/main/java/redis/clients/jedis/ClusterMaintenanceCoordinator.java
atakavci: will go through the comments and javadocs , looks like they are silently diverged from the implementation and guidelines i provided to AI

src/main/java/redis/clients/jedis/ClusterMaintenanceCoordinator.java
atakavci: some of the comments and javadocs are stale or misleading. Fixed them with AI, hope all is clear now.
the given assumption is completely wrong for the case above. 
`SMIGRATED`/`SMIGRATING` have different `seqId`.

src/main/java/redis/clients/jedis/ClusterMaintenanceCoordinator.java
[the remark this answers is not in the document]
atakavci: the suggested solution is legit for `SMIGRATING` events and addressed via checking against the highest `seqId` received/processed.  Also we keep a history of `SMIGRATED` events with this iteration, which addresses another issue related to out-of-order event processing.

src/main/java/redis/clients/jedis/ClusterMaintenanceCoordinator.java
atakavci: this is also fixed by the same set of changes with [this](https://github.com/redis/jedis/pull/4729#discussion_r3965546632).
- handling `SMIGRATING` in a different map and checking agains highest `seqId` processed
- SMIGRATED has a history of events to avoid re-processing while handling late-received/later-read events

src/main/java/redis/clients/jedis/ClusterMaintenanceCoordinator.java
```diff
@@ -0,0 +1,171 @@
...
+  public void onSMigrated(SMigratedEvent e, Connection c) {
+    long deadline = NanoClock.INSTANCE.getAsLong() + maxRelaxedDurationNanos;
+    boolean[] firstDelivery = { false };
+    operations.compute(e.identity(), (k, cur) -> { // atomic per identity
+      if (cur == null) {
+        // SMIGRATED without a preceding SMIGRATING (e.g. connected mid-event): still applied
+        firstDelivery[0] = true;
+        return new MigrationOperation(k, e.seq, deadline, false);
+      }
+      return cur; // duplicate broadcast delivery; retained to absorb the rest
```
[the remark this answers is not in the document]
atakavci: assumption on having same `seqId` for `SMIGRATING`/`SMIGRATED` is not correct. The issue is outdated with current implementation that holds them in different maps and apply delta from `SMIGRATED` even if they have same `seqId`.

src/main/java/redis/clients/jedis/ClusterMaintenanceCoordinator.java
atakavci: Approach now is to not apply older deltas other than one single case;
- when an `SMIGRATED` is received with a `seqId` never seen before but less than the last/highest `seqId` from `SMIGRATED` event.

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
atakavci: Approach now is to not apply older deltas other than one single case;
- when an `SMIGRATED` is received with a `seqId` never seen before but less than the last/highest `seqId` from `SMIGRATED` event.

src/main/java/redis/clients/jedis/ClusterMaintenanceCoordinator.java
atakavci: fixed the comments and `SMIGRATED` entries are not discarded outright anymore.

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
atakavci: addressed.

src/main/java/redis/clients/jedis/ClusterMaintenanceCoordinator.java
```diff
@@ -0,0 +1,312 @@
...
+      logger.info("Dropping late slot migration (seq={}): ordering context is ambiguous "
+          + "(retained range {}..{})",
+        seq, completedMigrations.firstKey(), newestSeq);
+      onDuplicateSMigrated(seq, c);
+      return;
+    }
+    discardOutdatedMigrationWindows(seq);
+    // the cache queues and applies atomically against its refresh lifecycle; a delta racing a
+    // refresh may be applied later, on the refreshing/draining thread
+    cache.applySlotMigration(toApply);
```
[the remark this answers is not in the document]
atakavci: `onSMigrated` implements a new approach where `seqId` is respected to keep the order of events in case multiple events queue up.

src/main/java/redis/clients/jedis/ClusterMaintenanceCoordinator.java
```diff
@@ -0,0 +1,171 @@
...
+ * Cluster maintenance coordinator — one per cluster client, the single dedup/apply point for the
+ * per-node SMIGRATING/SMIGRATED broadcast. Owns the seq-keyed operations table that (a) folds the
+ * N-connection broadcast into one client-wide operation and (b) drives the shared relax gate:
+ * timeouts stay relaxed while ANY operation is open, so overlapping migrations unrelax only when
+ * the last one closes, and a connection created mid-event relaxes from the moment its overlay is
+ * installed. On the first SMIGRATED delivery it hands the slot delta to
+ * {@link JedisClusterInfoCache#applySlotMigration}, which queues and applies it atomically against
+ * the refresh lifecycle — never blocking or spinning a read thread on a running refresh.
+ */
+final class ClusterMaintenanceCoordinator implements MaintenanceEventListener {
```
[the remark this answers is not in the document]
atakavci: this is subject of [another PR](https://github.com/redis/jedis/pull/4732).

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -365,6 +393,99 @@ public void assignSlotsToNode(List<Integer> targetSlots, HostAndPort targetNode)
...
+        });
```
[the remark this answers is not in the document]
atakavci: addressed;  `processSlotDelta()` runs `clientSideCache.flush()`

src/main/java/redis/clients/jedis/JedisClusterInfoCache.java
```diff
@@ -365,6 +393,99 @@ public void assignSlotsToNode(List<Integer> targetSlots, HostAndPort targetNode)
...
+    try {
+      for (SlotMigration migration : migrations) {
+        ConnectionPool destPool = setupNodeIfNotExist(migration.dest);
+        primaryNodesCache.put(getNodeKey(migration.dest), destPool);
+        migration.slots.forEachSlot(slot -> {
+          slots[slot] = destPool;
+          slotNodes[slot] = migration.dest;
+          if (replicaSlots != null) {
+            // the delta carries no replica placement; drop stale entries until the next refresh
+            replicaSlots[slot] = null;
```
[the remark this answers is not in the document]
atakavci: addressed; `processSlotDelta()` runs `clientSideCache.flush()`

src/main/java/redis/clients/jedis/ClusterMaintenanceCoordinator.java
```diff
@@ -0,0 +1,275 @@
...
+      return;
+    }
+    if (logger.isDebugEnabled()) {
+      logger.debug("Slot migration starting: {} (seq={}) conn={}", e.slots, e.seq,
+        c.toIdentityString());
+    }
+    long deadline = NanoClock.INSTANCE.getAsLong() + maxRelaxedDurationNanos;
+    migratingWindows.computeIfAbsent(e.seq, k -> new MigratingWindow(e.seq, deadline));
+    lastProcessedSeq.accumulateAndGet(e.seq, Math::max);
+    c.applyCurrentTimeout();
```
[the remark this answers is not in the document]
atakavci: addressed with [1033787](https://github.com/redis/jedis/pull/4729/commits/1033787692f7f9ee7f80966092940ef96d5fa8ad)

PR #4736, opened 2026-09-14
Fix user/password boundary for percent-encoded userinfo in JedisURIHelper

### LINKED ISSUE #4728: Percent-encoded colon in URI username shifts the user/password split

Rainmemery:
Fixes #4728.

## What changed

`JedisURIHelper.getUser()` and `getPassword()` used `URI.getUserInfo()`, which returns the already-decoded form, and then split on `':'` — so a percent-encoded colon inside the username (`redis://us%3Aer:pw@host`) had already become a literal `:` before the split, shifting the user/password boundary.

The split now happens on `URI.getRawUserInfo()` (still encoded, where the only `:` is the real separator), and each component is percent-decoded afterwards.

## Decoding details

- New private `decodeUserInfoComponent()` decodes `%XX` octets as UTF-8, matching what `URI.getUserInfo()` produces, and throws `IllegalArgumentException` on malformed escapes.
- It deliberately does **not** use `URLDecoder`, since userinfo is not form data and `URLDecoder` would turn a `+` into a space — a literal `+` in a password must survive (now covered by a test, including its `%2B` form).

## Behavior compatibility

For URIs without percent-escapes the raw and decoded userinfo are identical, so existing behavior is unchanged; all pre-existing tests pass untouched.

## Tests

Added to `JedisURIHelperTest`:

- `shouldKeepEncodedColonInUsername` — the repro from the issue (`us%3Aer:pw` → user `us:er`, password `pw`);
- `shouldDecodeEncodedColonInPassword` — encoded colon inside the password;
- `shouldKeepPlusSignInPassword` — literal `+` and `%2B` both stay `+`;
- `shouldDecodeNonAsciiUserAndPassword` — UTF-8 octet decoding matches `URI.getUserInfo()`.

Full `JedisURIHelperTest` (22 tests) passes on JDK 17 against Maven `master` (formatter validation clean on the touched files).

---

> [!NOTE]
> **Medium Risk**
> Changes how URI-derived username and password are parsed for all `Jedis` connections using credentials with percent-encoding; fixes mis-authentication but may differ from the previous incorrect split.
> 
> **Overview**
> **Fixes Redis URI user/password parsing when credentials contain percent-encoding** (e.g. `redis://us%3Aer:pw@host`).
> 
> `JedisURIHelper.getUser()` and `getPassword()` now split on `URI.getRawUserInfo()` and percent-decode each side via a new `decodeUserInfoComponent()`, so an encoded colon in the username no longer moves the separator. Decoding uses UTF-8 for `%XX` sequences, rejects malformed escapes, and **does not** treat `+` as space (unlike `URLDecoder`).
> 
> Release notes document this as an 8.1.0 behavior change; only apps that relied on the old wrong split for `%3A` in usernames may see different auth. Tests cover encoded colons, `+`/`%2B`, and non-ASCII userinfo.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a19732fa322d7d75ce16c552aa41b0f0c963f03c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Rainmemery:
Addressed the review findings in a19732f:

1. **Mixed literal non-ASCII + escapes corrupted (P2/Cursor)**: `decodeUserInfoComponent` wrote every unescaped character as a single octet and UTF-8-decoded the whole buffer, so `redis://é%20x:pw@host` decoded the user as `\uFFFD x` instead of `é x`. Unescaped characters are now kept as-is and only percent-escaped octet sequences are UTF-8-decoded. Added `shouldPreserveLiteralNonAsciiMixedWithEscapes` covering literal characters mixed with escapes in both user and password.

2. **Release notes (P1)**: documented the raw-userinfo split behavior change in `docs/release-notes/8.1.0.md` per AGENTS.md.

ggivo:
@Rainmemery 
Thanks for the contribution! 
Closing this one as it's already addressed by #4617
 It takes the same approach (split the raw userinfo,  decode after), but reuses the JDK `URLDecoder` instead of a hand-rolled percent-decoder, which leaves less code to maintain.

Issue #3839, opened 2024-05-17
BUG/Connection Leak：cannot close SentinelListener connection in a JedisSentineled/JedisSentinelPool mode when executing method close


peng-lin:
### Expected behavior

Close sentinelListener when excuting method jedisSentineled.close().

### Actual behavior

Cannot close sentinelListener thread

### Steps to reproduce:

Please create a reproducible case of your problem. Make sure
that case repeats consistently and it's not random
1. init JedisSentineled object and close it immediately in a while true loop or in a multi-threaded concurrent environment  
2. Root Cause：concurrent problems；thread safety
<img width="576" alt="problem-0" src="https://github.com/user-attachments/assets/1756b546-53cb-483d-bb6f-4f69fa6d50dc">

![problem-1](https://github.com/user-attachments/assets/41cd5b91-5a7f-4bce-8046-56b8c8923230)

![problem-2](https://github.com/user-attachments/assets/f0efbccd-9531-49ab-9e30-7304c79955ca)

### how to resolve：
see the pictures

![resolve](https://github.com/user-attachments/assets/685c146e-f270-4a30-8e3d-f2c1e0b60e00)

alexciuca:
Hi is it okay if I work on this issue. I am a junior developer and want to contribute to meaningful open source projects.

I’ll try to reproduce the SentinelListener connection leak, add tests, and open a PR if I can confirm the fix.
Thanks!

ggivo:
@alexciuca 
Sure!

kimsanhaa:
Here is a reproducer for the connection-initialization race addressed by #4739. It uses a real Redis master and Sentinel through both `RedisSentinelClient` and `JedisSentinelPool`, without mocks or reflection.

A latch in `JedisClientConfig.getClientName()` pauses the listener after the socket connects but before its new `Jedis` instance is assigned to the listener field. The main thread calls `close()` and then releases it. On `d6df23c5`, both variants leave a live listener and a Sentinel subscription (`CLIENT LIST` shows `sub=1 cmd=subscribe`). On `029e599d`, both terminate and close the connection. Master `PING` succeeds before closing in both versions.

Tested with Corretto 8u504 and Redis/Sentinel 7.4.11. This covers the connection-creation window, not every possible concurrent-close scenario.

Save these three files in the same directory:

<details>
<summary>SentinelCloseReproducer.java</summary>

```java
import java.util.Collections;
import java.util.UUID;
import java.util.concurrent.CountDownLatch;
[...]

PR #4739, opened 2026-09-17
Preserve Sentinel listener shutdown during startup

kimsanhaa:
Sentinel listeners can lose a shutdown request when `run()` starts after the client is closed. They can also enter the subscription loop with a connection created while shutdown was in progress.

Initialize the running flag before the thread starts and check it after publishing each new connection. The existing `finally` block then closes connections created during shutdown. Apply the change to both `SentineledConnectionProvider` and `JedisSentinelPool`.

Add deterministic tests for shutdown before startup, shutdown during connection creation, and normal subscription in both implementations.

Related to #3839.

Validated with Java 8: full unit suite (3,430 tests, 6 skipped, no failures), Javadoc build, and formatter checks. Redis integration tests were not run because the Docker daemon was unavailable.

---

> [!NOTE]
> **Medium Risk**
> Touches Sentinel failover listener lifecycle in both legacy pool and connection provider; behavior change is narrow but affects shutdown and connection timing on background threads.
> 
> **Overview**
> Fixes a race where Sentinel listener threads could ignore an early `shutdown()` and still open or subscribe on a Sentinel connection.
> 
> In **`JedisSentinelPool.MasterListener`** and **`SentineledConnectionProvider.SentinelListener`**, `running` now starts as `true` (removed the `running.set(true)` at the top of `run()`). Each loop iteration creates the Sentinel connection first, then bails out if `running` was cleared—so a shutdown that happens before `run()` or while the connection is being created skips `sentinelGetMasterAddrByName` / `subscribe`, and the existing `finally` closes any connection that was just created.
> 
> Adds **`MasterListenerShutdownTest`** and **`SentinelListenerShutdownTest`** with mocked connections covering shutdown before `run()`, shutdown during connection creation, and the normal subscribe path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c562b2bf96d72f44c4998219714fd9f92ff94d7a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ggivo:
@kimsanhaa 

Thanks for the PR.

The fix looks good. No release notes needed — this is a bug fix that requires no user code changes.
I would like to take another look at the provided tests &  before merging, I'd like to verify the fix against an actual reproducer. Could you attach a minimal reproducer to the related issue — code that triggers the bug before the fix and passes after — so we can confirm the end-to-end scenario is resolved?

kimsanhaa:
Removed the release note in c562b2bf and posted the runnable Sentinel reproducer in https://github.com/redis/jedis/issues/3839#issuecomment-5791826344. Both client variants fail before the fix and pass after it against Redis/Sentinel 7.4.11.

I also checked the failed doctests job. `HomeJsonExample` receives zero search results and throws at `paulDocs.get(0)` on line 104. It uses standalone `RedisClient`; that example and its search code are unchanged by this PR. Running the full doctests suite with Java 11 and Redis 8.10.2 passed all 37 tests on both the base commit d6df23c5 and this PR. I have not reproduced the CI failure, so the reason for the empty result is still unconfirmed.

Could you rerun the [doctests job](https://github.com/redis/jedis/actions/runs/35709944203/job/107075726520)? My rerun request was rejected with an Actions permission error.

ggivo:
Thanks

> Could you rerun the [doctests job](https://github.com/redis/jedis/actions/runs/35709944203/job/107075726520)? My rerun request was rejected with an Actions permission error.

Will do. There is an instability after bumping test infra to Redis 8.10 not related to this PR!

PR #4740, opened 2026-09-18
Compare SortingParams by pattern content

renechoi:
## Summary

`by()` and `get()` store their patterns as byte arrays in a `List<Object>`, so the
`Objects.equals(params, that.params)` in `equals()` compares them by reference identity: two
instances built from the same arguments are never equal and their hash codes differ. The `String`
overloads funnel into the byte array ones, so `by("weight_*")` is affected too.

## Key Decisions

`Arrays.deepEquals` and `Arrays.deepHashCode` dispatch to the array-aware comparison for the nested
byte arrays, which keeps the existing `List<Object>` layout and the wire format untouched.
`ScanParams` reaches the same result by wrapping its patterns in `ByteBuffer`.

## Testing

`SortingParamsTest` gains cases for string patterns, binary patterns and a differing pattern; the
first two fail on master. Full unit suite: 3299 tests, 0 failures.

---

> [!NOTE]
> **Low Risk**
> Localized fix to equals/hashCode contract for SORT params; no protocol or API surface changes beyond correct value semantics.
> 
> **Overview**
> Fixes **`SortingParams` equality and hashing** so two instances built with the same `by()` / `get()` patterns compare as equal instead of always differing because patterns live as distinct `byte[]` entries in the internal list.
> 
> `equals()` now uses **`Arrays.deepEquals`** on the param list contents and **`hashCode()`** uses **`Arrays.deepHashCode`**, so string and binary pattern overloads behave consistently without changing command wire format.
> 
> **`SortingParamsTest`** adds coverage for matching string patterns, matching encoded binary patterns, and unequal patterns (including hash contract where applicable).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c5244030536dc4e63475ceb943a163c326a96d89. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ggivo:
@renechoi 
 Before we review the fix, please open an issue describing the problem from the user's perspective. Include:
 
 - The concrete scenario: what are you trying to do and what goes wrong
 - A minimal reproducer — code or commands that trigger the issue against an actual Redis server
 - Expected vs actual output
 
 Without a reproducer we cannot verify the fix end-to-end or confirm the scenario is actually resolved.

PR #4742, opened 2026-09-19
Fix equality for GEO responses without coordinates

jakezwang:
## Summary

Comparing separate `GeoRadiusResponse` instances throws `NullPointerException` when `WITHCOORD` was omitted. Use `Objects.equals` for the optional coordinate, preserving equality and hash-code behavior when coordinates are present.

Regression tests cover equality, unequal members, mixed coordinate presence, hash-based lookup, and responses built with `WITHDIST`/`WITHHASH`. The 8.1 release notes document the correction.

## Testing

All four new tests fail on the original implementation. On JDK 8, `mvn -Dwith-param-names=true test jacoco:report -Djacoco.dataFile=target/jacoco-ut.exec javadoc:jar` passes: 3,428 tests, 6 skipped. Formatting validation passes for the new test file.

The first full run failed in the unrelated `MultiDbClientRebindMockTest.eachDatabaseHandlesItsOwnMovingAfterFailoverRebuild`; an isolated baseline run and the complete rerun passed. Redis-backed integration tests were not run.

---

> [!NOTE]
> **Low Risk**
> Small null-safety fix in equals/hashCode for an optional field, with regression tests and release-note documentation only.
> 
> **Overview**
> Fixes **`GeoRadiusResponse.equals()`** so comparing GEO radius results that omit **`WITHCOORD`** no longer throws **`NullPointerException`**. Coordinate comparison now uses **`Objects.equals`**, so two absent coordinates are equal and one absent vs one present is unequal; behavior when both sides have coordinates is unchanged.
> 
> Adds unit tests for equality, hash codes, **`HashSet`** lookup, mixed coordinate presence, and lists built via **`BuilderFactory.GEORADIUS_WITH_PARAMS_RESULT`**. Documents the behavior change in the **8.1.0** release notes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3da48d2569502e432522741a5a72c2c12c3483c2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ggivo:
@jakezwang 
 Before we review the fix, please open an issue describing the problem from the user's perspective. Include:
 
 - The concrete scenario: what are you trying to do and what goes wrong
 - A minimal reproducer — code or commands that trigger the issue against an actual Redis server
 - Expected vs actual output
 
 Without a reproducer, we cannot verify the fix end-to-end or confirm the scenario is actually resolved.

jakezwang:
This was independently discovered, not a production incident. I have now reproduced it against a real Redis 8.10.2 server with JDK 8 using GEOSEARCH WITHDIST WITHHASH without WITHCOORD. Two identical queries return responses with null coordinates; comparing their lists or doing a HashSet lookup throws NullPointerException on unpatched master. The new `GeoRadiusResponseIT` covers RESP2 and RESP3 and passes with the fix; the four unit tests also pass. The branch is updated to current master. Reproducer issue: https://github.com/redis/jedis/issues/4746

docs/release-notes/8.1.0.md
```diff
@@ -25,6 +25,14 @@ RESP3 connections; pass `MaintenanceNotificationsConfig.DISABLED` to opt out. Se
...
 not breaking API changes.
 
+### Comparing GEO responses without coordinates
+
+`GeoRadiusResponse.equals()` now compares responses that omit `WITHCOORD`
+without throwing `NullPointerException`. Two missing coordinates compare equal;
+a missing coordinate and a present coordinate compare unequal.
+
+**Action required:** none.
+
```
ggivo: no need for release not, it is an actual straightforward bug and no actions required fro the users

ggivo: The release notes entry is not required— a NPE fix requires no user action and isn't a behavior change.

src/test/java/redis/clients/jedis/misc/GeoRadiusResponseIT.java
ggivo: The integration test in `GeoRadiusResponseIT` would fit better as a test method in [`[GeoCommandsTestBase](https://claude.ai/epitaxy/src/test/java/redis/clients/jedis/commands/unified/GeoCommandsTestBase.java)`](src/test/java/redis/clients/jedis/commands/unified/GeoCommandsTestBase.java). That base class is already extended by `RedisClientGeoCommandsTest` and `ClusterGeoCommandsTest`, both parameterized over RESP2/RESP3, so the scenario runs against all client types and protocols

